PR #427, opened 2023-09-16
POC for RFC - Add exclude flag

### LINKED ISSUE #426: -exclude flag for targeted plan/apply
### Summary

This RFC details a new flag, `-exclude`, which would do the inverse of `-target` - that is, plan/apply all resources _except_ those specified by the exclude flag and their descendants.

### Problem Statement

Only for use in exceptional circumstances as is the case for `-target`, this makes it a lot easier to plan/apply all changes to an environment except ones with known issues eg where there is a problem with the infrastructure or the committed config for one module/resource, but you still want to apply everything else.

### User-facing description

Usage example:
`opentf plan -exclude module.something_broken` / `opentf apply -exclude module.something_broken`

The `-exclude` feature could be explained as an argument that allows you to plan/apply all resources except a specified resource.

### Technical Description

- `-exclude` flag added to `extendedFlagSet` when an operation is provided (same as `-target`)
- Passed through to `OperationRequest`
- `TargetsTransformer.Transform` should remove vertices that _are_ in `t.Excludes` if `t.Excludes` are provided (just like how it removes vertices that _aren't_ in `t.Targets`).
- If both targets and excludes are provided, use the most specific path eg:
  ``` 
  -exclude module.entire_environment \
  -include module.entire_environment.module.network \
  -include module.entire_environment.module.application \
  -exclude module.entire_environment.module.application.module.eks
  ```
  should exclude all of the entire_environment module but include the network and application modules within, _except_ the eks module within the application module. This should be the outcome regardless of the order of the flags.
- Return an error when -target and -exclude point at the exact same resource
- When destroying, if there are dependencies of an excluded resource, the dependencies should _also_ be kept (excluded from change)

The rationale is that this is a highly requested tf feature, which makes certain scenarios (eg a single resource in a module is broken, but you still need to apply the module) potentially easier to resolve since you can exclude the offending resource.

### Downsides

I can't think of any downsides, unless opentf is intended to support the exact same flags as tf.

### Unresolved Questions

In my poc branch I've added the flag, and added the `TargetsTransformer` change that should make excludes work. I've also followed through a few places where targets are passed around and added excludes there too.

However I'm not too familiar with the opentf architecture other than a brief look at the doc, so there may be things I've missed to make this work, and places I've added excludes that didn't really need it. Any pointers from people familiar with the code would be appreciated.

Also since we can't change TFE which is imported from a Hashicorp project, I presume there would be backends that we can't support this flag for? At the very least I guess this would require me to add errors when somebody tries to pass `-exclude` using unsupported backends? If not a complete blocker for this feature ever going in.

### Related Issues

https://github.com/hashicorp/terraform/issues/2253 this issue in the original Terraform repo requests an exclude flag that does the inverse of `-target`, and is very highly requested.

### Proof of Concept

https://github.com/opentffoundation/opentf/compare/main...JamesBalazs:opentf-excludes:add-exclude-flag is a fork (POC PR https://github.com/opentffoundation/opentf/pull/427) with what I _think_ is the necessary (surprisingly simple) change to `TargetsTransformer.Transform`, and the flag added to `extendedFlagSet`. I have not been able to test if this works yet since I don't have anything using tf on my home macbook, and I threw this together at 2am so it still needs unit tests etc adding.

JamesBalazs:
Add `-exclude` flag for apply/plan

Resolves #426

## Target Release

1.6.0

-

### DIFF
--- a/internal/backend/remote/backend_plan.go
+++ b/internal/backend/remote/backend_plan.go
@@ -110,7 +110,7 @@ func (b *Remote) opPlan(stopCtx, cancelCtx context.Context, op *backend.Operatio
 	// equivalent to an API version < 2.3.
 	currentAPIVersion, parseErr := version.NewVersion(b.client.RemoteAPIVersion())
 
-	if len(op.Targets) != 0 {
+	if len(op.Targets) != 0 || len(op.Excludes) != 0 {
 		desiredAPIVersion, _ := version.NewVersion("2.3")
 
 		if parseErr != nil || currentAPIVersion.LessThan(desiredAPIVersion) {
@@ -311,6 +311,13 @@ in order to capture the filesystem context the remote workspace expects:
 	if len(op.Targets) != 0 {
 		runOptions.TargetAddrs = make([]string, 0, len(op.Targets))
 		for _, addr := range op.Targets {
+			runOptions.TargetAddrs = append(runOptions.TargetAddrs, addr.String())
+		}
+	}
+
+	if len(op.Excludes) != 0 {
+		runOptions.TargetAddrs = make([]string, 0, len(op.Excludes))
+		for _, addr := range op.Excludes {
 			runOptions.TargetAddrs = append(runOptions.TargetAddrs, addr.String())
 		}
 	}
--- a/internal/opentf/transform_targets.go
+++ b/internal/opentf/transform_targets.go
@@ -17,6 +17,7 @@ import (
 // node must filter this list to targets considered relevant.
 type GraphNodeTargetable interface {
 	SetTargets([]addrs.Targetable)
+	SetExcludes([]addrs.Targetable)
 }
 
 // TargetsTransformer is a GraphTransformer that, when the user specifies a
@@ -24,12 +25,13 @@ type GraphNodeTargetable interface {
 // their dependencies.
 type TargetsTransformer struct {
 	// List of targeted resource names specified by the user
-	Targets []addrs.Targetable
+	Targets  []addrs.Targetable
+	Excludes []addrs.Targetable
 }
 
 func (t *TargetsTransformer) Transform(g *Graph) error {
 	if len(t.Targets) > 0 {
-		targetedNodes, err := t.selectTargetedNodes(g, t.Targets)
+		targetedNodes, err := t.selectTargetedNodes(g, t.Targets, false)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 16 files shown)


yaronya:
Hi @JamesBalazs 
Thank you for your contribution, we really appreciate it 🙏 

Please note that #426 is still "pending decision" (see label) and requires the project's steering committee to decide whether this enhancement is something that should be supported or not. 
I suggest we first wait on the GH issue to hear more thoughts before reaching implementation.

How that sounds?

internal/backend/remote/backend_plan.go
```diff
@@ -315,6 +315,13 @@ in order to capture the filesystem context the remote workspace expects:
 		}
 	}
 
+	if len(op.Excludes) != 0 {
```
Hefeweizen, line 318: This does the same thing for Excludes that is done for Includes on L311-314 above.  Is that right?  I would have expected some sort of "inverse" operation.

internal/opentf/transform_targets.go
```diff
@@ -42,13 +44,27 @@ func (t *TargetsTransformer) Transform(g *Graph) error {
 		}
 	}
 
+	if len(t.Excludes) > 0 {
+		targetedNodes, err := t.selectTargetedNodes(g, t.Excludes, true)
```
Hefeweizen, line 48: a variable name of `excludedNodes` would preserve cognitive context when reading through the code.

Throughout this change we have two categories:  targeted and excluded.  Here we seem to cross the streams, potentially creating confusion.

L54-57 becomes:
```
			if excludedNodes.Include(v) {
				log.Printf("[DEBUG] Removing %q, filtered by excludes.", dag.VertexName(v))
				g.Remove(v)
```
and later readers of the code don't have to ask themselves why targeted nodes are being removed.

Hefeweizen (commented):
Please describe the intended include/exclude behaviour, either on the PR message, in the usage flag, or in the changelog.  This will allow reviewers to better analyze the code.

JamesBalazs:
> Hi @JamesBalazs Thank you for your contribution, we really appreciate it 🙏
> 
> Please note that #426 is still "pending decision" (see label) and requires the project's steering committee to decide whether this enhancement is something that should be supported or not. I suggest we first wait on the GH issue to hear more thoughts before reaching implementation.
> 
> How that sounds?

Hi @yaronya - this PR is in response to the RFC template that states:

> Working experimental PoCs in a draft Pull Request are a very welcome attachment to an RFC, especially ones that are easy to check out and take for a spin.

Hence this wasn't supposed to be a working implementation, just a branch exploring how this would be implemented, or if it is even remotely possible. Hence why this PR is in draft, with POC in the title.

However it looks like a few months on, while that RFC has had a response from @Hefeweizen, it is still in `pending-decision`.

I don't know what the process is for a decision to be made on an RFC, but as you suggested, I am waiting for any movement on #426 before revisiting this PR and trying to get something working. It doesn't look like this would be too hard to implement once a decision is made though.

internal/backend/remote/backend_plan.go
JamesBalazs: I think I gave up with this change, since it's not possible to change `tfe.RunCreateOptions` to store `ExcludeAddrs`.

This is why I commented on the RFC that it might not be possible to support for certain backends (although I believe you went on to respond that the backend actually don't use the targets for anything).

When revisiting this / redoing this PR if the RFC were accepted, I would probably leave out all changes to this file if they're not needed.

PR #440, opened 2023-09-18
internal/states/statefile: TestReadErrNoState()

### LINKED ISSUE #594: Add Tests for internal/states/statefile ErrNoState
### OpenTofu Version

```shell
.
```

### Use Cases

.

### Attempted Solutions

.

### Proposal

The Read() function in the statefile package can return the custom ErrNoState in two places. First if the io.Reader that it takes as an argument is a nil file, secondly if the io.Reader is an empty file.

Neither of these conditions are currently tested.

### References

https://github.com/opentofu/opentofu/issues/439

alrs:
This adds `TestReadErrNoState()` to exercise the places where `Read()` can return an `ErrNoState` error.

There are no user-facing changes that warrant a CHANGELOG entry.

Fixes #594

### DIFF
--- a/internal/states/statefile/read_test.go
+++ b/internal/states/statefile/read_test.go
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package statefile
+
+import (
+	"errors"
+	"os"
+	"testing"
+)
+
+func TestReadErrNoState_emptyFile(t *testing.T) {
+	emptyFile, err := os.Open("testdata/read/empty")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer emptyFile.Close()
+
+	_, err = Read(emptyFile)
+	if !errors.Is(err, ErrNoState) {
+		t.Fatalf("expected ErrNoState, got %T", err)
+	}
+}
+
+func TestReadErrNoState_nilFile(t *testing.T) {
+	nilFile, err := os.Open("")
+	if err == nil {
+		t.Fatal("wrongly succeeded in opening non-existent file")
+	}
+
+	_, err = Read(nilFile)
+	if !errors.Is(err, ErrNoState) {
+		t.Fatalf("expected ErrNoState, got %T", err)
+	}
+}


eranelbaz:
Hey @alrs,
Thank you very much for all the pull requests you created to help us maintain and update the codebase.
Please for the next pull request, make sure to get community approval before you create one

internal/states/statefile/read_test.go
```diff
@@ -0,0 +1,32 @@
...
+	_, err = Read(emptyFile)
+	if !errors.Is(err, ErrNoState) {
+		t.Fatalf("expected ErrNoState, got %T", err)
+	}
+
+	nilFile, err := os.Open("")
+	if err == nil {
+		t.Fatal("wrongly succeeded in opening non-existent file")
+	}
+
```
eranelbaz, line 27: do we really need that check? as `os` is golang internal package

internal/states/statefile/read_test.go
```diff
@@ -0,0 +1,32 @@
...
+	emptyFile, err := os.Open("testdata/read/empty")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer emptyFile.Close()
+
+	_, err = Read(emptyFile)
+	if !errors.Is(err, ErrNoState) {
+		t.Fatalf("expected ErrNoState, got %T", err)
+	}
```
eranelbaz, line 21: can you split those into two different cases?

eranelbaz (commented):
Thank you for those tests, left couple of comments

internal/states/statefile/read_test.go
alrs: I didn't put it there to test `os`, I put it there to communicate to future devs what the test is up to. I could change this to a comment, though having it like this puts up a speed-bump to having the test accidentally neutered.

internal/states/statefile/read_test.go
alrs: Done.

eranelbaz (approved):
Thanks @alrs for helping us with better code coverage 🙏

PR #486, opened 2023-09-20
backend/s3: add allowed/forbidden account_ids argument

moabukar:
Adds `allowed_account_ids` and `forbidden_account_ids` parameters as convenience helpers to prevent unintended modifications to certain environments.

Fix https://github.com/opentofu/opentofu/issues/489

Enhancements

- `allowed_account_ids` and `forbidden_account_ids` arguments to prevent unintended modifications to specified environments.

### DIFF
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -61,6 +61,16 @@ func (b *Backend) ConfigSchema() *configschema.Block {
 				Optional:    true,
 				Description: "AWS region of the S3 Bucket and DynamoDB Table (if used).",
 			},
+			"allowed_account_ids": {
+				Type:        cty.Set(cty.String),
+				Optional:    true,
+				Description: "List of allowed AWS account IDs.",
+			},
+			"forbidden_account_ids": {
+				Type:        cty.Set(cty.String),
+				Optional:    true,
+				Description: "List of forbidden AWS account IDs.",
+			},
 			"dynamodb_endpoint": {
 				Type:        cty.String,
 				Optional:    true,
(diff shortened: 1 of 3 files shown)


eranelbaz:
Hey @moabukar 
Thank you for your contribution.
Make sure to read the [CONTIRBUTING](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md) guide - 
- First create an issue
- Seek community approval
- Only then create PR to resolve the issue

moabukar:
> Hey @moabukar Thank you for your contribution. Make sure to read the [CONTIRBUTING](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md) guide -
> 
> * First create an issue
> * Seek community approval
> * Only then create PR to resolve the issue

Hey, thank you!

Sorry about that, I have created the issue and the link here.

marcinwyszynski (changes requested):
I'm afraid that the code change does not correspond to the original intention.

First of all, this change is scoped to the S3 state backend, so most you could get out of it is making sure that you are not reading and writing to an S3 bucket hosted by a non-kosher account. You're definitely not solving the problem of changing AWS resources - this would need to be solved at the [AWS provider](https://github.com/hashicorp/terraform-provider-aws) level.

Second, you're only adding fields to the config but you're not using these fields in any way. So there's no actual mechanism to prevent what you're trying to prevent - this may be a good FYI for sure but so would be a simple comment in the code.

Third, the S3 backend is sometimes used with [S3-compatible services like Minio](https://pet2cattle.com/2021/06/minio-backend-terraform) where the concept of AWS account does not make sense. Given that the bucket name is meant to be globally unique, I would propose that - if we decide we really need this feature - we allowlist / blocklist bucket names rather than AWS accounts.

Last but not least, I am not 100% sure that this feature is universally useful. We're potentially adding complexity to a frequently used API, which puts some cognitive tax on its every user. If the feature is valuable to many users then the trade-off feels acceptable.

eranelbaz:
@marcinwyszynski we are talking over it here - #489 
to understand if we need that feature

cube2222:
closing due to inactivity

Issue #442, opened 2023-09-19
Cannot create temporary file to update credentials during `opentf login`


mermoldy:
### OpenTF Version

```shell
1.6.0
```

### OpenTF Configuration Files

any configuration

### Debug Output

➜ TF_LOG=trace opentf login app.terraform.io
2023-09-19T16:01:56.760+0300 [INFO]  OpenTF version: 1.6.0 dev
2023-09-19T16:01:56.760+0300 [DEBUG] using github.com/hashicorp/go-tfe v1.32.0
2023-09-19T16:01:56.760+0300 [DEBUG] using github.com/hashicorp/hcl/v2 v2.17.0
2023-09-19T16:01:56.760+0300 [DEBUG] using github.com/hashicorp/terraform-svchost v0.1.1
2023-09-19T16:01:56.760+0300 [DEBUG] using github.com/zclconf/go-cty v1.13.2
2023-09-19T16:01:56.760+0300 [INFO]  Go runtime version: go1.21.1
2023-09-19T16:01:56.760+0300 [INFO]  CLI args: []string{"opentf", "login", "app.terraform.io"}
2023-09-19T16:01:56.760+0300 [INFO]  This build of OpenTF allows using experimental features
2023-09-19T16:01:56.761+0300 [TRACE] Stdout is a terminal of width 94
2023-09-19T16:01:56.761+0300 [TRACE] Stderr is a terminal of width 94
2023-09-19T16:01:56.761+0300 [TRACE] Stdin is a terminal
2023-09-19T16:01:56.761+0300 [DEBUG] Attempting to open CLI config file: /Users/mermoldy/.opentfrc
2023-09-19T16:01:56.761+0300 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2023-09-19T16:01:56.761+0300 [DEBUG] ignoring non-existing provider search directory terraform.d/plugins
2023-09-19T16:01:56.761+0300 [DEBUG] ignoring non-existing provider search directory /Users/mermoldy/.terraform.d/plugins
2023-09-19T16:01:56.761+0300 [DEBUG] ignoring non-existing provider search directory /Users/mermoldy/Library/Application Support/io.terraform/plugins
2023-09-19T16:01:56.761+0300 [DEBUG] ignoring non-existing provider search directory /Library/Application Support/io.terraform/plugins
2023-09-19T16:01:56.761+0300 [INFO]  CLI command args: []string{"login", "app.terraform.io"}
[...]

PR #512, opened 2023-09-20
OpenTofu will now attempt to create the configuration directory ~/.terraform.d on startup

mermoldy:
Resolves https://github.com/opentofu/opentofu/issues/442


main.go
```diff
@@ -491,3 +500,28 @@ func extractChdirOption(args []string) (string, []string, error) {
 	copy(newArgs[argPos:], args[argPos+1:])
 	return argValue, newArgs, nil
 }
+
+// Creates the the configuration directory.
+// `configDir` should refer to `~/.terraform.d` or its equivalent
+// on non-UNIX platforms.
+func createConfigDir(configDir string) error {
+	configDirInfo, err := os.Stat(configDir)
+	if os.IsNotExist(err) {
```
ilmax, line 509: Isn't better to just try create the directory directly and handle the error? This if is not safe and the directory can be created (even though chances are low) after the IsNotExists and before the Mkdir

mermoldy: @ilmax Makes sense. I've removed the os.Stat call.

cube2222 (approved):

Yantrio (approved):

cube2222:
Thanks for working on this @mermoldy! Looks good to me, merging.

PR #397, opened 2023-09-13
don't indent JSON in state files

imjasonh:
My colleague @jonjohnsonjr has been doing quite a lot of profiling of `terraform`, and has found some possible areas where time is being wasted unnecessarily.

The following flamegraph shows where time is spent running `terraform apply` on an admittedly pretty pathologically large TF graph (aren't they all 🤷). It shows TF state being marshalled to JSON 4 times, taking up about 1/2 of the total time. This is done so that the previous/new state can be diffed, to determine if any changes actually need to be made.

Roughly half of each of those JSON marshallings is `appendIndent`, which is only called with `json.MarshalIndent`.

Screenshot 2023-09-13 at 1 43 20 PM

The theory is that dropping indentation would cut this 22s apply down to about 17s.

I am not a deep enough TF expert to know what other implications this change might have. Based on a brief look around, I don't believe it would have any impact on state diffing, since both sides of the diff go through this same code path to accomplish the diff (and so would both be unmarshalled or not).

There might be other cases where state is expected/required to be marshalled for some reason; AFAIK, no human is expected to read that directly, but maybe something somewhere cares about there being newlines and spaces. If nothing else I'd love to learn through this process what that might be, if this change isn't safe after all.

PR #579, opened 2023-09-25
statemgr: Make Filesystem impl "correct" and fast

### LINKED ISSUE #578: Managing local state could be much faster
### OpenTofu Version

```shell
OpenTofu v1.6.0-dev
on darwin_arm64
```

### Use Cases

We manage a few thousand resources that each contribute a nontrivial amount of state to our `tfstate` file. As the number of resources has grown, we have noticed that `tofu apply` has slowed down tremendously.

I would like to make it feasible to manage large numbers of resources and large amounts of state in a single `tfstate` file.

### Attempted Solutions

https://github.com/opentofu/opentofu/pull/397

We have experimented with sharding the state across multiple invocations using terragrunt, but that doesn't feel like a proper solution.

### Proposal

I have described the problem with a repro and possible solutions here: https://github.com/jonjohnsonjr/slow-tfstate

Fixing the `Filesystem` implementation such that it doesn't `PersistState` constantly is fairly straightforward and makes things remarkably faster.

In the linked repro, I have a `fast` directory and a `slow` directory. The slow directory reads a local file with 1MB of random data. That inflates the `tfstate` file to be ~5MB. When I apply this, it takes almost 3 minutes, spending most of its time in WriteState:

image

If I "fix" `Filesystem` to behave like other state backends, it only takes ~9 seconds:

image'

Now... taking 9 seconds in order to sleep for 1 second seems not great, but we can tackle that next.

### References

_No response_

jonjohnsonjr:
The existing Filesystem statemgr implementation claims to use in-memory transient storage and the local filesystem for persistent storage.

> // Filesystem is a full state manager that uses a file in the local filesystem
> // for persistent storage.
> //
> // The transient storage for Filesystem is always in-memory.

However, the actual WriteState impl admits to being "incorrect":

> // WriteState is an incorrect implementation of Writer that actually also
> // persists.

It appears to have been this way since 2018. Perhaps this incorrectness is load-bearing, but given that this is only one of many implementations of the state manager interface, I suspect it would be fine to fix the implementation to be "correct".

This change splits off the expensive parts of WriteState into the PersistState method, which was previously a no-op.

Without this change, the time we spend maintaining is roughly quadratic with the number of resources being managed because each resouce causes:

1. The statefile to grow by some small amount.
2. The entire statefile to be serialized 3x.

For a small number of resources, serializing the statefile to JSON and writing it to disk is neglibible compared to the actual work being done by providers.

Once you start dealing with thousands of resources and megabytes of state, the overall time spent is dominated by this WriteState method. This is made worse by the fact that access to the statefile is managed by a mutex, so throwing more parallelism at the problem solves nothing.

Resolves #578 

## Target Release

1.6.0

### ENHANCEMENTS

- Managing large `terraform.tfstate` files is now much faster.

### DIFF
--- a/internal/states/statemgr/filesystem.go
+++ b/internal/states/statemgr/filesystem.go
@@ -120,13 +120,8 @@ func (s *Filesystem) State() *states.State {
 	return s.file.DeepCopy().State
 }
 
-// WriteState is an incorrect implementation of Writer that actually also
-// persists.
+// WriteState is an implementation of Writer.
 func (s *Filesystem) WriteState(state *states.State) error {
-	// TODO: this should use a more robust method of writing state, by first
-	// writing to a temp file on the same filesystem, and renaming the file over
-	// the original.
-
 	defer s.mutex()()
 
 	if s.readFile == nil {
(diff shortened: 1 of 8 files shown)


internal/states/statemgr/filesystem.go
```diff
@@ -140,13 +135,45 @@ func (s *Filesystem) WriteState(state *states.State) error {
...
+
+// PersistState writes state to a tfstate file.
+func (s *Filesystem) PersistState(schemas *tofu.Schemas) error {
+	defer s.mutex()()
+
+	return s.persistState(schemas)
+}
+
+func (s *Filesystem) persistState(schemas *tofu.Schemas) error {
+	// TODO: this should use a more robust method of writing state, by first
```
marcinwyszynski, line 162: Is there a reason not to do this as part of this change?

marcinwyszynski: Oh, sorry, I missed the fact that this is simply copied from the previous implementation. Nevermind!

marcinwyszynski:
Did you have a chance to profile this on the same input that your colleague used in #397 ?

Also, if we're no longer saving state snapshots for each resource, does it not mean that a crash would have more impact and leave resources orphaned that would have otherwise been persisted?

jonjohnsonjr:
> Did you have a chance to profile this on the same input that your colleague used in https://github.com/opentofu/opentofu/pull/397?

Yep! That run was a small subset of the overall workload I'm targeting. I tried this patch against the entire workload and brought the runtime down from 45 minutes to ~10 minutes (which is about how long I'd expect it to take).

jonjohnsonjr:
Modified the target release to 1.7.0 as per https://github.com/opentofu/opentofu/issues/578#issuecomment-1735104366

cube2222:
@jonjohnsonjr Hey! Since we've now split off a 1.6 branch and main is targeting 1.7, we're good to review and merge this. Could you please add a changelog entry and rebase?

cam72cam (approved):

Yantrio (approved):

cam72cam:
I'll add the specified changelog entry from the description in a subsequent PR.

PR #600, opened 2023-09-27
Snapcraft login shouldn't be called when credentials are set via env var.

### LINKED ISSUE #591: Release action doesn't work - fail for `craft-store error: Credentials could not be parsed.`
Action link - https://github.com/opentofu/opentofu/actions/runs/6313959167/job/17143181189

Https://github.com/opentofu/opentofu/pull/418 added a step to the release process
With this note
> Note: This should not be merged without a token being added that logs in correctly or the CI for the project will start failing.

I guess we didn't add that token to our secrets, we need to fix it

cube2222:
Fixes #591 

It seems like snapcraft complains about `snapcraft login` if `SNAPCRAFT_STORE_CREDENTIALS` is set.

Thus, I've removed snapcraft login and just moved the env var to be available to the whole goreleaser invocation.

Example workflow: https://github.com/opentofu/opentofu/actions/runs/6313959167/job/17173820994

cc @jnsgruk

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,7 +53,6 @@ jobs:
       - name: Setup snapcraft
         run: |
           sudo snap install snapcraft --classic --channel=7.x/stable
-          snapcraft login
 
           # See https://github.com/goreleaser/goreleaser/issues/1715
           mkdir -p "$HOME/.cache/snapcraft/download"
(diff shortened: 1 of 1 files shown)


eranelbaz (approved):
QA ✅ 
https://github.com/opentofu/opentofu/actions/runs/6324501103/job/17174136246

jnsgruk:
Nice, sorry about that @cube2222!

Issue #584, opened 2023-09-26
Add code formatting guidelines, and the "Code Consistency Check" step to CONTRIBUTING.md


kislerdm:
### OpenTofu Version

```shell
-
```

## Problem

Different contributors have different IDE settings, which may lead to the source code re-formatting violating the OpenTofu guidelines. **Note** that there are no explicit linting/formatting guidelines beyond running `gofmt` and `goimports` yet.

_The consequence_: avoidable increase of the "time-to-merge" causing delivery delays.

For example, my IDE has re-formatted a couple of files causing CI check to [fail](https://github.com/opentofu/opentofu/pull/576#issuecomment-1735127229).

### Attempted Solutions

Run the [CI command](https://github.com/opentofu/opentofu/blob/daf275ffbe213d06837be8e50c8081e54b266d2c/.github/workflows/checks.yml#L185C11-L185C78) in local env:

```bash
make fmtcheck importscheck generate staticcheck exhaustive protobuf
```

### Proposal

1. Describe the formatting guidelines explicitly as part of the contribution instructions [CONTRIBUTING.md](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md)
2. Include corresponding step to the [instructions](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md#contributing-a-code-change)

In addition, we may consider use of pre-commit, or pre-push git hooks as a contribution prerequisite.

## Open questions

- Shall we opt in for using [golang-lint](https://golangci-lint.run/) to facilitate maintenance of the codebase formatting?

serdardalgic:
Thanks for this suggestion @kislerdm, I agree adding a pre-commit hook would save time and avoid confusion. 

Adding a `.pre-commit-config.yaml` like this to the root would work for now if I'm not mistaken. 

```shell
repos:
  - repo: local
    hooks:
      - id: code-consistency-checks
        name: code-consistency-checks
        description: This pre-commit hook runs the consistency checks
        entry: make fmtcheck importscheck generate staticcheck exhaustive protobuf
        language: system
        pass_filenames: false
```
[...]

PR #513, opened 2023-09-20
all: sort import order

zchee:
Sort import order according to Go language conventions。

PR #624, opened 2023-10-02
[Issue 584] Add linter: golangci-lint

Hefeweizen:
PR 513 seeks to correct import ordering.
This PR is meant to supplement that work, seeing that imports stay organized thereafter.

# Starting with the bare minimum for this specific request -- keep imports properly organized:
```
$ golangci-lint help linters | grep -E '^import:'
import: depguard, gci, goimports, gomodguard
$ golangci-lint help linters | grep -E '^(depguard|gci|goimports|gomodguard):'
depguard: Go linter that checks if package imports are in a list of acceptable packages [fast: true, auto-fix: false]
gci: Gci controls Go package import order and makes it always deterministic. [fast: true, auto-fix: false]
goimports: Check import statements are formatted according to the 'goimport' command. Reformat imports in autofix mode. [fast: true, auto-fix: true]
gomodguard: Allow and block list linter for direct Go module dependencies. This is different from depguard where there are different block types for example version constraints and module recommendations. [fast: true, auto-fix: false]
```

Relates to https://github.com/opentofu/opentofu/issues/584
Relates to https://github.com/opentofu/opentofu/pull/513

## Target Release

This is project organizational work, it should have no impact on program functionality.

1.6.0

## Draft CHANGELOG entry

n/a

no user facing changes
-

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,55 @@
+name: golangci-lint
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.21'
+          cache: false
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Require: The version of golangci-lint to use.
+          # When `install-mode` is `binary` (default) the value can be v1.2 or v1.2.3 or `latest` to use the latest version.
+          # When `install-mode` is `goinstall` the value can be v1.2.3, `latest`, or the hash of a commit.
+          version: v1.54
+
+          # Optional: working directory, useful for monorepos
+          # working-directory: somedir
+
+          # Optional: golangci-lint command line arguments.
+          #
+          # Note: By default, the `.golangci.yml` file should be at the root of the repository.
+          # The location of the configuration file can be changed by using `--config=`
+          # args: --timeout=30m --config=/my/path/.golangci.yml --issues-exit-code=0
+          args:  -p import --timeout=10m --allow-parallel-runners
+
+          # Optional: show only new issues if it's a pull request. The default value is `false`.
+          only-new-issues: true
+
+          # Optional: if set to true, then all caching functionality will be completely disabled,
+          #           takes precedence over all other caching options.
+          # skip-cache: true
+
+          # Optional: if set to true, then the action won't cache or restore ~/go/pkg.
+          # skip-pkg-cache: true
+
+          # Optional: if set to true, then the action won't cache or restore ~/.cache/go-build.
+          # skip-build-cache: true
+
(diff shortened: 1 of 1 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,55 @@
...
+
+          # Optional: golangci-lint command line arguments.
+          #
+          # Note: By default, the `.golangci.yml` file should be at the root of the repository.
+          # The location of the configuration file can be changed by using `--config=`
+          # args: --timeout=30m --config=/my/path/.golangci.yml --issues-exit-code=0
+          args:  -p import --timeout=10m --allow-parallel-runners
+
+          # Optional: show only new issues if it's a pull request. The default value is `false`.
+          only-new-issues: true
```
eranelbaz, line 42: Are we going to fix the current issues?
how does it know if there are new issues?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,55 @@
+name: golangci-lint
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
```
eranelbaz, line 9: where do you use this token?

eranelbaz (commented):
Hey @Hefeweizen thanks for this contribution
Do you plan to also fix current issues or do we wait for #513?

.github/workflows/golangci-lint.yml
Hefeweizen: It's advised in the default configuration of the github action, but you're right, I don't see why it would need it.  Reviewing the code, I don't see where it's used (though I'm no expert on js).

I'll experiment on my local copy and update accordingly.

.github/workflows/golangci-lint.yml
Hefeweizen: > Are we going to fix the current issues?
I was going to rely on #513 to fix the bulk of the current issues.
I would not make future PR authors responsible for the entire state of the repo; I would only make them responsible for the code they've changed.  Yes, this might leave lingering errors, but to do otherwise creates an impediment for those wanting to contribute.

1-off issues should be created to have this work done for the entire repository.

> how does it know if there are new issues?
I didn't update the needed permissions, so it can't.  : )  I'll fix that.
When it's working, it grabs the parent PR to limit its scope to the changes within that PR.
https://github.com/golangci/golangci-lint-action/blob/e09eda33b5d643674c0219d04a7b18e4ef933127/src/run.ts#L25

& thanks for using the fine-toothed comb; great questions!

Hefeweizen:
> Do you plan to also fix current issues or do we wait for #513?

This PR was made to supplement #513; no sense in duplicating the already-done work.

eranelbaz:
Cool @Hefeweizen 
Thank you very much for this PR
LMK if you need anything from me
Waiting for your local copy

serdardalgic (commented):
One small suggestion, but apart from that, it looks good to me. 

I would suggest we don't only stick to the `import` set of linters but also benefit from many other linters that golangci-lint provide in the future.

sirvantedbrl:
@eranelbaz Is there any active work this PR?

Yantrio:
@eranelbaz is busy lately so I'll take over this PR :D 

In my opinion, I would like to get this merged once we have our first stable release. This ensures that we can move quickly right now.

Once the stable release is out we can look at introducing linting and code style guidelines in a strongly enforceable way without worrying about accidental bugs being introduced.

sirvantedbrl:
@Yantrio 👍🏻🙌🏻

PR #640, opened 2023-10-03
move `build-opentofu-oss.yml` to different folder

### LINKED ISSUE #637: Move `build-opentofu-oss` from `.github/workflows` to `.github/scripts`
As we never run `build-opentofu-oss` alone, and it have no triggers.
It only help to other build workflows. It should be in `scripts` folder

eranelbaz:
Resolves #637
This isn't really a workflow but a helper to split one, we don't trigger it directly.
moved to `common` folder

## Target Release

1.6.0

-


eranelbaz:
Closing as it seems like you can reference only workflows under the `workflow` folder and not in sub folders

PR #652, opened 2023-10-04
fix: add an alias to the snap package

jnsgruk:
The current released snap requires OpenTofu to be invoked with `opentofu.tofu`. There is an alias request on the store which is now live which means people will just be able to invoke `tofu` in the future.

https://forum.snapcraft.io/t/classic-confinement-request-for-opentf/36706/12

Resolves #

## Target Release

1.6.0

### BUG FIXES
- Enable the `tofu` alias for the OpenTofu snap package.


cube2222:
Ah, thanks @jnsgruk!

I tested it out myself, and was just about to post an issue, as I've hit this problem.

jnsgruk:
Yep, pretty sure this will fix it :)

FWIW you can test the current revision by doing `snap alias opentofu.tofu tofu` - then the `tofu` command will work normally :)

cube2222 (approved):

PR #658, opened 2023-10-05
cmd: Demote Exported Functions and Variables

### LINKED ISSUE #657: cmd Package Export Demotions
### OpenTofu Version

```shell
.
```

### Use Cases

.

### Attempted Solutions

.

### Proposal

There are variables and functions in the `cmd` package that are being exported, but used nowhere else. Demote them.

### References

_No response_

alrs:
This demotes the exported functions `ExperimentsAllowed()` and `WorkingDir()`. This also demotes several exported variables relating to commands.

Fixes #657

### DIFF
--- a/cmd/tofu/experiments.go
+++ b/cmd/tofu/experiments.go
@@ -22,6 +22,6 @@ package main
 // open experiment.)
 var experimentsAllowed string
 
-func ExperimentsAllowed() bool {
+func experimentsAreAllowed() bool {
 	return experimentsAllowed != ""
 }
--- a/cmd/tofu/commands.go
+++ b/cmd/tofu/commands.go
@@ -82,7 +82,7 @@ func initCommands(
 		configDir = "" // No config dir available (e.g. looking up a home directory failed)
 	}
 
-	wd := WorkingDir(originalWorkingDir, os.Getenv("TF_DATA_DIR"))
+	wd := workingDir(originalWorkingDir, os.Getenv("TF_DATA_DIR"))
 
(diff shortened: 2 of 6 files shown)


cmd/tofu/experiments.go
```diff
@@ -22,6 +22,6 @@ package main
 // open experiment.)
 var experimentsAllowed string
 
-func ExperimentsAllowed() bool {
```
Hefeweizen, line 25: In having the function name match the variable name, we have better insight into the code when searching.  With the shortened name, using automation to find 'expAllowed' will miss instances of 'experimentsAllowed'.

It's a small detail, but I think it keeps the code base more understandable.

Consider retaining the longer function name.

alrs: I intentionally made it different to reduce confusion. Neither is officially exported, though the variable is getting touched by compile-time `ldflags` games, so it made sense to me to keep them distinct.

cube2222: I don't think we can easily keep the same name, cause Go doesn't allow that.

But maybe experimentsAreAllowed or something like that for the function? Maybe experimentalFeaturesAllowed?

Having it exported to avoid the collision makes sense to me too.

I agree that expAllowed isn't obvious as to what it means.

alrs: @cube2222 I went with `experimentsAreAllowed()`.

cube2222 (approved):
Thanks for the contribution!

Issue #655, opened 2023-10-04
Tofu crashes when variable is marked as sensitive and used in locals block assignment


loafoe:
### OpenTofu Version

```shell
OpenTofu v1.6.0-alpha1
on darwin_arm64
+ provider registry.opentofu.org/cloudfoundry-community/cloudfoundry v0.51.3
+ provider registry.opentofu.org/grafana/grafana v2.3.3
+ provider registry.opentofu.org/hashicorp/archive v2.4.0
+ provider registry.opentofu.org/hashicorp/local v2.4.0
+ provider registry.opentofu.org/hashicorp/random v3.5.1
+ provider registry.opentofu.org/loafoe/htpasswd v1.0.4
+ provider registry.opentofu.org/philips-software/hsdp v0.44.4
+ provider registry.terraform.io/cloudfoundry-community/cloudfoundry v0.51.3
+ provider registry.terraform.io/hashicorp/random v3.5.1
+ provider registry.terraform.io/loafoe/htpasswd v1.0.4
+ provider registry.terraform.io/philips-software/hsdp v0.44.0
```

### OpenTofu Configuration Files

```hcl
variable "oauth2_client_password" {
  type        = string
  default     = ""
  sensitive = true # Setting this to false prevents crash
}

locals {
  self_managed_client    = var.iam_application_id != ""
  oauth2_client_password = local.self_managed_client ? random_password.client_password[0].result : var.oauth2_client_password
}
```

### Debug Output

```
!!!!!!!!!!!!!!!!!!!!!!!!!!! OPENTOFU CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

value is marked, so must be unmarked first
goroutine 1814 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:24 +0x64
runtime/debug.PrintStack()
	runtime/debug/stack.go:16 +0x1c
github.com/opentofu/opentofu/internal/logging.PanicHandler()
	github.com/opentofu/opentofu/internal/logging/panic.go:58 +0x170
panic({0x1028bd520, 0x102dbcd50})
	runtime/panic.go:890 +0x258
github.com/zclconf/go-cty/cty.Value.assertUnmarked(...)
	github.com/zclconf/go-cty@v1.13.2/cty/marks.go:141
github.com/zclconf/go-cty/cty.Value.AsString({{{0x102de7ea0?, 0x1400011ed39?}}, {0x102abfee0?, 0x14001ac5938?}})
[...]

PR #661, opened 2023-10-05
add e2e test to cover the panic case of unknown conditional with sensitive

eranelbaz:
part of #655

## Target Release

1.6.0

-


Yantrio:
Would you also be able to include a test for other types please? (collections, maps, numbers etc)

Yantrio:
Also, whilst we're looking in this area, we should also cover destroy and apply test cases too. not just planning.

cube2222:
Maybe we can keep have one test for each of the types and also one for apply? To not create too many of them (something would've need to be very messed up for it to work in apply but fail in destroy)

eranelbaz:
Under the same `tf` file i'm adding the rest of the types
And I'll run it also for `apply`

eranelbaz:
@Yantrio @cube2222 9337e8b7

Yantrio (approved):

eranelbaz:
Renamed to folder - removed the `destroy` 
d0aa8172

cube2222 (approved):

PR #672, opened 2023-10-06
chore: moving from opentf to opentofu in CODE_OF_CONDUCT and Dockerfile

Cadienvan:
Found two places where OpenTF was still being used

## Target Release

1.6.0

### BUG FIXES

- Now both CODE_OF_CONDUCT and Dockerfile state OpenTofu instead of OpenTF


Cadienvan:
@Yantrio signed off this other PR also :D

cube2222 (approved):
Thanks for the contribution @Cadienvan!

PR #674, opened 2023-10-06
Fix wording in documentation

wrp:
As a native English speaker, I find the expression "recommend to use" very awkward.  I propose we eradicate it from the documentation.


eranelbaz:
Hey @wrp 
Thanks for this wording fix,
Can you please rename the pull request title hence the commit message for some like "Fix wording in documentation"
Thank you very much

eranelbaz (approved):
thanks for the fix!

PR #696, opened 2023-10-09
Consider missing `x-terraform-snapshot-interval` for State Snapshot

### LINKED ISSUE #687: Cloud: State snapshots are always enabled, despite missing `x-terraform-snapshot-interval` header
## Summary:

The `cloud` APIs for creating/listing versions could return a `x-terraform-snapshot-interval` header, which tells `tofu` how often is should save intermediate snapshots
If the `x-terraform-snapshot-interval` header is missing, then `tofu` should not save intermediate snapshots. However, currently that is not the case, and if `x-terraform-snapshot-interval` is missing - `tofu` still saves intermediate snapshots

eranelbaz:
Resolves #687 
Ported MPL code from https://github.com/hashicorp/terraform/pull/33834 - Seems like we ported some of it already
## Target Release

1.6.0

-

### DIFF
--- a/internal/cloud/state.go
+++ b/internal/cloud/state.go
@@ -34,6 +34,11 @@ import (
 	"github.com/opentofu/opentofu/internal/tofu"
 )
 
+const (
+	// HeaderSnapshotInterval is the header key that controls the snapshot interval
+	HeaderSnapshotInterval = "x-terraform-snapshot-interval"
+)
+
 // State implements the State interfaces in the state package to handle
 // reading and writing the remote state to TFC. This State on its own does no
 // local caching so every persist will go to the remote storage and local
@@ -248,6 +253,8 @@ func (s *State) ShouldPersistIntermediateState(info *local.IntermediateStatePers
 		return true
 	}
 
+	// This value is controlled by a x-terraform-snapshot-interval header intercepted during
+	// state-versions API responses
 	if !s.enableIntermediateSnapshots && info.RequestedPersistInterval == time.Duration(0) {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


internal/cloud/testing.go
```diff
@@ -437,7 +437,7 @@ func testServerWithSnapshotsEnabled(t *testing.T, enabled bool) *httptest.Server
 			t.Fatal(err)
 		}
 
-		w.Header().Set("content-type", "application/json")
+		w.Header().Set("content-type", tfe.ContentTypeJSONAPI)
```
eranelbaz, line 440: Without that tests are failing for unknown content type - 
should be the same as https://github.com/opentofu/opentofu/blob/87de080aa206f9f6e8cad92de3fd68aa7c141460/internal/cloud/state.go#L588

internal/cloud/state.go
```diff
@@ -248,6 +253,8 @@ func (s *State) ShouldPersistIntermediateState(info *local.IntermediateStatePers
 		return true
 	}
 
+	// This value is controlled by a x-terraform-snapshot-interval header intercepted during
+	// state-versions API responses
 	if !s.enableIntermediateSnapshots && info.RequestedPersistInterval == time.Duration(0) {
```
RLRabinowitz, line 258: You should remove `&& info.RequestedPersistInterval == time.Duration(0)`, otherwise applies will always fail this condition

internal/cloud/state.go
```diff
@@ -577,7 +584,13 @@ func clamp(val, min, max int64) int64 {
 }
 
 func (s *State) readSnapshotIntervalHeader(status int, header http.Header) {
-	intervalStr := header.Get("x-terraform-snapshot-interval")
+	contentType := header.Get("Content-Type")
```
RLRabinowitz, line 587: Consider adding the comment `// Only proceed if this came from tfe.v2 API` above this

RLRabinowitz (changes requested):

internal/cloud/state.go
eranelbaz: 57a2345b(cr comment)

Yantrio (approved):

Yantrio (approved):

Issue #688, opened 2023-10-08
Provider Schema Cache is never used


RLRabinowitz:
### Summary

The provider `SchemaCache` is never used, and so is never actually helping `tofu` run memory and execution time. 
If you take a look at [this code](https://github.com/opentofu/opentofu/blob/3860a486b122e7f5cce63f23ab0b81c53c50954f/internal/plugin/grpc_provider.go#L82-L86), you can see that the condition in the `if` will never be `true`, because the named return value `resp` has never been set

When fixing, we should actually run tests and verify that memory does not increase significantly when we are using a provider multiple times with a different alias

PR #719, opened 2023-10-12
Fix: Global provider schema cache is never used

RLRabinowitz:
Fixes https://github.com/opentofu/opentofu/issues/688

The global provider cache was never used, as the cache was queried only if `ServerCapabilities.GetProviderSchemaOptional` was true, even though the `GetSchema` request didn't yet happen, so this data was always `false`

I changed to logic to only **put** the schema in the cache if the `GetProviderSchemaOptional` capability is on, and always fetch the schema if it's in the cache

Also added unit tests for the bug (had to deal with the fact that currently the SchemaCache leaks between tests, and "luckily" it didn't yet affect the results of the unit tests)

## Target Release

1.6.0

-

### DIFF
--- a/internal/plugin/grpc_provider_test.go
+++ b/internal/plugin/grpc_provider_test.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/opentofu/opentofu/internal/addrs"
+
 	"github.com/golang/mock/gomock"
 	"github.com/google/go-cmp/cmp"
 	"github.com/opentofu/opentofu/internal/configs/hcl2shim"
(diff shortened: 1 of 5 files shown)


internal/plugin/grpc_provider_test.go
```diff
@@ -125,6 +126,100 @@ func TestGRPCProvider_GetSchema_GRPCError(t *testing.T) {
 	checkDiagsHasError(t, resp.Diagnostics)
 }
 
+func TestGRPCProvider_GetSchema_GlobalCacheEnabled(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	client := mockproto.NewMockProviderClient(ctrl)
+	// The SchemaCache is global and is saved between test runs
```
RLRabinowitz, line 132: The `SchemaCache` is being leaked between tests, as it is global

Yantrio (approved):

PR #720, opened 2023-10-13
Add compare versions script

### LINKED ISSUE #663: Block release if `version/VERSION` content isn't the same / bumped since last release
When we release a new version of OpenTofu, we should also bump the content of `version/VERSION` to contain the correct version.
The correct version is based on the tag name without the `v`
for example 

|  tag name | file content  | 
|---|---|
| v1.6.0-alpha1  | 1.6.0-alpha1  |  
| v1.6.0-alpha2  | 1.6.0-alpha2  |  

Add a check to our release process to make sure we indeed bumped the file content.

Note that this check shouldn't run if the branch name doesn't start with `v` if we would like to check the release process (also might be the same rule as the [go release snapshot rule](https://github.com/opentofu/opentofu/blob/86866bff0c7f2b8997fb95a23e3b4827219efc10/.github/workflows/release.yml#L67))

Jake-Mok-Nelson:
## Description

I'm adding a check to the release deployment workflow that will compare the triggering tag against the contents of the version/VERSION file.

Read the tag being passed in by GITHUB_REF context and compare it the version file. People rely on these versions being identical and since the update of the file is manual it could be missed.

I've tested this process in another repo. Keep in mind that if you modify default behaviour like continue-on-error in this repo, this may behave differently.

## Test Cases

- [x] version/VERSION file is missing

```bash
./.github/scripts/compare-release-version.sh
  shell: /usr/bin/bash -e {0}
The VERSION file could not be found. Please create a VERSION file in the version/ directory. The contents of version should match the tag without the v prefix.
Error: Process completed with exit code 1.
```

- [x] Contents mismatch -  Github trigger tag minus v prefix does not match the contents of version/VERSION

```bash
Run ./.github/scripts/compare-release-version.sh
The VERSION file does not match the tag. Please update the version/VERSION file to match the tag without the v prefix.
The VERSION file contains: 1.6.0-alpha2 but the tag is: 0.1.12.
Error: Process completed with exit code 1.
```

- [x] Release won't run without the compare step passing first

image

- [x] Matching version passes

```bash
Run ./.github/scripts/compare-release-version.sh
The VERSION file matches the tag. Proceeding with the release of 1.6.0-alpha2.
```

Resolves #663 

## Target Release

N/A I believe this can go in whenever.

### ENHANCEMENTS

 

-  The release workflow will check that the tag version and the version in version/VERSION are consistent.

### DIFF
--- a/.github/scripts/compare-release-versions.sh
+++ b/.github/scripts/compare-release-versions.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Compare the semver tag against the current release in the VERSION file
+
+set -uo pipefail
+
+# Bail if VERSION cannot be found
+if [[ ! -f version/VERSION ]]; then
+    echo "The VERSION file could not be found. Please create a VERSION file in the version/ directory. The contents of version should match the tag without the v prefix."
+    exit 1
+fi
+
+# Bail if the input was not a tag
+if [[ ! "$GITHUB_REF_TYPE" == "tag" ]]; then
+    echo "This action only runs on tags. Please create a tag and try again."
+    exit 1
+fi
+
+# Create a clean semver tag without the v prefix
+CLEAN_TAG=$(echo "$GITHUB_REF_NAME" | sed 's/^v//')
+
+VERSION=$(cat version/VERSION)
+if [[ "$VERSION" != "$CLEAN_TAG" ]]; then
+    echo "The VERSION file does not match the tag. Please update the version/VERSION file to match the tag without the v prefix."
+    echo "The VERSION file contains: $VERSION but the tag is: $CLEAN_TAG."
+    exit 1
+fi
+
+echo "The VERSION file matches the tag. Proceeding with the release of $VERSION."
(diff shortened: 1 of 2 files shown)


Yantrio (approved):

Yantrio:
Hi Again @Jake-Mok-Nelson! Thanks for this contirbution 🙌  However, I realized post-merging that this file was missing a license at the top.

Normally we ask people to add `# SPDX-License-Identifier: MPL-2.0` under the shebang. Would you mind putting in another quick PR to add this? 🙏 

Sorry for this!

Jake-Mok-Nelson:
> Hi Again @Jake-Mok-Nelson! Thanks for this contirbution 🙌  However, I realized post-merging that this file was missing a license at the top.
> 
> Normally we ask people to add `# SPDX-License-Identifier: MPL-2.0` under the shebang. Would you mind putting in another quick PR to add this? 🙏 
> 
> Sorry for this! 

I'll get onto this shortly. I saw the licence in the other files but it always accompanied a hashicorp copyright and I forgot to ask. I'll fix it up 👍

PR #726, opened 2023-10-13
internal/registry: Prune and Demote Exported Functions

### LINKED ISSUE #725: Demote and Prune Functions in internal/registry
### OpenTofu Version

```shell
.
```

### Use Cases

.

### Attempted Solutions

.

### Proposal

There are functions within and under `internal/registry` that are exported and either unused or only used within their local package. Remove or demote them.

### References

.

alrs:
There are numerous exported functions in `internal/registry` that are either unused or only called locally. This PR removes unused functions (both exported and local) and demotes functions, and it demotes those needlessly exported.

Fixes #725 

There are no user-facing changes that warrant a CHANGELOG entry.

### DIFF
--- a/internal/registry/regsrc/friendly_host_test.go
+++ b/internal/registry/regsrc/friendly_host_test.go
@@ -78,7 +78,7 @@ func TestFriendlyHost(t *testing.T) {
 		// Matrix each test with prefix and total match variants
 		for _, sfx := range []string{"", "/", "/foo/bar/baz"} {
 			t.Run(tt.name+" suffix:"+sfx, func(t *testing.T) {
-				gotHost, gotRest := ParseFriendlyHost(tt.source + sfx)
+				gotHost, gotRest := parseFriendlyHost(tt.source + sfx)
 
 				if gotHost == nil {
 					if tt.wantHost != "" {
@@ -121,7 +121,7 @@ func TestFriendlyHost(t *testing.T) {
 }
 
 func TestInvalidHostEquals(t *testing.T) {
-	invalid := NewFriendlyHost("NOT_A_HOST_NAME")
+	invalid := newFriendlyHost("NOT_A_HOST_NAME")
 	valid := PublicRegistryHost
 
 	// invalid hosts are not comparable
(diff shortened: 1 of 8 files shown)


internal/registry/regsrc/friendly_host_test.go
```diff
@@ -133,8 +133,8 @@ func TestInvalidHostEquals(t *testing.T) {
 		t.Fatalf("%q is not equal to %q", valid, invalid)
 	}
 
-	puny := NewFriendlyHost("xn--s-fka0wmm0zea7g8b.xn--o-8ta85a3b1dwcda1k.io")
-	display := NewFriendlyHost("ʎɹʇsıƃǝɹ.ɯɹoɟɐɹɹǝʇ.io")
+	puny := newFriendlyHost("xn--s-fka0wmm0zea7g8b.xn--o-8ta85a3b1dwcda1k.io")
+	display := newFriendlyHost("ʎɹʇsıƃǝɹ.ɯɹoɟɐɹɹǝʇ.io")
```
maunzCache, line 137: Just because this is not found by regular search & replace. You could replace `ɯɹoɟɐɹɹǝʇ` with `nɟoʇuǝdo` which is upsidedown `terraform`/`opentofu`

serdardalgic: Woah! What a weird string there… 🤯

alrs: I think this is worthy of a second issue and PR, which I'm happy to take on once this one is merged or rejected.

For the record, `https://ʎɹʇsıƃǝɹ.nɟoʇuǝdo.io` translates to `xn--s-fka0wmm0zea7g8b.xn--noudo-6yb12ggj.io`.

cube2222: I love this suggestion (feel free to do it in a separate PR of course)

cube2222 (commented):
Thanks for the PR.

Honestly, I have mixed feelings on these changes.

Getting rid of `ModuleFromModuleSourceAddr` makes sense to me, that part's an unused legacy shim function, so no reason to keep it.

But the others seem to intentionally be library functions that might or might not be useful outside of this package, even if they're not currently used.

Take ParseFriendlyHost for instance - the struct is exported, and `NewFriendlyHost` and `ParseFriendlyHost` are its constructors. So it's only logical for them to be exported as well.

Same for NewPaginationMeta.

While Display() seems to be a reasonable helper function to have when operating on this struct (though weird that it doesn't satisfy an interface).

I'd be leaning toward keeping those parts.

Yantrio:
> Getting rid of ModuleFromModuleSourceAddr makes sense to me, that part's an unused legacy shim function, so no reason to keep it.

I agree completely with @cube2222 on this one here

As for the demotion of the other methods: I am on the fence (50/50). One one hand if we ever chose to expose this package externally, or if we want to use those functions we can promote them again.

I think this depends on how you look at this project. Is it one tool that's only consumed by it's self, or is this a collection of self-contained packages that combine into the project.

If you're to take the approach of self-contained packages then I think that demoting these methods is not needed. But if you were to take the contrary approach that this is one tool, then I think demotion works here.

So in this case, I don't think there's a direct right/wrong answer, but In my personal opinion, doing work for work's sake here by demoting these is not really needed. I feel like it's fine to be cleaning up, but demoting functions feels unnecessary.

I'm sorry I don't have a more concrete opinion on this one, but I would love to hear other people's PoV on this.

cam72cam:
Closing per discussion in https://github.com/opentofu/opentofu/issues/725

PR #732, opened 2023-10-16
Fix improper certificate validation vulnerability in configureTLS()

DharunKumar04:
This pull request fixes a security vulnerability in the `configureTLS()` function, where the client can be configured to ignore TLS verification. This can make the client vulnerable to man-in-the-middle attacks.

To fix this vulnerability, the pull request validates the **server's certificate** against a list of trusted certificates. This ensures that the client is only communicating with the intended server.

### DIFF
--- a/internal/backend/remote-state/http/backend.go
+++ b/internal/backend/remote-state/http/backend.go
@@ -132,46 +132,42 @@ type Backend struct {
 
 // configureTLS configures TLS when needed; if there are no conditions requiring TLS, no change is made.
 func (b *Backend) configureTLS(client *retryablehttp.Client, data *schema.ResourceData) error {
-	// If there are no conditions needing to configure TLS, leave the client untouched
-	skipCertVerification := data.Get("skip_cert_verification").(bool)
-	clientCACertificatePem := data.Get("client_ca_certificate_pem").(string)
-	clientCertificatePem := data.Get("client_certificate_pem").(string)
-	clientPrivateKeyPem := data.Get("client_private_key_pem").(string)
-	if !skipCertVerification && clientCACertificatePem == "" && clientCertificatePem == "" && clientPrivateKeyPem == "" {
-		return nil
-	}
-	if clientCertificatePem != "" && clientPrivateKeyPem == "" {
-		return fmt.Errorf("client_certificate_pem is set but client_private_key_pem is not")
-	}
-	if clientPrivateKeyPem != "" && clientCertificatePem == "" {
-		return fmt.Errorf("client_private_key_pem is set but client_certificate_pem is not")
-	}
+    // If there are no conditions needing to configure TLS, leave the client untouched
+    skipCertVerification := data.Get("skip_cert_verification").(bool)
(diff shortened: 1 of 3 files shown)


internal/backend/remote-state/http/backend.go
```diff
@@ -132,46 +132,42 @@ type Backend struct {
...
-	// TLS configuration is needed; create an object and configure it
-	var tlsConfig tls.Config
-	client.HTTPClient.Transport.(*http.Transport).TLSClientConfig = &tlsConfig
+    // TLS configuration is needed; create an object and configure it
+    var tlsConfig tls.Config
+    client.HTTPClient.Transport.(*http.Transport).TLSClientConfig = &tlsConfig
 
-	if skipCertVerification {
-		// ignores TLS verification
-		tlsConfig.InsecureSkipVerify = true
```
serdardalgic, line 156: If I understand correct, this `tlsConfig.InsecureSkipVerify = true` assignment is removed in this PR. Do we want to completely remove this functionality where the practitioner might prefer to skip the TLS verification for various purposes (testing, development etc.)? 

Also, if we decide to remove this functionality to be able to skip the verification, it needs to be added to the Changelog, If I'm not mistaken.

serdardalgic (commented):
Hi @DharunKumar04! Thanks for your PR! 
I'm a bit confused about the removal of the skipCertVerification, that's why a question here. I'm not familiar with this part of the code, but I feel like this change is only skipping the ability to skip the certificate verification, which might be useful in cases like testing and development. Can you describe how it's a vulnerability? 

Also, my personal opinion, It would be helpful to enhance the tests in backend_test.go that cover the fix. That would help to understand the issue and secure the code for the changes in the future.

DharunKumar04:
Hi @serdardalgic,

Thanks for the feedback!

I understand your concern about removing the `skipCertVerification` option. However, I believe that it is a necessary step to improve the security of the code.

Skipping TLS verification makes the client vulnerable to man-in-the-middle attacks. In a man-in-the-middle attack, an attacker intercepts the communication between the client and the server and impersonates one of them. If TLS verification is skipped, the attacker can impersonate the server without being detected.

I understand that there may be cases where it is necessary to skip TLS verification, such as for testing or development. However, I believe that it is better to have a separate mechanism for doing so, such as a command-line flag or environment variable. This would allow users to explicitly choose to skip TLS verification, rather than having it enabled by default.

I agree that it would be helpful to enhance the tests in `backend_test.go` to cover the fix. I will work on adding additional tests before merging this PR.

I will also add a changelog entry to document the removal of the `skipCertVerification` option.

Thanks again for the feedback! I appreciate it.

Yantrio:
I would've preferred if we could discuss this before a PR was created. Something similar like GPG validation [here](https://github.com/opentofu/opentofu/issues/673) could've been made to discuss our options here and understand the background a little more.

cube2222:
Closing due to the same reason as the issue (https://github.com/opentofu/opentofu/issues/738), feel free to reopen.

PR #734, opened 2023-10-16
build: setup go action

### LINKED ISSUE #290: ci: simplify go version handling
Currently the ci uses composite action to get the go version: https://github.com/opentffoundation/opentf/blob/main/.github/actions/go-version/action.yml

This is used in many places in CI and sometimes makes spawning a job and checking out the code just to get the version to be sent to other jobs: https://github.com/opentffoundation/opentf/blob/b24c89c4b2f5698b04469a645baa5ff451e9c85e/.github/workflows/build.yml#L58-L68

Setup-go v4 supports the `go-version-file` https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file which supports either `go.mod` or raw version file like we have in `.go-version`. Ref: https://github.com/actions/setup-go/blob/883490dfd06f396ebe0b738bc313a53cf9d851e5/src/installer.ts#L420-L432

I suggest we simplify the workflows and get rid of the composite action.

I'm happy to make those changes in a PR

zekth:
Resolves #290
I've still kept the `go-version` action because we need it within the `hashicorp/actions-go-build` action

## Target Release

N/A

## Draft CHANGELOG entry

- build: simplify go version setup

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -235,4 +235,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-go 1.21
+go 1.21.3
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -1,0 +1,10 @@
+
+name: 'Setup go toolchain'
+description: 'Sets up the go toolchain ato use for any go related actions'
+runs:
+  using: composite
+  steps:
+  - name: Install Go toolchain
+    uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
+    with:
+      go-version-file: 'go.mod'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -160,7 +160,6 @@ jobs:
       e2e-cache-path: ${{ steps.set-cache-values.outputs.e2e-cache-path }}
     needs:
       - get-product-version
-      - get-go-version
     strategy:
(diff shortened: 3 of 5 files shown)


go.mod
```diff
@@ -235,4 +235,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-go 1.21
+go 1.21.3
```
Jake-Mok-Nelson, line 238: Not a reviewer but just curious, was pinning the minor version here intentional?

zekth: to be in sync with `.go-version`

zekth: if it's a blocker i can put back the `go-version` action but use it in the `setup-go` composite

cube2222: No strong opinion here, I think this is fine.

serdardalgic: I would vote for keeping it at the minor version(`1.21`) instead of the patch version(`1.21.3`), unless there is a necessity to stick to that specific patch version. I think that makes it easier to maintain, my 2 cents. 

I believe that's more of a personal preference rather than a right or wrong question, so, it's completely fine to agree to disagree :)

cube2222: @serdardalgic Fair point, I think I agree. Let's use the minor version if it's possible @zekth

I'd also use that in .go-version, if we want them to stay in sync.

cube2222:
Closing due to lack of activity, feel free to reopen @zekth

PR #737, opened 2023-10-16
Add comparison to Terraform link

### LINKED ISSUE #736: Compare Terraform to OpenTofu in docs
### OpenTofu Version

```shell
n/a
```

### Use Cases

n/a

### Attempted Solutions

Go direct to https://opentofu.org/faq/#opentofu-terraform-differences

### Proposal

Update https://opentofu.org/docs/intro/vs/ to help readers compare OpenTofu to Terraform. Hyperlink to https://opentofu.org/faq/#opentofu-terraform-differences

### References

_No response_

sftim:
Update https://opentofu.org/docs/intro/vs/ to help readers compare OpenTofu to Terraform.

Resolves #736

### DIFF
--- a/website/docs/intro/vs/index.mdx
+++ b/website/docs/intro/vs/index.mdx
@@ -17,5 +17,6 @@ Learn how OpenTofu compares to:
 
 - [Chef, Puppet, etc.](/docs/intro/vs/chef-puppet)
 - [CloudFormation, Heat, etc.](/docs/intro/vs/cloudformation)
-- [Boto, Fog, etc.](/docs/intro/vs/boto)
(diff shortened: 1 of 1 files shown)


serdardalgic:
IMO, I'm not so sure about the practicality of this change for now, as the content of the link is only the following:
```
There will be no differences between Terraform (versions prior to 1.5.x) and OpenTofu. As new versions are released, this will change.
```

But in the feature, it might make sense to have a dedicated documentation page when there are actual differences between OpenTofu and its predecessor.

cube2222 (approved):

cube2222 (approved):
We should probably include a docs page soon-ish to at least address the license and governance differences.

Issue #747, opened 2023-10-18
Backend/s3: Add support for `assume_role` block


tomasmik:
## Summary

Terraform currently supports both configuration for the remote backend latter one being deprecated:

```
terraform {
  backend "s3" {
    bucket = "terraform-state-prod"
    key    = "network/terraform.tfstate"
    region = "us-east-1"
    assume_role {
      role_arn = "arn:aws:iam::PRODUCTION-ACCOUNT-ID:role/Terraform"
    }
  }
}
```

```
terraform {
  backend "s3" {
    bucket = "terraform-state-prod"
    key    = "network/terraform.tfstate"
    region = "us-east-1"
    role_arn = "arn:aws:iam::PRODUCTION-ACCOUNT-ID:role/Terraform"
  }
}
```

We only support the deprecated one and need to update our configuration to also support the `assume_role` block

PR #754, opened 2023-10-19
Backend/S3: Extract `assume_role` as a separate block

tomasmik:
This adds support for both assume_role configurations:
```hcl
terraform {
  backend "s3" {
    bucket = "opentofu-state-prod"
    key    = "network/opentofu.tfstate"
    region = "us-east-1"
    assume_role = {
      role_arn = "arn:aws:iam::ACCOUNT-ID:role/Opentofu"
    }
  }
}
```

```hcl
terraform {
  backend "s3" {
    bucket = "opentofu-state-prod"
    key    = "network/opentofu.tfstate"
    region = "us-east-1"
    role_arn = "arn:aws:iam::ACCOUNT-ID:role/Opentofu"
  }
}
```

Resolves https://github.com/opentofu/opentofu/issues/747

## Target Release

1.6.0

-

### DIFF
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -557,7 +557,7 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 	}
 
 	if value := obj.GetAttr("assume_role"); !value.IsNull() {
-		cfg.AssumeRole = configureAssumeRole(value)
+		cfg.AssumeRole = configureNestedAssumeRole(obj)
 	} else if value := obj.GetAttr("role_arn"); !value.IsNull() {
 		cfg.AssumeRole = configureAssumeRole(obj)
 	}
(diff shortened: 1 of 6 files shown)


internal/backend/remote-state/s3/backend.go
```diff
@@ -509,6 +612,40 @@ func getS3Config(obj cty.Value) func(options *s3.Options) {
...
 
+func configureNestedAssumeRole(obj cty.Value) *awsbase.AssumeRole {
+	assumeRole := awsbase.AssumeRole{}
+
+	obj = obj.GetAttr("assume_role")
+	if val, ok := stringAttrOk(obj, "role_arn"); ok {
+		assumeRole.RoleARN = val
+	}
+	if val, ok := stringAttrOk(obj, "duration"); ok {
+		dur, _ := time.ParseDuration(val)
```
cube2222, line 623: Shouldn't this error be handled?

cube2222: Oh, now I see it's handled in the validation function.

In that case, I'd panic here on error.

cube2222 (approved):
One nite

internal/backend/remote-state/s3/backend.go
tomasmik: We know that this is always possible to parse because we check it beforehand: https://github.com/opentofu/opentofu/pull/754/files#diff-c6f102657948029362a41861b9418f65f3830155d90e53b4a5141ddb2d60c2b4R80

cube2222: unless it is misused, which is why I propose the panic, just in case

PR #759, opened 2023-10-20
Backend/s3: Add support for `assume_role_with_web_identity` block

### LINKED ISSUE #689: Backend/s3: Add support `assume_role_with_web_identity`
## Summary

AWS provider [supports assuming a role with web identity](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume_role_with_web_identity-configuration-block), but the S3 backend does not

It would be good to have this option for the S3 backend as well, for feature parity

## References

- AWS provider code for reference - https://github.com/hashicorp/terraform-provider-aws/blob/3a33e810b7999e0f891c17940125831cf0ef49c2/internal/provider/provider.go#L229
- Related issue in Terraform Repository: https://github.com/hashicorp/terraform/issues/31244

mbialon:
Adds support for `assume_role_with_web_identity` block.

```
terraform {
  backend "s3" {
    bucket = "mybucket"
    key    = "my/key.tfstate"
    region = "us-east-1"
    assume_role_with_web_identity = {
      role_arn           = "arn:aws:iam::ACCOUNT-ID:role/Opentofu"
      web_identity_token = "<token value>"
    }
  }
}
```

Resolves #689 

## Target Release

1.6.0

-

### DIFF
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -180,54 +180,46 @@ func (b *Backend) ConfigSchema(context.Context) *configschema.Block {
 				Description: "The external ID to use when assuming the role",
 				Deprecated:  true,
 			},
-
 			"assume_role_duration_seconds": {
 				Type:        cty.Number,
 				Optional:    true,
 				Description: "Seconds to restrict the assume role session duration.",
 				Deprecated:  true,
 			},
-
 			"assume_role_policy": {
 				Type:        cty.String,
 				Optional:    true,
 				Description: "IAM Policy JSON describing further restricting permissions for the IAM Role being assumed.",
 				Deprecated:  true,
 			},
-
 			"assume_role_policy_arns": {
 				Type:        cty.Set(cty.String),
 				Optional:    true,
 				Description: "Amazon Resource Names (ARNs) of IAM Policies describing further restricting permissions for the IAM Role being assumed.",
 				Deprecated:  true,
 			},
-
 			"assume_role_tags": {
 				Type:        cty.Map(cty.String),
 				Optional:    true,
 				Description: "Assume role session tags.",
 				Deprecated:  true,
 			},
-
 			"assume_role_transitive_tag_keys": {
 				Type:        cty.Set(cty.String),
 				Optional:    true,
 				Description: "Assume role session tag keys to pass to any subsequent sessions.",
 				Deprecated:  true,
 			},
-
 			"workspace_key_prefix": {
 				Type:        cty.String,
 				Optional:    true,
 				Description: "The prefix applied to the non-default state path inside the bucket.",
 			},
-
 			"force_path_style": {
 				Type:        cty.Bool,
 				Optional:    true,
 				Description: "Force s3 to use path style api.",
 			},
-
 			"max_retries": {
 				Type:        cty.Number,
 				Optional:    true,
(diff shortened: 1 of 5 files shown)


website/docs/language/settings/backends/s3.mdx
```diff
@@ -224,6 +224,41 @@ terraform {
...
+The following `assume_role_with_web_identity` configuration block is optional:
+
+* `role_arn` - (Required) Amazon Resource Name (ARN) of the IAM Role to assume.
+Can also be set with the `AWS_ROLE_ARN` environment variable.
+* `duration` - (Optional) The duration individual credentials will be valid.
+Credentials are automatically renewed up to the maximum defined by the AWS account.
+Specified using the format `<hours>h<minutes>m<seconds>s` with any unit being optional.
+For example, an hour and a half can be specified as `1h30m` or `90m`.
+Must be between 15 minutes (15m) and 12 hours (12h).
+* `policy` - (Optional) IAM Policy JSON describing further restricting permissions for the IAM Role being assumed.
```
cube2222, line 238: Would it be possible for you to provide an example of using this?

E.g. constraining the policy to only create s3 buckets.

cube2222 (approved):
one nit

tomasmik (approved):

internal/backend/remote-state/s3/backend.go
```diff
@@ -674,6 +717,27 @@ func configureAssumeRole(obj cty.Value) *awsbase.AssumeRole {
...
+func configureAssumeRoleWithWebIdentity(obj cty.Value) *awsbase.AssumeRoleWithWebIdentity {
+	cfg := &awsbase.AssumeRoleWithWebIdentity{
+		RoleARN:              stringAttrDefaultEnvVar(obj, "role_arn", "AWS_ROLE_ARN"),
+		Policy:               stringAttr(obj, "policy"),
+		PolicyARNs:           stringSliceAttr(obj, "policy_arns"),
+		SessionName:          stringAttrDefaultEnvVar(obj, "session_name", "AWS_ROLE_SESSION_NAME"),
+		WebIdentityToken:     stringAttr(obj, "web_identity_token"),
+		WebIdentityTokenFile: stringAttrDefaultEnvVar(obj, "web_identity_token_file", "AWS_WEB_IDENTITY_TOKEN_FILE"),
+	}
+	if val, ok := stringAttrOk(obj, "duration"); ok {
```
marcinwyszynski, line 729: Would it make sense to validate if the duration makes sense in the 1st place ([15 minutes to 12 hours](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role-with-web-identity.html#session-duration))?

marcinwyszynski: Sorry being stupid here.

internal/backend/remote-state/s3/backend.go
```diff
@@ -674,6 +717,27 @@ func configureAssumeRole(obj cty.Value) *awsbase.AssumeRole {
 	return &assumeRole
 }
 
+func configureAssumeRoleWithWebIdentity(obj cty.Value) *awsbase.AssumeRoleWithWebIdentity {
+	cfg := &awsbase.AssumeRoleWithWebIdentity{
+		RoleARN:              stringAttrDefaultEnvVar(obj, "role_arn", "AWS_ROLE_ARN"),
+		Policy:               stringAttr(obj, "policy"),
```
marcinwyszynski, line 723: I guess if the value is not provided here, it then defaults to an empty string. Is the empty string a valid input to this call?

internal/backend/remote-state/s3/backend.go
```diff
@@ -674,6 +717,27 @@ func configureAssumeRole(obj cty.Value) *awsbase.AssumeRole {
...
+		PolicyARNs:           stringSliceAttr(obj, "policy_arns"),
+		SessionName:          stringAttrDefaultEnvVar(obj, "session_name", "AWS_ROLE_SESSION_NAME"),
+		WebIdentityToken:     stringAttr(obj, "web_identity_token"),
```
marcinwyszynski, line 726: Why wouldn't we want to be able to pass this via an envvar? We do it for the token file, but not for the token itself, which looks inconsistent and potentially cumbersome to use?

cube2222: Afaik the idea behind the new APIs is to move away from environment-variable-based configuration, and instead use e.g. the aws config file present on the machine.

marcinwyszynski: Well but all the other fields have corresponding envvars.

cube2222: Fair.

This actually seems to correspond to the AWS CLI though.

The AWS CLI also doesn't expose that one as an env var, but does the others: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

cube2222: So it might be just consistency with that.

internal/backend/remote-state/s3/backend.go
mbialon: I believe it's fine. An empty string is the same value as the zero value for the `Policy` field.

internal/backend/remote-state/s3/backend.go
mbialon: I used the env variables from the [guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc) but I can't think of a reason not to add it.

cube2222: I think we can add it, then.

website/docs/language/settings/backends/s3.mdx
mbialon: @cube2222 Added an example of using a policy. WDYT?

PR #766, opened 2023-10-23
Improve logging of global provider schema cache hits.

cube2222:
Right now there's no way to know if the cache was hit, which makes debugging harder.

### DIFF
--- a/internal/plugin/grpc_provider.go
+++ b/internal/plugin/grpc_provider.go
@@ -82,7 +82,7 @@ func (p *GRPCProvider) GetProviderSchema() (resp providers.GetProviderSchemaResp
 	// check the global cache
 	if !p.Addr.IsZero() {
 		if resp, ok := providers.SchemaCache.Get(p.Addr); ok {
-			logger.Trace("GRPCProvider: GetProviderSchema: serving from global schema cache")
+			logger.Trace("GRPCProvider: GetProviderSchema: serving from global schema cache", "address", p.Addr)
 			return resp
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/plugin/grpc_provider.go
```diff
@@ -81,6 +82,7 @@ func (p *GRPCProvider) GetProviderSchema() (resp providers.GetProviderSchemaResp
 	// check the global cache
 	if !p.Addr.IsZero() {
 		if resp, ok := providers.SchemaCache.Get(p.Addr); ok {
+			logger.Trace("GRPCProvider: GetProviderSchema: serving from global schema cache")
```
mbialon, line 85: Wouldn't it be useful to include `p.Addr` in the log entry?

internal/plugin6/grpc_provider.go
```diff
@@ -81,6 +82,7 @@ func (p *GRPCProvider) GetProviderSchema() (resp providers.GetProviderSchemaResp
 	// check the global cache
 	if !p.Addr.IsZero() {
 		if resp, ok := providers.SchemaCache.Get(p.Addr); ok {
+			logger.Trace("GRPCProvider: GetProviderSchema: serving from global schema cache")
```
mbialon, line 85: Same as in `internal/plugin/grpc_provider.go`.

Yantrio (approved):

internal/plugin/grpc_provider.go
cube2222: Good point, will do!

cube2222: Good idea!

mbialon (approved):

Issue #763, opened 2023-10-20
executing 'tofu init' on an existing configuration 


binnsr:
### OpenTofu Version

```shell
$ tofu version
OpenTofu v1.6.0-alpha3
on linux_amd64
```

### OpenTofu Configuration Files

we have quite a lot of code ( ~7k lines of active terraform code across ~200 files )
```hcl
...tofu config...
```

### Debug Output

tofu  init -backend-config="conn_str=${TERRAFORM_STATE_BACKEND}" -upgrade && tofu select workspace <name> && tofu plan
https://gist.github.com/binnsr/a6789f676849e191858288d1d8b1b205

second tofu  init -backend-config="conn_str=${TERRAFORM_STATE_BACKEND}" -upgrade && tofu plan
https://gist.github.com/binnsr/434bd52428052c5ecadfd844984a174a

### Expected Behavior

No changes. Your infrastructure matches the configuration.

OpenTofu has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

### Actual Behavior

╷
│ Error: Failed to load plugin schemas
│ 
│ Error while loading schemas for plugin components: 3 problems:
│ 
│ - Failed to obtain provider schema: Could not load the schema for provider registry.terraform.io/cloudflare/cloudflare: failed to instantiate provider "registry.terraform.io/cloudflare/cloudflare" to obtain
│ schema: unavailable provider "registry.terraform.io/cloudflare/cloudflare".
│ - Failed to obtain provider schema: Could not load the schema for provider registry.terraform.io/hashicorp/local: failed to instantiate provider "registry.terraform.io/hashicorp/local" to obtain schema:
│ unavailable provider "registry.terraform.io/hashicorp/local".
│ - Failed to obtain provider schema: Could not load the schema for provider registry.terraform.io/hashicorp/random: failed to instantiate provider "registry.terraform.io/hashicorp/random" to obtain schema:
│ unavailable provider "registry.terraform.io/hashicorp/random"..

### Steps to Reproduce

1. `rm -rf .terraform .terraform.lock.hcl`
2. `tofu  init -backend-config="conn_str=${TERRAFORM_STATE_BACKEND}" -upgrade`
3. `tofu workspace select $(cat .workspace_name)`
4. `tofu plan`

### Additional Context

Running `tofu init` and `tofu plan` a second time results in a valid plan.

[...]

PR #773, opened 2023-10-24
Stop reading providers from the HashiCorp registry based on the statefile if not specified in code.

cube2222:
If you're migrating from Terraform to OpenTofu and run `tofu init` on a codebase and statefile which references providers like `hashicorp/aws` then the code will be interpreted as `registry.opentofu.org/hashicorp/aws` while in the statefile you'll still have `registry.terraform.org/hashicorp/aws`. In this case, `tofu init` will fetch both copies, one from each of the registries.

That is not intuitive and not what the user would expect, so this code makes it so that if a provider in the old registry is addressed in the statefile but not in the code, then we assume there was no intention to fetch from the HashiCorp registry, and translate the providers in-memory. You can still specify the full path in the code to override this behavior.

The statefile will be updated on the first `tofu apply`, or can also manually be updated with the `replace-provider` command.

Related to #763

### DIFF
--- a/internal/tofumigrate/tofumigrate.go
+++ b/internal/tofumigrate/tofumigrate.go
@@ -1,0 +1,45 @@
+package tofumigrate
+
+import (
+	"github.com/opentofu/opentofu/internal/configs"
+	"github.com/opentofu/opentofu/internal/states"
+)
+
+// MigrateStateProviderAddresses can be used to update the in-memory view of the state to use registry.opentofu.org
+// provider addresses. This only applies for providers which are *not* explicitly referenced in the configuration in full form.
+// For example, if the configuration contains a provider block like this:
+//
+//	terraform {
+//	 required_providers {
+//	   random = {}
+//	 }
+//	}
+//
+// we will migrate the in-memory view of the statefile to use registry.opentofu.org/hashicorp/random.
+// However, if the configuration contains a provider block like this:
+//
+//	terraform {
+//	 required_providers {
+//	   random = {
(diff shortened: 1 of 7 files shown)


internal/tofumigrate/tofumigrate.go
```diff
@@ -0,0 +1,52 @@
...
+	diags = diags.Append(hclDiags)
+	if hclDiags.HasErrors() {
+		return nil, diags
+	}
+
+	for _, module := range stateCopy.Modules {
+		for _, resource := range module.Resources {
+			_, referencedInConfig := providers[resource.ProviderConfig.Provider]
+			if resource.ProviderConfig.Provider.Hostname == "registry.terraform.io" && !referencedInConfig {
+				resource.ProviderConfig.Provider.Hostname = "registry.opentofu.org"
```
Yantrio, line 46: Can you switch this to use `tfaddr.DefaultRegistryHost` instead of the hardcoded string?

That way when/if we do #618 we get this for free here.

cube2222: Great point!

Yantrio (changes requested):
one small nit. otherwise LGTM

Yantrio (approved):

Yantrio (approved):

dacut:
For for those (like me) blindly copying/pasting: the values for using the Terraform registry should be `registry.terraform.io/hashicorp/aws`, not ~`registry.terraform.org/hashicorp/aws`~, etc.

Issue #753, opened 2023-10-19
Backend/S3: Properly propagate the context in backend and client interfaces


tomasmik:
## Summary

After upgrading the S3 backend in https://github.com/opentofu/opentofu/pull/701 most calls to S3 now require a context to be passed. The pull request used context.TODO() as that was the easiest option not requiring any changes to interface of the backend or the client used.

Alter the `backend.Backend` and `remote.Client` interfaces to accept contexts and properly propagate it through out the codebase dropping the `context.TODO()`

marcinwyszynski:
This is quite a bit of work so in order to keep the review manageable I will go over the interface methods one by one.

PR #776, opened 2023-10-24
Make backend.ConfigSchema accept a context

marcinwyszynski:
Given that many state backend methods use contexts in their calls, we will make the related interfaces accept context as an input to all methods. Since it's a ton of work, we will go one by one, starting with a method where the context is not yet used (ConfigSchema) in any of the implementations.

related to #753

## Target Release

1.6.0

## Draft CHANGELOG entry

### DIFF
--- a/internal/builtin/providers/tf/data_source_state.go
+++ b/internal/builtin/providers/tf/data_source_state.go
@@ -4,6 +4,7 @@
 package tf
 
 import (
+	"context"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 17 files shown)


internal/builtin/providers/tf/data_source_state.go
```diff
@@ -222,7 +223,9 @@ func getBackend(cfg cty.Value) (backend.Backend, cty.Value, tfdiags.Diagnostics)
 		config = cty.ObjectVal(config.AsValueMap())
 	}
 
-	schema := b.ConfigSchema()
+	ctx := context.TODO()
```
marcinwyszynski, line 226: Note: I'm adding these TODOs as markers where I think the host function should be accepting a context itself.

tomasmik (approved):

PR #782, opened 2023-10-24
Make backend.DeleteWorkspace accept a context

marcinwyszynski:
Given that many state backend methods use contexts in their calls, we will make the related interfaces accept context as an input to all methods. Since it's a ton of work, we go one by one. This time it's a method (DeleteWorkspace) where the context is fairly widely used.

## Target Release

1.6.0

## Draft CHANGELOG entry

### DIFF
--- a/internal/backend/remote-state/http/client.go
+++ b/internal/backend/remote-state/http/client.go
@@ -243,7 +243,7 @@ func (c *httpClient) Put(data []byte) error {
 
 func (c *httpClient) Delete(ctx context.Context) error {
 	// Make the request
-	resp, err := c.httpRequest(context.TODO(), "DELETE", c.URL, nil, "delete state")
+	resp, err := c.httpRequest(ctx, "DELETE", c.URL, nil, "delete state")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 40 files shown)


internal/backend/remote-state/http/client.go
```diff
@@ -240,9 +241,9 @@ func (c *httpClient) Put(data []byte) error {
 	}
 }
 
-func (c *httpClient) Delete() error {
+func (c *httpClient) Delete(ctx context.Context) error {
 	// Make the request
-	resp, err := c.httpRequest("DELETE", c.URL, nil, "delete state")
+	resp, err := c.httpRequest(context.TODO(), "DELETE", c.URL, nil, "delete state")
```
mbialon, line 246:
```suggestion
	resp, err := c.httpRequest(ctx, "DELETE", c.URL, nil, "delete state")
```

mbialon (approved):
Tiny nit.

marcinwyszynski:
Good catch @mbialon, thanks! I will be making one more pass after I'm done with everything, so that I can catch all these stray `TODO`'s and `Background`'s.

PR #784, opened 2023-10-24
Make backend.Workspaces accept a context

marcinwyszynski:
Given that many state backend methods use contexts in their calls, we will make the related interfaces accept context as an input to all methods. Since it's a ton of work, we go one by one. This time it's a method (Workspaces) where the context is fairly widely used.

## Target Release

1.6.0

## Draft CHANGELOG entry

### DIFF
--- a/internal/command/workspace_delete.go
+++ b/internal/command/workspace_delete.go
@@ -75,7 +75,7 @@ func (c *WorkspaceDeleteCommand) Run(args []string) int {
 	// This command will not write state
 	c.ignoreRemoteVersionConflict(b)
 
-	ctx := context.Background()
+	ctx := context.TODO()
 
 	workspaces, err := b.Workspaces(ctx)
 	if err != nil {
--- a/internal/command/workspace_delete.go
+++ b/internal/command/workspace_delete.go
@@ -75,7 +75,9 @@ func (c *WorkspaceDeleteCommand) Run(args []string) int {
 	// This command will not write state
 	c.ignoreRemoteVersionConflict(b)
 
-	workspaces, err := b.Workspaces()
+	ctx := context.Background()
+
+	workspaces, err := b.Workspaces(ctx)
 	if err != nil {
 		c.Ui.Error(err.Error())
 		return 1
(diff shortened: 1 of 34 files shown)


internal/command/workspace_delete.go
```diff
@@ -75,7 +75,9 @@ func (c *WorkspaceDeleteCommand) Run(args []string) int {
 	// This command will not write state
 	c.ignoreRemoteVersionConflict(b)
 
-	workspaces, err := b.Workspaces()
+	ctx := context.Background()
```
mbialon, line 78: Why not use `context.TODO()`?

internal/command/workspace_list.go
```diff
@@ -58,7 +59,9 @@ func (c *WorkspaceListCommand) Run(args []string) int {
 	// This command will not write state
 	c.ignoreRemoteVersionConflict(b)
 
-	states, err := b.Workspaces()
+	ctx := context.Background()
```
mbialon, line 62: Why not use `context.TODO()`?

internal/command/workspace_new.go
```diff
@@ -90,7 +90,9 @@ func (c *WorkspaceNewCommand) Run(args []string) int {
 	// This command will not write state
 	c.ignoreRemoteVersionConflict(b)
 
-	workspaces, err := b.Workspaces()
+	ctx := context.Background()
```
mbialon, line 93: Why not use `context.TODO()`?

internal/command/workspace_select.go
```diff
@@ -83,7 +83,9 @@ func (c *WorkspaceSelectCommand) Run(args []string) int {
 		return 1
 	}
 
-	states, err := b.Workspaces()
+	ctx := context.Background()
```
mbialon, line 86: Why not use `context.TODO()`?

internal/command/workspace_delete.go
marcinwyszynski: Why not indeed. Fixed.

internal/command/workspace_new.go
marcinwyszynski: Same.

internal/command/workspace_list.go
marcinwyszynski: Fixed.

mbialon (approved):

PR #786, opened 2023-10-24
Pass context to all remote.Client operations

marcinwyszynski:
Given that many state backend methods use contexts in their calls, we will make the related interfaces accept context as an input to all methods. Since it's a ton of work, we go slowly. In this PR state reading and writing (as they're heavily intertwined) are taught to use contexts.

## Target Release

1.6.0

## Draft CHANGELOG entry

### DIFF
--- a/internal/backend/local/backend.go
+++ b/internal/backend/local/backend.go
@@ -348,9 +348,14 @@ func (b *Local) opWait(
 	case <-stopCtx.Done():
 		view.Stopping()
 
+		// We want to have a context that's guaranteed to be active that can be
+		// used to persist the state. Otherwise, if the operation is canceled
+		// or stopped before we can persist the state, we'll lose the state.
+		persistCtx := context.Background()
+
 		// try to force a PersistState just in case the process is terminated
 		// before we can complete.
-		if err := opStateMgr.PersistState(cancelCtx, nil); err != nil {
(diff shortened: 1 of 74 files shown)


internal/backend/local/backend.go
```diff
@@ -350,7 +350,7 @@ func (b *Local) opWait(
 
 		// try to force a PersistState just in case the process is terminated
 		// before we can complete.
-		if err := opStateMgr.PersistState(nil); err != nil {
+		if err := opStateMgr.PersistState(cancelCtx, nil); err != nil {
```
marcinwyszynski, line 353: This is one that I'm not sure about. Perhaps we should actually use a totally separate context here?

mbialon: We should. Canceling that operation was not possible before. It looks like the last line of defense against losing state changes. Canceling it midway could have bad consequences, right?

tomasmik: I would either go with a separate context or maybe `stopCtx`? Though separate sounds like a better choice

mbialon (approved):

tomasmik (approved):

internal/backend/local/backend.go
marcinwyszynski: I created a separate context and documented this choice.

tomasmik (approved):

PR #788, opened 2023-10-25
Backend/S3: Add support for `skip_requesting_account_id`

### LINKED ISSUE #774: Backend/S3: Add support for `skip_requesting_account_id`
`awsbase.Config` Has a setting `SkipRequestingAccountId` which should be set by a flag `skip_requesting_account_id`

mbialon:
Adds the `skip_requesting_account_id` attribute.

Resolves #774 

## Target Release

1.6.0

-


internal/backend/remote-state/s3/backend.go
```diff
@@ -161,6 +161,11 @@ func (b *Backend) ConfigSchema(context.Context) *configschema.Block {
 				Optional:    true,
 				Description: "Skip static validation of region name.",
 			},
+			"skip_requesting_account_id": {
```
marcinwyszynski, line 164: How does it work if the list of allowed or disallowed IDs is explicitly passed? I understand that the magic is happening inside a 3rd party library but we're the "contact surface" here, and I think it would be good to explain the expected behavior.

tomasmik (approved):
LGTM as long as other comments are addressed.

internal/backend/remote-state/s3/backend.go
mbialon: Passing this option will cause an empty account ID to be returned. The allowlist verification will work as expected. The forbidden account verification will "silently" fail (an empty account ID won't match any account IDs).

At the same time, the purpose of this option is to be used with non-AWS APIs. That's why I don't see a reason to use it with account verification options.

We could make it conflict with the account verification arguments though. WDYT? @marcinwyszynski @tomasmik

marcinwyszynski: > the purpose of this option is to be used with non-AWS APIs

Gotcha, makes sense.

marcinwyszynski (approved):

Yantrio (approved):

PR #792, opened 2023-10-26
Proper context propagation in Consul state backend implementation

marcinwyszynski:
While working on the context propagation in state backend implementations, I noticed that Consul was missing context handling entirely. I investigated the API a bit and figured out how to handle this properly. Here's my attempt on propagating the context to all Consul operations.

Related to #753

## Target Release

1.6.0

## Draft CHANGELOG entry

n/a since this is a pure housekeeping change

### DIFF
--- a/internal/backend/remote-state/consul/backend.go
+++ b/internal/backend/remote-state/consul/backend.go
@@ -18,76 +18,76 @@ import (
 func New() backend.Backend {
 	s := &schema.Backend{
 		Schema: map[string]*schema.Schema{
-			"path": &schema.Schema{
+			"path": {
 				Type:        schema.TypeString,
 				Required:    true,
 				Description: "Path to store state in Consul",
 			},
 
-			"access_token": &schema.Schema{
+			"access_token": {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: "Access token for a Consul ACL",
 				Default:     "", // To prevent input
 			},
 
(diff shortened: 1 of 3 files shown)


tomasmik (approved):

mbialon (approved):
Nit. What if we had one consistent way of creating options?

marcinwyszynski:
@mbialon can you please elaborate? I think different calls require different option types, so my feeling is that we're as consistent as it gets?

mbialon:
> @mbialon can you please elaborate? I think different calls require different option types, so my feeling is that we're as consistent as it gets?

One option is with the `var` expr like:

```
var queryOpts consulapi.QueryOptions
queryOpts.WithContext(ctx)

var writeOpts consulapi.WriteOptions
writeOpts.WithContext(ctx)
```

And the other is building the options in a single expr like:

```
writeOpts := (&consulapi.WriteOptions{}).WithContext(ctx)
```

marcinwyszynski:
Oh I see @mbialon, thanks for the explanation. I'm actually going to do something slightly different and will make these options part of the struct. This way if someone wants to add support for more options for the backend they will have a nice central place to put it.

mbialon (approved):
Nice work!

tomasmik (approved):

marcinwyszynski:
We are temporarily rolling back the context propagation on the backend. Once it's back in, we will merge it.

Issue #778, opened 2023-10-24
Backend/S3: Optionally set `ChecksumAlgorithm` during `Put` 


tomasmik:
AWS Docs: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

Let's use Sha256 and add an argument which would exclude it from the input in `Put` as not every service providing S3 supports checksum algorithms

PR #795, opened 2023-10-27
Backend/S3: Update the `Put` method

tomasmik:
* Switched to use the uploader as it's a better way to upload objects
* Added the checksum header
* Added a way to force not include the checksum header

Resolves https://github.com/opentofu/opentofu/issues/778

## Target Release

1.6.0

-


mbialon (approved):

r0b2g1t (approved):

internal/backend/remote-state/s3/client.go
```diff
@@ -184,7 +191,8 @@ func (c *RemoteClient) Put(ctx context.Context, data []byte) error {
 
 	log.Printf("[DEBUG] Uploading remote state to S3: %#v", i)
 
-	_, err := c.s3Client.PutObject(ctx, i)
+	uploader := manager.NewUploader(c.s3Client, func(u *manager.Uploader) {})
```
cube2222, line 194: The second argument is optional, you can skip it entirely.

cube2222 (approved):
With one nit

PR #840, opened 2023-11-08
Address s3 compatible remote state issues + logging

### LINKED ISSUE #821: S3, IAM, DynamoDB and STS endpoints do not allow bare hostname
### OpenTofu Version

```shell
1.6.0-alpha3
```

```hcl
terraform {
  backend "s3" {
    endpoint = "s3.us-east-1.amazonaws.com"
  }
}
```

### Debug Output

.

### Expected Behavior

Should initialize successfully

### Actual Behavior

```
Initializing the backend...
╷
│ Warning: Deprecated Parameter
│
│   on main.tf line 3, in terraform:
│    3:     endpoint = "s3.us-east-1.amazonaws.com"
│
│ The parameter "endpoint" is deprecated. Use parameter "endpoints.s3" instead.
╵

╷
│ Error: Invalid Value
│
│   on main.tf line 3, in terraform:
│    3:     endpoint = "s3.us-east-1.amazonaws.com"
│
│ The value must be a valid URL containing at least a scheme and hostname. Had "s3.us-east-1.amazonaws.com"
╵
```

### Steps to Reproduce

`terraform init -backend-config=bucket=my_bucket -backend-config=encrypt=true -backend-config=key=qwe.tfstate -backend-config=region=us-east-1`

### Additional Context

The following behaviour happens for S3, IAM, DynamoDB and STS endpoints as well

### References

- Related issue in the Terraform repository: https://github.com/hashicorp/terraform/issues/33981

cam72cam:
This PR started out addressing #821, but ended up being an investigation of the workflow itself instead of the not-adopted validation changes.

* Add in the relevant context hooks for aws-go-sdk-base logging.  This is required to easily debug the rest of the changes.
* Missing proto:// backwards compatibility
  - BaseEndpoint -> EndpointResolver fixed many of the issues related to non-proto endpoints.
  - Unfortunately it did not cover all cases, the main one I identified was in the aws-go-sdk-base middleware.
  - A workaround was added such that any non-proto endpoints would automatically have the https:// prefix added.
  - This is hopefully a sane default, and comes with a debug message that should help track down if it ever causes issues.
* PutObject checksum calculation
  - `skip_s3_checksum` was added in https://github.com/opentofu/opentofu/pull/795/files to partially work around s3 compatible endpoints not supporting the default sha256 PARTIAL headers.
  - This can be turned off, but should be avoided if possible.
  - Since we already have the payload as a `[]byte` in memory, it's simple to compute the hash ahead of time and bypass the streaming PARTIAL header issue.
  - In testing with DigitalOcean, providing the full sha256 instead of messing with PARTIAL fixed the issue.  Issues here and on related repositories suggest that this should fix the issue for multiple providers.
  - More reading: https://github.com/aws/aws-sdk-go-v2/issues/1689
* Added a HeadObject check to state.get to work around a MinIO issue where the GetObject request would return the wrong error message and fail.  I'm not sure that this is worth keeping in, though it might solve issues with other providers?

TODO:
- [x] Add tests to detecting / adding `proto://`
- [x] Decide if removing the new HeadObject call is the right call.

Resolves #821

## Target Release

1.6.0

### BUG FIXES
* S3 backend endpoints without `proto://` now default to `https://` instead of failing.
* Most S3 compatible remote state backends should now work without checksum errors / 400s by default.
* Logging has been re-added to S3 remote state calls.

### DIFF
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	awsbase "github.com/hashicorp/aws-sdk-go-base/v2"
+	baselogging "github.com/hashicorp/aws-sdk-go-base/v2/logging"
 	"github.com/opentofu/opentofu/internal/backend"
 	"github.com/opentofu/opentofu/internal/configs/configschema"
 	"github.com/opentofu/opentofu/internal/httpclient"
@@ -693,11 +694,13 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 		}
 	}
 
+	ctx := context.TODO()
+	ctx, baselog := baselogging.NewHcLogger(ctx, logging.HCLogger().Named("s3-backend"))
+
 	cfg := &awsbase.Config{
 		AccessKey:               stringAttr(obj, "access_key"),
 		CallerDocumentationURL:  "https://opentofu.org/docs/language/settings/backends/s3",
 		CallerName:              "S3 Backend",
-		SuppressDebugLog:        logging.IsDebugOrHigher(),
 		IamEndpoint:             customEndpoints["iam"].String(obj),
 		MaxRetries:              intAttrDefault(obj, "max_retries", 5),
 		Profile:                 stringAttr(obj, "profile"),
@@ -719,6 +722,7 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 		CustomCABundle:                 stringAttrDefaultEnvVar(obj, "custom_ca_bundle", "AWS_CA_BUNDLE"),
 		EC2MetadataServiceEndpoint:     stringAttrDefaultEnvVar(obj, "ec2_metadata_service_endpoint", "AWS_EC2_METADATA_SERVICE_ENDPOINT"),
 		EC2MetadataServiceEndpointMode: stringAttrDefaultEnvVar(obj, "ec2_metadata_service_endpoint_mode", "AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE"),
+		Logger:                         baselog,
 	}
 
 	if val, ok := boolAttrOk(obj, "use_legacy_workflow"); ok {
@@ -772,7 +776,6 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 		cfg.RetryMode = mode
 	}
 
-	ctx := context.TODO()
 	_, awsConfig, awsDiags := awsbase.GetAwsConfig(ctx, cfg)
 
 	for _, d := range awsDiags {
@@ -829,7 +832,8 @@ func verifyAllowedAccountID(ctx context.Context, awsConfig aws.Config, cfg *awsb
 func getDynamoDBConfig(obj cty.Value) func(options *dynamodb.Options) {
 	return func(options *dynamodb.Options) {
 		if v, ok := customEndpoints["dynamodb"].StringOk(obj); ok {
-			options.BaseEndpoint = aws.String(v)
+			// EndpointResolver does not require scheme://
+			options.EndpointResolver = dynamodb.EndpointResolverFromURL(v)
 		}
 	}
 }
@@ -837,7 +841,8 @@ func getDynamoDBConfig(obj cty.Value) func(options *dynamodb.Options) {
 func getS3Config(obj cty.Value) func(options *s3.Options) {
 	return func(options *s3.Options) {
 		if v, ok := customEndpoints["s3"].StringOk(obj); ok {
-			options.BaseEndpoint = aws.String(v)
+			// EndpointResolver does not require scheme://
+			options.EndpointResolver = s3.EndpointResolverFromURL(v)
 		}
 		if v, ok := boolAttrOk(obj, "force_path_style"); ok {
 			options.UsePathStyle = v
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -7,7 +7,9 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"log"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -1176,6 +1178,15 @@ func (e customEndpoint) String(obj cty.Value) string {
 	return v
 }
 
+func includeProtoIfNessesary(endpoint string) string {
+	//if !strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
+	if matched, _ := regexp.MatchString("[a-z]*://.*", endpoint); !matched {
+		log.Printf("[DEBUG] Adding https:// prefix to endpoint '%s'", endpoint)
+		endpoint = fmt.Sprintf("https://%s", endpoint)
+	}
+	return endpoint
+}
+
 func (e customEndpoint) StringOk(obj cty.Value) (string, bool) {
 	for _, path := range e.Paths {
 		val, err := path.Apply(obj)
@@ -1183,12 +1194,12 @@ func (e customEndpoint) StringOk(obj cty.Value) (string, bool) {
 			continue
 		}
 		if s, ok := stringValueOk(val); ok {
-			return s, true
+			return includeProtoIfNessesary(s), true
 		}
 	}
 	for _, envVar := range e.EnvVars {
 		if v := os.Getenv(envVar); v != "" {
-			return v, true
+			return includeProtoIfNessesary(v), true
 		}
 	}
 	return "", false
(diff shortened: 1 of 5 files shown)


Yantrio:
I've converted this to draft whilst we still have items here in TODO

internal/backend/remote-state/s3/backend.go
```diff
@@ -1171,19 +1178,28 @@ func (e customEndpoint) String(obj cty.Value) string {
...
 }
 
+func includeProtoIfNessesary(endpoint string) string {
+	//if !strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
+	if matched, _ := regexp.MatchString("[a-z]*://.*", endpoint); !matched {
+		log.Printf("[DEBUG] Adding https:// prefix to endpoint '%s'", endpoint)
+		endpoint = fmt.Sprintf("https://%s", endpoint)
+	}
+	return endpoint
+}
```
Yantrio, line 1188: I think instead of using regex, I would prefer if we parse the url and check the scheme directly.

Something like this should work I think

```suggestion
func includeProtoIfNecessary(endpoint string) (string, error) {
	parsedUrl, err := url.Parse(endpoint)
	if err != nil {
		return endpoint, err
	}

	if parsedUrl.Scheme == "" {
		log.Printf("[DEBUG] Adding https:// prefix to endpoint '%s'", endpoint)
		endpoint = "https://" + endpoint
	}

	return endpoint
}
```

Yantrio: You would have to add the errorhandling higher up though for the failure to parse

internal/backend/remote-state/s3/client.go
```diff
@@ -170,6 +197,17 @@ func (c *RemoteClient) Put(data []byte) error {
...
+
+		// There is a conflict in the aws-go-sdk-v2 that prevents it from working with many s3 compatible services
+		// Since we can pre-compute the hash here, we can work around it.
+		// ref: https://github.com/aws/aws-sdk-go-v2/issues/1689
+		algo := sha256.New()
+		algo.Write(data)
+		sum256 := algo.Sum(nil)
+		sum64 := make([]byte, base64.StdEncoding.EncodedLen(len(sum256)))
+		base64.StdEncoding.Encode(sum64, sum256)
+		sum64str := string(sum64)
```
Yantrio, line 209: You can use base64.StdEncoding.EncodeToString here to save a bit of memory

Yantrio (changes requested):
LGTM so far, just a couple nitpicks 👍

internal/backend/remote-state/s3/client.go
```diff
@@ -109,6 +112,30 @@ func (c *RemoteClient) get(ctx context.Context) (*remote.Payload, error) {
...
+			return nil, fmt.Errorf(errS3NoSuchBucket, err)
+		}
+
+		var nk *types.NotFound
+		if errors.As(err, &nk) {
+			return nil, nil
+		}
+
+		return nil, err
+	}
```
kislerdm, line 137: **(nitpick)** It may be a matter of taste, but I'd like to suggest the following refactoring 😃 

```suggestion
	// Head works around some s3 compatible backends not handling missing GetObject requests correctly (ex: minio Get returns Missing Bucket)
	_, err = c.s3Client.HeadObject(ctx, &s3.HeadObjectInput{
		Bucket: &c.bucketName,
		Key:    &c.path,
	})
	
	switch err.(type) {
	case nil:
	case *types.NotFound:
		return nil, nil
	case *types.NoSuchBucket:
		return nil, fmt.Errorf(errS3NoSuchBucket, err)
	default:
		return nil, err
	}
```

internal/backend/remote-state/s3/backend.go
cam72cam: Unfortunately, go's url.Parse(s) does not handle proto-less urls in a good way.  In many cases, it will set Scheme to the hostname, among other things.  That's why terraform's similar check looks at both Scheme != "" and Host != "".  I think the current approach is cleaner as we are not relying on a edge case in url.Parse.

internal/backend/remote-state/s3/client.go
cam72cam: I tried to match the current style, IMO there's a log of potential clean up that could happen in this backend.  The question is how much to do now and how much might be broken during a refactor.

internal/backend/remote-state/s3/backend_state.go
```diff
@@ -36,10 +38,14 @@ func (b *Backend) Workspaces() ([]string, error) {
 		MaxKeys: maxKeys,
 	}
 
+	ctx := context.TODO()
+
+	ctx, baselog := baselogging.NewHcLogger(ctx, logging.HCLogger().Named("s3-backend"))
+	ctx = baselogging.RegisterLogger(ctx, baselog)
+
```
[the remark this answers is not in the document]
cam72cam: Added

internal/backend/remote-state/s3/backend.go
Yantrio: Sounds good to me then! 👍

internal/backend/remote-state/s3/client.go
Yantrio: I'm personally not a huge fan of handling errors in this way, especially as we're only trying to check for 2 types. I'm happy with the handling as it is now.

Yantrio: I'm also 50/50 on inlining the parameter of the headObjectInput. Feel free to go with whichever you guys prefer 👍

Yantrio (approved):
So far so good, I'll tentitively approve for now, please ping me again when the final TODO items are done and it's ready for review again. Nice work!

internal/backend/remote-state/s3/client.go
```diff
@@ -109,6 +110,29 @@ func (c *RemoteClient) get(ctx context.Context) (*remote.Payload, error) {
...
```
Yantrio, line 134: I'm pretty happy with this approach of using HeadObject. But my experience with the s3 api is a bit lacking. 

@tomasmik @marcinwyszynski @mbialon what do you guys think here? Is this an acceptable approach?

cam72cam: I initially did this to work around an issue with MinIO not handing GetObject's correctly.  I was going to remove it, but noticed that Terraform was updated to do the same as part of their S3 Download Manager work (probably for buffer size allocation).

I could easily back it out, and am wondering if it's worth keeping as MinIO has a number of other issues.

tomasmik: >what do you guys think here? 

I have no strong feelings about this. If it doesn't break compatibility, I think this is fine.

internal/backend/remote-state/s3/client.go
tomasmik: >I'm personally not a huge fan of handling errors in this way

Yes. I agree.

My suggestion would be to just abstract those if checks in to a function `func sanitizeErr(err error) err` or something with a better name. Check and convert the error there and reuse it in other places.

kislerdm: 2 vs. 1 - democracy wins 😄

kislerdm:
**(question)** Shall we also cover custom endpoint for STS and IAM as per the issue description:

> The following behaviour happens for S3, IAM, DynamoDB and STS endpoints as well

For context, see the details:
- [IAM endpoint configuration](https://github.com/opentofu/opentofu/blob/aa6b6a2fec9924dbc70e15e04aa1ccdd514e7e16/internal/backend/remote-state/s3/backend.go#L706)
- [STS endpoint configuration](https://github.com/opentofu/opentofu/blob/aa6b6a2fec9924dbc70e15e04aa1ccdd514e7e16/internal/backend/remote-state/s3/backend.go#L713)

cc: @RLRabinowitz @Yantrio @cam72cam

cam72cam:
> **(question)** Shall we also cover custom endpoint for STS and IAM as per the issue description:
> 
> > The following behaviour happens for S3, IAM, DynamoDB and STS endpoints as well
> 
> For context, see the details:
> 
>     * [IAM endpoint configuration](https://github.com/opentofu/opentofu/blob/aa6b6a2fec9924dbc70e15e04aa1ccdd514e7e16/internal/backend/remote-state/s3/backend.go#L706)
> 
>     * [STS endpoint configuration](https://github.com/opentofu/opentofu/blob/aa6b6a2fec9924dbc70e15e04aa1ccdd514e7e16/internal/backend/remote-state/s3/backend.go#L713)
> 
> 
> cc: @RLRabinowitz @Yantrio @cam72cam

The https:// correction happens for all of the endpoints in the StringOk function.

internal/backend/remote-state/s3/backend.go
```diff
@@ -829,6 +840,8 @@ func verifyAllowedAccountID(ctx context.Context, awsConfig aws.Config, cfg *awsb
 func getDynamoDBConfig(obj cty.Value) func(options *dynamodb.Options) {
 	return func(options *dynamodb.Options) {
 		if v, ok := customEndpoints["dynamodb"].StringOk(obj); ok {
+			// EndpointResolver does not require scheme://
```
cube2222, line 843: What are those comments about?

internal/backend/remote-state/s3/backend.go
```diff
@@ -837,6 +850,8 @@ func getDynamoDBConfig(obj cty.Value) func(options *dynamodb.Options) {
 func getS3Config(obj cty.Value) func(options *s3.Options) {
 	return func(options *s3.Options) {
 		if v, ok := customEndpoints["s3"].StringOk(obj); ok {
+			// EndpointResolver does not require scheme://
```
cube2222, line 853: Same

kislerdm (approved):

cube2222 (approved):

PR #848, opened 2023-11-09
Allow expired provider pgp keys, with warning

### LINKED ISSUE #822: Do not fail initialization if signing key has expired
### OpenTofu Version

```shell
1.6.0-alpha3
```

```hcl
...tofu config...
```

### Debug Output

.

### Expected Behavior

Running `tofu init` should not fail if one of the providers' signing keys are expired

### Actual Behavior

Running `tofu init` does fails if one of the providers' signing keys are expired

### Steps to Reproduce

`tofu init` on a configuration that includes a provider that has been signed by expired signing keys

### Additional Context

`go-crypto` is making the GPG key expiry validation for `tofu`. We'd need to either pick a different package, or to somehow skip this validation / display a warning instead of an error.
An expired key should not fail `tofu init`

### References

- Related issue in the Terraform repository: https://github.com/hashicorp/terraform/issues/33984
- Related issue regarding the current behaviour of having the GPG key validation not be mandatory - https://github.com/opentofu/opentofu/issues/266

cam72cam:
Example:
```
- Installing hashicorp/tfcoremock v0.2.0...
[WARN] Provider hashicorp/tfcoremock (registry.opentofu.org) gpg key expired, this will fail in future versions of OpenTofu: 'https://github.com/opentofu/terraform-provider-tfcoremock/releases/download/v0.2.0/terraform-provider-tfcoremock_0.2.0_linux_amd64.zip'
```

Resolves #822

## Target Release

1.6.0

### BUG FIXES
* Allow providers with expired gpg keys (with a warning)

### DIFF
--- a/internal/getproviders/package_authentication.go
+++ b/internal/getproviders/package_authentication.go
@@ -491,7 +491,7 @@ func (s signatureAuthentication) findSigningKey() (*SigningKey, string, error) {
 		if err == openpgpErrors.ErrSignatureExpired || err == openpgpErrors.ErrKeyExpired {
 			// Internally openpgp will *only* return the Expired errors if all other checks have succeded
 			// This is currently the best way to work around expired provider keys
-			fmt.Printf("[WARN] Provider %s/%s (%v) gpg key expired, this will fail in future versions of OpenTofu: '%s'\n", s.Meta.Provider.Namespace, s.Meta.Provider.Type, s.Meta.Provider.Hostname, s.Meta.Location.String())
+			fmt.Printf("[WARN] Provider %s/%s (%v) gpg key expired, this will fail in future versions of OpenTofu\n", s.Meta.Provider.Namespace, s.Meta.Provider.Type, s.Meta.Provider.Hostname)
 			err = nil
 		}
 
--- a/internal/getproviders/package_authentication.go
+++ b/internal/getproviders/package_authentication.go
@@ -29,6 +29,7 @@ const (
 
 const (
 	enforceGPGValidationEnvName = "OPENTOFU_ENFORCE_GPG_VALIDATION"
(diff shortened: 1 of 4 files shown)


internal/getproviders/package_authentication.go
```diff
@@ -491,7 +487,13 @@ func (s signatureAuthentication) findSigningKey() (*SigningKey, string, error) {
 			return nil, "", fmt.Errorf("error decoding signing key: %w", err)
 		}
 
-		entity, err := openpgp.CheckDetachedSignature(keyring, bytes.NewReader(s.Document), bytes.NewReader(s.Signature), openpgpConfig)
+		entity, err := openpgp.CheckDetachedSignature(keyring, bytes.NewReader(s.Document), bytes.NewReader(s.Signature), nil)
+		if err == openpgpErrors.ErrSignatureExpired || err == openpgpErrors.ErrKeyExpired {
+			// Internally openpgp will *only* return the Expired errors if all other checks have succeded
+			// This is currently the best way to work around expired provider keys
+			fmt.Printf("[WARN] Provider %s/%s (%v) gpg key expired, this will fail in future versions of OpenTofu: '%s'\n", s.Meta.Provider.Namespace, s.Meta.Provider.Type, s.Meta.Provider.Hostname, s.Meta.Location.String())
```
Yantrio, line 494: I think we don't need to print the location here, Printing the provider name is enough is enough.

Yantrio (approved):
One tiny nitpick, otherwise LGTM.

Maybe if we get time we could add some tests too?

cam72cam:
This is actually tested by default as many of the example keys are already expired, I could add a note?

kislerdm (approved):

PR #267, opened 2023-09-05
Fix `opentf test` crash on nil output

eranelbaz:
Fixes #257

Since OpenTF does not store `nil` outputs in it's state we fail to get the value.
Added a check of `if output == nil` and if true returning `cty.NilVal` 

Added relevant tests - 
- `opentf test` does not crash for `null` outputs
- Made sure we don't some how pass undeclared outputs and set them as `null`

## Target Release

1.6.0-alpha

### BUG FIXES

 

-  `opentf test` does not crash for `null` outputs

PR #850, opened 2023-11-10
Avoid mutating input config in node_resource_validate

### LINKED ISSUE #815: `tofu test` performs progressively slower for multiple `run` blocks when using `provisioner`
### OpenTofu Version

```shell
1.6.0-alpha3
```

**main.tf:**
```hcl
resource "terraform_data" "provision" {
  connection {
    host = "localhost"
  }
  provisioner "remote-exec" {
    inline = ["echo test"]
  }
}
```
**main.tftest.hcl:**
```hcl
run "test 01" { command = plan }
run "test 02" { command = plan }
run "test 03" { command = plan }
run "test 04" { command = plan }
run "test 05" { command = plan }
run "test 06" { command = plan }
run "test 07" { command = plan }
run "test 08" { command = plan }
run "test 09" { command = plan }
run "test 10" { command = plan }
run "test 11" { command = plan }
run "test 12" { command = plan }
run "test 13" { command = plan }
run "test 14" { command = plan }
run "test 15" { command = plan }
run "test 16" { command = plan }
run "test 17" { command = plan }
run "test 18" { command = plan }
run "test 19" { command = plan }
run "test 20" { command = plan }
run "test 21" { command = plan }
run "test 22" { command = plan }
run "test 23" { command = plan }
run "test 24" { command = plan }
run "test 25" { command = plan }
run "test 26" { command = plan }
run "test 27" { command = plan }
run "test 28" { command = plan }
run "test 29" { command = plan }
run "test 30" { command = plan }

```

### Debug Output

After running around ~20 tests:
```
2023-11-05T14:11:24.053+0200 [TRACE] Executing graph transform *tofu.ReferenceTransformer
2023-11-05T14:14:24.037+0200 [DEBUG] ReferenceTransformer: "terraform_data.provision (expand)" references: []
```

### Expected Behavior

Each `run` should pretty much take the same amount of time

### Actual Behavior

Each `run` gets progressively slows, and doubles on execution time

### Steps to Reproduce

1. `tofu init`
2. `tofu test`

### Additional Context

_No response_

### References

- Related issue in Terraform repository: https://github.com/hashicorp/terraform/issues/34009

tarruda:
The `Execute()` method of `NodeValidateResource` merges the connection config of each provisioner with the base connection config.

The problem was that it used to replace the original provisioner connection with the result of the merge, so on each invocation it was merging with previous results.

Fix the issue by creating a local copy of the provisioner and connection to store the merge.

Resolves #815

## Target Release

1.6.0

- `tofu test`: Fixed exponential slowdown with multiple consecutive `run` blocks


kislerdm:
@tarruda  Hey Thiago! Thank you for your contribution! Could you please add a test covering the change's behaviour? Thanks!

cube2222 (approved):
Looks good, and works great!

Thanks a lot for the contribution @tarruda!

cube2222:
@tarruda @kislerdm Agree a test would be nice. I think the easiest and most valuable would be an end to end test that basically contains the reproduction from the issue.

cube2222:
Here's a previous PR adding an e2e test for `tofu test`, that might be useful as a reference: https://github.com/opentofu/opentofu/pull/267/files

tarruda:
> Here's a previous PR adding an e2e test for `tofu test`, that might be useful as a reference: https://github.com/opentofu/opentofu/pull/267/files

That's helpful, thanks @cube2222 

I tried adding the reproduction, but the test fails with the following message:

```
            Failure! 0 passed, 1 failed, 29 skipped.
            ╷
            │ Error: Missing required provider
            │ 
            │ This configuration requires built-in provider
            │ terraform.io/builtin/terraform, but that provider isn't available in this
            │ OpenTofu version.
            ╵
            ╷
            │ Error: Missing required provisioner plugin
            │ 
            │ This configuration requires provisioner plugin "remote-exec", which isn't
            │ available. If you're intending to use an external provisioner plugin, you
            │ must install it manually into one of the plugin search directories before
            │ running OpenTofu.
```

Any hints on how to fix that?

I noticed that the test runner invokes the `TestCommand` API directly:

```
			c := &TestCommand{
				Meta: Meta{
					testingOverrides: metaOverridesForProvider(provider.Provider),
					View:             view,
				},
			}

			code := c.Run(tc.args)
```
Do we have some e2e testing infrastructure that invokes `tofu` program directly in a directory containing the fixture? I'm assuming that invoking `TestCommand` directly is not loading the required provider.

cube2222:
@tarruda The first error is easy, we can use use a resource from the test provider.

```
resource "test_resource" "foo" {
  id    = "constant_value"
  value = "bar"

  provisioner "local-exec" {
    inline = ["echo test"]
  }
}
```

But that will still fail on the provisioner plugin. Still looking for a solution to that one. Maybe there's a test provisioner?

tarruda:
I think I made a mistake, I tried adding it to `internal/command/test_test.go`, but it should be in the `e2etest` directory which actually invokes the program

cube2222:
Ok, it seems like the only provisioner test that actually runs on .tf files is the fully e2e test in `internal/command/e2etest/provisioner_test.go`

Ok, so it seems there are two kinds of provisioner tests. One is fully e2e (`internal/command/e2etest/provisioner_test.go`) and ones are more integration (e.g. TestContext2Apply_provisionerDestroy in internal/tofu/context_apply_test.go). I think ideally we'd do it similar to the latter here.

cube2222:
@tarruda Ok, so it seems there are two kinds of provisioner tests. One is fully e2e (internal/command/e2etest/provisioner_test.go) and ones are more integration (e.g. TestContext2Apply_provisionerDestroy in internal/tofu/context_apply_test.go). I think ideally we'd do it similar to the latter here.

cube2222:
@tarruda alright, figured it out. We need to add a mock provisioner, similar to the latter case above.

Change `metaOverridesForProvider` in `internal/command/command_test.go` to
```
func metaOverridesForProvider(p providers.Interface) *testingOverrides {
	return &testingOverrides{
		Providers: map[addrs.Provider]providers.Factory{
			addrs.NewDefaultProvider("test"):                                           providers.FactoryFixed(p),
			addrs.NewProvider(addrs.DefaultProviderRegistryHost, "hashicorp2", "test"): providers.FactoryFixed(p),
		},
		Provisioners: map[string]provisioners.Factory{
			"shell": func() (provisioners.Interface, error) {
				provisioner := new(tofu.MockProvisioner)
				provisioner.GetSchemaResponse = provisioners.GetSchemaResponse{
					Provisioner: &configschema.Block{
						Attributes: map[string]*configschema.Attribute{
							"command": {
								Type:     cty.String,
								Optional: true,
							},
							"order": {
								Type:     cty.String,
								Optional: true,
							},
							"when": {
								Type:     cty.String,
								Optional: true,
							},
						},
					},
				}
				return provisioner, nil
			},
		},
	}
}
```
(basically adding a provisioner)

And then you can use it in the `internal/command/test_test.go` tests like so
```
resource "test_resource" "foo" {
  id    = "constant_value"
  value = "bar"

  provisioner "shell" {
    command = "echo test"
  }
}
```

tarruda:
@cube2222  I've added an e2e test that invokes the `tofu` command in a separate goroutine and fails if it takes more than 3 seconds to complete, can you check if it is sufficient?

tarruda:
```
resource "test_resource" "foo" {
  id    = "constant_value"
  value = "bar"

  provisioner "shell" {
    command = "echo test"
  }
}
```

Not sure this would reproduce the bug, as it only happens when merging connection configurations with remote-exec

tarruda:
Also verified that the test fails with `timed out` error when reverting the fix

cube2222:
The test looks excellent!

> Not sure this would reproduce the bug, as it only happens when merging connection configurations with remote-exec

Haven't realized that, thanks for the explanation.

Will be merging this now, thanks again for the contribution @tarruda!

PR #855, opened 2023-11-11
keep init wording consistent in internal/command/init.go

haojue:
Resolves #

## Target Release

1.6.0

-

### DIFF
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -264,7 +264,7 @@ func (c *InitCommand) Run(args []string) int {
 	}
 
 	// We've passed the core version check, now we can show errors from the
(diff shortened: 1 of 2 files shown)


cam72cam (changes requested):
Several tests will need to be updated, for example:
https://github.com/opentofu/opentofu/blob/main/internal/command/init_test.go#L2644

haojue:
> Several tests will need to be updated, for example: https://github.com/opentofu/opentofu/blob/main/internal/command/init_test.go#L2644

@cam72cam , thank you for the review! I've updated these per your comment.

cam72cam (approved):

PR #861, opened 2023-11-13
Remove draft CHANGELOG section of PR template

cam72cam:
This primarily caused confusion of where the CHANGELOG was supposed to be updated.  The current bot that requests the actual CHANGELOG.md file be updated is a much better reminder.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -30,15 +30,3 @@ all targeted releases.
 -->
 
 1.6.0
-
-<!--
-
-Write a short description of the user-facing change. Examples:
-
-- `tofu show -json`: Fixed crash with sensitive set values.
-- When rendering a diff, OpenTofu now quotes the name of any object attribute whose string representation is not a valid identifier.
-- The local token configuration in the cloud and remote backend now has higher priority than a token specified in a credentials block in the CLI configuration.
-
---> 
-
--  
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -31,16 +31,6 @@ all targeted releases.
 
 1.6.0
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -31,14 +31,6 @@ all targeted releases.
...
 
-## Draft CHANGELOG entry
-
-<!--
-
-Choose a category, delete the others:
-
--->
-
 ### NEW FEATURES | UPGRADE NOTES | ENHANCEMENTS | BUG FIXES | EXPERIMENTS
```
Yantrio, line 35: I think this line needs removing too
```suggestion
```

cam72cam: Fixed!

kislerdm: I feel like the comment below may be removed as well
```markdown 
<!--

Write a short description of the user-facing change. Examples:

- `tofu show -json`: Fixed crash with sensitive set values.
- When rendering a diff, OpenTofu now quotes the name of any object attribute whose string representation is not a valid identifier.
- The local token configuration in the cloud and remote backend now has higher priority than a token specified in a credentials block in the CLI configuration.

--> 
-
```

cam72cam:
> I feel like these [lines](https://github.com/opentofu/opentofu/blob/cc3727a0420ccf3fb05448ac3b762c125621b635/.github/pull_request_template.md?plain=1#L34-L44) can be removed as well, most likely they will be ignored when writing PR description IMO. WDYT?

I think so?  I was not 100% sure if that was supposed to be attached to the Draft CHANGELOG section or not.  I'm pretty sure you are correct.

kislerdm (approved):

PR #867, opened 2023-11-13
Fix workspace name overwrite with env var TF_WORKSPACE

### LINKED ISSUE #814: `tofu init` with `cloud` block and `TF_WORKSPACE` doesn't work
### OpenTofu Version

```shell
1.6.0-alpha3
```

```hcl
terraform {
  cloud {
    organization = "<MY_CLOUD_ORGANIZATION>"
    hostname     = "<MY_CLOUD_BACKEND_HOSTNAME>"

    workspaces {
      tags = ["<MY_WORKSPACE_TAGS>"]
    }
  }
}
```

### Debug Output

None relevant

### Expected Behavior

Should initialize successfully

### Actual Behavior

Initializing cloud backend...
╷
│ Error: Invalid workspaces configuration
│
│   on main.tf line 2, in terraform:
│    2:   cloud {
│
│ Only one of workspace "tags" or "name" is allowed.
│
│ The 'workspaces' block configures how OpenTofu CLI maps its workspaces for this single
│ configuration to workspaces within a cloud backend organization. Two strategies are available:
│
│ tags - A set of tags used to select remote cloud backend workspaces to be used for this single
│ configuration. New workspaces will automatically be tagged with these tag values. Generally, this
│ is the primary and recommended strategy to use. This option conflicts with "name".
│
│ name - The name of a single cloud backend workspace to be used with this configuration.
│ When configured, only the specified workspace can be used. This option conflicts with "tags".

### Steps to Reproduce

`TF_WORKSPACE="any_workspace" tofu init`

### Additional Context

Using `TF_WORKSPACE` should completely overwrite the workspace used, and overwrite the `workspaces` block configuration

### References

- Relevant issue in Terraform repository: https://github.com/hashicorp/terraform/issues/33976

kislerdm:
Resolves #814 

## Target Release

1.6.0-alpha6

## What changed

- Fixes cloud backend configuration order: `TF_WORKSPACE` overwrites `workspaces` attributes
- Removed deadcode `WithEnvVars`
- Enhanced tests for the cloud backend method `setConfigurationFields` and `PrepareConfig`

## Why do we need it

- To follow expected `tofu init` behaviour

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ BUG FIXES:
 * Fixed regression where `tofu` was not ignoring GPG signing key expiration by default ([822](https://github.com/opentofu/opentofu/issues/822))
   - Added experimental env variable to enforce checking of GPG signing key expiration `OPENTOFU_ENFORCE_GPG_EXPIRATION=true`
   - This is planned to default to `true` in a future release.
-* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814))
+* cloud: fixed a bug related to `workspaces` configuration not correctly recognizing TF_WORKSPACE environment variable ([#814](https://github.com/opentofu/opentofu/issues/814))
 
 S3 BACKEND:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ BUG FIXES:
 * Fixed regression where `tofu` was not ignoring GPG signing key expiration by default ([822](https://github.com/opentofu/opentofu/issues/822))
   - Added experimental env variable to enforce checking of GPG signing key expiration `OPENTOFU_ENFORCE_GPG_EXPIRATION=true`
   - This is planned to default to `true` in a future release.
-* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814)). Now, if both, `workspaces.name` in the cloud block and `TF_WORKSPACE` are defined, the `TF_WORKSPACE` value takes precedence.
+* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814))
 
 S3 BACKEND:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ BUG FIXES:
 * Fixed regression where `tofu` was not ignoring GPG signing key expiration by default ([822](https://github.com/opentofu/opentofu/issues/822))
   - Added experimental env variable to enforce checking of GPG signing key expiration `OPENTOFU_ENFORCE_GPG_EXPIRATION=true`
   - This is planned to default to `true` in a future release.
+* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814)). Now, `workspaces` attributes can be overwritten by the environment variables `TF_WORKSPACE` and `TF_CLOUD_PROJECT`.
+  **Note** that `TF_WORKSPACE` overwrites `name` and removes `tags` from the backend's workspaces configuration.
 
 S3 BACKEND:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,8 +52,7 @@ BUG FIXES:
 * Fixed regression where `tofu` was not ignoring GPG signing key expiration by default ([822](https://github.com/opentofu/opentofu/issues/822))
   - Added experimental env variable to enforce checking of GPG signing key expiration `OPENTOFU_ENFORCE_GPG_EXPIRATION=true`
   - This is planned to default to `true` in a future release.
-* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814)). Now, `workspaces` attributes can be overwritten by the environment variables `TF_WORKSPACE` and `TF_CLOUD_PROJECT`.
-  **Note** that `TF_WORKSPACE` overwrites `name` and removes `tags` from the backend's workspaces configuration.
+* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814)). Now, if both, `workspaces.name` in the cloud block and `TF_WORKSPACE` are defined, the `TF_WORKSPACE` value takes precedence.
 
 S3 BACKEND:
 
--- a/internal/cloud/backend.go
+++ b/internal/cloud/backend.go
@@ -438,8 +438,15 @@ func (b *Cloud) setConfigurationFields(obj cty.Value) tfdiags.Diagnostics {
 	b.WorkspaceMapping = newWorkspacesMappingFromFields(obj)
 
 	// Overwrite workspaces config from env variable
-	// NOTE that any workspace attribute can be overwritten using env variables
-	reconcileWorkspaceMappingEnvVars(&b.WorkspaceMapping)
+	if err := reconcileWorkspaceMappingEnvVars(&b.WorkspaceMapping); err != nil {
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			err.Error(),
+			`Find details in https://opentofu.org/docs/cli/cloud/settings#environment-variables.`,
+		))
+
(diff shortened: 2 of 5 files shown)


internal/cloud/backend.go
```diff
@@ -424,40 +435,44 @@ func (b *Cloud) setConfigurationFields(obj cty.Value) tfdiags.Diagnostics {
...
+}
 
-		// PrepareConfig checks that you cannot set both of these.
-		if val := workspaces.GetAttr("name"); !val.IsNull() {
-			b.WorkspaceMapping.Name = val.AsString()
+func reconcileWorkspaceMappingEnvVars(w *WorkspaceMapping) tfdiags.Diagnostic {
+	// See: https://github.com/opentofu/opentofu/issues/814
+	if v := os.Getenv("TF_WORKSPACE"); v != "" && w.Name == "" {
+		if len(w.Tags) > 0 && !workspaceInTags(w.Tags, v) {
+			return invalidWorkspaceConfigMisconfigurationEnvVar
```
cube2222, line 456: I think we just want to only use the workspace and tag env vars if the workspace block isn't there, no? Basically, only look at env var if workspace configuration isn't in code.

kislerdm: It's slightly misleading because it'd contradict to the behaviour illustrated in the documentation:

> The name of a single cloud backend workspace. OpenTofu reads this when workspaces is omitted from the cloud block. The cloud backend will not create a new workspace from this variable; the workspace must exist in the specified organization. You can set TF_WORKSPACE if the cloud block uses tags. However, the value of TF_WORKSPACE must be included in the set of tags.

It's a bit cryptic, however, I read it as following: `TF_WORKSPACE` is used in two cases:
- `workspaces` config is not present in the `cloud` block;
- `workspaces.tags` is set only.

I find the logic implemented in 1.5.7 which we agreed to use a reference contradicting that understanding, hence I adjusted it accordingly.

WDYT?

internal/cloud/backend.go
```diff
@@ -424,40 +435,44 @@ func (b *Cloud) setConfigurationFields(obj cty.Value) tfdiags.Diagnostics {
...
+		w.Project = v
 	}
 
-	// Determine if we are forced to use the local backend.
-	b.forceLocal = os.Getenv("TF_FORCE_LOCAL_BACKEND") != ""
+	return nil
+}
 
-	return diags
+func workspaceInTags(tags []string, workspace string) bool {
```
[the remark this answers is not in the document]
kislerdm: Sorry if it's misleading, but tags after parsing are stored as a slice in the `WorkspacesMapping` attributed of the backend object `b`.

image

It's also in line with the interface [definition](https://github.com/opentofu/opentofu/blob/637c65f6f2d7af65a9ad967be5ce489f2aaac05a/internal/backend/backend.go#L119).

internal/cloud/backend.go
cube2222: You're explanation, and especially the docs excerpt clear things up. LGTM, thanks!

internal/cloud/backend.go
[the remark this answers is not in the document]
cube2222: You've cleared up my confusion in the other discussion.

cube2222 (approved):

cam72cam (approved):

internal/cloud/backend_test.go
```diff
@@ -273,340 +273,43 @@ func TestCloud_PrepareConfigWithEnvVars(t *testing.T) {
...
-					"project": cty.NullVal(cty.String),
-				}),
-			}),
-			vars: map[string]string{
-				"TF_WORKSPACE": "shire",
-			},
-			expectedWorkspaceName: "", // No error is raised, but workspace is not set
-		},
-		"project specified": {
+		"with workspace env var overwrite tags config value": {
```
RLRabinowitz, line 276: Maybe I missed something here, but shouldn't this test cause an error? If workspace overwrite is not in the tags, then there should be an error

There's no `expectedErr` in this test case

kislerdm: Well spotted! Thank you Arel! I fixed the test definition. Unfortunately, I feel like most of the tests (no relevant to this issue) with inherited validation will require thorough cross-check as a followup.

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ BUG FIXES:
 * Fixed regression where `tofu` was not ignoring GPG signing key expiration by default ([822](https://github.com/opentofu/opentofu/issues/822))
   - Added experimental env variable to enforce checking of GPG signing key expiration `OPENTOFU_ENFORCE_GPG_EXPIRATION=true`
   - This is planned to default to `true` in a future release.
+* cloud: fixed a bug related to `workspaces` configuration using environment variables ([#814](https://github.com/opentofu/opentofu/issues/814))
```
Yantrio, line 55: I think people may come along and hit `ctrl+F` and search for `TF_WORKSPACE` in our changelog if they did hit this issue, maybe we can be a bit more verbose here?

Maybe something like
```suggestion
* cloud: fixed a bug related to `workspaces` configuration not correctly recognizing TF_WORKSPACE env var ([#814](https://github.com/opentofu/opentofu/issues/814))
```

kislerdm: @Yantrio thank you for suggestion James! I agree with it 👍🏻

internal/cloud/backend.go
```diff
@@ -210,6 +201,26 @@ func (b *Cloud) PrepareConfig(obj cty.Value) (cty.Value, tfdiags.Diagnostics) {
...
+		if val := workspaces.GetAttr("tags"); !val.IsNull() {
+			err := gocty.FromCtyValue(val, &WorkspaceMapping.Tags)
+			if err != nil {
+				log.Panicf("An unexpected error occurred: %s", err)
+			}
+		}
+		if val := workspaces.GetAttr("project"); !val.IsNull() && val.AsString() != "" {
+			WorkspaceMapping.Project = val.AsString()
+		}
+	}
```
Yantrio, line 220: Whilst I'm aware we're moving existing code here, we have a lot of nesting `if` statements and a chance to clean some of this whilst we're here to improve readability and reduce cognitive load overall.

```suggestion
  workspaces := obj.GetAttr("workspaces")
  if workspaces.IsNull() {
      return WorkspaceMapping{}
  }
  
  nameVal := workspaces.GetAttr("name")
  if !nameVal.IsNull() {
      WorkspaceMapping.Name = nameVal.AsString()
  }
  
  tagsVal := workspaces.GetAttr("tags")
  if !tagsVal.IsNull() {
      err := gocty.FromCtyValue(tagsVal, &WorkspaceMapping.Tags)
      if err != nil {
          log.Panicf("An unexpected error occurred: %s", err)
      }
  }
  
  projectVal := workspaces.GetAttr("project")
  if !projectVal.IsNull() && projectVal.AsString() != "" {
      WorkspaceMapping.Project = projectVal.AsString()
  }
```


Maybe something like this is a bit easier to handle? Early returns and explicit steps about what each bit is handling helps the reader understand what's going on.

(**Note:** Formatting may be a little off, I wrote this in the github comment ui)

internal/cloud/errors.go
```diff
@@ -38,6 +38,13 @@ var (
 		fmt.Sprintf("Only one of workspace \"tags\" or \"name\" is allowed.\n\n%s", workspaceConfigurationHelp),
 		cty.Path{cty.GetAttrStep{Name: "workspaces"}},
 	)
+
+	invalidWorkspaceConfigMisconfigurationEnvVar = tfdiags.AttributeValue(
+		tfdiags.Error,
+		"Invalid workspaces configuration",
+		fmt.Sprintf("The workspace defined using env variable \"TF_WORKSPACE\" does not belong to \"tags\".\n\n%s", workspaceConfigurationHelp),
```
Yantrio, line 45:
```suggestion
		fmt.Sprintf("The workspace defined using the environment variable \"TF_WORKSPACE\" does not belong to \"tags\".\n\n%s", workspaceConfigurationHelp),
```

Yantrio (changes requested):

internal/cloud/errors.go
kislerdm: Applied

internal/cloud/backend.go
kislerdm: Thanks for the suggestion James! I agree, such structure might be subjectively more human-friendly. The suggestion was applied.

Yantrio (approved):

RLRabinowitz (approved):

PR #876, opened 2023-11-15
Fix S3 Remote State Backend Validation

### LINKED ISSUE #875: Failed to read terraform_remote_state data
### OpenTofu Version

```shell
OpenTofu v1.6.0-alpha5
on linux_amd64
```

```hcl
data "terraform_remote_state" "vpc" {
  backend = "s3"
  config = {
    region = "us-west-2"
    bucket = "terragrunt-test-bucket-2ttejd"
    key    = "mgmt/vpc/terraform.tfstate"
  }
}

```

### Debug Output

https://gist.github.com/denis256/341a3042d60cd885d3fc918446acce2d

### Expected Behavior

Read remote state as data

Terraform is reading it, Tofu fails

### Actual Behavior

╷
│ Error: Invalid backend configuration
│ 
│   with data.terraform_remote_state.vpc,
│   on main.tf line 3, in data "terraform_remote_state" "vpc":
│    3:   config = {
│    4:     region = "us-west-2"
│    5:     bucket = "terragrunt-test-bucket-2ttejd"
│    6:     key    = "mgmt/vpc/terraform.tfstate"
│    7:   }
│ 
│ The given configuration is not valid for backend "s3": attribute "assume_role" is required.

### Steps to Reproduce

1. tofu init
2. tofu plan

### Additional Context

_No response_

### References

_No response_

cam72cam:
This is the first part of #875 and fixes the main issue reported.  The schema validation for backends has two paths, one from data blocks, the other from backend blocks.  At the moment, backend blocks pre-populates empty maps as shown in the issue.  This will be fixed in a followup PR.

This PR fixes the schema optional fields, adds a test for the minimum set of fields required, and also improves the user facing errors when they are missing multiple fields (stable key sort).

Resolves #875

## Target Release

1.6.0

### DIFF
--- a/internal/configs/configschema/coerce_value.go
+++ b/internal/configs/configschema/coerce_value.go
@@ -5,6 +5,7 @@ package configschema
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/convert"
@@ -57,7 +58,15 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 
 	attrs := make(map[string]cty.Value)
 
-	for name, attrS := range b.Attributes {
+	// Stable sort keys for consistent error reports
+	attrKeys := make([]string, 0, len(b.Attributes))
+	for key, _ := range b.Attributes {
+		attrKeys = append(attrKeys, key)
+	}
+	sort.Strings(attrKeys)
+
+	for _, name := range attrKeys {
+		attrS := b.Attributes[name]
 		attrType := impliedType.AttributeType(name)
 		attrConvType := convType.AttributeType(name)
 
@@ -78,7 +87,15 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 		attrs[name] = val
 	}
 
-	for typeName, blockS := range b.BlockTypes {
+	// Stable sort keys for consistent error reports
+	typeKeys := make([]string, 0, len(b.BlockTypes))
+	for key, _ := range b.BlockTypes {
+		typeKeys = append(typeKeys, key)
+	}
+	sort.Strings(typeKeys)
+
+	for _, typeName := range typeKeys {
+		blockS := b.BlockTypes[typeName]
 		switch blockS.Nesting {
 
 		case NestingSingle, NestingGroup:
--- a/internal/configs/configschema/coerce_value.go
+++ b/internal/configs/configschema/coerce_value.go
@@ -60,7 +60,7 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 
 	// Stable sort keys for consistent error reports
 	attrKeys := make([]string, 0, len(b.Attributes))
-	for key, _ := range b.Attributes {
+	for key := range b.Attributes {
 		attrKeys = append(attrKeys, key)
 	}
 	sort.Strings(attrKeys)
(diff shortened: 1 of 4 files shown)


internal/configs/configschema/coerce_value.go
```diff
@@ -57,7 +58,15 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 
 	attrs := make(map[string]cty.Value)
 
-	for name, attrS := range b.Attributes {
+	// Stable sort keys for consistent error reports
```
cube2222, line 61: The second instance of this makes sense, but in this case no errors are possible, so I'd leave it out.

internal/configs/configschema/coerce_value.go
```diff
@@ -78,7 +87,15 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 		attrs[name] = val
 	}
 
-	for typeName, blockS := range b.BlockTypes {
+	// Stable sort keys for consistent error reports
```
cube2222, line 90: I'm kind of split on this.

On the one hand, consistent error messages make sense.

On the other hand, this will require an effort on our side in all such places in the codebase (I expect there to be a bunch) and I'm not sure the (very limited imo) value is worth the effort.

On the other other hand, we could also use Diagnostics here instead of failing on the first error. Basically, before the for loop create a diagnostics struct, and then append the errors to it, continuing the loop. This way we'll get all the errors. But if this is the approach we wanna take, I'd put in an issue and delay it past the stable release.

kislerdm: I see the value in collecting all error messages. It's especially useful when user validates the configuration using `tofu validate`.

internal/configs/configschema/coerce_value.go
cam72cam: I don't follow, attrs are the main error path in the linked ticket?

internal/configs/configschema/coerce_value.go
cam72cam: I think that wherever we are iterating through maps where there are user facing messages, we should try to sort if possible.  I see this as an ongoing cleanup when we encounter it.

As far as tfdialogs,  I think that's the right long term solution, but it might drastically baloon the scope of this bugfix.  I'll take a look into it this morning and give a rough estimate on how hard it would be.  Alternatively I could return a more complex error type as a stopgap.

cam72cam: It's definitely a non-trivial amount of work to introduce a complex error type / diags and make sure functionality is not broken.  I'm going to recommend we keep the sort and add a ticket to the backlog.

internal/configs/configschema/coerce_value.go
cube2222: my bad, github collapsed the relevant part

internal/configs/configschema/coerce_value.go
cam72cam: Created https://github.com/opentofu/opentofu/issues/881 to track the follow up work proposed here.

internal/configs/configschema/coerce_value.go
cam72cam: Spoke with @cube2222, simple misunderstanding.

internal/configs/configschema/coerce_value.go
cam72cam: The remaining question is should we back out the stable sort to be consistent or set precedence that we fix these when we encounter them, either with a stable sort or better use of tfdiags/complex errors

cube2222 (approved):

internal/configs/configschema/coerce_value.go
cube2222: > I think that wherever we are iterating through maps where there are user facing messages, we should try to sort if possible.

I don't care too much either way, so happy to do it the way you propose

kislerdm (approved):

PR #877, opened 2023-11-15
Draft: Don't allow nulls in required backend maps

cam72cam:
The tricky part of this PR is that previously nulls were allowed in required cty maps.  I believe that this functionality should not be allowed.  It led directly to #875 since the validation paths for data{} and backend{} are distinct.

The schema validation part of this PR should probably go in regardless of the null check as it validates that the current backend config matches the current schema.  It could be cherry picked out if the null check proves problematic.

After these changes the backend{} and data{} blocks function nearly identically, though the error message is slightly cryptic:
```
│ Error: Invalid backend configuration
│ 
│ The given configuration is not valid for backend "s3": attribute "endpoints" is required, null provided.
```
Notice the ", null provided".  It's probably good to have that in the error message, but it creates confusion because it defaults to null (see #875).  I think that this is probably good enough for now, but we probably want to remove the null defaults at some point as mentioned in #875 .

Related to #875

## Target Release

1.6.0

### DIFF
--- a/internal/configs/configschema/coerce_value.go
+++ b/internal/configs/configschema/coerce_value.go
@@ -71,6 +71,11 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 			return cty.UnknownVal(impliedType), path.NewErrorf("attribute %q is required", name)
 		}
 
+		// This might break backwards compatability
+		if !attrS.Computed && !attrS.Optional && val.IsNull() {
+			return val, path.NewErrorf("attribute %q is required, null provided", name)
+		}
+
 		val, err := convert.Convert(val, attrConvType)
 		if err != nil {
 			return cty.UnknownVal(impliedType), append(path, cty.GetAttrStep{Name: name}).NewError(err)
--- a/internal/command/meta_backend.go
+++ b/internal/command/meta_backend.go
@@ -1399,6 +1399,19 @@ func (m *Meta) backendInitFromConfig(c *configs.Backend) (backend.Backend, cty.V
 		return nil, cty.NilVal, diags
 	}
 
+	// Try to coerce the provided value into the desired configuration type.
+	configVal, err := schema.CoerceValue(configVal)
+	if err != nil {
+		diags = diags.Append(tfdiags.AttributeValue(
+			tfdiags.Error,
+			"Invalid backend configuration",
+			fmt.Sprintf("The given configuration is not valid for backend %q: %s.", c.Type,
+				tfdiags.FormatError(err)),
+			cty.Path(nil).GetAttr("backend"),
+		))
+		return nil, cty.NilVal, diags
+	}
+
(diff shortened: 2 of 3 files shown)


internal/configs/configschema/coerce_value.go
```diff
@@ -71,6 +71,11 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 			return cty.UnknownVal(impliedType), path.NewErrorf("attribute %q is required", name)
 		}
 
+		// This might break backwards compatability
+		if !attrS.Computed && !attrS.Optional && val.IsNull() {
```
kislerdm, line 75: **(question)** What if a required attribute is allowed to be `nil`/`null`? 

IMO there are two things separate validations:
- Attribute is provided, or not
- Provided attribute is nil/null, or not

I feel that nilabilitiy validation shall be at the level of backend configuration instead.

**(nitpick)** AFAIK `!attrS.Computed && !attrS.Optional` is equivalent to `attrS.Required`. The taler is easier to understand when reading IMO.

cam72cam: Unfortunately, due to the confusion of Optional vs Required, we can't rely on Required to be set correctly.  It's a very tricky situation.

cam72cam: Part of the problem is that the backend configuration sets all fields to null by default, so this bypasses validation for the most common cases.  Since that is not easy to fix, I'm curious if there is a real use case for allowing null for a required field.  To me, that seems invalid.

kislerdm: The proposal when we shall not expect users setting attributes to `null` makes absolute sense. An illustration:
```terraform
terraform {
  cloud {
    organization = "<MY_CLOUD_ORGANIZATION>"
    hostname     = "<MY_CLOUD_BACKEND_HOSTNAME>"

    workspaces {
      name = "foo"
      tags = null
    }
  }
}
```
will result in the error `Invalid workspaces configuration: Only one of workspace "tags" or "name" is allowed.`.

In other words, this thread can be resolved :) Thank you!

internal/configs/configschema/coerce_value.go
```diff
@@ -71,6 +71,11 @@ func (b *Block) coerceValue(in cty.Value, path cty.Path) (cty.Value, error) {
 			return cty.UnknownVal(impliedType), path.NewErrorf("attribute %q is required", name)
 		}
 
+		// This might break backwards compatability
```
Yantrio, line 74: We should mention it in the changelog if it breaks backwards compat

cube2222:
My main worry is whether we shouldn't delay this until after the stable release, as I can see a change like this having unexpected consequences.

cam72cam:
Closing as I'm pretty sure this was due to a misunderstanding early on in my exploring of the codebase.

Issue #182, opened 2023-08-25
Clarify `tofu test -test-directory -filter` filter behavior in documentation.


eranelbaz:
### OpenTF Configuration Files

```hcl
...opentf config...
```

### Expected Behavior

to read and filter relevant files from the `test-directory`

### Actual Behavior

Running from the root folder of this repo, when using `-filter` flag it reads no files

```
opentf test -test-directory=/internal/command/testdata/test/multiple_files -filter=one.tftest.hcl                                                                     

Success! 0 passed, 0 failed.
```

eranelbaz:
So we have two ways to tackle that issue
1. alter `--help` that explains `filter` only filter from the current directory
2. change the `filter` flag behavior to filter the files can be read by using `--test-directory`

WDYT? would like to have some more ideas

cipherboy:
My 2c. but given the commentary under the `test-directory` help text flag, I'd say the behavior is correct if confusing:

```
  -test-directory=path  Set the OpenTofu test directory, defaults to "tests". When set, the
                        test command will search for test files [[[[in the current directory and]]]]
                        in the one specified by the flag.
```

(`[[[[emphasis]]]]` mine)

In particular, `-test-directory` is not the same as e.g., a `tar -C` option, and really means _this+that_... `-filter` behaving as it does thus allows disambiguating `$PWD/file` vs `$TESTDIR/file`, seemingly following the same ordering as `-test-directory`.

eranelbaz:
@cipherboy I didn't manage to follow, but the issue isn't the `-test-directory` but the `-filter` which filter only from the test directory

serialseb:
Just an experience report on current usage:
I looked back into my scripts, and I used `-filter` with the full path to the exact test I was currently executing, without using `-test-directory`, so it makes sense to me as you may want, say from an IDE, test a subset of tests. I also absolutely don't remember finding this complicated to figure out, in the absence of `-test-directory` that I haven't remembered ever using or discovering, but this may be due to the documentation at the time being non-existent.

cipherboy:
[...]

PR #886, opened 2023-11-16
add test command definition

### LINKED ISSUE #292: Testing feature documentation
Write technical documentation about how testing feature works

kislerdm:
Resolves #292 

## Target Release

1.6.0

## What changed

- Added content of the page describing `tofu test` command
- Removed outdated document about the modules testing experiment

## Why do we need it

To have a clear description of the feature shared with the users.

## Followup

@Yantrio and I discussed that it'd make sense to make a blog post, or a dedicate tutorial page with testing examples which would illustrate a real case to share "best practices" and discuss rationale of why one would need to mock/overwrite variables etc.

TODO

- [x] Validate if `variables` defined in one *.tftest.hcl file are shared between all runs defined in all test files.
- [x] Create a followup issue to enhance the schemas relevant for the `tofu test`, [details](https://github.com/opentofu/opentofu/blob/c5170df91a6d4e511c0742ea6baefa97232b6448/internal/configs/test_file.go#L524-L576).


website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+sequentially run the test blocks to perform conditional checks and assertions against attributes of provisioned resources.
+Upon tests completion, OpenTofu will attempt to clean up created resources.
+
+_Note_: monitor the test output for successful cleanup.
+
+## General Options
+
+* `-test-directory=path` Set the test directory (default: "tests"). OpenTofu will search for test files in the set directory
+and in the directly of the test command execution.
+*Note* that the path shall be relative to the directory where `tofu test` is called.
```
Yantrio, line 29:
```suggestion
* `-test-directory=path` Set the test directory (default: "tests"). OpenTofu will search for test files in the set directory
as well as in the directly in which `tofu test` was executed.
*Note* that the path should be relative to the directory in which `tofu test` is called.
```

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+
+* `-filter=testfile` Set the test file *path* within the `test-directory`. The option can be used multiple times to specify several files.
+*Note* that the path shall be relative to the directory where `tofu test` is called.
```
Yantrio, line 32:
```suggestion
* `-filter=testfile` Set the test file *path* within the `test-directory`. The option can be used multiple times to specify several files.
*Note* that the path should be relative to the directory in which `tofu test` is called.
```

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+Usage: `tofu test [options]`
+
+The `test` command executes the automated integration tests defined in `.tftest.hcl` files which OpenTofu will read
+from the current configuration and testing directories.
+
+OpenTofu will create real infrastructure according to the definition in the tested configuration. Then, it will
+sequentially run the test blocks to perform conditional checks and assertions against attributes of provisioned resources.
+Upon tests completion, OpenTofu will attempt to clean up created resources.
+
+_Note_: monitor the test output for successful cleanup.
```
Yantrio, line 23: We use `*Note*` in some places, and `_Note_` in others, let's be consistent

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+
+## General Options
+
+* `-test-directory=path` Set the test directory (default: "tests"). OpenTofu will search for test files in the set directory
+and in the directly of the test command execution.
+*Note* that the path shall be relative to the directory where `tofu test` is called.
+
+* `-filter=testfile` Set the test file *path* within the `test-directory`. The option can be used multiple times to specify several files.
```
Yantrio, line 31:
```suggestion
* `-filter=testfile` Specify an individual test file *path* within the `test-directory` to use for testing. The option can be used multiple times to specify several files.
```

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+- modules/module_bar/foo.tftest.hcl,
+- modules/module_bar/main.tftest.hcl.
+
+**WHEN** `tofu test -test-directory=modules/module_bar -filter=bar.tftest.hcl` is executed in the project root directory,
+
+**THEN** the tests defined in the file `bar.tftest.hcl` will be run only.
+
+**WHEN** `tofu test -test-directory=modules/module_bar -filter=modules/module_bar/bar.tftest.hcl` is executed in the project root directory,
+
+**THEN** the tests defined in the file `modules/module_bar/bar.tftest.hcl` will be run only.
```
Yantrio, line 142: I understand we want to be comprehensive, but this is a very complex example, could we simplify this a bit more? I don't think we need foo/bar in every directory maybe?

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+
+## Definition syntax
+
+TBD
+
+## Variables
+
+Tests can be defined using variables specified in several ways.
+
+The variable values will be resolved according to the following order:
```
Yantrio, line 152:
```suggestion
Variable values will be resolved in the following order of precedence:
```

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+
+1. Environment variables with the prefix `TF_VAR_`;
+2. tfvar files specified in terraform.tfvars, *.auto.tfvars;
+3. Files specified using the flag `-var-file`;
+4. Variables defined using the flag `-var`;
+5. Variables from the `variable block` in test file;
+6. Variables from the variable block in run block.
+
+_Note_ that the variable's value found at the later step takes precedence over the value at the earlier step.
```
Yantrio, line 161: This can be dropped if we go with the suggestion above.

Yantrio (changes requested):
Overall I think it reads really well and explains the parameters nicely. Just a few tiny comments. Nice work! 👍

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+
+* `-json` Set for test outputs to be formatted as JSON.
+
+* `-no-color` Disable color codes in the command output.
+
+* `-verbose` Print the plan or state for each test run block as it executes.
+
+## Usage examples
+
+### Flat structure
```
RLRabinowitz, line 47: I feel like the docs now focus too much on the folder structure and about the combinations of `-filter` and `-test-directory`. It's probably enough to explain the interaction between them in a line or two, instead of creating a project structure mainly for that

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+Tests can be defined using variables specified in several ways.
+
+The variable values will be resolved according to the following order:
+
+1. Environment variables with the prefix `TF_VAR_`;
+2. tfvar files specified in terraform.tfvars, *.auto.tfvars;
+3. Files specified using the flag `-var-file`;
+4. Variables defined using the flag `-var`;
+5. Variables from the `variable block` in test file;
+6. Variables from the variable block in run block.
```
RLRabinowitz, line 159: I think that we should already have docs on variable precedence. Maybe we can link there instead

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,165 @@
...
+
+**WHEN** `tofu test -test-directory=modules/module_bar -filter=modules/module_bar/bar.tftest.hcl` is executed in the project root directory,
+
+**THEN** the tests defined in the file `modules/module_bar/bar.tftest.hcl` will be run only.
+
+## Definition syntax
+
+TBD
+
+## Variables
```
RLRabinowitz, line 148: Here I'd probably give an example of overriding variables in a test file / run block, and explain why one might want to do that

RLRabinowitz (commented):
Personally I'd try and incorporate real-world usages here. Some things I feel like we need to focus on in the docs

1. What is this feature for? What type of testing is it for, and why?
2. Real world usages - For example, overriding providers, running modules, expecting failures, etc.
3. Detailed explanations on what blocks are supported in a `tftest.hcl` file, and in a `run` block, with deep explanation of every field there

For example - explain provider overriding and why you'd need to do that

website/docs/cli/commands/test.mdx
kislerdm: Hey Arel! Thanks for your feedback. The docu is WIP, however I wanted to share enough examples following GWT pattern as a response to this [feedback](https://github.com/opentofu/opentofu/issues/182). IMO if something is not intuitive, it shall be clearly illustrated to facilitate users onboarding. I feel that it's also influenced by the experience level of working with the testing feature. I'm trying to write docu assuming a reader who tries the tool for the first time.

website/docs/cli/commands/test.mdx
kislerdm: Thanks for your feedback. I'll add a couple of illustrations/use cases. 

Could you please expand on this

> explain why one might want to do that

For context, I find that the main focus of this docu page shall be more practice oriented than explanation oriented (leaning more towards the top half of the [diagram](https://documentation.divio.com/) below).

image

 WDYT?

website/docs/cli/commands/test.mdx
kislerdm: Could you please confirm if you mean this [subsection](http://localhost:3000/docs/language/values/variables/#variable-definition-precedence)? Thanks!

website/docs/cli/commands/test.mdx
kislerdm: Thanks for suggestion! It was implemented in line with this [subsection](https://opentofu.org/docs/language/values/variables/#variable-definition-precedence).

website/docs/cli/commands/test.mdx
kislerdm: Thanks for idea - I adjusted the docu accordingly.

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,284 @@
...
+Usage: `tofu test [options]`
+
+The `test` command executes the automated integration tests defined in `.tftest.hcl` files which OpenTofu will read
+from the current configuration and testing directories.
+
+OpenTofu will create real infrastructure according to the definition in the tested configuration. Then, it will
+sequentially run the test blocks to perform conditional checks and assertions against attributes of provisioned resources.
+Upon tests completion, OpenTofu will attempt to clean up created resources.
+
+_Note_: monitor the test output for successful cleanup.
```
cube2222, line 23: Should this maybe be an info box instead of a Note?

kislerdm: Thanks for suggestion Kuba! Implemented.

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,297 @@
...
+
+* `-json` Set for test outputs to be formatted as JSON.
+
+* `-no-color` Disable color codes in the command output.
+
+* `-verbose` Print the plan or state for each test run block as it executes.
+
+## How to define a test suite
+
+The subsection describes how to filter test files when defining a test `suite`.
```
Yantrio, line 65:
```suggestion
In this subsection we will illustrate how to filter test files when defining a test `suite`.
```

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,297 @@
...
+├── bar.tf
+├── bar.tftest.hcl
+├── outputs.tf
+└── variables.tf
+```
+
+- When you execute `tofu test` in the project directory, the test `suite` will consist of the files:
+`bar.tftest.hcl`, `foo.tftest.hcl`, `main.tftest.hcl`.
+
+- To filter some files from the `suite`, use the `-filter` option.
```
Yantrio, line 88:
```suggestion
- To select some specific files from the `suite`, use the `-filter` option.
```

website/docs/cli/commands/test.mdx
```diff
@@ -1,12 +1,297 @@
...
+
+<!-- details: /internal/configs/test_file.go:239 decodeTestRunBlock -->
+
+The `run` block defines a _single test_. It specifies the operation which OpenTofu will execute,
+sets mocks for providers,
+and defines the checks to assess the state after the OpenTofu operation was executed.
+
+The following non-mutually exclusive blocks and attributes can be defined within the scope of the `run` block to configure the test.
+
+| Name              | Type   | Multiple allowed | Description                                                                                                                          |
```
Yantrio, line 157:
```suggestion
| Name              | Type   | Multiple Allowed | Description                                                                                                                          |
```

Yantrio (approved):

PR #897, opened 2023-11-20
Fixes #539: Setting both "Type" and "NestedType" removes other validation error messages

### LINKED ISSUE #539: Setting both "Type" and "NestedType" removes other validation error messages.
### OpenTofu Version

```shell
1.6.0
```
NOTE: To be honest, I have not installed OpenTofu. I report this only by reading the code.

N/A

I might add configuration files later if they are necessary.

### Debug Output

N/A

I might add debug outputs later if they are necessary.

### Expected Behavior

Validation error message "Type and NestedType cannot both be set" appears with other validation error messages.

### Actual Behavior

Validation error message "Type and NestedType cannot both be set" will remove other validation error messages.
https://github.com/opentofu/opentofu/blob/d2eacd93dc435df5eaab1b69c100ed8ca480b04f/internal/configs/configschema/internal_validate.go#L129-L132
(Seeing it carefully, original `err` looks to go away)

### Steps to Reproduce

To be honest, I have not installed OpenTofu. I report this only by reading the code.
I might investigate reproduction steps later if they are necessary.

### Additional Context

_No response_

### References

_No response_

janosdebugs:
Resolves #539

Note: the target file seems to be not gofmt'd. I left it as-is for now.

## Target Release

1.6.0

### DIFF
--- a/internal/configs/configschema/internal_validate_test.go
+++ b/internal/configs/configschema/internal_validate_test.go
@@ -134,32 +134,32 @@ func TestBlockInternalValidate(t *testing.T) {
 			},
 			[]string{"foo: either Type or NestedType must be defined"},
 		},
-        "attribute with both type and nestedtype": {
-            &Block{
-                Attributes: map[string]*Attribute{
-                    "foo": {
-                        // These properties are here to make sure other errors are also reported.
-                        Optional: true,
-                        Required: true,
-                        // Here's what we actually want to validate:
-                        Type: cty.String,
-                        NestedType: &Object{
-                            Nesting: NestingSingle,
-                            Attributes: map[string]*Attribute{
-                                "foo": {
-                                    Type:     cty.String,
-                                    Required: true,
-                                },
-                            },
-                        },
-                    },
-                },
-            },
-            []string{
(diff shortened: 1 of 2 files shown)


kislerdm (approved):
@janosdebugs Hey! Thank you for a prompt resolution, well done! 🚀

internal/configs/configschema/internal_validate_test.go
```diff
@@ -134,6 +134,32 @@ func TestBlockInternalValidate(t *testing.T) {
 			},
 			[]string{"foo: either Type or NestedType must be defined"},
 		},
+		"attribute with both type and nestedtype": {
```
Yantrio, line 137: tiny tiny tiny nitpick: Could we make the test name a little more verbose to describe what it's doing here?

Maybe something like `attribute with both type and nestedtype should not suppress other validation messages`

I know it doesn't really fit with the naming of the rest of these, but I feel like it would help us understand why we added this test in the future if we ever regress.

I would also be open to a comment that describes what you want to do for this individual test.

Yantrio (approved):

PR #901, opened 2023-11-20
Align CLI logic to be compliant with the stable registry module protocol

### LINKED ISSUE #900: Modify get module logic to be compliant with the OpenTofu registry protocol
## Problem

The OpenTofu registry "v1" protocol will be implemented for the stable OpenTofu registry:

_GIVEN_ that a client attempts to download the module `foo/terraform-bar-baz@v0.0.1`

_WHEN_ CLI calls the corresponding URL `https://registry.opentofu.org/v1/modules/foo/baz/bar/0.0.1/download` and the module is found in the registry

_THEN_ the following response is expected:

|Attribute|OpenTofu BW implementation|OpenTofu V1|
|:---|:---|:---|
|Status code|204|200|
|Body||`{"location": "git::https://github.com/foo/terraform-bar-baz?ref=v0.0.1"}`|
|Header: `X-Terraform-Get`|`git::https://github.com/foo/terraform-bar-baz?ref=v0.0.1`||

## Proposed solution

Adjust the OpenTofu CLI to be compatible with the OpenTofu registry "v1" protocol.

## Acceptance criteria

- [x] OpenTofu CLI is compatible with the "v1" protocol, i.e. it reads the module location from the response body.
- [x] OpenTofu CLI is BW compatible with other registry implementation, i.e. it reads the location from the response header in the absence of body.
- [x] The documentation is [adjusted](https://opentofu.org/docs/internals/module-registry-protocol/#sample-response-1).

kislerdm:
Resolves #900 

## Target Release

1.6.0

## What changed

- Added handling of the registry response in line with the server's [contract](https://github.com/opentofu/registry-stable/blob/c54f44a6b912bcb134c9727ab594af0be0e2e5bd/src/internal/v1api/responses.go#L5-L9).
- The http response status codes 200 and 204 are treated separately now: 200 - read location from the body, 204 - read location from the header.

## Why do we need it

For the OpenTofu CLI to be compliant with the OpenTofu registry "v1" protocol.

### DIFF
--- a/internal/registry/client_test.go
+++ b/internal/registry/client_test.go
@@ -5,6 +5,8 @@ package registry
 
 import (
 	"context"
+	"errors"
+	"io"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/registry/client_test.go
```diff
@@ -370,3 +372,122 @@ func TestLookupModuleNetworkError(t *testing.T) {
 		t.Fatal("unexpected error, got:", err)
 	}
 }
+
+func TestModuleLocation_readRegistryResponse(t *testing.T) {
+	server := test.Registry()
+	defer server.Close()
+
+	cases := map[string]struct {
```
janosdebugs, line 380: None of these test cases seem to test the X-Terraform-Get header. Is that tested elsewhere?

internal/registry/client_test.go
```diff
@@ -370,3 +372,122 @@ func TestLookupModuleNetworkError(t *testing.T) {
...
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if err != nil && err.Error() != tc.wantErrorStr {
+				t.Fatalf("unexpected error content: want=%s, got=%v", tc.wantErrorStr, err)
+			}
+		})
+	}
+}
+
+type mockRoundTripper struct {
```
janosdebugs, line 471: This looks like a test utility that could be unified somewhere maybe?

internal/registry/test/mock_registry.go
```diff
@@ -167,9 +172,22 @@ func mockRegHandler() http.Handler {
 			location = fmt.Sprintf("file://%s/%s", wd, location)
 		}
 
-		w.Header().Set("X-Terraform-Get", location)
-		w.WriteHeader(http.StatusNoContent)
-		// no body
+		switch strings.HasPrefix(matches[0], "alpha/") {
```
janosdebugs, line 175: The `alpha` identifier here is repeated at least twice across the code and I assume it goes away at some point, which will break the tests. Maybe there should be a flag for it?

janosdebugs (approved):
Looks good to me, although I can't verify the registry protocol itself yet. A few minor notes.

internal/registry/client_test.go
kislerdm: Hey János, thanks for the feedback. [This case](https://github.com/opentofu/opentofu/blob/6967669d7939c5781809bcbdd8d15320bd2d5c1f/internal/registry/client_test.go#L388-L390) does cover the alpha registry protocol, i.e. the location read from the header.

internal/registry/client_test.go
kislerdm: I feel it's a matter of taste. My preference is to colocate self-made mocks/stubs next to the tests they are dedicated to. Overall, I think that everyone on the @opentofu/core-engineers team shares the feeling refactoring to the codebase would be beneficial after the stable release. I'd scope tests reworking as part of that activity. WDYT?

internal/registry/test/mock_registry.go
kislerdm: You are right the `alpha` is an indicator of the alpha registry protocol to mock the response accordingly. I decided not to make over-invasive changes/optimisations to prevent blocking registry delivery. Please feel free to raise your ideas about refactoring - we will definitely come back to them after the stable release. Thanks!

Yantrio (changes requested):
Overall I like this approach. I agree with @janosdebugs 's comments too.

Overall though I think we should get rid of the usages of "alpha" and "stable" registry because in the future this will make no sense.

I would be tempted to not even mention old/new legacy/current alpha/stable here, and just mention that we have an order of precedence that we are grabbing from the body first, and then checking the header. (Same for the PR title/description)

kislerdm:
@Yantrio Hey James! Thanks for the feedback. Did I get it right that we intend to maintain both protocols, i.e. body+header moving forward? I'd assume the motivation here is BW compatibility to allow users to reuse their custom registries which they used with terraform before, right? Thanks!

internal/registry/client_test.go
janosdebugs: I'm more concerned with touching the existing logic despite there being no test cases for it. Typically, a stabilization test would make sure that the existing (legacy) functionality doesn't break.

Yantrio:
@kislerdm yes, we do intend to support both. Mainly because only the opentofu registry is going to be returning the download in the body, whilst private registries will be returning it in a header.

My comment is mainly just around naming of "alpha"/"stable". I don't want to treat it as old + new, but instead we should treat it as 2 different approaches that we will support.

kislerdm:
@Yantrio thanks for clarification! 👍🏻

internal/registry/client_test.go
kislerdm: For context, the test cases cover the status quo behaviour as well as the one in line with the OpenTofu v1 protocol (location in the response body).

internal/registry/test/mock_registry.go
kislerdm: Thanks for the suggestion - I implemented flags as variadic arguments to the mock registry initialiser :)

janosdebugs (approved):

website/docs/internals/module-registry-protocol.mdx
```diff
@@ -191,25 +191,37 @@ This endpoint downloads the specified version of a module for a single target sy
...
+{"location": "git::https://github.com/foo/terraform-baz-bar?ref=v0.0.1"}
+```
+
+In the absence of a response body, OpenTofu will use the `X-Terraform-Get` header as the module location:
+
+```text
+HTTP/2 204 No Content
 Content-Length: 0
-X-Terraform-Get: https://api.github.com/repos/hashicorp/terraform-aws-consul/tarball/v0.0.1//*?archive=tar.gz
+X-Terraform-Get: git::https://github.com/foo/terraform-baz-bar?ref=v0.0.1
```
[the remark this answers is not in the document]
kislerdm: Thanks for the question. Yes, it is intentional. 

For example, you can call the registry:

```commandline
curl -i 'https://registry.opentofu.org/v1/modules/hashicorp/consul/aws/0.0.1/download'
```

to receive an output like:

```commandline
HTTP/2 204 
date: Tue, 21 Nov 2023 17:01:20 GMT
x-terraform-get: git::https://github.com/hashicorp/terraform-aws-consul?ref=v0.0.1
<aws-headers-are-removed>
```

website/docs/internals/module-registry-protocol.mdx
```diff
@@ -191,25 +191,37 @@ This endpoint downloads the specified version of a module for a single target sy
...
+A successful response contains the location from which the module version's source can be downloaded.
+
+It is expected to be found in the JSON encoded body as the value for the key `location`:
+
 ```text
-HTTP/1.1 204 No Content
+HTTP/2 200
+Content-Length: 81
+
+{"location": "git::https://github.com/foo/terraform-baz-bar?ref=v0.0.1"}
```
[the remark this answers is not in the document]
kislerdm: See the response [here](https://github.com/opentofu/opentofu/pull/901#discussion_r1400907935).

website/docs/internals/module-registry-protocol.mdx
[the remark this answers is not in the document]
kislerdm: You can also check this [test case](https://github.com/opentofu/opentofu/blob/b2acc5c1bf89b0ec18f1e79c8a2b1a900efc0d57/internal/addrs/module_source_test.go#L160-L166) as an illustration of how such address will be prepared to be used to fetch the module from the storage (GitHub). 

The whole execution path involving the URI:
1. Call the registry to obtain the location of the module's version: [internal/initwd/module_install.go:583](https://github.com/opentofu/opentofu/blob/b2acc5c1bf89b0ec18f1e79c8a2b1a900efc0d57/internal/initwd/module_install.go#L583)
2. Parse the address: [internal/initwd/module_install.go:593](https://github.com/opentofu/opentofu/blob/b2acc5c1bf89b0ec18f1e79c8a2b1a900efc0d57/internal/initwd/module_install.go#L593)
3. Fetch the module and store to the installation path: [internal/initwd/module_install.go:623](https://github.com/opentofu/opentofu/blob/b2acc5c1bf89b0ec18f1e79c8a2b1a900efc0d57/internal/initwd/module_install.go#L623)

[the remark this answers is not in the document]
janosdebugs: Thanks, I just wanted to make sure, it looked weird.

website/docs/internals/module-registry-protocol.mdx
```diff
@@ -191,25 +191,37 @@ This endpoint downloads the specified version of a module for a single target sy
...
 ```
 
-A successful response has no body, and includes the location from which the
-module version's source can be downloaded in the `X-Terraform-Get` header.
-The value of this header accepts the same values as the `source` argument
-in a `module` block in OpenTofu configuration, as described in
-[Module Sources](/docs/language/modules/sources),
-except that it may not recursively refer to another module registry address.
+:::warning
+OpenTofu will read the response body content if both, the body and the `X-Terraform-Get` header, are received from the registry server.
```
Yantrio, line 219:
```suggestion
OpenTofu will prioritize reading the response body content if both, the body and the `X-Terraform-Get` header, are received from the registry server.
```

kislerdm: Thank you! Implemented.

Yantrio (approved):

Issue #745, opened 2023-10-18
Automate running e2e tests requiring external credentials.


cube2222:
There are some tests in the OpenTofu codebase, most notable the remote state backend tests, that require external services to run.

Those are only executed if the TF_ACC environment variable is set. An example of those tests is the `internal/backend/remote-state/s3/backend_test.go` file. Right now this doesn't happen in our CI/CD.

We should set up actions that will run each of these test suites with the credentials required, and set up any accounts required for that.

This issue is partially a blocker for the stable release. That is, we shouldn't make a stable release until we verify that these tests pass, even if we run them outside of CI/CD for now.

Here's the relevant subtasks for now:

- [ ] Azure
- [ ] Consul (we can deploy this ephemerally in a GitHub Action)
- [ ] Tencent Cloud
- [ ] Google Cloud
- [ ] Kubernetes (maybe we could deploy this ephemerally in a GitHub Action)
- [ ] Alibaba Cloud
- [ ] Postgres (we can deploy this ephemerally in a GitHub Action)
- [ ] AWS S3

Jake-Mok-Nelson:
I'm guessing that due to credentials this will be for core devs only. If you need help with this one anyway though, I have CI/CD and cloud experience.

Jake-Mok-Nelson:
Also, totally agree with not running this action by default on PRs. That would be a huge security issue.
Perhaps a runnable action like you suggest or perhaps use a PR comment (issue_comment) with some filters on permission/role/group.  [There's an action to help with this](https://github.com/marketplace/actions/command-action). `/test-e2e`

cube2222:
Thanks for the offer to help @Jake-Mok-Nelson!

As you say, most of this is actually just creating the accounts and setting up the credentials, which is core team only.

However, I'll let you know if there's any way you can contribute here!

cube2222:
Alibaba Cloud and Tencent Cloud were previously externally maintained. We won't be spinning up accounts for those.

In practice, there shouldn't be any issues with them (we haven't changed anything notable), and feel free to reach out if you hit any.

If anybody wants to help with running those, that's welcome too.

Everything else has been successfully tested, and either documented or automated via the Makefile.

PR #924, opened 2023-11-24
Add Azure integration test runbook.

cube2222:
Part of #745

Due to the number of steps involved and what they do behind the scenes, the guide is in form of a runbook. We may want to fully automate it eventually.

One test is removed with a relevant commit message. I don't think it's worth spending more time on it.

### DIFF
--- a/internal/backend/remote-state/azure/backend_test.go
+++ b/internal/backend/remote-state/azure/backend_test.go
@@ -126,38 +126,6 @@ func TestAccBackendOIDCBasic(t *testing.T) {
 	backend.TestBackendStates(t, b)
 }
 
-func TestAccBackendAzureADAuthBasic(t *testing.T) {
-	// cube2222: This doesn't seem to be a recommended mode of authentication and I have found no guide on how to configure it.
-	//           I've decided to skip this test in the suite, rather than spending more time to make it work.
-	t.Skip()
-
-	testAccAzureBackend(t)
-	rs := acctest.RandString(4)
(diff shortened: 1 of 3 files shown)


internal/backend/remote-state/azure/backend_test.go
```diff
@@ -127,6 +127,10 @@ func TestAccBackendOIDCBasic(t *testing.T) {
 }
 
 func TestAccBackendAzureADAuthBasic(t *testing.T) {
+	// cube2222: This doesn't seem to be a recommended mode of authentication and I have found no guide on how to configure it.
+	//           I've decided to skip this test in the suite, rather than spending more time to make it work.
+	t.Skip()
```
janosdebugs, line 132: Shouldn't this test be removed entirely rather than permaskipped? If it's not recommended, we could just not support it. If someone wants it, we can add it back later.

janosdebugs (approved):
One question, other than that it looks good.

internal/backend/remote-state/azure/backend_test.go
cube2222: I thought that this might be a better indication of intent, but I'm happy with removing it completely, too.

cube2222: I'll mention it in the commit message.

cube2222: Done.

Yantrio (approved):

kislerdm (approved):

Issue #930, opened 2023-11-28
[FAILED] Failed to run aws-provider-4.64.0 with openTofu


zhenrong-wang:
This is a follow-up test, previous test with aliyun-provider, please check this issue: https://github.com/opentofu/opentofu/issues/929

I followed the same method and tested aws just now. It turns out the aws provider failed to start, and caused my Ubuntu system halt.

**Versions: openTofu (1.6.0 alpha5), terraform (1.6.2), aws-provider (4.64.0)**

**Results**

**A. [hpcopr](https://github.com/zhenrong-wang/hpc-now) with terraform:**

Creating a stack cost 123 seconds, while destroying it took 230 seconds. 

Somehow the trace log of terraform is very large (35 MB). So I zipped them, please check this out: 
[terraform-trace-creating-aws.zip](https://github.com/opentofu/opentofu/files/13482390/terraform-trace-creating-aws.zip)

**B. [hpcopr](https://github.com/zhenrong-wang/hpc-now) with openTofu:**

The provider started communicating with AWS API after 230 seconds of the `apply` command. And finally caused the Ubuntu system stop responding, and I had to restart the system by force (I used VMwarkstation, and the Ubuntu vm has 2 cores + 4 GB Memory). 

Check the screenshot below:

![屏幕截图 2023-11-28 092924](https://github.com/opentofu/opentofu/assets/59274212/1730909f-d47c-4c12-b87b-f726ac17ae3b)

After rebooting my system, I exported the failed-log, please check this out:

[tofu-trace-aws-creating-FAILED.zip](https://github.com/opentofu/opentofu/files/13482455/tofu-trace-aws-creating-FAILED.zip)

**Some interesting behaviors:**

a. There are too many provider processes, see the window at the right-bottom.
b. The trace log shows that the provider didn't work in a continuous way. It exited and restarted frequently during the `apply` process.

During the test, nothing but the tf binary changed. Because my HCL resource files were automatically-generated by the [hpcopr](https://github.com/zhenrong-wang/hpc-now) , so the target resource stack should be the same.

Could you please help on this issue? Thanks!

zhenrong-wang:
Follow-up:

I also tried to run **tofu plan** command without [hpcopr](https://github.com/zhenrong-wang/hpc-now), and it turns out the provider and tofu failed to output plan to the terminal in the first 3 minutes.Check the screenshot below:

[...]

Issue #929, opened 2023-11-27
The provider is slow to creating/destroying resources


zhenrong-wang:
Hi opentofu developers,

I am switching my workload from terraform to opentofu. When I use 1.6.0-alpha5 and aliyun cloud provider 1.213.0, the provider is really slow to orchestrate the resources.

I opened the DEBUG mode, and it turns out the following message is streaming out continously.

Is there anything wrong with the provider or openTofu ? Thanks a lot!

```
2023-11-27T19:09:42.665+0800 [DEBUG] provider: plugin process exited: path=.terraform/providers/registry.opentofu.org/aliyun/alicloud/1.213.0/linux_amd64/terraform-provider-alicloud_v1.213.0 pid=30472
2023-11-27T19:09:42.665+0800 [DEBUG] provider: plugin exited
2023-11-27T19:09:42.665+0800 [DEBUG] created provider logger: level=debug
2023-11-27T19:09:42.665+0800 [INFO]  provider: configuring client automatic mTLS
2023-11-27T19:09:42.675+0800 [DEBUG] provider: starting plugin: path=.terraform/providers/registry.opentofu.org/aliyun/alicloud/1.213.0/linux_amd64/terraform-provider-alicloud_v1.213.0 args=[".terraform/providers/registry.opentofu.org/aliyun/alicloud/1.213.0/linux_amd64/terraform-provider-alicloud_v1.213.0"]
2023-11-27T19:09:42.676+0800 [DEBUG] provider: plugin started: path=.terraform/providers/registry.opentofu.org/aliyun/alicloud/1.213.0/linux_amd64/terraform-provider-alicloud_v1.213.0 pid=30479
2023-11-27T19:09:42.676+0800 [DEBUG] provider: waiting for RPC address: path=.terraform/providers/registry.opentofu.org/aliyun/alicloud/1.213.0/linux_amd64/terraform-provider-alicloud_v1.213.0
2023-11-27T19:09:42.762+0800 [INFO]  provider.terraform-provider-alicloud_v1.213.0: configuring server automatic mTLS: timestamp="2023-11-27T19:09:42.761+0800"
2023-11-27T19:09:42.793+0800 [DEBUG] provider.terraform-provider-alicloud_v1.213.0: plugin address: address=/tmp/plugin3192086639 network=unix timestamp="2023-11-27T19:09:42.793+0800"
2023-11-27T19:09:42.793+0800 [DEBUG] provider: using plugin: version=5
[...]

PR #954, opened 2023-12-01
Fix global schema caching

### LINKED ISSUE #944: The execution of tofu plan and tofu apply becomes stuck
### OpenTofu Version

```shell
OpenTofu v1.6.0-beta1
on linux_amd64
```

https://github.com/hashicorp/learn-terraform-provision-eks-cluster

### Debug Output

[plan.txt](https://github.com/opentofu/opentofu/files/13508290/plan.txt)

### Expected Behavior

[expected-results.txt](https://github.com/opentofu/opentofu/files/13508332/expected-results.txt)

### Actual Behavior

After execute `tofu plan` or `tofu apply`, it's getting stuck for a long time and don't return anything

### Steps to Reproduce

Follow the same instructions described in: https://developer.hashicorp.com/terraform/tutorials/kubernetes/eks

1. `aws configure` then fill the required fields
2. comment out the cloud block in the `terraform.tf` file 
3. `tofu init`
4. **OPTIONAL:** `export TF_LOG=trace` 
5. `tofu plan` or `tofu apply`

### Additional Context

_No response_

### References

_No response_

kislerdm:
Resolves #944, #929, #930 

## Target Release

1.6.0

## What changed

- Revert the global cache setting condition introduced in #719 
- `GetProviderSchema`: add the condition to validate if schema is required to be read after provider's instantiation

## Why do we need it

- To fix global caching flow

### DIFF
--- a/internal/providers/provider.go
+++ b/internal/providers/provider.go
@@ -124,7 +124,7 @@ type ServerCapabilities struct {
 	// normally, and the caller can used a cached copy of the provider's
 	// schema.
 	// In other words, the providers for which GetProviderSchemaOptional is false
-	// require their schema to be read after EVERY initialisation to function normally.
+	// require their schema to be read after EVERY instantiation to function normally.
 	GetProviderSchemaOptional bool
 }
 


internal/providers/provider.go
```diff
@@ -123,6 +123,8 @@ type ServerCapabilities struct {
 	// provider does not require calling GetProviderSchema to operate
 	// normally, and the caller can used a cached copy of the provider's
 	// schema.
+	// In other words, the providers for which GetProviderSchemaOptional is false
+	// require their schema to be read after EVERY initialisation to function normally.
```
cube2222, line 127: initialisation => instantiation

PR #958, opened 2023-12-01
Improve comments around the global and local provider schema cache.

cube2222:


janosdebugs (approved):
I approve of more documentation.

internal/plugin/grpc_provider.go
```diff
@@ -142,13 +147,21 @@ func (p *GRPCProvider) GetProviderSchema() (resp providers.GetProviderSchemaResp
 		resp.ServerCapabilities.GetProviderSchemaOptional = protoResp.ServerCapabilities.GetProviderSchemaOptional
 	}
 
-	// set the global cache if we can
+	// Set the global provider cache so that future calls to this provider can use the cached value.
+	// Crucially, this doesn't look at GetProviderSchemaOptional, because the layers above could use this cache
+	// *without* creating an instance of this provider. And if there is no instance,
+	// then we don't need to set up anything (cause there is nothing to set up), so we need no call
+	// to the providers GetSchema rpc.
```
kislerdm, line 154:
```suggestion
        // Set the global provider cache so that future calls to this provider can use the cached value.
	// Crucially, this doesn't look at GetProviderSchemaOptional, because the provider was instantiated at this point already, and its schema can be stored to the cache for further use without the need to instantiate provider. 
```

internal/tofu/context_plugins.go
```diff
@@ -70,6 +70,11 @@ func (cp *contextPlugins) ProviderSchema(addr addrs.Provider) (providers.Provide
 	// This cache is only written by the provider client, and transparently
 	// used by GetProviderSchema, but we check it here because at this point we
 	// may be able to avoid spinning up the provider instance at all.
+	//
+	// It's worth noting that ServerCapabilities.GetProviderSchemaOptional is ignored here.
+	// That is because we're checking *prior* to the provider's instantiation.
+	// GetProviderSchemaOptional only says that *if we instantiate a provider*,
+	// then we need to run the get schema call at least once.
```
kislerdm, line 77: Sorry, but I don't get the point of this comment. AFAIU `contextPlugins.ProviderSchema` does not keep track of *prior* initialisation(s) for decision on whether `GetProviderSchemaOptional` value is relevant for cache. On that basis, I find this comment confusing here.

cube2222: @kislerdm The provider is getting instantiated a few lines later, and this is an early return.

So this is clarifying that it's ok to read this even if GetProviderSchemaOptional is false, because the provider hasn't yet been instantiated (so doesn't require setup) and we have the schema leftover from a previous provider instantiation (one that is possibly now no more running).

internal/plugin/grpc_provider.go
cube2222: I think you're suggesting that we don't look at GetProviderSchemaOptional because for future calls this will now be instantiated and already called at least once.

But that's not the case, as this instance of the provider could be stopped (and that actually does indeed happen) right after this call finishes. So if we created a new instance of the provider, it would again require initialization.

kislerdm: Ah, right. Thanks for clarification.

PR #994, opened 2023-12-11
WIP: Fixes #979: Replace convenience script

### LINKED ISSUE #979: Replace convenience script
As mentioned in #978, the convenience script currently installs the `packagecloud.io` URL despite being called from `packages.opentofu.org`. This is not ideal and we should bring the convenience script in line with the manual install instructions anyway, so here's an issue to track just that. :)

janosdebugs:
This PR adds a new convenience script that should work on all Linux/UNIX distributions, and even on Windows with Powershell.

Resolves #979

## Target Release

1.6.0

## Screenshot

image

### DIFF
--- a/install/brew.sh
+++ b/install/brew.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -ex
+
+apt-get update
+apt-get install -y curl git build-essential gcc procps curl file
+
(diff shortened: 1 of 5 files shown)


janosdebugs:
Closing this PR in favor of opentofu/get.opentofu.org#1

Issue #898, opened 2023-11-20
Swap `hashicorp`, or `terraform` CLI identification occurrences in the codebase with `OpenTofu`


kislerdm:
## Problem

The codebase contains several places where `tofu` is identified as `terraform`. Also, `app.hashicorp.io` is used as URL for tests, and URL.

Examples:
- [cmd/tofu/telemetry.go](https://github.com/opentofu/opentofu/blob/d077a33939ddbe07f8a71d75d8d61c127fee6cd1/cmd/tofu/telemetry.go#L65)
- [internal/backend/remote/testing.go](https://github.com/opentofu/opentofu/blob/d077a33939ddbe07f8a71d75d8d61c127fee6cd1/internal/backend/remote/testing.go#L73-L74)
- [internal/cloud/backend.go](https://github.com/opentofu/opentofu/blob/d077a33939ddbe07f8a71d75d8d61c127fee6cd1/internal/cloud/backend.go#L43-L46)
- [internal/registry/regsrc/friendly_host_test.go](https://github.com/opentofu/opentofu/blob/d077a33939ddbe07f8a71d75d8d61c127fee6cd1/internal/registry/regsrc/friendly_host_test.go#L36-L43)

**Note** please do no change occurrences mentioned in #530, for example:
- `terraform.tfstate.d`
- `terraform.tfstate`
- `X-Terraform-Get`
- Please do not change [internal/version/version.go](https://github.com/opentofu/opentofu/blob/d077a33939ddbe07f8a71d75d8d61c127fee6cd1/version/version.go#L54)

## Proposed solution

- Replace terraform occurrences with `OpenTofu`
- Fix the links to the OpenTofu documentation (see #891)

Yantrio:
We should also replace usages of things like this: https://github.com/opentofu/opentofu/blob/d077a33939ddbe07f8a71d75d8d61c127fee6cd1/internal/registry/regsrc/friendly_host_test.go#L36-L43

lplazas:
Is there any place you are collecting examples of this? I'd be happy to start working on this if there isn't anyone on the task yet.

janosdebugs:
To document what's going on, we rolled back this change because it had side effects for people referencing providers without a namespace (e.g. `aws` instead of `hashicorp/aws`).

cam72cam:
[...]

PR #997, opened 2023-12-12
Add gate for TFE tests that use the network

cam72cam:
We don't want opentofu to connect to app.terraform.io during testing, unless explicitly specified.

Validation was performed by unplugging my network cable and adding the skipIfTFENotEnabled(t) where nessesary.

Related to https://github.com/opentofu/opentofu/issues/898

## Target Release

1.6.0


RLRabinowitz (approved):

cam72cam:
Adding a quick note before merge: A lot of the cloud package uses dependencies that configure their own instances of cleanhttp behind the scenes.  It is possible to mock these out, but are out of scope for this initial bandaid.

PR #1009, opened 2023-12-13
Add a warning when multiple likely forks of a provider are detected

### LINKED ISSUE #1008: Warn when there are potentially conflicting providers between hashicorp and opentofu namespaces
As the opentofu namespace now exists, we should now warn users when they are using both `hashicorp/aws` and `opentofu/aws` as they will not be merged into the same set of provider configuration/version requirements.

This is a footgun that we should add a light warning to before 1.6.0

cam72cam:
When we attempted to change the default namespace from "hashicorp" to "opentofu" we realized that there are no guardrails for when multiple potential duplicates of a provider are specified.

Example:
main.tf requires company/bar with configuration A
module/child.tf requires company/bar with configuration B tofu init finds a single provider company/bar is represented with configuration A + B

company/bar is forked into alt/bar with some additional fixes

main.tf is updated to alt/bar, and module/child.tf is forgotten. tofu init finds two providers, alt/bar with configuration A and company/bar with configuration B

This can cause all sorts of havoc depending on what provider configuration lives where.  It is also nearly invisible to the end user, which is why we added a warning message.

It is possible that someone may deliberately configure two identically typed providers from different namespaces.  Therefore this is a warning for a potential foot-gun instead of an outright error.

image

Resolves #1008

## Target Release

1.6.0


RLRabinowitz (approved):

Yantrio:
I like that the warning is clear here, can you also update the CHANGELOG please?

cube2222 (approved):
I'm not sure if we want that for namespaces other than hashicorp and opentofu?

I suppose you could check in the provider registry if there are some name duplicates in different namespaces.

cam72cam:
@cube2222 I think so, there are definitely a lot of identical provider names in different namespaces.  Also, given the fork example above I think it's worth it.

Issue #1013, opened 2023-12-14
Tofu Test: A module referring a `data` source from main configuration - returns error in teardown


RLRabinowitz:
### OpenTofu Version

```shell
OpenTofu v1.6.0-dev
```

### OpenTofu Configuration Files

```hcl
# main.tf
resource "local_file" "example" {
  content  = var.file_content
  filename = var.file_name
}

variable "file_name" {
  type = string
}
variable "file_content" {
  type = string
}
```

```hcl
# verify/main.tf
data "local_file" "foo" {
  filename = var.file_name
}
variable "file_name" {
  type = string
}

```

```hcl
# tests/file_test.tftest.hcl
variables {
    file_name = "example.txt"
    file_content = "test"
}

run "test_example" {
}

run "verify" {
    module {
        source = "./verify"
    }

    assert {
        condition = data.local_file.foo.content == "test"
        error_message = "wrong file content"
    }
}
```

### Debug Output

```
tests/main.tftest.hcl... pass
  run "test_example"... pass
  run "verify"... pass
OpenTofu encountered an error destroying resources created while executing tests/main.tftest.hcl/verify.
╷
│ Error: Read local file data source error
│
│   with data.local_file.foo,
│   on verify/main.tf line 2, in data "local_file" "foo":
│    2: data "local_file" "foo" {
│
│ The file at given path cannot be read.
│
│ +Original Error: open example.txt: no such file or directory
╵

Success! 2 passed, 0 failed.
```

### Expected Behavior

The resources that the data source references should be destroyed before attempting to destroy the module's resources

### Actual Behavior

Main resources are destroyed first, and then other `run` blocks of modules are being destroyed, starting with the last `run`.
So, `tofu test` attempts to destroy the module's resources **after** the main configuration is destroyed

### Steps to Reproduce

1. `tofu init`
2. `tofu test`

### Additional Context

The expected order of teardown should be by run should be different. The states should be destroyed by `run` block order

If the runs where "Main configuration" -> "Module A" -> "Main configuration" -> "Module B" -> "Module A", the clean up should teardown the states in the following order: "Module B" -> "Module A" -> "Main configuration"

**Note** - This is technically not fullproof, as the modules might have conditional `data` source interdependencies, making the determination of `run` order for destroy hard to figure out. But this should work for most tofu configurations

**Another note** - Maybe it makes more sense to run the `destroy`s with `-refresh=false` always. At least the `data` sources would not be refetched during the teardown process

### References

[...]

PR #1043, opened 2023-12-20
Fix test command teardown order

Evi1Pumpkin:
## The problem
Today, the `tofu test` command creates resources and then tears them down. It first destroys the resources in the state of the "main" run, and then goes over the the states of any additional modules and destroys them in reverse to their run order.

So if the run blocks order is:
"Module A" -> "Main configuration" -> "Module B"
The cleanup order will be:
"Main configuration" -> "Module B" -> "Module A"
If "Module B" has a data resource referencing "Main configuration", we'll face an error on cleanup (as discussed [here](https://github.com/opentofu/opentofu/issues/1013)).

## The solution
Change the cleanup order. Don't destroy the "main" state first, instead destroy based on running order:
 "Module B"  -> "Main configuration" -> "Module A".

Resolves https://github.com/opentofu/opentofu/issues/1013

## Target Release

1.7.0


internal/command/test_test.go
```diff
@@ -780,58 +780,98 @@ can remove the provider configuration again.
 	}
 }
 
-func TestTest_NestedSetupModules(t *testing.T) {
-	td := t.TempDir()
-	testCopyDir(t, testFixturePath(path.Join("test", "with_nested_setup_modules")), td)
-	defer testChdir(t, td)()
+func TestTest_Modules(t *testing.T) {
```
Evi1Pumpkin, line 783: I merged a few similar tests together - `TestTest_OnlyExternalModules`, `TestTest_NestedSetupModules` and the new `with_verify_module`

cube2222: Do I understand correctly that this new test is a regression test that fails before the change?

cube2222 (approved):
LGTM!

internal/command/test_test.go
Evi1Pumpkin: Indeed 🙃

cam72cam (approved):

internal/command/testdata/test/with_verify_module/main.tftest.hcl
```diff
@@ -0,0 +1,23 @@
...
+
+run "test" {
+}
+
+run "verify" {
+  module {
+    source = "./verify"
+  }
+
+ assert {
```
janosdebugs, line 14: The formatting seems to be slightly off here?

janosdebugs (approved):

internal/command/testdata/test/with_verify_module/main.tftest.hcl
Evi1Pumpkin: You are sharp 👁️ 
Fixed https://github.com/opentofu/opentofu/pull/1043/commits/d3d39f2b3b743ffadfca90c8172d2a536f8363dc

Issue #1044, opened 2023-12-20
Provider acceptance test are slow, not using the Provider Schema cache


RLRabinowitz:
### OpenTofu Version

```shell
1.6.0-dev
```

### OpenTofu Configuration Files

Running acceptance tests for providers takes longer than in Terraform. It seems that the reason is that acceptance tests don't use the Provider Schema cache, causing the tests to run significantly slower

### Debug Output

.

### Expected Behavior

Running the `terraform-provider-null` provider using the "terraform" binary took me about 90s

### Actual Behavior

Doing the same with the tofu binary took me about 300s

### Steps to Reproduce

Run acceptance tests of a provider, like `terraform-provider-null`, with the following environment variables:
```
TF_ACC_TERRAFORM_PATH="/path/to/opentofu"
TF_ACC_PROVIDER_NAMESPACE="hashicorp"
TF_ACC_PROVIDER_HOST="registry.opentofu.org"
```

### Additional Context

Seems like unmanaged providers are not using the Provider Schema cache

RLRabinowitz:
Running some provider acceptance tests multiple times, I notice that the difference in execution time between Terraform and OpenTofu isn't that significant. 

However, in OpenTofu, we never get a `GRPCProvider: GetProviderSchema: serving from global schema cache` log. That means that the provider schema cache isn't being used at all in OpenTofu

PR #1054, opened 2023-12-26
Fix Global Schema Cache not working in provider acceptance tests

RLRabinowitz:
When running provider acceptance tests with the OpenTofu binary, the Global Schema Cache is never used. This causes the acceptance tests to take longer than they should

The reason for that is that unmanaged plugins don't have the `Addr` attribute, so the Global Schema Cache code never attempts to store or fetch the schema from cache

Resolves https://github.com/opentofu/opentofu/issues/1044

## Target Release

1.7.0

### DIFF
--- a/internal/command/meta_providers.go
+++ b/internal/command/meta_providers.go
@@ -376,22 +376,33 @@ func providerFactory(meta *providercache.CachedProvider) providers.Factory {
 			return nil, err
 		}
 
-		// store the client so that the plugin can kill the child process
 		protoVer := client.NegotiatedVersion()
-		switch protoVer {
-		case 5:
-			p := raw.(*tfplugin.GRPCProvider)
-			p.PluginClient = client
-			p.Addr = meta.Provider
-			return p, nil
-		case 6:
-			p := raw.(*tfplugin6.GRPCProvider)
-			p.PluginClient = client
-			p.Addr = meta.Provider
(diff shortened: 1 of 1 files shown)


internal/command/meta_providers.go
```diff
@@ -376,22 +376,33 @@ func providerFactory(meta *providercache.CachedProvider) providers.Factory {
...
-		case 6:
-			p := raw.(*tfplugin6.GRPCProvider)
-			p.PluginClient = client
-			p.Addr = meta.Provider
-			return p, nil
-		default:
-			panic("unsupported protocol version")
+		p, err := initializeProviderInstance(raw, protoVer, client, meta.Provider)
+		if err != nil {
+			panic(err)
```
cube2222, line 382: Could we use a sentinel error here? I'd prefer not to blanket panic on an error received from a function,

That is, introduce an errUnsuportedProtocolVersion global, and then we can return it from the function, and check for it here (and panic only for it).

I know right now it's the only error case, but someone could presumably add a non-panic error condition to the function and we could miss it here.

RLRabinowitz: Done in f37c60439e37c2a56a4547e460acbde35b0aee0b

cube2222 (approved):
LGTM!

cam72cam (approved):

Issue #1030, opened 2023-12-18
Implement State Encryption


cube2222:
Issue to track the implementation of #874 

## Subtasks

- [ ] #1063
  - [x] #1160
    - [x] #1161
    - [x] #1162
    - [x] #1163
    - [x] #1164
    - [x] #1165
  - [x] #1166
    - [x] #1167
    - [x] #1168
    - [x] #1169
      - [x] #1170
    - [ ] #1171
    - [x] #1172
    - [x] #1173
    - [x] #1174
- [ ] #1175

StephanHCB:
A Comment

Issue #874, opened 2023-11-14
RFC: Client Side State Encryption


StephanHCB:
## Summary

This feature adds the option to encrypt local state files, remote state, and plan files. Encryption is off-by-default.
Partial encryption, when enabled, only encrypts values marked as sensitive to protect credentials contained in the
state. Full encryption, when enabled, protects against any information disclosure from leaked state or plans.

## Problem Statement

OpenTofu state and plans contain lots of sensitive information.

The most obvious example are **credentials** such as primary access keys to storage, but even ignoring any credentials
state often includes a **full map of your network**, including every VM, kubernetes cluster, database, etc.
That is a treasure trove for an attacker who wishes to orient themselves in your private network.

Unlike runtime information processed by OpenTofu, which only lives in memory and is discarded when the run ends,
state and plans are persisted. In large installations, state is not (just) stored in local files because multiple
users need access to it. **Remote state backend** options include simple storage (such as storage accounts, various
databases, ...), meaning these storage options do not "understand" the state, but there are also **extended backends**,
which do wish to gain information from state. The **persistent nature** and (often) **cloud storage** of state increases
the risk of it falling into the wrong hands.

Large corporations and financial institutions have compliance requirements for storage of sensitive information.
One frequent requirement is **encryption at rest** using a **customer managed key**. This is exactly what this feature
provides, and if you use it intelligently, even the cloud provider storing your state will not have access to the
encryption key at all.

## User-facing description

OpenTofu masks sensitive values in its printed output, but those very same sensitive values are written to state:

```
example snippet from a statefile for an Azure storage account with the primary access key (of course not a real one)
```

Pay particular attention to the line listing the primary access key. The storage account listed here doesn't exist,
but if it did, the primary access key would give an attacker full access to all the data on the storage account.

### Getting Started

_Note: the exact format of the configuration is likely to change as we test out the implementation and figure out
the precise details. So don't rely too much on exact field names or format of the contents at this point in time._

With the feature this RFC is about, you could simply set an environment variable before running OpenTofu:

```
export TF_STATE_ENCRYPTION='{"backend":{"method":{"name":"full"},"key_provider":{"name":"passphrase","config":{"passphrase":"foobarbaz"}}}}'
```

For readability let's spell out the value of the environment variable even though you wouldn't normally set it like this:

```
export TF_STATE_ENCRYPTION='''{
  "backend": {
    "method": {
      "name": "full"
    },
    "key_provider": {
      "name": "passphrase",
[...]

PR #1063, opened 2024-01-02
[State Encryption] Implement configuration layer and top-level functions

StephanHCB:
This PR, when ready, will close the first subtask of #1030. 

Step 1 towards client-side state encryption
- configuration data structures 
- configuration handling
- external interface and top-level functions to allow parallelization of the development effort

Specification and approved RFC see #874 

Subtasks planning see Slack Channel and #1030.

part of #1030 

## Target Release

1.7.0


internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,90 @@
...
+var FallbackConfigEnvName = "TF_STATE_DECRYPTION_FALLBACK"
+
+type KeyProviderName string
+
+const (
+	KeyProviderPassphrase KeyProviderName = "passphrase" // derive key from config field "passphrase"
+	KeyProviderDirect     KeyProviderName = "direct"     // key is explicitly specified in config field "key"
+)
+
+type KeyProviderConfig struct {
```
janosdebugs, line 66: I wonder if adding a receiver function to formally validate the KeyProviderConfig would possibly prevent user confusion. Something along these lines:

```go
func (k KeyProviderConfig) Validate() error {
    // Validate the Name enum here.
    // ...
}
```

This function could also be used by a `UnmarshalJSON` function to ensure we are not unmarshalling invalid settings.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,90 @@
...
+//
+// Note that decryption will always try the relevant configuration specified in TF_STATE_ENCRYPTION first.
+// Only if decryption fails with that, it will try this configuration.
+//
+// Why is this useful?
+//   - key rotation (put the old key here until all state has been migrated)
+//   - decryption (leave TF_STATE_ENCRYPTION unset, but set this variable, and your state will be decrypted on next write)
+var FallbackConfigEnvName = "TF_STATE_DECRYPTION_FALLBACK"
+
+type KeyProviderName string
```
janosdebugs, line 59: Since this is an enum, it would be helpful to have a function that validates if a string actually complies with this enum.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,90 @@
...
+
+type EncryptionMethodName string
+
+const (
+	EncryptionMethodFull EncryptionMethodName = "full" // full state encryption
+)
+
+type EncryptionMethodConfig struct {
+	// Name specifies which encryption method to use.
+	Name EncryptionMethodName `json:"name"`
```
janosdebugs, line 84: Same validation concerns as above.

internal/states/encryption/flow/interface.go
```diff
@@ -0,0 +1,102 @@
+// Package flow contains the top-level flow for client-side state encryption and the interfaces an encryption
+// or key derivation method need to implement.
+package flow
+
+import "github.com/opentofu/opentofu/internal/states/encryption/encryptionconfig"
+
+type ConfigurationSource string
+
+const (
+	ConfigurationSourceEnvDefault ConfigurationSource = "env-default"
```
janosdebugs, line 10: Same enum validation concerns as above.

internal/states/encryption/flow/interface.go
```diff
@@ -0,0 +1,102 @@
...
+	// if the payload was not actually encrypted.
+	DecryptState(payload []byte) ([]byte, error)
+
+	// EncryptState encrypts plaintext state.
+	//
+	// There are special configurations that will not actually encrypt
+	// state. This happens when you only configure a decryption fallback,
+	// but not encryption. This is not an error.
+	//
+	// state is a json document passed in as a []byte.
```
janosdebugs, line 44: I'm debating with myself about this. Theoretically, someone could pass in any JSON to this function to undefined results. I wonder if the format that needs to be passed in should be explicitly documented here and then enforced. This may be especially important if future developers change the format of the state file or something along those lines as this may break the encryption.

I'm thinking about cases where new fields are added, etc.

internal/states/encryption/flow/mockup.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/opentofu/opentofu/internal/states/encryption/encryptionconfig"
+)
+
+// MockUpLoggingFlow will be removed when we replace it with the real implementation.
+type MockUpLoggingFlow struct {
+	configKey string
+	logger    hclog.Logger
+}
+
+func New(configKey string) Flow {
```
janosdebugs, line 15: To make sure nobody mistakes this for an actual production flow:

```suggestion
func NewMock(configKey string) Flow {
```

internal/states/encryption/instance.go
```diff
@@ -0,0 +1,50 @@
...
+//
+//	data "terraform_remote_state" "foo" {}
+//
+// For enumerated resources, the format is "terraform_remote_state.foo[17]" or
+// "terraform_remote_state.foo[key]" (no quotes around the for_each key).
+//
+// The first time a particular instance is requested, Instance may bail out due to invalid configuration.
+//
+// See also RemoteStateInstance, StatefileInstance, PlanfileInstance.
+func Instance(configKey string) flow.Flow {
```
janosdebugs, line 22: This function should maybe be called `NewInstance` to follow Go conventions?

internal/states/encryption/instance.go
```diff
@@ -0,0 +1,50 @@
...
+	// TODO validate "at least one .", bail out if wrong
+	return instance(configKey, true)
+}
+
+// RemoteStateInstance obtains the instance of the encryption flow that is intended for our own remote
+// state backend, as opposed to terraform_remote_state data sources.
+func RemoteStateInstance() flow.Flow {
```
janosdebugs, line 29: Same comment as above.

internal/states/encryption/flow/interface.go
StephanHCB: I am hoping we can keep the currently planned encryption methods (full, partial = all the string leaf nodes) agnostic of the rest of the json contents (unless someone introduces "encryption" at top level, which would cause be a collision).

I'll add documentation that the json can't have a key "encryption". That at least should cause an error.

internal/states/encryption/instance.go
StephanHCB: Once all TODO comments are implemented, it will only return a new instance if one wasn't found in the cache (for each configKey, the instance is a singleton so it can keep track of the configuration for this configKey).

NewOrCachedInstance? My train of thought was, it's "THE" instance for this configKey anyway.

internal/states/encryption/flow/interface.go
```diff
@@ -0,0 +1,102 @@
...
+	// encryption configuration. If this fails, it tries the decryption
+	// fallback configuration. If neither produces a valid result it fails.
+	//
+	// payload must be a json document passed in as a []byte.
+	// This can be encrypted or unencrypted state.
+	//
+	// If no error is returned, then the first return value will always
+	// be a json document, possibly the same one passed in as payload
+	// if the payload was not actually encrypted.
+	DecryptState(payload []byte) ([]byte, error)
```
Yantrio, line 36: Is there a reason we specifically wanted DecryptState AND DecryptPlan instead of just a generic Encrypt/Decrypt methods?

internal/states/encryption/instance.go
```diff
@@ -0,0 +1,50 @@
...
+// If the key is "terraform_remote_state.foo", the returned Flow is intended for
+//
+//	data "terraform_remote_state" "foo" {}
+//
+// For enumerated resources, the format is "terraform_remote_state.foo[17]" or
+// "terraform_remote_state.foo[key]" (no quotes around the for_each key).
+//
+// The first time a particular instance is requested, Instance may bail out due to invalid configuration.
+//
+// See also RemoteStateInstance, StatefileInstance, PlanfileInstance.
```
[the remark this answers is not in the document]
StephanHCB: No, I forgot the (), and Goland didn't notify me. Will fix.

internal/states/encryption/flow/interface.go
StephanHCB: I tried to lump these two together like I did in the PoC, but then I added the comments and they became really unclear.

I think this is probably an indication these should be separate functions on the outward facing interface.

The differences are (maybe need to be clearer about those in the comments):
- plaintext planfile is binary (zip format) while state is json
- planfile does not need a decryption fallback (not retained long term) while state does
- planfile can only use full encryption while state can use all encryption methods

Other than easier-to-read comments and differences in validation, the two implementations will share most of their functionality.

internal/states/encryption/instance.go
janosdebugs: I'm generally not a fan of side effects. Caching in this way, without an injected cache layer, would make testing very hard since the behavior would depend on the caching. Generally, I prefer my caching to be an overlay that implements the same function as the underlying implementation, but then again, I tend tk write Go in a fairly OOP way which may not be the most appropriate.

internal/states/encryption/flow/interface.go
janosdebugs: I wonder if that special key could be made configurable. That would make the API more explicit about what it does.

StephanHCB: Good idea. Makes it explicitly visible in the code rather than just the comments. I'll do that.

internal/states/encryption/instance.go
StephanHCB: Not 100% sure, but in this case I think it's not too bad, because the only thing being cached is the configuration, and given an instance you can override all parts of the configuration by calling `EncryptionConfiguration(...)` and `DecryptionFallbackConfiguration(...)` with the relevant values of `source` on it before calling the code to be tested. In this way, tests can also avoid having to set environment variables, unless parsing them is the specific thing being tested, which is part of the unit tests of package `encryptionconfig` and `encryption`.

The tradeoff is that we can avoid passing around `Flow` instances between where they'll likely be configured ...
- parsing the `terraform {}` block, see `internal/configs/parser_config.go`
- remote_state data source configuration, see `internal/builtin/providers/tf`

... and where they'll likely be used
- `internal/states/remote/state.go`
- `internal/states/statefile/(read.go,version4.go)`
- `internal/plans/planfile/(reader.go,writer.go)`

We will see if this tradeoff is worth it when we implement the callsites (subtasks 2 and dependent in #1030). If it turns out to be doable to store the `Flow` instances externally without touching large bodies of existing code, we can remove the caching layer.

More reason to keep it isolated and small.

AmerAlkaki (approved):

internal/states/encryption/instance.go
StephanHCB: I have renamed the Instance...(...) functions to "GetSingleton...(...)". I have also added a comment that explains the reason for having the singleton instance cache.

internal/states/encryption/instance.go
StephanHCB: same as above. Let's discuss above.

internal/states/encryption/instance.go
janosdebugs: I believe we sorted this out.

internal/states/encryption/environment.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ParseEnvironmentVariables checks the state encryption environment variables for structure and syntax errors.
+//
+// Call ParseEnvironmentVariables early during OpenTofu's initialization to ensure that the configuration from
+// environment variables can be correctly parsed.
+//
+// It is not an error if either of the environment variables has an empty value or is unset.
+//
+// If this function returns nil, GetSingleton(), GetRemoteStateSingleton(), GetStatefileSingleton(), and
+// GetPlanfileSingleton() are much less likely to fail at inopportune times.
```
Yantrio, line 17:
```suggestion
// If this function returns nil then the subsequent calls to GetSingleton(), GetRemoteStateSingleton(), GetStatefileSingleton(), and
// GetPlanfileSingleton() are much less likely to fail at inopportune times.
```

internal/states/encryption/example_test.go
```diff
@@ -0,0 +1,182 @@
...
+    }`
+	TF_STATE_DECRYPTION_FALLBACK := `{
+      "backend": {
+		"key_provider": {
+			"config": {
+				"passphrase": "the previous passphrase for key rotation"
+			}
+		}
+	  }
+    }`
```
Yantrio, line 39: tiny nitpick: this is a bit hard to review in github due to the mixing of tabs and spaces.

internal/states/encryption/instance.go
```diff
@@ -0,0 +1,119 @@
...
+//   - there is one instance for each remote state data source
+//
+// One way to configure state/plan encryption is adding settings to the terraform {} block
+// (or the terraform_remote_state block, respectively). These blocks are parsed by code far removed
+// from where they are used, and they are parsed multiple times by OpenTofu.
+//
+// Combine that with procedural code in internal/states/statefile (local state file) and
+// internal/plans (local plan file), which is called from all over the place.
+//
+// This is why caching singleton instances in this package is the less painful option.
```
Yantrio, line 26: Whilst I understand it's parsed multiple times and in multiple places and that's why you wanted to have the singleton, is it worth the technical debt of this layer to introduce a singleton/cache/etc instead of parsing it each time?

I would argue that we're maybe over complicating things here with this layer.

Yantrio (changes requested):
Thanks for submitting this. Overall the code looks solid, but I've put in a few comments about naming and questioning a couple things.

However, I'm not a huge fan of the instance/singletonecache approach here. I don't see it being needed and I feel like it makes the code less approachable. Maybe I'm just not understanding the need for this yet because I don't have the full context.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+//
+// There can be more than one configuration with different encryption or key derivation methods.
+type Config struct {
+	KeyProvider KeyProviderConfig `json:"key_provider"`
+
+	Method EncryptionMethodConfig `json:"method"`
+
+	// Required forces encryption operations to fail if they would
+	// result in an unencrypted output.
+	Required bool `json:"required"`
```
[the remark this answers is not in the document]
cam72cam: I'm also in favor of enforced

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+type ConfigEnvJsonStructure map[string]Config
+
+// ConfigEnvName is the name of the environment variable used to configure encryption and decryption
+//
+// Set this environment variable to a json representation of ConfigEnvJsonStructure, or leave it unset/blank
+// to disable encryption (default behaviour).
+//
+// Note: With rare exceptions, you should avoid setting the state encryption environment variables in tests,
+// as this may make tests depend on each other. See the comments on encryption.ParseEnvironmentVariables().
+var ConfigEnvName = "TF_STATE_ENCRYPTION"
```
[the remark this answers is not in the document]
janosdebugs: No strong opinion, `TF` covers Tofu just fine for me.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+//
+//	data "terraform_remote_state" "foo" {}
+//
+// For enumerated resources, the format is "terraform_remote_state.foo[17]" or "terraform_remote_state.foo[key]"
+// (no quotes around the for_each key).
+type ConfigEnvJsonStructure map[string]Config
+
+// ConfigEnvName is the name of the environment variable used to configure encryption and decryption
+//
+// Set this environment variable to a json representation of ConfigEnvJsonStructure, or leave it unset/blank
```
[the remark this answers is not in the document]
janosdebugs: If we switched to environment variables, this would look as follows:

```
TF_STATE_ENCRYPTION_ENFORCED=1
TF_STATE_ENCRYPTION_KEY_PROVIDER_NAME=bla
TF_STATE_ENCRYPTION_KEY_PROVIDER_SOMECONFIG=somevalue
TF_STATE_ENCRYPTION_METHOD_NAME=bla
TF_STATE_ENCRYPTION_METHOD_SOMECONFIG=somevalue
```

This is not even taking into account separate configuration for remote state data sources. Btw. JSON in env vars was decided in the RFC, if we change this we should update the RFC too.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+// FallbackConfigurationsFromEnv parses the decryption fallback configuration from the environment variable
+// configured in FallbackConfigEnvName.
+//
+// It is not an error if the environment variable is unset or empty, that just means no decryption fallback
+// configuration is provided via environment.
+func FallbackConfigurationsFromEnv() (ConfigEnvJsonStructure, error) {
+	return parseEnvJsonStructure(FallbackConfigEnvName, "fallback decryption configuration")
+}
+
+type KeyProviderName string
```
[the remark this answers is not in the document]
janosdebugs: `KeyProvider` will be the same name as an interface in a subsequent PR, this distinction seems important for readability.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+type KeyProviderConfig struct {
+	// Name specifies which key provider to use.
+	Name KeyProviderName `json:"name"`
+
+	// Config configures the key provider.
+	//
+	// The available values are key provider dependent.
+	Config map[string]string `json:"config"`
```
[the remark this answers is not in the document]
janosdebugs: Typically, no other types will be needed. Specifying an `interface{}` type will complicate JSON-unmarshaling and processing the configuration from .tf.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return nil
+}
+
+func parseJsonStructure(jsonValue string) (ConfigEnvJsonStructure, error) {
+	parsed := make(ConfigEnvJsonStructure)
+
+	decoder := json.NewDecoder(strings.NewReader(jsonValue))
+	decoder.DisallowUnknownFields()
```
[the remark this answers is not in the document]
janosdebugs: Yes, to avoid typos.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,213 @@
...
+	err := decoder.Decode(&parsed)
+	if err != nil {
+		logging.HCLogger().Trace("json parse error", "details", err.Error())
+		return nil, errors.New("json parse error, wrong structure, or unknown fields - details omitted for security reasons (may contain key related settings)")
```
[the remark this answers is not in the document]
janosdebugs: Removing this because the error may contain sensitive values and anything above trace would be irresponsible. The structure is fairly simple, so users should be able to work with a better error message (which we are adding).

internal/states/encryption/encryptionconfig/config.go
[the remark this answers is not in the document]
Yantrio: >Typically, no other types will be needed.

But are we certain of that? I suspect quite strongly that someone could want a boolean somewhere. (ie, going to vault and disabling certificate validation etc)

internal/states/encryption/encryptionconfig/config.go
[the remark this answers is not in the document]
Yantrio: That's fair enough

internal/states/encryption/encryptionflow/interface_example_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/opentofu/opentofu/internal/states/encryption/encryptionflow"
+)
+
+func ExampleFlow() {
+	var builder encryptionflow.FlowBuilder = &base64UnsafeFlowBuilder{}
+
+	flow, err := builder.Build()
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
janosdebugs: We removed this and there is a comprehensive example in example_test.go.

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/opentofu/opentofu/internal/states/encryption/encryptionconfig"
+	"github.com/opentofu/opentofu/internal/states/encryption/encryptionflow"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+	"sync"
+)
+
+// New creates a new Encryption object. You can use this object to feed in encryption configuration and then create
+// an encryptionflow.StateFlow or encryptionflow.PlanFlow as needed.
+//
+// Note: due to a lot of procedural code you probably want to use GetSingleton() to obtain the Encryption implementation
```
cube2222, line 15: I don't understand this note and what is meant with procedural.

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,263 @@
...
+		logger:                    logger,
+	}
+}
+
+// Encryption is the main interface to feed the encryption configuration and obtain the encryptionflow.Flow for running
+// the actual encryption.
+//
+// Note: due to a lot of procedural code you probably want to use GetSingleton() to obtain the Encryption implementation
+// that's globally scoped.
+type Encryption interface {
```
cube2222, line 31: Since there is only a single implementation, could you please clarify why this is an interface?

Is it the ability to mock?

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,263 @@
...
+	// ApplyEnvConfigurations applies a configuration coming from the operating system environment. The
+	// encryptionconfig.Config structures embedded in the parameters should have the source set to Environment in order
```
cube2222, line 33: It doesn't seem like the "source" is part of the encryptionconfig.Config structure anymore

cube2222: Yeah, seems like the method handles that now by itself.

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,270 @@
...
+	// for the priorities to be applies correctly.
+	ApplyEnvConfigurations(
+		encryption map[encryptionconfig.Key]encryptionconfig.Config,
+		decryptionFallback map[encryptionconfig.Key]encryptionconfig.Config,
+	) error
+
+	// ApplyHCLEncryptionConfiguration applies a single encryption configuration coming from HCL. The config should
+	// have its source set to HCL.
```
cube2222, line 42: Same, the source is handled by the method already.

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,270 @@
...
+	//   with the environment taking precedence. The default configuration should be ignored.
+	//
+	// Tip: encryptionconfig.ConfigMap.Merge implements these precedence rules.
+	PlanFile() (encryptionflow.PlanFlow, error)
+}
+
+type encryption struct {
+	encryptionConfigs         encryptionconfig.ConfigMap
+	decryptionFallbackConfigs encryptionconfig.ConfigMap
+	mutex                     *sync.Mutex
```
cube2222, line 128: Nit: as long as the encryption struct is passed by reference, there's no need to make this a pointer.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,239 @@
...
+				),
+			)
+		}
+		return nil
+	}
+}
+
+// IsRemoteDataSource returns true if the specified key is a valid remote data source key.
+func (k Key) IsRemoteDataSource() bool {
+	return strings.Contains(string(k), ".")
```
cube2222, line 173: I think it would make sense to check for inclusion of `terraform_remote_state.`.

I could see users setting it for wrong resource types otherwise due to some misunderstanding.

internal/states/encryption/singleton.go
```diff
@@ -0,0 +1,34 @@
...
+	defer lock.Unlock()
+	if instance == nil {
+		instance = New(logging.HCLogger())
+	}
+	return instance
+}
+
+// ClearSingleton wipes the singleton. You can use this function to reset the encryption state for tests that involve
+// procedural function calls. Remember, using the singleton makes the tests using it impossible to parallelize, so
+// consider using the New() function above to create and inject a separate instance of the Encryption interface.
```
cube2222, line 26: It's not longer "above" because it's in a different file.

internal/states/encryption/encryptionconfig/config.go
janosdebugs: I'm not sure this piece of code should be the arbiter of what is and isn't a remote state.

internal/states/encryption/encryption.go
janosdebugs: It's for the ability to mock and dependency inversion.

internal/states/encryption/encryption.go
janosdebugs: Tofu has a bunch of procedural code so the singleton serves as a global scope. I guess this could do with rewriting and better explanation.

internal/states/encryption/encryptionconfig/deepmerge.go
```diff
@@ -0,0 +1,97 @@
...
+
+func valOrDefault[T comparable](val T, empty T, defaultVal T) T {
+	if val == empty {
+		return defaultVal
+	} else {
+		return val
+	}
+}
+
+func mapMerge[K comparable, V any](addon map[K]V, base map[K]V) map[K]V {
```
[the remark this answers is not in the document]
janosdebugs: We checked, unfortunately we can't replace the entire function with it.

cube2222 (approved):
Reviewed it all, LGTM.

The code is straightforward, well-tested and the abstractions make sense to me.

Excellent job!

internal/states/encryption/encryptionconfig/config.go
cube2222: I mean, we could at least use a slightly better heuristic than just the period while we keep it here.

internal/states/encryption/encryption.go
cube2222: Yeah, a slight clarification could be of use here.

internal/states/encryption/encryptionconfig/config.go
janosdebugs: A regexp for resource names?

internal/states/encryption/singleton_test.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+	}()
+	defer ClearSingleton()
+
+	if instance == nil {
+		t.Fatalf("GetSingleton did not create the singleton, it was still nil")
+	}
+}
+
+/*
```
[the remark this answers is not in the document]
janosdebugs: Nope, this is old stuff, removing it.

internal/states/encryption/encryptionconfig/config.go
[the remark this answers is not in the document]
janosdebugs: I'm ok with booleans, strings, floats and ints, but I'm not happy allowing maps or lists in there, those are tricky to validate.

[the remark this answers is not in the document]
janosdebugs: @opentofu/core-engineers your input is needed here, please.

internal/states/encryption/encryptionconfig/config_key_provider.go
```diff
@@ -0,0 +1,35 @@
+package encryptionconfig
```
Evi1Pumpkin, line 1: I believe we are missing a test for this file?

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,242 @@
...
+						string(KeyDefaultRemote), string(KeyBackend), string(KeyPlanFile), string(KeyStateFile),
+					}, ", ",
+				),
+			)
+		}
+		return nil
+	}
+}
+
+var remoteDatasourceRe = regexp.MustCompile(`^[a-z0-9_]+\.[a-z0-9_]+(|\[(\d+|[a-z0-9_]+)])$`)
```
Evi1Pumpkin, line 172: I'm not very familiar with the terraform_remote_state data source, but doesn't it always start with the string `terraform_remote_state` (I saw we don't support `tofu_remote_state`)? So why are we searching for a regex that contains a dot instead of a regex that fits `terraform_remote_state.`?

Evi1Pumpkin (approved):
Amazing job! Very well-written and clear.

internal/states/encryption/encryptionconfig/config.go
janosdebugs: I don't know if currently or in the future there may be another way to access the remote state. However, even if there is only this one way and we don't want to change it, the validation of that fact these start with `terraform_remote_state` shouldn't be done in this package, but rather in a central place in the codebase. This may be something for a future refactor PR.

internal/states/encryption/encryptionconfig/config_key_provider.go
janosdebugs: This is tested from `config_test.go`. There is no explicit test for this file.

internal/states/encryption/encryptionconfig/config.go
cube2222: I would just propose `strings.Contains(string(k), "terraform_remote_state.")`.

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,271 @@
...
+		decryptionFallback map[encryptionconfig.Key]encryptionconfig.Config,
+	) error
+
+	// ApplyHCLEncryptionConfiguration applies a single encryption configuration coming from HCL.
+	ApplyHCLEncryptionConfiguration(key encryptionconfig.Key, config encryptionconfig.Config) error
+	// ApplyHCLDecryptionFallbackConfiguration applies a single decryption fallback configuration coming from HCL.
+	ApplyHCLDecryptionFallbackConfiguration(key encryptionconfig.Key, config encryptionconfig.Config) error
+
+	// Validate ensures that the previously-applied configuration is valid. You can call this function after you
+	// applied all configurations in order to verify that all configurations are valid.
```
Yantrio, line 49:
```suggestion
	// Validate ensures that the previously-applied configuration is valid. You should call this function after you
	// applied all configurations in order to verify that all configurations are valid.
```

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,271 @@
...
+	return e.build(encryptionconfig.KeyStateFile)
+}
+
+func (e *encryption) PlanFile() (encryptionflow.PlanFlow, error) {
+	return e.build(encryptionconfig.KeyPlanFile)
+}
+
+func (e *encryption) RemoteStateDatasource(configKey encryptionconfig.Key) (encryptionflow.StateFlow, error) {
+	if !configKey.IsRemoteDataSource() {
+		return nil, fmt.Errorf("the specified configuration key is not a valid remote data source key (this is likely a bug, did you want to call RemoteState(), StateFile(), or PlanFile()?)")
```
Yantrio, line 251: could this be a panic? There's a chance that if this is an implementation error that it shouldn't be ignorable.

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,242 @@
...
+			configOrNil(c, Meta{SourceEnv, configKey}),
+		)
+	} else {
+		merged = mergeConfigs(
+			nil,
+			configOrNil(c, Meta{SourceHCL, configKey}),
+			configOrNil(c, Meta{SourceEnv, configKey}),
+		)
+
+	}
```
Yantrio, line 75:
```suggestion
		)
	}
```

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,242 @@
...
+				k,
+				strings.Join(
+					[]string{
+						string(KeyDefaultRemote), string(KeyBackend), string(KeyPlanFile), string(KeyStateFile),
+					}, ", ",
+				),
+			)
+		}
+		return nil
+	}
```
Yantrio, line 169: tiny tiny nitpick: a switch/case here doesn't help the readability too much, how about just a simple if statement?

```suggestion
	if k == KeyDefaultRemote || k == KeyBackend || k == KeyPlanFile || k == KeyStateFile || k.IsRemoteDataSource() {
        return nil
    }

    return fmt.Errorf(
        ....
    )
```

internal/states/encryption/encryptionconfig/deepmerge.go
```diff
@@ -0,0 +1,97 @@
...
+		current = deepMergeConfig(deepMergeMe, current)
+	}
+	if current != nil {
+		return current
+	} else {
+		return defaultIfAllOthersNil
+	}
+}
+
+func deepMergeConfig(addon *Config, base *Config) *Config {
```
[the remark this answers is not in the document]
Yantrio: Nothing changed here, why was this resolved?

internal/states/encryption/encryptionconfig/source.go
```diff
@@ -0,0 +1,15 @@
+package encryptionconfig
+
+type Source string
+
+const (
+	// SourceHCL indicates that the configuration has been obtained by parsing the HCL source.
+	SourceHCL Source = "hcl"
```
Yantrio, line 7: Food for thought: We don't just support HCL, we also support JSON configuration format (https://opentofu.org/docs/language/syntax/json/).

Do we need another way to define that it came from the config files and not from the env?

internal/states/encryption/singleton.go
```diff
@@ -0,0 +1,46 @@
...
+//	    encryption := encryption.GetSingleton()
+//	}
+func GetSingleton() Encryption {
+	lock.Lock()
+	defer lock.Unlock()
+	if instance == nil {
+		instance = New(logging.HCLogger())
+	}
+	return instance
+}
```
Yantrio, line 34: Do you think it's worth adding a TODO/FIXME for added visibility about removing this.

Yantrio (approved):
Overall this PR seems great, I've left a couple small tiny nitpicks but they are not blocking this being merged.

This is great work and I am super thankful that we took the time to work on this so much and get it to the point that it's at today! 

Thanks @StephanHCB & @janosdebugs <3

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,271 @@
...
+	// ApplyEnvConfigurations applies a configuration coming from the operating system environment.
+	ApplyEnvConfigurations(
+		encryption map[encryptionconfig.Key]encryptionconfig.Config,
+		decryptionFallback map[encryptionconfig.Key]encryptionconfig.Config,
+	) error
+
+	// ApplyHCLEncryptionConfiguration applies a single encryption configuration coming from HCL.
+	ApplyHCLEncryptionConfiguration(key encryptionconfig.Key, config encryptionconfig.Config) error
+	// ApplyHCLDecryptionFallbackConfiguration applies a single decryption fallback configuration coming from HCL.
+	ApplyHCLDecryptionFallbackConfiguration(key encryptionconfig.Key, config encryptionconfig.Config) error
```
RLRabinowitz, line 46: I think it makes more sense for these methods to be independent of what's the source of configuration, and get it as a dependency. There's no actual difference in behaviour between these methods (which is great, as there shouldn't be).

WDYT about `ApplyConfigurations` + `ApplyEncryptionConfiguration` + `ApplyDecryptionFallbackConfiguration`, all of these getting the source as an argument? I think that this approach also makes at least one of these methods unnecessary (probably `ApplyConfigurations`, depending on usage)

Would appreciate to know what you think about this approach

internal/states/encryption/encryptionconfig/config.go
```diff
@@ -0,0 +1,242 @@
...
+// to providing the configuration in the .tf files directly.
+//
+// Set this environment variable to a JSON representation of ConfigEnvStructure, or leave it unset/blank
+// to disable encryption (default behaviour). If you do not specify a configuration but "enforced" is set to true, tofu
+// will refuse to function. If you specify an invalid JSON, the entire tofu run will fail regardless of the "enforced"
+// setting.
+//
+// Note: With rare exceptions, you should avoid setting the state encryption environment variables in tests,
+// as this may make tests depend on each other. See the comments on encryption.ParseEnvironmentVariables().
+const ConfigEnvName = "TF_STATE_ENCRYPTION"
```
RLRabinowitz, line 189: The comment here and the comment over `FallbackConfigEnvName` to refer to some things that I can't seem to find

I couldn't find `encryption.ParseEnvironmentVariables` nor `ConfigEnvStructure`

internal/states/encryption/encryptionflow/flow.go
```diff
@@ -0,0 +1,123 @@
...
+	// it fails.
+	DecryptState(payload []byte) ([]byte, error)
+
+	// EncryptState encrypts plaintext state.
+	//
+	// To encrypt the state, pass the unencrypted state in JSON format to this function. This function will attempt to
+	// encrypt it with the encryption key and return a valid, but encrypted JSON.
+	//
+	// If no encryption is configured, but "enforced" is enabled, the encryption will fail. Otherwise, this function
+	// will return the state unmodified.
```
RLRabinowitz, line 64: I think this statement regarding `"enforced"` is also correct for `EncryptPlan` and also for the decrypt methods. Is that correct?
If so, we should probably add this statement to the relative comments

RLRabinowitz (approved):
Looks good @StephanHCB , thank you for your contribution!
I've left a few comments, mostly nits or asks for clarification 🙏🏻

internal/states/encryption/encryptionconfig/source.go
RLRabinowitz: Or possibly change the source to "config" from "hcl"?

internal/states/encryption/encryption.go
```diff
@@ -0,0 +1,271 @@
...
+				fmt.Sprintf("Invalid state encryption configuration for configuration key %s", key),
+				err.Error(),
+			))
+		}
+	}
+
+	return diags
+}
+
+func (e *encryption) allConfigKeys() map[encryptionconfig.Key]struct{} {
```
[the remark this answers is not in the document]
janosdebugs: Because a `map[encryptionconfig.Key]struct{}` has an O(1) access time, whereas a slice has a n `O(n)` access time in Go and also guarantees that no duplicate items are present. As ugly as it is, it's the Go way of creating sets.

internal/states/encryption/encryption.go
janosdebugs: There is a fixed ruleset in what order these configurations need to be applied, see the description of the `Merge` function in `config.go`. If we remove the specialization in the API, we need to make the caller pass along the priority of that configuration, which would increase the overhead of the calling side.

internal/states/encryption/encryptionconfig/config.go
janosdebugs: Thank you, this is a leftover from the refactoring, fixing it.

internal/states/encryption/encryption.go
janosdebugs: Happy to add this change, but you'll get the same error eventually if you call functions like `Statefile`. This only makes sure you check all possible configurations all at once.

internal/states/encryption/encryptionconfig/deepmerge.go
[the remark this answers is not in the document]
janosdebugs: That's an oversight, fixing.

internal/states/encryption/encryption.go
janosdebugs: This can also be a configuration error if we let users specify this, depending on how the calling side is implemented.

internal/states/encryption/singleton.go
janosdebugs: I don't think we should litter the code with TODOs, it will be come unmanagable very quickly. If we get rid of globally scoped things, this will fall out automatically.

internal/states/encryption/encryption.go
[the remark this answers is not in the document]
janosdebugs: I think that we might even want to get rid of this function as it is called only in one place.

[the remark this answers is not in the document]
janosdebugs: Removed and also made sure that `Validate()` works on a single lock.

internal/states/encryption/encryptionflow/flow.go
cube2222: I believe the enforce flag is a config setting in the state backend or remote data source block. In other words, I don't think it would affect the plan, cause there's no place to set this enforcing for a plan.

janosdebugs: Then the current naming and structure is incorrect, we shouldn't be allowing the `enforced` value for configurations that are not for these purposes.

StephanHCB: I would not introduce multiple different structures. That'll make the code really complex.

It's just a case of an invalid configuration, which should lead to an error being returned by `Validate()`. 

There are lots of ways configurations can be invalid:
- setting key_provider.name to an unregistered key provider name
- setting a parameter under key_provider.config that does not exist
- not providing a required parameter under key_provider.config that is required by a key provider
- setting a parameter under key_provider.config to a value that a key provider does not allow
- setting method.name to an unregistered method name
- setting a parameter under method.confit that does not exist
- not providing a required parameter under method.config that is required by a method
- setting a parameter under method.config to a value that an encryption method does not allow
- setting enforced on a decryption fallback configuration
- setting enforced on an encryption configuration for plan files (encryption will fail for plan files anyway if they can't be encrypted)

Any of these needs to end the run with a non-zero exit code and cause an error to be printed for the user before any processing has taken place.

All that is important are good error messages that explain to the user what is wrong.

The final versions of the various validation functions will emerge when the flow, the methods, and the key providers are actually implemented.

cam72cam:
Closing as we have built on top of this work in https://github.com/opentofu/opentofu/pull/1227

PR #1074, opened 2024-01-07
shutdown OAuth2 callback server when user hits Ctrl+C

### LINKED ISSUE #1058: `tofu login` cannot be interrupted
### OpenTofu Version

```shell
OpenTofu v1.7.0-dev
on linux_amd64
```

N/A

### Debug Output

https://gist.github.com/leg100/ac9548f268fba721d692b7928409f46b

### Expected Behavior

When running `tofu login <login.server>`, sending it a Ctrl-C is expected to terminate the process

### Actual Behavior

Nothing happens.

### Steps to Reproduce

1. Run a service that advertises [OAuth endpoints via remote service discovery](https://developer.hashicorp.com/terraform/internals/login-protocol), for example:

```golang
package main

import (
	"net/http"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-type", "application/json")
		w.Write([]byte(`{
	"login.v1": {
		"client": "terraform-cli",
		"grant_types": ["authz_code"],
		"authz": "/oauth/authorization",
		"token": "/oauth/token",
		"ports": [10000, 10010]
	}
}`))
	})
	http.ListenAndServeTLS("localhost:8080", "cert.pem", "key.pem", nil)
}
```
2. Run `tofu login` and enter `yes`:

```
137> tofu login localhost:8080
OpenTofu will request an API token for localhost:8080 using OAuth.

This will work only if you are able to use a web browser on this computer to
complete a login process. If not, you must obtain an API token by another
means and configure it in the CLI configuration manually.

If login is successful, OpenTofu will store the token in plain text in
the following file for use by subsequent commands:
    /home/louis/.terraform.d/credentials.tfrc.json

Do you want to proceed?
  Only 'yes' will be accepted to confirm.

  Enter a value: yes
```
3. `tofu` opens a browser; ignore and instead attempt to terminate the `tofu` process by entering `Ctrl-C`. Nothing will happen.
4. You are forced to open another terminal and send the `tofu` process a SIGKILL.

```
1> pkill -9 tofu
```

```
OpenTofu will now wait for the host to signal that login was successful.

^C^C^CKilled
```

### Additional Context

This has long been a problem with `terraform login` as well.

The problem does _not_ occur when the given login server hostname is `app.terraform.io` because the code has a special branch for that hostname that skips OAuth.

### References

_No response_

maksym-nazarenko:
This PR introduces code to handle `Ctrl+C` signal during `tofu login <host>` where host is different from `app.terraform.io`

Resolves #1058 

## Target Release

1.7.0


cam72cam:
The only comment I have for now is to wonder if there's a reasonable way to test this.  I did some quick searching and didn't see anything too promising.

maksym-nazarenko:
@cam72cam thank you for looking into this one
no worries about timings - `1.6` is worth waiting :) 

regarding tests - I was thinking about the same, but didn't have a chance to check the possibility. I'll try to write something these days, if possible at all.

cam72cam:
Thanks for taking a stab at testing this, I think I'm happy accepting with manual testing in the worst case scenario.

maksym-nazarenko:
putting it back to draft state, as I think it's possible to test new behavior

internal/command/login.go
```diff
@@ -515,6 +529,12 @@ func (c *LoginCommand) interactiveGetTokenByCode(hostname svchost.Hostname, cred
 		log.Printf("[WARN] login: callback server can't shut down: %s", err)
 	}
 
+	if code == "" {
+		// empty code is not possible in happy path as it is validated in the HTTP handler of our callback server
+		// so it means, the current command was interrupted byt the shutdown signal
```
cube2222, line 534: typo - byt

cube2222 (approved):
Thanks for the contribution!

Reviewed code and tested manually. One nit comment left, LGTM overall!

Happy to merge based on manual test alone, but having an automatic test would be even better.

maksym-nazarenko:
added automated test to cover this change - not it's fully ready for review.

(just realized that my previous comment was never published because of connection hiccup 😬 )

internal/command/login.go
maksym-nazarenko: eagle eye :) 
changed

cam72cam (approved):

cube2222 (approved):
LGTM!

PR #1106, opened 2024-01-11
Return false when assertBackendStateLocked fails and update comments for it and assertBackendStateUnlocked

jjshanks:
`assertBackendStateLocked` was always returning true. It has been updated to return false when it fails. The comments for it and `assertBackendStateUnlocked` have also been updated for clarity.

## Target Release

1.7.0

### DIFF
--- a/internal/backend/local/testing.go
+++ b/internal/backend/local/testing.go
@@ -201,9 +201,9 @@ func mustResourceInstanceAddr(s string) addrs.AbsResourceInstance {
 	return addr
 }
 
-// assertBackendStateUnlocked attempts to lock the backend state. Failure
-// indicates that the state was locked and false is returned. True is
-// returned if a lock was obtained.
+// assertBackendStateUnlocked attempts to lock the backend state for a test.
+// Failure indicates that the state was locked and false is returned.
+// True is returned if a lock was obtained.
 func assertBackendStateUnlocked(t *testing.T, b *Local) bool {
 	t.Helper()
 	stateMgr, _ := b.StateMgr(backend.DefaultStateName)
--- a/internal/backend/local/testing.go
+++ b/internal/backend/local/testing.go
@@ -202,8 +202,8 @@ func mustResourceInstanceAddr(s string) addrs.AbsResourceInstance {
 }
 
 // assertBackendStateUnlocked attempts to lock the backend state. Failure
-// indicates that the state was indeed locked and therefore this function will
-// return true.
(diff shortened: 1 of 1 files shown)


janosdebugs (approved):
As discussed, this change looks good to me.

internal/backend/local/testing.go
```diff
@@ -202,27 +202,31 @@ func mustResourceInstanceAddr(s string) addrs.AbsResourceInstance {
 }
 
 // assertBackendStateUnlocked attempts to lock the backend state. Failure
-// indicates that the state was indeed locked and therefore this function will
-// return true.
+// indicates that the state was locked and false is returned. True is
+// returned is a lock obtained.
```
cam72cam, line 206:
```suggestion
// returned if a lock was obtained.
```

internal/backend/local/testing.go
```diff
@@ -202,27 +202,31 @@ func mustResourceInstanceAddr(s string) addrs.AbsResourceInstance {
 }
 
 // assertBackendStateUnlocked attempts to lock the backend state. Failure
```
cam72cam, line 204: I liked how you added "for a test" below, would you mind adding that here?

cam72cam:
Looking more at where this function is used, I think we could get rid of the boolean return all together and simplify it further.

cam72cam:
This is probably outside the scope of this PR, but these asserts have side effects that should probably be documented or removed, like trying to unlock if the lock was successful.

cam72cam:
@jjshanks Thanks for finding and fixing this!

internal/backend/local/testing.go
jjshanks: Good catch, I meant to update it as well

internal/backend/local/testing.go
```diff
@@ -202,27 +202,31 @@ func mustResourceInstanceAddr(s string) addrs.AbsResourceInstance {
...
 	t.Helper()
 	stateMgr, _ := b.StateMgr(backend.DefaultStateName)
 	if _, err := stateMgr.Lock(statemgr.NewLockInfo()); err != nil {
+		// lock was not obtained
 		return true
 	}
 	t.Error("unexpected success locking state")
-	return true
+	// lock was obtained
+	return false
```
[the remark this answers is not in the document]
jjshanks: Verified no callers were using the return value

cam72cam (approved):

PR #1107, opened 2024-01-11
Allow static evaluation of locals and variables for module sources and backends

### LINKED ISSUE #1042: Allow evaluation of constant locals and variables prior to module configuration
### Summary

Many users expect to be able to DRY up their large configurations by using locals/variables in module sources and backend configurations (among other places).  This is currently prohibited and users are forced to do a lot of manual work or use additional tooling to work around this limitation.  We should be able to allow this functionality, with some caveats and guardrails.

### Problem Statement

One of the more frustrating features with the OpenTofu configuration language is the inability to use variables and locals in many common areas.  Two of the most requested being module sources and backed configurations.  This limitation exists as many parts of the configuration can only be evaluated during the plan and apply stages.  However, certain variables and locals *can* be evaluated in a static fashion prior to plan/apply and should be able to be used in many currently restricted areas.  This represents a massive improvement for many common use cases.

## Module Sources
The most common use case I have seen requested is the ability to specify module versions as a local variable.  Without the ability to modify the source attribute from a constant expression, bumping the version of org/my-utils is a tedious and error prone process.
```hcl
locals {
    util_version = 1.5.2
}

# Consider 20-30 similar instances of this module in half a dozen files
module {
    source = "git@github.com:org/my-utils?ref=${local.util_version}"
}
```

## Backend Configurations
Additionally users frequently request the ability to configure backends via locals/variables
```hcl
locals {
    region = "us-east-1"
}

terraform {
    backend "s3" {
        region = locals.region
    }
    required_providers {
        "aws_primary" = {
            source = "hashicorp/aws"
            region = locals.region
        }
    }
}
```

There are additional use cases, such as lifecycles, that are mentioned in the referenced issues and could be an extension off of this initial work.

## Pass-through Constant Variables
Some users have a legitimate use case for also allowing for some static fields to be passed in via variables.  A common use case is switching to a development version of a given module.

### main.tf
```hcl
locals {
    util_dev = true
}

module "submod" {
    source = "./submod/"
    util_dev = local.util_dev
}
```
### ./submod/main.tf
```hcl
variable "util_dev" {
    type = "bool"
}

locals {
    util_version = "1.5.2${var.util_dev ? "-dev" : ""}"
}

module {
    source = "git@github.com:org/my-utils?ref=${local.util_version}"
}
```

## ~~Label Interpolation (out-of-scope, technical discovery only)~~

~~Additionally, some users requested the ability to parameterize block labels.  This is possible, but I'm not sure we want to tackle it as part of this RFC.~~
```hcl
locals {
   prefix  = "org-"
}

module "{local.prefix}-mod" {
  source = "./src"
}
```
~~Note that the `$` is omitted from the interpolation as it is forbidden in the current HCL parser.~~

### Technical Description

The current workflow for modules is:
* For each path, create a ModuleFile:
  - Read the file contents
  - Parse the contents into a generic hcl structure
  - For each block:
    - Validate the block structure
    - Add a representation to the ModuleFile struct
* Merge ModuleFiles together into Module
* Determine child modules and load them via the same workflow

The ModuleFile initialization can be split up and handled in stages.

An alternative workflow could look like:
* An additional parameter is required that represents variable values passed in from the caller
* For each path, create a ModuleFile
  - Read the file contents
  - Parse the contents into a generic hcl structure
* Create an EvaluationContext to allow for evaluation of expressions by holding var/local values as they are discovered
* For each ModuleFile:
  - For each local/variable block
    - Validate the block
    - Add a representation to the ModuleFile struct
* Until no new variables or locals can be evaluated:
  - For each ModuleFile:
    - For each Variable in the file:
      - Look for the variable in the values passed in via the new parameter.
      - Otherwise try to evaluate the variable's default expression given the current EvaluationContext
      - If either is successful, add the discovered value to the EvaluationContext
    - For each Local in the file:
      - Try to evaluate the local's expression given the current EvaluationContext
      - If successful, add the discovered value to the EvaluationContext
  - NOTE: There is probably a better method to build a dependency tree between locals and variables to avoid the iterative try loop.
* For each ModuleFile:
  - For each non variable/local block:
    - Validate the block structure
    - Add a representation to the ModuleFile struct
* Merge ModuleFiles together into Module
* For each child module:
  - Evaluate the parameters (where possible) to the child module given the current EvaluationContext
  - Use the same workflow to load the child module, passing in the parameters that were able to be evaluated
  - Care has to be taken here to differentiate "parameter not able to be computed yet" and "parameter not specified, use the default"

There are many cases not supported by this workflow.  Any dynamically computed value will not be available during this evaluation flow.  Examples include but are not limited to: resource outputs, module outputs, data outputs.

One thing not discussed thus-far is the ability to use variables in the "root" module.  These are usually specified via tfvars, -var flags, or environment variables during the plan and apply stage.  There is a case to be made to allow these in the root module, though I am not sure how reasonable of a use case that is.

There has been much discussion on this feature over nearly a decade.  Many proposal have been created and this RFC takes inspiration from many of them.

The main alternative usually proposed is to add a new constant construct to the language.  This adds clarity for what properties are guaranteed static vs potentially dynamic.  In an earlier hack and slash session in the codebase, I one variant of that [here](https://github.com/opentofu/opentofu/compare/main...cam72cam:opentofu:const_block_experiment).  It was a more straightforward change and I did appreciate the clarity of "only a constant can be used here".  Given that it is requested for the statically computed values to be able to be passed as module variables and preserve their constant nature, I felt that it was possibly too limited.  Additionally, it adds dramatically new language features that would break existing linters / code tools.  It is still an option if the current proposal is too "radical", but I think the downsides are much greater with the explicitly const approach.

### Downsides

Although syntactically identical to Terraform, this would mark a clear divergence between the feature sets of Terraform / OpenTofu configuration language.  My hope is that HashiCorp would make similar changes on their side to match our feature set, but that is not guaranteed.  It would mean that certain modules would not function if they opt-in to this new functionality.  However it is a clear error when terraform tries to parse any of these dynamic properties.

We will also need to make sure that this does not incur a performance impact for large configurations.  My initial testing has not suggested this, but it is additional complexity that will need to be considered.

### Unresolved Questions

* How should this work be broken up / divided into sub-tasks?
* ~~Should label name interpolation be considered as part of this work?~~ No
* ~~Should variables be recognized from the Environment / CLI during init for static evaluation?~~ Yes

## OpenTofu:
* https://github.com/opentofu/opentofu/issues/286
* https://github.com/opentofu/opentofu/issues/450
* https://github.com/opentofu/opentofu/issues/1017

## Terraform

* https://github.com/hashicorp/terraform/issues/13603
* https://github.com/hashicorp/terraform/issues/24187
* https://github.com/hashicorp/terraform/issues/24796
* https://github.com/hashicorp/terraform/issues/31993
  - This has been implemented in my PoC, but I am not sure if this should be included in the first pass of this feature.
* https://github.com/hashicorp/terraform/issues/32041

### Modules
* https://github.com/hashicorp/terraform/issues/1439
* https://github.com/hashicorp/terraform/issues/1846
* https://github.com/hashicorp/terraform/issues/2940
* https://github.com/hashicorp/terraform/issues/8165
* https://github.com/hashicorp/terraform/issues/14745
* https://github.com/hashicorp/terraform/issues/15613
* https://github.com/hashicorp/terraform/issues/15614
* https://github.com/hashicorp/terraform/issues/17994
* https://github.com/hashicorp/terraform/issues/18063
* https://github.com/hashicorp/terraform/issues/21891
* https://github.com/hashicorp/terraform/issues/23948
* https://github.com/hashicorp/terraform/issues/25587
* https://github.com/hashicorp/terraform/issues/26544
* https://github.com/hashicorp/terraform/issues/26619
* https://github.com/hashicorp/terraform/issues/26888
* https://github.com/hashicorp/terraform/issues/27624
* https://github.com/hashicorp/terraform/issues/28912
* https://github.com/hashicorp/terraform/issues/31701
* https://github.com/hashicorp/terraform/issues/33793

### Backends
* https://github.com/hashicorp/terraform/issues/13022
* https://github.com/hashicorp/terraform/issues/13289
* https://github.com/hashicorp/terraform/issues/17288
* https://github.com/hashicorp/terraform/issues/18632
* https://github.com/hashicorp/terraform/issues/22088
* https://github.com/hashicorp/terraform/issues/26592

### Lifecycle
* https://github.com/hashicorp/terraform/issues/3116
* https://github.com/hashicorp/terraform/issues/5346
* https://github.com/hashicorp/terraform/issues/5446
* https://github.com/hashicorp/terraform/issues/5481
* https://github.com/hashicorp/terraform/issues/10730

### Proof of Concept

This code is *incredibly* rough and implements the majority of features described in this RFC.  https://github.com/opentofu/opentofu/compare/main...cam72cam:opentofu:pre_evaluate_variables_locals.

However, some of the features (like backend configurations) were not ported from my earlier work with the const block experimentation https://github.com/opentofu/opentofu/compare/main...cam72cam:opentofu:const_block_experiment.  In the end though, the flow for those components would be nearly identical.

This only implements a subset of the features mentioned in the related issues.  However, the module versioning and backend configuration are the hardest IMO and a solid design there will allow the rest to be added over time with less effort.

cam72cam:
This is the initial implementation of a static evaluation context that can be used to evaluate expression that do not rely on any state data (vars/locals only). 

This pretty closely follows the plan detailed in RFC #1042 as I had already started experimenting with this code and knew roughly what I needed to change.  All of the examples detailed in the RFC should be functional with this PR.

Current state of this work: Technical Preview.  Most components should be functional, but may have unexpected bugs or limitations.

TODO:
* [x] Circular references
* [ ] Fix a *lot* of tests
* [ ] Write a *lot* of tests
* [ ] Dramatically improve diagnostics.  The required information is there, but not well formatted.
* [ ] Make sure all module calls are passing the correct parameter context.  This was ad-hoc to start.
* [ ] Clean up how raw input variables are handled
* [ ] Update `tofu init` help text to include vars/var-file
* [ ] Pass workspace into root module context
* [ ] Changelog
* [ ] Docs

Open Questions:
* [ ] What are the performance implications?
* [x] What does lazy evaluation look like?  Does that have a dramatic performance benefit?
  - Added lazy evaluation, code is dramatically cleaner and only does the hcl evaluation when required
* [ ] How much type checking should be done for variables?

Resolves #1042

## Target Release

The corresponding RFC has not been discussed by the Technical Steering Committee.  Depending on timing and approval, this could potentially make it into 1.7.0 at the earliest.

### DIFF
--- a/internal/backend/backend.go
+++ b/internal/backend/backend.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"os"
 
+	"github.com/hashicorp/hcl/v2"
 	svchost "github.com/hashicorp/terraform-svchost"
 	"github.com/mitchellh/go-homedir"
 	"github.com/opentofu/opentofu/internal/addrs"
@@ -327,6 +328,22 @@ func (o *Operation) Config() (*configs.Config, tfdiags.Diagnostics) {
 	config, hclDiags := o.ConfigLoader.LoadConfig(o.ConfigDir)
 	diags = diags.Append(hclDiags)
 	return config, diags
+}
+
+func (o *Operation) RawVariables() configs.RawVariables {
+	// TODO copy paste from meta_config.go
+	vars := make(configs.RawVariables)
+	for k, v := range o.Variables {
+		v := v
+		vars[k] = func(mode configs.VariableParsingMode) (cty.Value, hcl.Diagnostics) {
+			parsed, _ := v.ParseVariableValue(mode)
+			// TODO DIAGS
+			return parsed.Value, nil
+
+		}
+	}
+
+	return vars
 }
 
 // ReportResult is a helper for the common chore of setting the status of
--- a/internal/backend/local/backend_local.go
+++ b/internal/backend/local/backend_local.go
@@ -142,7 +142,7 @@ func (b *Local) localRunDirect(op *backend.Operation, run *backend.LocalRun, cor
 	var diags tfdiags.Diagnostics
 
 	// Load the configuration using the caller-provided configuration loader.
-	config, configSnap, configDiags := op.ConfigLoader.LoadConfigWithSnapshot(op.ConfigDir)
+	config, configSnap, configDiags := op.ConfigLoader.LoadConfigWithSnapshot(op.ConfigDir, op.RawVariables())
 	diags = diags.Append(configDiags)
 	if configDiags.HasErrors() {
 		return nil, nil, diags
--- a/internal/backend/remote/backend_common.go
+++ b/internal/backend/remote/backend_common.go
@@ -221,7 +221,7 @@ func (b *Remote) waitForRun(stopCtx, cancelCtx context.Context, op *backend.Oper
 // remote system's responsibility to do final validation of the input.
 func (b *Remote) hasExplicitVariableValues(op *backend.Operation) bool {
 	// Load the configuration using the caller-provided configuration loader.
-	config, _, configDiags := op.ConfigLoader.LoadConfigWithSnapshot(op.ConfigDir)
+	config, _, configDiags := op.ConfigLoader.LoadConfigWithSnapshot(op.ConfigDir, op.RawVariables())
 	if configDiags.HasErrors() {
 		// If we can't load the configuration then we'll assume no explicit
 		// variable values just to let the remote operation start and let
--- a/internal/cloud/backend_plan.go
+++ b/internal/cloud/backend_plan.go
@@ -254,7 +254,7 @@ in order to capture the filesystem context the remote workspace expects:
 		}
 	}
 
-	config, _, configDiags := op.ConfigLoader.LoadConfigWithSnapshot(op.ConfigDir)
+	config, _, configDiags := op.ConfigLoader.LoadConfigWithSnapshot(op.ConfigDir, op.RawVariables())
 	if configDiags.HasErrors() {
 		return nil, fmt.Errorf("error loading config with snapshot: %w", configDiags.Errs()[0])
 	}
--- a/internal/command/apply.go
+++ b/internal/command/apply.go
@@ -95,6 +95,8 @@ func (c *ApplyCommand) Run(rawArgs []string) int {
 	// object state for now.
 	c.Meta.parallelism = args.Operation.Parallelism
 
+	c.GatherVariables(args.Vars)
+
 	// Prepare the backend, passing the plan file if present, and the
 	// backend-specific arguments
 	be, beDiags := c.PrepareBackend(planFile, args.State, args.ViewType)
@@ -109,7 +111,7 @@ func (c *ApplyCommand) Run(rawArgs []string) int {
 	diags = diags.Append(opDiags)
 
 	// Collect variable value and add them to the operation request
-	diags = diags.Append(c.GatherVariables(opReq, args.Vars))
+	diags = diags.Append(c.StuffVariables(opReq))
 
 	// Before we delegate to the backend, we'll print any warning diagnostics
 	// we've accumulated here, since the backend will start fresh with its own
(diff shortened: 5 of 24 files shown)


cam72cam:
Closing this for now as this branch is incredibly out of date and will be replaced by a much more methodical approach in #1042

PR #1123, opened 2024-01-15
Ignore version conflicts when unlocking state

### LINKED ISSUE #1026: tofu force-unlock should not fail on tofu version mismatch with remote backend
### OpenTofu Version

```shell
OpenTofu v1.6.0-dev
```

Any configuration with remote backend

### Debug Output

.

### Expected Behavior

`tofu force-unlock lock_id` should work even if local tofu version mismatches the version in remote backend

### Actual Behavior

`tofu force-unlock` removes the lock but fails:

```
│ Error: OpenTofu version mismatch
│
│ The local OpenTofu version (XXX) does not match the configured version for remote workspace WORKSPACE (YYY). If you're sure you want to upgrade the state, you can force OpenTofu to continue using the
│ -ignore-remote-version flag. This may result in an unusable workspace.
```

### Steps to Reproduce

1. `tofu force-unlock lock_id` on a state lock

### Additional Context

_No response_

### References

Related issue in the Terraform repository - https://github.com/hashicorp/terraform/issues/28853

jjshanks:
Added call in unlock command to ignore remote version conflicts. The same code used in other read-only commands was added to the unlock command. This bug only happens with the `remote` and `cloud` backends which have special checks for version conflicts and why the test case was added to the remote backend test.

Unlocking does not alert any state only meta data about the lock so there is no risk of upgrading the state which makes this safe.

Resolves #1026

## Target Release

1.7.0


cam72cam (approved):

cube2222 (approved):
LGTM!

PR #1126, opened 2024-01-16
docs(install): add missing steps in Debian family when facing hardened umask

nsballmann:
When facing a hardened umask (e.g. `0027`) the current installation steps aren't sufficient on a Debian family OS.

This PR adds the missing steps.

### DIFF
--- a/website/docs/intro/install/deb-step2.sh
+++ b/website/docs/intro/install/deb-step2.sh
@@ -1,4 +1,4 @@
 sudo install -m 0755 -d /etc/apt/keyrings
 curl -fsSL https://get.opentofu.org/opentofu.gpg | sudo tee /etc/apt/keyrings/opentofu.gpg >/dev/null
 curl -fsSL https://packages.opentofu.org/opentofu/tofu/gpgkey | sudo gpg --no-tty --batch --dearmor -o /etc/apt/keyrings/opentofu-repo.gpg >/dev/null
-sudo chmod a+r /etc/apt/keyrings/opentofu.gpg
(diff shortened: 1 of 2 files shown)


janosdebugs:
Hey @nsballmann thank you for this. I believe these instructions should be made consistent across all operating systems and this should also be tested by creating the umask condition in the test script, otherwise this may be broken in the future. Would you be up for extending this PR to that effect? No worries if not, then we can add that in a subsequent PR.

As a side note for later, this may also affect the [install script](https://github.com/opentofu/get.opentofu.org).

nsballmann:
> Hey @nsballmann thank you for this. I believe these instructions should be made consistent across all operating systems and this should also be tested by creating the umask condition in the test script, otherwise this may be broken in the future. Would you be up for extending this PR to that effect? No worries if not, then we can add that in a subsequent PR.
> 
> As a side note for later, this may also affect the [install script](https://github.com/opentofu/get.opentofu.org).

I just faced this when switching over from Terraform and snug in the time to create the PR in between meetings so I don't forget. I would like to contribute like this now and then, when I notice something and my time allows it.

So for this first PR I would like to just go with the current scope to get to know this projects processes. Do you ask because of how to scope the (probably mandatory) issue?

That said, I agree that somebody should have a look, check and include tests for this scenario. As a developer I tend to measure project maturity on how it deals with (CIS-)hardened `umask`s.

janosdebugs:
Hey @nsballmann no worries, I'll open an issue and we'll sort out the other distros and testing later. Thank you for your help, this is much appreciated.

janosdebugs (approved):

cam72cam (approved):

nsballmann:
@janosdebugs @cam72cam thank you for the quick and uncomplicated merge. :+1:

PR #1129, opened 2024-01-16
Run block variable references

### LINKED ISSUE #1045: `tofu test` - support variables in `run` block referring to outputs of previous `run` blocks
### OpenTofu Version

```shell
OpenTofu v1.6.0-dev
```

### Use Cases

Variables in `tofu test` `run` blocks cannot reference outputs from previously executed `run` blocks. This feature can be a big help in writing tests.
If we use a setup `run.module` block at the beginning of the test to create needed resources that are not created in the "main" `run`. I can think of many cases we would like to reference these resources' Ids and use them as a variable for the the main `run` block. 

### Attempted Solutions

Cannot be done today, we don't support it.

### Proposal

If we implement this suggestion we can do the following:

```
# setup/main.tf
variable "content" {
  type = string
}

resource "random_string" "file_name" {
  length           = 16
  special          = false
}

resource "local_file" "foo" {
  filename = random_string.file_name.result
  content  = var.content
}

output "file_name" {
  value = local_file.foo.filename
}
```

```
# main.tftest.hcl
variables {
  content = "some value"
}

run "setup" {
  module {
	source = "./setup"
  }
}

run "test" {
  variables {
    file_name = run.setup.file_name <---
  }
}
```

### References

It's possible in Terrafrom - here's a reference to it mentioned in the [docs](https://developer.hashicorp.com/terraform/language/tests#variable-references).

cam72cam:
Builds an EvalContext which represents the output state of previously run steps to be used when evaluating variable expressions.

Resolves #1045

## Target Release

1.7.0, 1.6.1

### DIFF
--- a/internal/command/testdata/test/variables_reference/main.tftest.hcl
+++ b/internal/command/testdata/test/variables_reference/main.tftest.hcl
@@ -4,7 +4,7 @@ variables {
 
 run "setup" {
   module {
-        source = "./setup"
+    source = "./setup"
   }
 }
 
--- a/internal/command/testdata/test/variables_reference/main.tftest.hcl
+++ b/internal/command/testdata/test/variables_reference/main.tftest.hcl
@@ -1,0 +1,15 @@
+variables {
+  content = "some value"
+}
+
+run "setup" {
+  module {
+        source = "./setup"
+  }
+}
+
+run "test" {
+  variables {
+    file_name = run.setup.file_name
+  }
+}
--- a/internal/command/testdata/test/variables_reference/setup/main.tf
+++ b/internal/command/testdata/test/variables_reference/setup/main.tf
@@ -1,0 +1,17 @@
+variable "content" {
+  type = string
+}
+/*
+resource "random_string" "file_name" {
+  length           = 16
+  special          = false
+}
+
+resource "local_file" "foo" {
+  filename = random_string.file_name.result
+  content  = var.content
+}*/
+
+output "file_name" {
+  value = "output_value"
+}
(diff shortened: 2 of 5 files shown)


cam72cam:
Current error when trying to reference a block that does not exist or has not yet been run:
```
main.tftest.hcl... fail
  run "test"... fail
╷
│ 
│   on main.tftest.hcl line 7, in run "test":
│    7:     file_name = run.setup.file_name
│ 
│ This object does not have an attribute named "setup".
╵
  run "setup"... skip

Failure! 0 passed, 1 failed, 1 skipped.
```

I think this is reasonable for now and not worth improving yet.

internal/command/testdata/test/variables_reference/setup/main.tf
```diff
@@ -0,0 +1,17 @@
+variable "content" {
+  type = string
+}
+/*
```
cube2222, line 4: Why is this test commented out, could you write a comment about it please?

internal/command/testdata/test/variables_reference/main.tftest.hcl
```diff
@@ -0,0 +1,15 @@
+variables {
+  content = "some value"
+}
+
+run "setup" {
+  module {
+        source = "./setup"
```
cube2222, line 7: Something with the whitespace here is broken.

Yantrio (approved):
LGTM apart from the nits from @cube2222 =)

internal/command/testdata/test/variables_reference/setup/main.tf
cam72cam: Was an artifact of the test I used as reference.  Removed.

PR #1145, opened 2024-01-18
Draft: Support the XDG Base Directory Specification

### LINKED ISSUE #303: Support the XDG Base Directory Specification
Currently, OpenTofu will default to storing a data in the user's home directory, i.e. `~/.opentofu.d` (could still be `~/.terraform.d`). The correct way to store cache, config and state on a user's machine is to follow a platform's best practices - in Linux's case most Linux desktops support and/or adhere to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), a specification for environment variables and default storage locations for certain types of data a program might want to store in dotfiles. Following this spec would allow Linux and MacOS users the ability to specify correct directories for their environment. 

(Note: MacOS by default doesn't follow the XDG spec - they have [their own](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1), however it doesn't refer to environment variables (i.e. supporting MacOS's standard is a separate issue.))

avorima:
This PR introduces support for XDG environment variables. `XDG_CONFIG_HOME` for config and `XDG_DATA_HOME` for plugins.

If the `~/.terraform.d` directory doesn't exist, config is assumed to live in `XDG_CONFIG_HOME/opentofu` with `XDG_CONFIG_HOME` defaulting to `~/.config`.
`~/.local/share/opentofu` and `~/.terraform.d` are always used for looking up plugins with `XDG_DATA_HOME` as third option when it is set to a non-default value.

This ensures that existing config and plugins will still get used, but allow users to gradually migrate to the new structure, with and without setting `XDG_*` environment variables.

The CLI ensures that the intermediate config directory exists, so either `$XDG_CONFIG_HOME/opentofu` or `~/.config/opentofu` will be created if neither `~/.terraformrc`, `~/.tofurc` or `~/.terraform.d` exist. The lookup behavior remains unchanged after this directory was created.

Only affects unix* platforms.

Resolves #303

## Target Release

1.7.0

### DIFF
--- a/internal/command/cliconfig/config_unix.go
+++ b/internal/command/cliconfig/config_unix.go
@@ -45,7 +45,7 @@ func configDir() (string, error) {
 	return configDir, nil
 }
 
-func pluginDir() (string, error) {
+func dataDir() (string, error) {
 	dir, err := homeDir()
 	if err != nil {
 		return "", err
--- a/internal/command/cliconfig/config_unix.go
+++ b/internal/command/cliconfig/config_unix.go
@@ -45,6 +45,20 @@ func configDir() (string, error) {
 	return configDir, nil
 }
 
+func pluginDir() (string, error) {
+	dir, err := homeDir()
+	if err != nil {
+		return "", err
+	}
+
+	configDir := filepath.Join(dir, ".terraform.d")
+	if xdgDir := os.Getenv("XDG_DATA_HOME"); xdgDir != "" && !pathExists(configDir) {
+		configDir = filepath.Join(xdgDir, "opentofu")
+	}
+
+	return configDir, nil
+}
+
 func homeDir() (string, error) {
 	// First prefer the HOME environmental variable
 	if home := os.Getenv("HOME"); home != "" {
--- a/internal/command/cliconfig/config_unix.go
+++ b/internal/command/cliconfig/config_unix.go
@@ -13,6 +13,13 @@ import (
 	"path/filepath"
 )
 
+// Default directories to use when XDG_* is not defined
+// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+const (
+	defaultConfigDir = ".config"
+	defaultDataDir   = ".local/share"
+)
+
 func configFile() (string, error) {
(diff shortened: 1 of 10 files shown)


internal/command/cliconfig/config_unix.go
```diff
@@ -31,7 +37,26 @@ func configDir() (string, error) {
...
-	return filepath.Join(dir, ".terraform.d"), nil
+	configDir := filepath.Join(dir, ".terraform.d")
+	if xdgDir := os.Getenv("XDG_CONFIG_HOME"); xdgDir != "" && !pathExists(configDir) {
+		configDir = filepath.Join(xdgDir, "opentofu")
+	}
+
+	return configDir, nil
+}
+
+func pluginDir() (string, error) {
```
cam72cam, line 48: For clarity, this should probably be called dataDir() to match configDir().

cam72cam:
You've added in some interesting backwards compatibility logic here.  I think that for configuration files that could make sense, but we might want to approach plugin search path a bit differently.  Would it make sense to use the HOME path and the XDG_DATA path in the plugin search location?

avorima:
> You've added in some interesting backwards compatibility logic here. I think that for configuration files that could make sense, but we might want to approach plugin search path a bit differently. Would it make sense to use the HOME path and the XDG_DATA path in the plugin search location?

Do you mean to always use both paths instead of a fallback? If so, then yes, I think that would make moving from one to the other easier.

cam72cam:
> Do you mean to always use both paths instead of a fallback? If so, then yes, I think that would make moving from one to the other easier.

Yep, though I'd like some of the other @opentofu/core-engineers to weigh in before requesting you make significant changes to your PR.  A lot of them may be wrapping up for day and may not see this until tomorrow.

avorima:
Sorry, I wanted to amend a commit and deleted the branch by accident. Re-opened at https://github.com/opentofu/opentofu/pull/1200

PR #1154, opened 2024-01-19
Improve stability on 32-bit architectures

### LINKED ISSUE #1069: Some tests fail for 32-bit builds in Alpine Linux CI
### OpenTofu Version

```shell
1.6.0_rc1
```

n/a

### Debug Output

https://gitlab.alpinelinux.org/pj/aports/-/jobs/1233584
https://gitlab.alpinelinux.org/pj/aports/-/jobs/1233584/raw

### Expected Behavior

All tests pass

### Actual Behavior

Some tests fails

### Steps to Reproduce

A bit hard to reproduce since it relies on AlpineLinux CI :)

### Additional Context

Runnings tests for 32-bit architectures in CI fails for seemingly random tests

### References

_No response_

Yantrio:
This PR attempts to fix the failing 32 bit unit tests throughout the codebase. 

It makes the following changes:

### Improve 32 bit processing of String hashing inside schema's Set type
    
This type is only really used in the remote backend packages lately, but nonetheless we should be fixing this.
This fix ensures that we squash the value to 32 bits (signed) and embrace the overflow to introduce consistency between 32 and 64 bit systems.

### Fix TestResourceDataSet test on 32-bit systems
    
The previous function used in the test was causing issues with clashing IDs. This commit attempts to fix this by re-working how the unit test generates IDs for the set elements in a way that's both friendly to 32 bit and 64 bit systems

### Use 64 bit integer conversions when parsing keys from v2 statefiles
This fixes an issue where we were using strconv.Atoi, and instead we should be enforcing 64-bit conversions here.

### Ensure we use 64 bit integers for durations when assuming roles in the s3 backend

We had a bug in the s3 backend role assumption logic that caused an overflow if used on a 32 bit system.

### Misc
- Reworked some of the test case names in TestResourceDataSet so it's clearer what is failing

Resolves #1069 

## Target Release

1.7.0, 1.6.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ BUG FIXES:
 * Don't check for version conflicts when doing a force-unlock ([#1123](https://github.com/opentofu/opentofu/pull/1123))
 * Fix Global Schema Cache not working in provider acceptance tests ([#1054](https://github.com/opentofu/opentofu/pull/1054))
 * Fix `tofu show` and `tofu state show` not working with state files referencing Terraform registry providers in some instances ([#1141](https://github.com/opentofu/opentofu/pull/1141))
+* Improved stability on 32-bit architectures ([#1154](https://github.com/opentofu/opentofu/pull/1154))
 
 ## Previous Releases
 
--- a/internal/legacy/helper/hashcode/hashcode.go
+++ b/internal/legacy/helper/hashcode/hashcode.go
@@ -10,19 +10,26 @@ import (
 )
 
 // String hashes a string to a unique hashcode.
-//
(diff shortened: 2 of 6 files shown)


cube2222 (approved):

cube2222 (commented):
Target release should include 1.6.1

Yantrio:
Converting to draft as I need to do more fixes, and add a changelog

cam72cam:
"squash the value to 32 bits" -> "squash the value to 32 bits (signed)"

cube2222 (approved):

cam72cam (approved):

Issue #454, opened 2023-09-20
Update copyright notice


nemobis:
At least the primary LICENSE file should be updated from
> Copyright (c) 2014 HashiCorp, Inc.

to a generic attribution/copyright holder, for example:

> Copyright (c) 2014-2023 HashiCorp, Inc.
> Copyright (c) 2023- OpenTofu contributors

See also https://reuse.software/spec/#format-of-copyright-notices

To avoid any confusion it could also be useful to explicitly state that contributors are not assigning or transferring any copyrights to HashiCorp.

cube2222:
Hey @nemobis!

Agreed it would be good to clarify the copyright notices. We'll be working on the details of this, and how to exactly express this, with lawyers. Until then they'll have to remain unchanged.

kislerdm:
Hey folks! I noticed that the forked files contain the copyright header which we may want to change as well. For example, [build.sh](https://github.com/opentofu/opentofu/blob/0930a908227376f0647f759cdfbadfc71f32fad5/scripts/build.sh#L1-L4). WDYT?

cube2222:
Yeah, it's in most of the repo. We still have to figure out (by talking with lawyers) what exactly we want to put there.

I like the approach proposed above and it seems reasonable to me:
> Copyright (c) 2014-2023 HashiCorp, Inc.
Copyright (c) 2023- OpenTofu contributors

but we need to make sure that it's sound first.

nemobis:
BTW I've just noticed that many hashicorp repositories have pending pull requests to add copyright headers which would declare HashiCorp as the only copyright holder, which sounds unlikely to be correct in repositories where there was no CLA (and may be copyfraud if intentionally misleading).

Hopefully it's just a misunderstanding or some unfinished work which they're willing to revise, but see https://github.com/hashicorp/copywrite/issues/100

So just to be clear my suggestion to use
> 2014-2023 HashiCorp, Inc.

was only an example, as something more vague (either in time or copyright holder naming) may be required on repositories with less clear copyright status.

cube2222:
Just to provide an update here, the updated notice should look as follows:
```
// Copyright (c) The OpenTofu Authors
// SPDX-License-Identifier: MPL-2.0
// Copyright (c) 2023 HashiCorp, Inc.
// SPDX-License-Identifier: MPL-2.0
```

cube2222:
We should also apply this to the registry-address repo.

namgyalangmo:
[...]

PR #1158, opened 2024-01-21
Add support for removed block

### LINKED ISSUE #1032: Add support for `removed` block
### OpenTofu Version

```shell
1.6.0-dev
```

### Use Cases

Similarly to the `import` and `moved` blocks, which help manage changing the state through IaC without making any changes in the infrastructure or without using explicit `tofu` commands, it would be helpful if OpenTofu could support a `removed` block, which would be used to remove an object from the tofu state (doing `tofu state rm`)

### Attempted Solutions

Today, the only way to remove an object from state is to `tofu state rm`, and cannot be done in pure IaC

### Proposal

Similarly to `import` or `moved`, create a `removed` block:

```
removed {
  from = "resource.address"
}
```

`tofu` could check that there are no references to the resource address, and remove it from state during `apply`

### References

Related issues - https://github.com/hashicorp/terraform/issues/33226, https://github.com/opentofu/opentofu/issues/996
Related issue in the Terraform repository - https://github.com/hashicorp/terraform/issues/34402,

Evi1Pumpkin:
Adds `removed` block to OpenTofu:
```
removed {
  from = null_resource.example
}
```

The `removed` block is a feature that allows users to remove resources or modules from the state without destroying them. This feature is similar to the `tofu state rm` [command](https://opentofu.org/docs/cli/commands/state/rm/). 
The `removed` block only supports resource and module addresses without key instances. This means that you cannot use `from = module.example[0]`, but you can use `from = module.example`. 
If you want to use the `removed` block, you must remove the configuration block of the resource or module that you want to forget.

## Solution explanation:
1. We have a new struct describing the removed block in the configuration in `configs/removed.go`.  
2. We parse the user’s configuration in `configs/parser_config.go` and add the removed block as part of each file and module in `configs/module.go`.
3. When we parse the removed block in the previous step, we also parse the `removeEndpoint` address (`addrs/remove_endpoint.go`). We decided we’ll save the address as a `ConfigRemovable` interface (`addrs/removable.go`), that describes a resource or module address without an instance key (e.g., `module.hi[0]` is not supported). `removeEndpoint` is heavily based on `moveEndpoint`.
4. Until now, we have never parsed the address of a resource/module with the constrictions of not using instance keys. So we had to create `parseModulePrefix()` and `parseResourceUnderModule()` to do it. They are very similar to `parseModuleInstancePrefix()` and `parseResourceInstanceUnderModule()`, and I extracted some parts to shared functions. Here, we also do the check that the user didn’t provide any instance keys, and if he did, we throw an error.
5. In `tofu/context_plan.go`, before we build the graph of the plan, we use `refactoring/remove_statements.go` to go over all the removed blocks in each config module recursively and construct their absolute address. We also validate that the resource/module referenced in the removed block doesn’t have a configuration block in the config. We pass this data to `tofu/graph_builder_plan.go`, which will pass this data to the appropriate nodes.
6. Because the resource/module referenced in the removed block doesn’t have a configuration block in the config, it means that the relevant resources can only produce orphan or deposed nodes in the plan graph (`NodePlannableResourceInstanceOrphan` and `NodePlanDeposedResourceInstanceObject`). We added a logic that instead of always destroying these resources, we first check if they are included in `EndpointsToForget` (which we calculated in the previous step). If they are, we run the `planForget()` function. The `planForge()` produces a diff for the resources with the `forget` action.
7. Later, in the apply stage, based on the diff of the plan we create `tofu/node_resource_forget.go`. Executing this node will remove the resource from the state but not destroy it. It's heavily based on `tofu/node_resource_destroy.go`.

Resolves #1032

## Target Release

1.7.0


internal/addrs/module.go
```diff
@@ -168,3 +171,107 @@ func (m Module) Ancestors() []Module {
...
+				Name:     tt.Name,
+				SrcRange: tt.SrcRange,
+			}
+		}
+	}
+
+	return module, retRemain, diags
+}
+
+func getModuleName(remain hcl.Traversal) (moduleName string, isValidModule bool, diags tfdiags.Diagnostics) {
```
cam72cam, line 227: Is there any reason to return isValidModule when all callers check diags.HasErrors() ?

cam72cam: Additionally, would it make sense to return the remaining traversal and remove the duplicated [2:] in both call sites?

internal/command/jsonformat/plan.go
```diff
@@ -455,6 +463,14 @@ func resourceChangeComment(resource jsonplan.ResourceChange, action plans.Action
...
 			buf.WriteString("\n  # (left over from a partially-failed replacement of this instance)")
 		}
+	case plans.Forget:
+		buf.WriteString(fmt.Sprintf("[bold]  # %s[reset] will be removed from the OpenTofu state [bold][red]but will not be destroyed[reset]", dispAddr))
+
+		if len(resource.Deposed) != 0 {
+			// Some extra context about this unusual situation.
+			buf.WriteString("\n  # (left over from a partially-failed replacement of this instance)")
+		}
+		break
```
cam72cam, line 473: Calling out the github consistency check, unnecessary break

internal/refactoring/remove_statement.go
```diff
@@ -0,0 +1,122 @@
...
+		case addrs.Module:
+			// Get the absolute address of the module by appending the module config address
+			// to the module itself
+			var absModule = make(addrs.Module, 0, len(modAddr)+len(FromAddress))
+			absModule = append(absModule, modAddr...)
+			absModule = append(absModule, FromAddress...)
+			removedEndpoint = &RemoveStatement{From: absModule, DeclRange: tfdiags.SourceRangeFromHCL(rc.DeclRange)}
+
+		default:
+			panic(fmt.Sprintf("unhandled address type %T", FromAddress))
```
cam72cam, line 63: Would this be hit if a data block was referenced or is that handled by ConfigResource?

internal/tofu/context_plan.go
```diff
@@ -664,9 +664,18 @@ func (c *Context) planWalk(config *configs.Config, prevRunState *states.State, o
 }
 
 func (c *Context) planGraph(config *configs.Config, prevRunState *states.State, opts *PlanOpts) (*Graph, walkOperation, tfdiags.Diagnostics) {
+	var diags tfdiags.Diagnostics
```
cam72cam, line 667: This whole block can be simplified by not pre-creating diags/graph and returning nil, walkPlan, forgetDiags below.  That would be equivalent to this current refactoring.

cam72cam:
Overall, this looks really good.  Looking forward to seeing all the added test cases and documentation!

internal/refactoring/remove_statement.go
Evi1Pumpkin: It'll not. However, I did miss the validation for the `data` block. And I added it [here](https://github.com/opentofu/opentofu/pull/1158/commits#:~:text=if%20removed%20target%20is%20data%20source%2C%20throw%20an%20error)

internal/addrs/module.go
Evi1Pumpkin: > Is there any reason to return isValidModule when all callers check diags.HasErrors() ?

It might just not be a module at all. If the address is `null_resource.example`, then I don't want to throw an error, but I do want to tell the caller functions they need to stop looping on the address, trying to parse the module address.
Maybe the naming here is off; if I call It `isModule` instead of `isValidModule,` will it make more sense?

> Additionally, would it make sense to return the remaining traversal and remove the duplicated [2:] in both call sites?

Makes sense. I'll do it now.

internal/addrs/module.go
```diff
@@ -168,3 +171,107 @@ func (m Module) Ancestors() []Module {
...
+	remain := traversal
+	var module Module
+	var diags tfdiags.Diagnostics
+
+	for len(remain) > 0 {
+		moduleName, isValidModule, moduleNameDiags := getModuleName(remain)
+		diags = diags.Append(moduleNameDiags)
+
+		if !isValidModule || diags.HasErrors() {
+			break
```
Yantrio, line 186: instead of breaking, we could probably just return here, and remove the `if len(remain) > 0` later on.

internal/addrs/module_instance.go
```diff
@@ -85,61 +85,22 @@ func parseModuleInstancePrefix(traversal hcl.Traversal) (ModuleInstance, hcl.Tra
...
-				Detail:   "Module address prefix must be followed by dot and then a name.",
-				Subject:  remain[0].SourceRange().Ptr(),
-			})
-			break LOOP
-		}
-
-		if next != "module" {
-			break
-		}
+		moduleName, isValidModule, moduleNameDiags := getModuleName(remain)
```
Yantrio, line 89: i love that we're abstracting out this logic 👍

internal/addrs/removable.go
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) HashiCorp, Inc.
```
Yantrio, line 1: why are we adding new files with the hashicorp copyright?

internal/addrs/remove_endpoint.go
```diff
@@ -0,0 +1,99 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package addrs
+
+import (
+	"github.com/hashicorp/hcl/v2"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// RemoveEndpoint is to ConfigMoveable what Target is to Targetable:
```
Yantrio, line 11:
```suggestion
// RemoveEndpoint is to ConfigRemovable what Target is to Targetable:
```

internal/addrs/remove_endpoint.go
```diff
@@ -0,0 +1,99 @@
...
+// a wrapping struct that captures the result of decoding an HCL
+// traversal representing a relative path from the current module to
+// a removable object. It is very similar to MoveEndpoint.
+//
+// Its purpose is for the "from" address in a "removed" statement
+// in the configuration.
```
Yantrio, line 17: this bit of the comment doesn't make sense. Is it copied from the move stuff?

internal/addrs/remove_endpoint.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// the representation of our relative address as a ConfigRemovable
+	RelSubject ConfigRemovable
+}
+
+func (e *RemoveEndpoint) String() string {
+	// Our internal pseudo-AbsMoveable representing the relative
+	// address (either ModuleInstance or AbsResourceInstance) is
+	// a good enough proxy for the relative move endpoint address
+	// serialization.
```
Yantrio, line 35: this comment seems to be for move stuff, not remove. 

is it okay for us to re-use this comment here? 

Also, do we need to implement a psuedo-AbsRemovable in module_call.go?

internal/command/views/json/hook.go
```diff
@@ -317,6 +317,10 @@ func startActionVerb(action plans.Action) string {
 		// This is not currently possible to reach, as we receive separate
 		// passes for create and delete
 		return "Replacing"
+	case plans.Forget:
+		// This is not currently possible to reach, as we receive separate
+		// passes for create and delete
```
Yantrio, line 322: comment doesn't match logic.

internal/configs/removed.go
```diff
@@ -0,0 +1,46 @@
...
+
+package configs
+
+import (
+	"github.com/hashicorp/hcl/v2"
+	"github.com/opentofu/opentofu/internal/addrs"
+)
+
+type Removed struct {
+	From *addrs.RemoveEndpoint
```
Yantrio, line 12: The variable name `From` feels strange here, maybe because we're mimicing some of the `move` logic.

Maybe the noun to use here as the target of removal is `Target` or `Address` or `Resource`?

internal/refactoring/remove_statement.go
```diff
@@ -0,0 +1,122 @@
...
+type RemoveStatement struct {
+	From      addrs.ConfigRemovable
+	DeclRange tfdiags.SourceRange
+}
+
+// GetEndpointsToForget recurses through the modules of the given configuration
+// and returns an array of all "removed" addresses within, in a
+// deterministic but undefined order.
+// We also validate that the removed modules/resources configuration blocks were removed.
+func GetEndpointsToForget(rootCfg *configs.Config) ([]addrs.ConfigRemovable, tfdiags.Diagnostics) {
```
Yantrio, line 24: We seem to flip between Remove and `Forget` here, and I understand there's a nuance, but It's a bit confusing when we're calling it `Removable` here.

Maybe we should just stick to `GetEndpointsToRemove`?

internal/refactoring/remove_statement.go
```diff
@@ -0,0 +1,122 @@
...
+		// validate that a resource/module with this address doesn't exist in the config
+		switch fromAddr := fromAddr.(type) {
+		case addrs.ConfigResource:
+			moduleConfig := cfg.Descendent(fromAddr.Module)
+			if moduleConfig != nil && moduleConfig.Module.ResourceByAddr(fromAddr.Resource) != nil {
+				diags = diags.Append(&hcl.Diagnostic{
+					Severity: hcl.DiagError,
+					Summary:  "Removed resource block still exists",
+					Detail: fmt.Sprintf(
+						"This statement declares a removal of the resource %s, but it still exists in the configuration. Please remove the resource block.",
```
Yantrio, line 98:
```suggestion
					Summary:  "Removed resource block still exists",
					Detail: fmt.Sprintf(
						"This statement declares a removal of the resource %s, but this resource block still exists in the configuration. Please remove the resource block.",
```

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -346,6 +346,55 @@ func (n *NodeAbstractResourceInstance) writeResourceInstanceStateImpl(ctx EvalCo
...
+	// If there is no state or our attributes object is null then the resource
+	// was already forgotten.
+	if currentState == nil || currentState.Value.IsNull() {
+		// We still need to generate a NoOp change, because that allows
+		// outside consumers of the plan to distinguish between us affirming
+		// that we checked something and concluded no changes were needed
+		// vs. that something being entirely excluded e.g. due to -target.
+		noop := &plans.ResourceInstanceChange{
+			Addr:        absAddr,
+			PrevRunAddr: n.prevRunAddr(ctx),
```
Yantrio, line 365: Just for clarification, can we rename AND remove at the same time? This feels a little complicated. If we didn't have to handle renames and deposed keys it would be simpler.

Yantrio (changes requested):

internal/addrs/module.go
Evi1Pumpkin: I tested this option, and it throws an error. I need to keep [this](https://github.com/opentofu/opentofu/blob/ca1d39b88bf7d9c7522bf5a0c73807aa87d332a1/internal/addrs/module.go#L209-L222) logic. 
Let's say I have the following address `module.example.null_resource.example_resource`.  The `TraverseRoot` in this case is `module`. 
After I parse the module part of the address, I want to return the remaining `null_resource.example_resource`. If I won't set `null_resource` as the new root, later in `ParseRemoveEndpoint()` -> `parseResourceUnderModule()` I won't be able to access the Root [here](https://github.com/opentofu/opentofu/blob/ca1d39b88bf7d9c7522bf5a0c73807aa87d332a1/internal/addrs/parse_target.go#L128) and it'll throw an error instead.

internal/addrs/remove_endpoint.go
Evi1Pumpkin: ✅

internal/addrs/remove_endpoint.go
Evi1Pumpkin: No, sorry, it's just poor English. Fixed it.

internal/command/views/json/hook.go
Evi1Pumpkin: Removed it.

internal/configs/removed.go
Evi1Pumpkin: The `From` variable name here references the `From` attribute of the removed block:
```
removed {
  from = module.example.null_resource.example_resource
}
```

It's using the same convention as in `moved.go` and `import.go`. So, IMO it makes sense, but let me know if you think otherwise.

internal/refactoring/remove_statement.go
Evi1Pumpkin: ✅

internal/addrs/removable.go
Evi1Pumpkin: I waited for the [final copywrite notice](https://github.com/opentofu/opentofu/issues/454#issuecomment-1904213569).
Updated all the newly added files with the OpenTofu and HashiCorp headers, as [discussed on Slack](https://opentofucommunity.slack.com/archives/C05R0FD0VRU/p1706012969801869?thread_ts=1705997490.521719&cid=C05R0FD0VRU).

internal/tofu/node_resource_abstract_instance.go
ronnyorot: > Just for clarification, can we rename AND remove at the same time? 

That's a good question. I had a todo to check all possible moved block options with the removed block, and I found one scenario in which you can move and delete at the same time.
From this situation:
```
resource "null_resource" "example_resource" {
   provisioner "local-exec" {
     command = "echo 'Hello from null resource'"
   }
 }
```
To this:
```
removed {
   from = null_resource.example_resource2
}

moved {
   from = null_resource.example_resource
   to = null_resource.example_resource2
}

/* resource "null_resource" "example_resource" {
   provisioner "local-exec" {
     command = "echo 'Hello from null resource'"
   }
 } */

  /* resource "null_resource" "example_resource2" {
	   provisioner "local-exec" {
	     command = "echo 'Hello from null resource'"
	   }
   } */
```

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -346,6 +346,54 @@ func (n *NodeAbstractResourceInstance) writeResourceInstanceStateImpl(ctx EvalCo
 	return nil
 }
 
+// planForget returns a removed from state diff.
+func (n *NodeAbstractResourceInstance) planForget(ctx EvalContext, currentState *states.ResourceInstanceObject, deposedKey states.DeposedKey) (*plans.ResourceInstanceChange, tfdiags.Diagnostics) {
+	var diags tfdiags.Diagnostics
+	var plan *plans.ResourceInstanceChange
+
+	// TODO: remove - this is an unreachable code, the state of this resource cannot be null
```
Evi1Pumpkin, line 354: I'll remove it before I merge the PR. This is an unreachable code because this scenario doesn't make sense. The forgotten resources must be orphans or deposed, which means they exist in the state.
But I'm still trying to determine if I missed any relevant scenario.

internal/tofu/context_plan.go
Evi1Pumpkin: ✅

internal/addrs/module.go
Evi1Pumpkin: > Additionally, would it make sense to return the remaining traversal and remove the duplicated [2:] in both call sites?

Although, on second thought, I'm not a fan of having 4 return values for this function 😬

internal/tofu/node_resource_deposed_test.go
```diff
@@ -217,3 +215,82 @@ func TestNodeDestroyDeposedResourceInstanceObject_ExecuteMissingState(t *testing
...
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+
+	if !state.Empty() {
+		t.Fatalf("resources left in state after forget")
+	}
+}
+
+func initMockEvalContext(resourceAddrs string, deposedKey states.DeposedKey) (*MockEvalContext, *MockProvider) {
```
Evi1Pumpkin, line 246: Extracted common logic from here.

internal/tofu/node_resource_abstract_instance.go
Evi1Pumpkin: Removed it.

internal/tofu/node_resource_abstract_instance.go
Evi1Pumpkin: Also, I added a test for the `moved` and the `removed` blocks at the same time - [here](https://github.com/opentofu/opentofu/pull/1158/files#diff-29c9b1b9ff61214376c78ddd5be7b929d5864b8f131aa9cd386c56c9585c0a2fR5807).

internal/addrs/module.go
```diff
@@ -170,3 +173,109 @@ func (m Module) Ancestors() []Module {
 func (m Module) configMoveableSigil() {
 	// ModuleInstance is moveable
 }
+func (m Module) configRemovableSigil() {}
+
+func parseModulePrefix(traversal hcl.Traversal) (Module, hcl.Traversal, tfdiags.Diagnostics) {
```
janosdebugs, line 178: Maybe I'm not well-versed enough in the codebase, but I don't know what a module prefix is and what the naming rules would be to parse them by. It may be useful to add a comment here explaining what the naming rules are we are parsing by.

internal/addrs/removable.go
```diff
@@ -0,0 +1,26 @@
...
+// ConfigRemovable is an interface implemented by address types that represents
+// the destination of a "removed" statement in configuration.
+//
+// Note that ConfigRemovable might represent:
+//  1. An absolute address relative to the root of the configuration.
+//  2. A direct representation of these in configuration where the author gives an
+//     address relative to the current module where the address is defined.
+type ConfigRemovable interface {
+	Targetable
+	configRemovableSigil()
```
janosdebugs, line 17: What's the reason behind adding a non-exported member in an exported interface? I'm personally not a fan of exported interfaces with non-exported members as it makes it impossible to implement this interface outside of this package, in which case an interface may not be needed.

internal/addrs/resource.go
```diff
@@ -459,6 +459,9 @@ func (v ConfigResource) CheckableKind() CheckableKind {
 	return CheckableResource
 }
 
+func (r ConfigResource) configRemovableSigil() {
```
janosdebugs, line 462: What is this function supposed to do?

internal/command/jsonformat/plan.go
```diff
@@ -453,6 +461,13 @@ func resourceChangeComment(resource jsonplan.ResourceChange, action plans.Action
 		case jsonplan.ResourceInstanceDeleteBecauseEachKey:
 			buf.WriteString(fmt.Sprintf("\n  # (because key [%s] is not in for_each map)", resource.Index))
 		}
+		if len(resource.Deposed) != 0 {
+			// Some extra context about this unusual situation.
```
janosdebugs, line 465: I'd like to know more about how this situation is unusual. Maybe an example would help?

internal/plans/internal/planproto/planfile.proto
```diff
@@ -116,6 +116,7 @@ enum Action {
     DELETE = 5;
     DELETE_THEN_CREATE = 6;
     CREATE_THEN_DELETE = 7;
+    FORGET = 8;
```
janosdebugs, line 119: I wonder if this breaks compatibility with TACOS and/or makes planfiles unreadable by Terraform?

janosdebugs (commented):
I'm sorry, I don't feel even remotely qualified to pass judgement on this PR so I only have a few nitpicks to offer. Overall, there is nothing that jumps out from the code of being somehow wrong.

internal/addrs/module.go
Evi1Pumpkin: You are correct! I wanted to add an explanation here and forgot. I'll add it.

internal/addrs/resource.go
Evi1Pumpkin: We can see those empty sigil functions in different places in the codebase (for reference, the [absMovableSigil()](https://github.com/opentofu/opentofu/blob/00320dcd00c7011d1025373d63bc4c3002a306d1/internal/addrs/moveable.go#L17)). They are always empty, and from what I understand, their usage is to make sure only specific structs can implement the interface. 
For example, although on a technical level, different address types might be able to implement the `AbsMoveable` interface, logically it's not the case. Only some address types can really be treated as moveable addresses. The address type [ModuleCall](https://github.com/opentofu/opentofu/blob/00320dcd00c7011d1025373d63bc4c3002a306d1/internal/addrs/module_call.go#L14) can technically fall under the `AbsMoveable` interface, but in implementation `tofu` doesn't want to treat this address as a movable address. That's why it doesn't define the `absMoveableSigil()`.
Same logic for the `ConfigRemovable` interface.
I'm not a GO professional, so I'm not entirely sure if this is considered a best practice or not.

janosdebugs: I believe you are correct, let's keep it this way even if I personally feel somewhat uncomfortable with this technique. (It feels like a hack to get a `final` keyword-equivalent functionality in Go.)

internal/command/jsonformat/plan.go
Evi1Pumpkin: I altered the comment to contain a better explanation.
But I'll add a more extended version here:
Let's say you have a resource set with a `create_before_destroy` [lifecycle configuration](https://opentofu.org/docs/language/meta-arguments/lifecycle/#syntax-and-arguments). You do a change that requires replacing your resource. Because the `create_before_destroy` is set, tofu first sets the existing resource as deposed (so it can destroy it later), creates the new instance of the resource, and then fails when it tries to destroy the deposed instance.
In this scenario, the deposed instance will be saved as a leftover in the state. On the next `apply`, tofu will try to destroy/forget it, and it'll mention it specifically in the plan.

internal/addrs/remove_endpoint.go
```diff
@@ -0,0 +1,78 @@
...
+	diags = diags.Append(moreDiags)
+	if diags.HasErrors() {
+		return nil, diags
+	}
+
+	if riAddr.Resource.Mode == DataResourceMode {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Data source address is not allowed",
+			Detail:   "Data sources cannot be destroyed, and therefore, 'removed' blocks are not allowed to target them. To remove data sources from the state, you can remove the data source block from the configuration.",
```
Yantrio, line 67: Another tiny tiny nitpick
```suggestion
			Detail:   "Data sources cannot be destroyed, and therefore, 'removed' blocks are not allowed to target them. To remove data sources from the state, you should remove the data source block from the configuration.",
```

internal/addrs/resource.go
Yantrio: Maybe we could add a comment here to mention why it's there. So it doesn't get removed in the future.

Maybe something like `empty function to fulfil the requirements of X interface`

internal/command/jsonplan/plan.go
Yantrio, line 1: There are multiple places in this file that discuss actions, and list them.

For example: https://github.com/opentofu/opentofu/blob/d79ddf27bfb6a065384508e4fc9ca03df9d24fcf/internal/command/jsonplan/plan.go#L94

We should probably add Forget to these comments

internal/plans/internal/planproto/planfile.proto
Yantrio: As far as I'm aware, most TACOS read the json output using a library similar to [terraform-json](https://github.com/hashicorp/terraform-json) and work from the json representation of the plan and not this.

Sidenote: once this is in, we should put in a PR to that repo to add the Forget action

Yantrio (changes requested):
Overall this is looking great and will be an awesome addition!

First, I've left some small nitpick comments.

Secondly: I've reached out internally, but I also wanted to raise it here. I'm concerned we're not handling the path of Create + Forget in a single run.

I suspect that some users may want to do this to use their config to create objects but never store them in their state. Do we have a plan for this?

internal/command/jsonformat/plan.go
```diff
@@ -454,7 +462,20 @@ func resourceChangeComment(resource jsonplan.ResourceChange, action plans.Action
...
 		if len(resource.Deposed) != 0 {
-			// Some extra context about this unusual situation.
+			// In the case where we partially failed to replace a resource
+			// configured with 'create_before_destroy' in a previous apply and
+			// the deposed instance is still in the state, we give some extra
+			// context about this unusual situation.
+			buf.WriteString("\n  # (left over from a partially-failed replacement of this instance)")
+		}
+	case plans.Forget:
+		buf.WriteString(fmt.Sprintf("[bold]  # %s[reset] will be removed from the OpenTofu state [bold][red]but will not be destroyed[reset]", dispAddr))
```
cam72cam, line 472: Nit: I know different people have thoughts on case fallthrough in go, but that might be a easy way to combine a lot of these two blocks?

Happy to merge without this!

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -348,6 +348,32 @@ func (n *NodeAbstractResourceInstance) writeResourceInstanceStateImpl(ctx EvalCo
 	return nil
 }
 
+// planForget returns a removed from state diff.
+func (n *NodeAbstractResourceInstance) planForget(ctx EvalContext, currentState *states.ResourceInstanceObject, deposedKey states.DeposedKey) (*plans.ResourceInstanceChange, tfdiags.Diagnostics) {
```
cam72cam, line 352: Why does this return diags?  I don't see anywhere that they are used?  Is this following the prototype of another similar function?

internal/addrs/parse_target.go
```diff
@@ -80,17 +80,97 @@ func parseResourceInstanceUnderModule(moduleAddr ModuleInstance, remain hcl.Trav
...
+			Severity: hcl.DiagError,
+			Summary:  "Invalid address",
+			Detail:   "Unexpected extra operators after address.",
+			Subject:  remain[1].SourceRange().Ptr(),
+		})
+		return AbsResourceInstance{}, diags
+	}
+}
+
+func parseResourceUnderModule(moduleAddr Module, remain hcl.Traversal) (ConfigResource, tfdiags.Diagnostics) {
```
cube2222, line 126: Could you please add comments to this function and the one below describing what their purpose is? I *think* I understand them, but I'm not entirely sure, even after reading it 3 times.

cube2222: Fwiw, after reading more files, I now see this is expected to be used preceded by the module prefix parsing function, so it's clearer to me. But I still think a comment indicating that (that this only parses the last part and expects no module prefix) would be useful.

internal/command/jsonformat/plan.go
cube2222: Please, please, don't :P

I'm worried someone will add a case not noticing the fallthrough and wreak havoc.

internal/plans/action_string.go
```diff
@@ -15,26 +15,32 @@ func _() {
 	_ = x[DeleteThenCreate-8723]
 	_ = x[CreateThenDelete-177]
 	_ = x[Delete-45]
+	_ = x[Forget-46]
 }
 
 const (
 	_Action_name_0 = "NoOp"
 	_Action_name_1 = "Create"
-	_Action_name_2 = "Delete"
```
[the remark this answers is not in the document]
cube2222: What the heck is this, haha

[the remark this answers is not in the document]
cube2222: Maybe it does it for codepoints which are directly next to each-other?

internal/tofu/context_plan.go
```diff
@@ -694,6 +699,7 @@ func (c *Context) planGraph(config *configs.Config, prevRunState *states.State,
 			ExternalReferences: opts.ExternalReferences,
 			ImportTargets:      opts.ImportTargets,
 			GenerateConfigPath: opts.GenerateConfigPath,
+			EndpointsToRemove:  endpointsToRemove,
```
cube2222, line 702: To be fair, just to stay consistent with the remaining code here (specifically ImportTargets) I would propose moving endpointsToRemove into opts.EndpointsToRemove and initializing it in the plan function in line 303, similar to how imports are prepared there.

internal/tofu/transform_diff.go
```diff
@@ -211,6 +213,26 @@ func (t *DiffTransformer) Transform(g *Graph) error {
...
+					NodeAbstractResourceInstance: abstract,
+					DeposedKey:                   dk,
+				}
+				log.Printf("[TRACE] DiffTransformer: %s will be represented for removal from the state by %s", addr, dag.VertexName(node))
+			} else {
+				node = &NodeForgetDeposedResourceInstanceObject{
+					NodeAbstractResourceInstance: abstract,
+					DeposedKey:                   dk,
+				}
+				log.Printf("[TRACE] DiffTransformer: %s deposed object %s will be represented for for removal from the state by %s", addr, dk, dag.VertexName(node))
```
cube2222, line 230: "for for" typo

cube2222 (approved):
This looks excellent overall!

I've left a couple nit comments. Also, once you're documenting this, please also mention the deposed functionality and how you can forget deposed resources instead of destroying them.

internal/plans/action_string.go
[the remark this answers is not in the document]
Evi1Pumpkin: Yes, the `-` and `.` ASCII decimal values are 45 and 46, respectively.
When I changed the Forget action symbol to something other than `.`, it was generated normally.
But it doesn't really matter, I didn't see any impact on the running code.

internal/addrs/module.go
```diff
@@ -170,3 +173,109 @@ func (m Module) Ancestors() []Module {
...
+			}
+		}
+	}
+
+	return module, retRemain, diags
+}
+
+func getModuleName(remain hcl.Traversal) (moduleName string, isModule bool, diags tfdiags.Diagnostics) {
+	var next string
+	switch tt := remain[0].(type) {
```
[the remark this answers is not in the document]
Evi1Pumpkin: Great suggestion, I added a safeguard.

internal/command/jsonplan/plan.go
Evi1Pumpkin: ✅

internal/command/jsonformat/plan.go
Evi1Pumpkin: It might reduce a few lines, but I agree with @cube2222 on this one. Plus, I think it's clearer when it's more explicit.

internal/tofu/node_resource_abstract_instance.go
Evi1Pumpkin: They are actually used [here](https://github.com/opentofu/opentofu/pull/1158/files/f0e276135196719cfde459ef80bdf64346d0d1ba#diff-846b5b253854eec7652df68e77a9ed83ff5346529c138439120a9998a2bf98fdR152) and [here](https://github.com/opentofu/opentofu/pull/1158/files/f0e276135196719cfde459ef80bdf64346d0d1ba#diff-f305d1d715be433c5e7d826409091c8f34b27dfebf25d488b3fadcb6145c640aR155) 🙃

internal/addrs/parse_target.go
Evi1Pumpkin: ✅

internal/tofu/transform_diff.go
Evi1Pumpkin: ✅

internal/tofu/context_plan.go
Evi1Pumpkin: Done ✅

internal/tofu/node_resource_abstract_instance.go
cam72cam: Thanks!

cam72cam: Sorry, I should have specified that diags is never set within this function and always returns null

Yantrio (approved):

schue:
Does this InfoWorld article have a legitimate complaint?
https://www.infoworld.com/article/3714980/opentofu-may-be-showing-us-the-wrong-way-to-fork.html

Where did the MPL2 version of the file come from?
The article was written by the developer relations manager for MongoDB.

dylanmtaylor:
> Does this InfoWorld article have a legitimate complaint?
> https://www.infoworld.com/article/3714980/opentofu-may-be-showing-us-the-wrong-way-to-fork.html
> 
> Where did the MPL2 version of the file come from?
> The article was written by the developer relations manager for MongoDB.

If so, this is alarming and this PR needs to be reverted. OpenTofu can't just steal code from Terraform.

marcinwyszynski:
@schue @dylanmtaylor 

The team is aware of these accusations and is currently working on an official response. Please stay tuned.

Issue #1144, opened 2024-01-18
Random race condition in test workflows (in package internal/command)


StephanHCB:
### OpenTofu Version

```shell
commit fcd241c0e4 (mainline from a few days ago)
```

### OpenTofu Configuration Files

n/a

### Debug Output

```
==================
WARNING: DATA RACE
Write at 0x00c00032e670 by goroutine 6749:
  github.com/opentofu/opentofu/internal/tofu.(*MockProvider).ConfigureProvider()
      /home/runner/work/opentofu/opentofu/internal/tofu/provider_mock.go:247 +0xe4
  github.com/opentofu/opentofu/internal/tofu.(*BuiltinEvalContext).ConfigureProvider()
      /home/runner/work/opentofu/opentofu/internal/tofu/eval_context_builtin.go:204 +0x430
  github.com/opentofu/opentofu/internal/tofu.(*NodeApplyableProvider).ConfigureProvider()
      /home/runner/work/opentofu/opentofu/internal/tofu/node_provider.go:166 +0x1141
  github.com/opentofu/opentofu/internal/tofu.(*NodeApplyableProvider).Execute()
      /home/runner/work/opentofu/opentofu/internal/tofu/node_provider.go:45 +0x7a8
  github.com/opentofu/opentofu/internal/tofu.(*ContextGraphWalker).Execute()
      /home/runner/work/opentofu/opentofu/internal/tofu/graph_walk_context.go:143 +0x159
  github.com/opentofu/opentofu/internal/tofu.(*Graph).walk.func1()
      /home/runner/work/opentofu/opentofu/internal/tofu/graph.go:78 +0x38f
  github.com/opentofu/opentofu/internal/dag.(*Walker).walkVertex()
      /home/runner/work/opentofu/opentofu/internal/dag/walk.go:384 +0x426
  github.com/opentofu/opentofu/internal/dag.(*Walker).Update.func3()
      /home/runner/work/opentofu/opentofu/internal/dag/walk.go:307 +0x5d

Previous read at 0x00c00032e670 by goroutine 6745:
  github.com/opentofu/opentofu/internal/tofu.(*MockProvider).ApplyResourceChange()
      /home/runner/work/opentofu/opentofu/internal/tofu/provider_mock.go:402 +0x12f
  github.com/opentofu/opentofu/internal/tofu.(*NodeAbstractResourceInstance).apply()
[...]

PR #1182, opened 2024-01-24
Fix data race condition due to shared slice (#1144)

desmaraisp:
Addresses a possible data race due to sharing a single slice across multiple goroutines

Relates To #1144 

## Target Release

1.7.0


.github/workflows/checks - race.yml
```diff
@@ -0,0 +1,32 @@
+name: Quick Checks (race)
+
+on:
+  workflow_dispatch: 
+
+jobs:
+  race-tests:
```
cam72cam, line 7: Do we need this dedicated workflow?  Can this be removed now that the intermittent issue has been identified?

desmaraisp: I figured it would be useful for the review process, but you're right, I can remove it. I'll just past a copy in the related issue so that if we ever face a similar hard-to-reproduce issue, we can simply add it back to the repo, run it, then remove it

desmaraisp: Workflow was deleted as per feedback.

cam72cam (approved):

Yantrio:
Hi @desmaraisp , would you mind signing off on your commits please to fulfil the DCO requirement.

you can find instructions on how to this by pressing `Details` next to the failed DCO check below.

Thanks!

desmaraisp:
Done. Commits are now signed. Sorry about missing that part of the contrib instructions

cam72cam:
Thanks!

PR #1197, opened 2024-01-25
Terraform compatibility table

janosdebugs:


cam72cam:
Should also mention functions added since 1.6

PR #1200, opened 2024-01-26
Support the XDG Base Directory Specification

### LINKED ISSUE #303: Support the XDG Base Directory Specification

avorima:
This PR introduces support for XDG environment variables. `XDG_CONFIG_HOME` for config and `XDG_DATA_HOME` for plugins.

If `XDG_CONFIG_HOME` is set and `~/.tofurc` or `~/.terraformrc` don't exist, `XDG_CONFIG_HOME/opentofu/tofurc` will be used as config file.
If `XDG_DATA_HOME` is set, plugins will be searched in `XDG_DATA_HOME/opentofu` as well as the default `~/.terraform.d`.

This ensures that existing config and plugins will still get used, but allow users to gradually migrate to the new structure, with and without setting `XDG_*` environment variables.

The CLI ensures that the intermediate config directory exists, so if `XDG_CONFIG_HOME` is set and the default `~/.terraform.d` directory doesn't exist `XDG_CONFIG_HOME/opentofu` is created.

Only affects unix* platforms.

Resolves #303

## Target Release

1.7.0

### DIFF
--- a/internal/command/cliconfig/config_unix.go
+++ b/internal/command/cliconfig/config_unix.go
@@ -45,7 +45,7 @@ func configDir() (string, error) {
 	return configDir, nil
 }
 
-func pluginDir() (string, error) {
+func dataDir() (string, error) {
 	dir, err := homeDir()
 	if err != nil {
 		return "", err
--- a/internal/command/cliconfig/config_unix.go
+++ b/internal/command/cliconfig/config_unix.go
@@ -45,6 +45,20 @@ func configDir() (string, error) {
 	return configDir, nil
 }
 
+func pluginDir() (string, error) {
+	dir, err := homeDir()
+	if err != nil {
+		return "", err
+	}
+
+	configDir := filepath.Join(dir, ".terraform.d")
+	if xdgDir := os.Getenv("XDG_DATA_HOME"); xdgDir != "" && !pathExists(configDir) {
+		configDir = filepath.Join(xdgDir, "opentofu")
+	}
+
+	return configDir, nil
+}
+
 func homeDir() (string, error) {
 	// First prefer the HOME environmental variable
 	if home := os.Getenv("HOME"); home != "" {
--- a/internal/command/cliconfig/config_unix.go
+++ b/internal/command/cliconfig/config_unix.go
@@ -13,6 +13,13 @@ import (
 	"path/filepath"
 )
 
+// Default directories to use when XDG_* is not defined
+// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+const (
+	defaultConfigDir = ".config"
+	defaultDataDir   = ".local/share"
+)
+
 func configFile() (string, error) {
 	dir, err := homeDir()
 	if err != nil {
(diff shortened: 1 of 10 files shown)


internal/command/cliconfig/config_unix.go
```diff
@@ -13,6 +13,13 @@ import (
 	"path/filepath"
 )
 
+// Default directories to use when XDG_* is not defined
+// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+const (
+	defaultConfigDir = ".config"
+	defaultDataDir   = ".local/share"
```
cam72cam, line 20: Per https://github.com/opentofu/opentofu/issues/303#issuecomment-1906209913  I think we don't want to add the non XDG_ fallbacks at this point.

avorima: Sure. I only touched `configDir` because it's the only function that has a side-effect when things aren't found. I reverted back to explicit XDG lookup.

cam72cam (approved):
Thanks @avorima !

internal/command/cliconfig/config_unix.go
```diff
@@ -22,6 +22,11 @@ func configFile() (string, error) {
 	newConfigFile := filepath.Join(dir, ".tofurc")
 	legacyConfigFile := filepath.Join(dir, ".terraformrc")
 
+	if xdgDir := os.Getenv("XDG_CONFIG_HOME"); xdgDir != "" && !pathExists(legacyConfigFile) && !pathExists(newConfigFile) {
+		// a fresh install should not use terraform naming
+		return filepath.Join(xdgDir, "opentofu", "tofurc"), nil
```
cube2222, line 27: Is the omission of a dot in `.tofurc` intentional here?

cube2222 (approved):
Makes sense to me, thanks for working on this!

Left one question.

internal/command/cliconfig/config_unix.go
avorima: Yes. From my understanding the dot is not needed for files in XDG because they (usually) are already in a dot directory, e.g. `~/.config`. For example, git will use `XDG_CONFIG_HOME/git/config`.

cube2222: Makes sense, thanks!

PR #1202, opened 2024-01-26
Run unit tests on various OSes and Architectures on pull request

### LINKED ISSUE #1135: Run unit tests on all architectectures from CI
Currently we are only running [e2e tests](https://github.com/opentofu/opentofu/blob/main/.github/workflows/build.yml) for all architectures. As prompted by #1069, we should also be running unit tests on all architectures on PRs.

IgnorantSapient:
Here is a run of this action in my fork: https://github.com/IgnorantSapient/opentofu/actions/runs/7673240823/job/20915441788?pr=1

Added test matrix for various OSes and Architectures on pull request

Resolves #1135

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -33,13 +33,31 @@ permissions:
 
 jobs:
   unit-tests:
-    name: "Unit Tests"
-    runs-on: ubuntu-latest
-
+    name: Unit tests for ${{ matrix.goos }}_${{ matrix.goarch }}
+    runs-on: ${{ matrix.runson }}
     env:
       TF_APPEND_USER_AGENT: Integration-Test
-
+      GOOS: ${{ matrix.goos }}
+      GOARCH: ${{ matrix.goarch }}
+    strategy:
+      matrix:
+        include:
+          - { runson: ubuntu-latest, goos: linux, goarch: "arm64" }
+          - { runson: ubuntu-latest, goos: linux, goarch: "amd64" }
+          - { runson: ubuntu-latest, goos: linux, goarch: "386" }
+          - { runson: ubuntu-latest, goos: linux, goarch: "arm" }
+          - { runson: macos-latest, goos: darwin, goarch: "amd64" }
+          # - { runson: windows-latest, goos: windows, goarch: "amd64" }
+          # https://github.com/opentofu/opentofu/issues/1201 if fixed
+          #  ^ un-comment the  windows-latest line
+      fail-fast: false
     steps:
+      # 👇🏾 GH actions supports only "AMD64 arch", so we are using this action
+      # for testing on non amd64 envs like 386, arm64 etc...
+      - name: "Set up QEMU"
+        if: matrix.goos == 'linux' && matrix.goarch != 'amd64'
+        uses: docker/setup-qemu-action@v3
+
       - name: "Fetch source code"
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
 
@@ -234,7 +252,6 @@ jobs:
         with:
           path: ".licensei.cache"
           key: licensei-cache-${{ hashFiles('go.sum') }}
-
 #  installation-instructions:
 #    name: "Test Installation Instructions"
 #    runs-on: ubuntu-latest


janosdebugs (approved):
Thank you!

lkwg82:
please add darwin arm64, thx

janosdebugs (changes requested):
Sorry, I have to retract my previous approval. It seems that this doesn't do anything because it only runs on ubuntu-latest, but the `goos` and `goarch` are only used in the name of the test, but don't get passed anywhere to the go runner.

IgnorantSapient:
Hey @janosdebugs, sorry for the delay in getting back to you, I was able to find sometime today to look into this issue, YES! you are right, the code does not do anything to change GOOS and GOARCH besides the title, great catch!

So I got around to fixing it [here](https://github.com/IgnorantSapient/opentofu/commit/967946374bbfc4e534281f458cf143a0a3e91d38) 

Here is the corresponding [actions run](https://github.com/IgnorantSapient/opentofu/actions/runs/7878705885)

But we can not support multiple architectures as the default GitHub runners provided by GitHub actions does not support ARM 64 yet ([Source](https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/))

So I advise that we just go ahead with `{ runson: ubuntu-latest, goos: linux, goarch: "amd64" }`, `{ runson: macos-latest, goos: darwin, goarch: "amd64" }` and `{ runson: windows-latest, goos: windows, goarch: "amd64" }` as these are the only 3 options supported by cloud hosted GitHub actions at this point. Another alternative would be for us to go with self-hosted runners, is this a path that you want me to explore?

After ARM 64 becomes generally available, we can add it.

janosdebugs:
Hi @IgnorantSapient thank you for your work on this. I agree as far as ARM is concerned, but 386 (16 bit). @cube2222 @Yantrio one of you mentioned an example on how to do this, could you please share?

Yantrio:
I'm not sure what options we have for 386 and arm with github action. Maybe we will have to look at hosting something ourselves or shopping around? I'll investigate more!

janosdebugs:
There's some sort of qemu action for Docker, maybe @cube2222 mentioned it?

cube2222:
@janosdebugs Yeah, there's https://github.com/docker/setup-qemu-action which then allows you to easily run docker containers for other architectures.

I'm not sure what you mean with 16-bit 386. The 386 arch builds are 32 bit iirc.

janosdebugs:
@cube2222 Whoops, you're right, my brain was going too far back in time. :D

@IgnorantSapient do you think you could take a look at setup-qemu-action?

IgnorantSapient:
@janosdebugs will experiment with it now.....

IgnorantSapient:
Hey @janosdebugs, and thanks for the pointer @cube2222, I managed to setup, the QEMU action, The great news is, it works well for `amd64, arm64`, but the `386` [tests are failing](https://github.com/IgnorantSapient/opentofu/actions/runs/7891328945/job/21535425132). I think we should go ahead with whatever Architectures that work and add failing architectures to the issue backlog and fix it one at a time.

What are your thoughts?

cube2222:
That's an excellent find @IgnorantSapient (the broken tests)!

I think it makes sense to only include the working ones for now.

Ideally, though I'm not a fan of commented-out code in the codebase, you could leave the 386-specific bit commented out, so that whoever wants to tackle the broken test-cases, can just uncomment it, though not sure what @janosdebugs thinks. I suppose it's just an entry in a matrix.

janosdebugs:
@cube2222 we could mark those tests as not required for merging for now.

cube2222:
@janosdebugs Is that easily doable for separate matrix runs (some runs required, some optional)? Fwiw, I'm not a huge fan of making main red with this. But no strong opinion, really.

janosdebugs:
I believe so. I'd like to make it red so we actually have a verifiable incentive to fix it.

IgnorantSapient:
@janosdebugs I think this is finally ready to be reviewed and merged!

janosdebugs:
Hi @IgnorantSapient thanks for the update, I'm not entirely sure where it tells Docker to run on ARM or x86 here, maybe @cube2222 you could take a look and point us in the right direction?

cam72cam (approved):

cube2222:
Yeah, we have to move the running of unit tests into a docker container with the right platform.

So with the `--platform` argument or something like that. But yeah, unit tests need to be ran in a docker container now.

cam72cam:
@IgnorantSapient apologies for letting this fall behind.  Are you still working on this and do you need any help with it?

IgnorantSapient:
Hey @cam72cam I ran into a few failing tests in the `i386` when running it in these new ENVs. Let us connect on Monday and see if you have any idea on why some fail.

cam72cam (dismissed):

cam72cam:
Looking at the test results on your branch, it clearly shows the failing i386 tests being fixed after we applied the required patch. :+1: 

DM'd about the DCO signoff

cam72cam (approved):

IgnorantSapient:
> Hi @IgnorantSapient thanks for the update, I'm not entirely sure where it tells Docker to run on ARM or x86 here, maybe @cube2222 you could take a look and point us in the right direction?

Hey @janosdebugs, @cam72cam found the answer here:
https://ctrl-c.us/posts/test-goarch.html

cube2222 (approved):

Issue #1033, opened 2023-12-18
Add support for `for_each` in the `import` block


RLRabinowitz:
### OpenTofu Version

```shell
1.6.0-dev
```

### Use Cases

The `import` block could benefit greatly from supporting `for_each`. 

Currently, importing resources that are using `for_each` is not possible / hard using the `import` block for them. Right now, to import those you'd need to use a script running `tofu import` commands

### Attempted Solutions

No possibility of importing `for_each` resources nicely using a single `import` block

### Proposal

By adding `for_each` for the `import` block, you could do the following:

```
import {
  for_each = local.bucket_ids
  id = each.value.id
  to = aws_s3_bucket.bucket[each.value.key]
}

resource "aws_s3_bucket" "bucket" {
  for_each = local.bucket_ids
  ...
}
```

### References

Related issue in the Terraform repository - https://github.com/hashicorp/terraform/issues/33624

PR #1207, opened 2024-01-29
Import ForEach: Prerequisite - Prepare codebase for dynamic addresses for `ImportTarget`s

RLRabinowitz:
Related to https://github.com/opentofu/opentofu/issues/1033

Current plan for implementing `for_each` for the `import` block:
1. Prepare `ImportTarget` types and codebase for more dynamic addresses (addresses that are not fully known from the get-go) **<--- You are here**
2. Allow for having a more dynamic address in the `to` field of an `import` block
3. Add `for_each` field for `import` block, and adapt the code for resolving multiple import targets from the same import configuration

### Background - What this PR does
Once we implement `for_each` for `import`, there are a couple of things that will not be the same for `ImportTarget`, as they are today:
- The address of the import target is not necessarily fully known from the get go (as it could rely on data that would only be evaluated later on)
- A single import block could result in multiple import targets

This PR starts to adapt `ImportTarget`s and the rest of the code for when the addresses might be more dynamic, and for when the number of actual targets to import could not be determined from the get-go

### In detail

- Create a better separation between import targets originating from a `tofu import` command vs import targets originating from an `import` block - See `ImportTarget` and `CommandLineImportTarget`, alongside `EvaluatedConfigImportTarget`
- In order to preserve the validation that all `import` blocks have targeted an actual resource - Store all import targets as they are resolved when walking the graph in `ResolvedImports`, and use these in `postPlanValidateImports`
- For imports originating from the `tofu import` - remove `legacyImportMode`, as there is now no need for it
- Validations
  - Added a validation to make sure all imports are pointing to modules that are actually in the configuration (these imports were simply ignored in 1.6.0)
  - Align all "configuration missing" validation errors with Terraform
## Target Release

1.7.0

### DIFF
--- a/internal/tofu/context_plan.go
+++ b/internal/tofu/context_plan.go
@@ -540,18 +540,8 @@ func (c *Context) postPlanValidateImports(resolvedImports *ResolvedImports, allI
 			return addrParseDiags
 		}
 
-		// TODO - validate behaviour with generateConfig
-
-		// TODO make this error not Sourceless
 		if !allInst.HasResourceInstance(address) {
-			diags = diags.Append(tfdiags.Sourceless(
-				tfdiags.Error,
-				"Cannot import to non-existent resource address", // TODO - align the error message with terraform?
-				fmt.Sprintf(
-					"Importing to resource address %s is not possible, because that address does not exist in configuration. Please ensure that the resource key is correct, or remove this import block.",
-					address,
-				),
-			))
+			diags = diags.Append(importResourceWithoutConfigDiags(address, nil))
 		}
 	}
 	return diags
--- a/internal/tofu/context_plan.go
+++ b/internal/tofu/context_plan.go
@@ -6,6 +6,7 @@ package tofu
 import (
 	"bytes"
 	"fmt"
+	"github.com/hashicorp/hcl/v2"
 	"log"
 	"sort"
 	"strings"
@@ -304,6 +305,8 @@ func (c *Context) plan(config *configs.Config, prevRunState *states.State, opts 
 	}
 
 	opts.ImportTargets = c.findImportTargets(config, prevRunState)
+	importTargetDiags := c.validateImportTargets(config, opts.ImportTargets)
+	diags = diags.Append(importTargetDiags)
 	plan, walkDiags := c.planWalk(config, prevRunState, opts)
 	diags = diags.Append(walkDiags)
 
@@ -537,13 +540,6 @@ func (c *Context) postPlanValidateImports(resolvedImports *ResolvedImports, allI
 			return addrParseDiags
 		}
 
-		// We only care about import target addresses that have a key.
-		// If the address does not have a key, we don't need it to be in config
-		// because are able to generate config.
-		if address.Resource.Key == nil {
-			continue
-		}
-
 		// TODO - validate behaviour with generateConfig
 
 		// TODO make this error not Sourceless
@@ -573,6 +569,23 @@ func (c *Context) findImportTargets(config *configs.Config, priorState *states.S
 		}
 	}
 	return importTargets
+}
+
+func (c *Context) validateImportTargets(config *configs.Config, importTargets []*ImportTarget) (diags tfdiags.Diagnostics) {
+	for _, imp := range importTargets {
+		staticAddress := imp.StaticAddr()
+		descendantConfig := config.Descendent(staticAddress.Module)
+		if descendantConfig == nil {
+			diags = diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Cannot import to non-existent resource address",
+				Detail:   fmt.Sprintf("Importing to resource address %s is not possible, because that address does not exist in configuration. Please ensure that the resource key is correct, or remove this import block.", staticAddress),
+				Subject:  imp.Config.DeclRange.Ptr(),
+			})
+			return
+		}
+	}
+	return
 }
 
 func (c *Context) planWalk(config *configs.Config, prevRunState *states.State, opts *PlanOpts) (*plans.Plan, tfdiags.Diagnostics) {
--- a/internal/tofu/context_plan.go
+++ b/internal/tofu/context_plan.go
@@ -526,30 +526,38 @@ func (c *Context) postPlanValidateMoves(config *configs.Config, stmts []refactor
 // All import target addresses with a key must already exist in config.
 // When we are able to generate config for expanded resources, this rule can be
 // relaxed.
-func (c *Context) postPlanValidateImports(config *configs.Config, importTargets []*ImportTarget, allInst instances.Set) tfdiags.Diagnostics {
-	// FIXME - Fix this post plan validation that imports have all been satisfied by a resource
-	//   From now on import targets can represent multiple response instances, so we'd need a way to figure out what
-	//   those instances were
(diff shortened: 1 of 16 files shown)


internal/tofu/context_plan.go
```diff
@@ -526,23 +526,34 @@ func (c *Context) postPlanValidateMoves(config *configs.Config, stmts []refactor
...
 	var diags tfdiags.Diagnostics
-	for _, it := range importTargets {
+	for resolvedImport := range resolvedImports.imports {
 		// We only care about import target addresses that have a key.
 		// If the address does not have a key, we don't need it to be in config
 		// because are able to generate config.
-		if it.Addr.Resource.Key == nil {
+		address, addrParseDiags := addrs.ParseAbsResourceInstanceStr(resolvedImport.AddrStr)
+		if addrParseDiags.HasErrors() {
+			return addrParseDiags
```
cam72cam, line 537: Nit: add addrParseDiags to diags and use the next continue instead of checking HasErrors()?

RLRabinowitz: I personally don't like that approach much, as it relies on implicit behaviour of `ParseAbsResourceInstanceStr`

Also, there's a change the `if` afterwards will be removed, as it's technically just a way to skip an unnecessary iteration of the for loop, but it's not really needed. Once that `if` is removed, we will lose error handling for `ParseAbsResourceInstanceStr`

RLRabinowitz: I actually ended up removing that `if`, as a I made to how config generation works resulted in it skipping validation for some imports when it shouldn't have

internal/tofu/context_plan2_test.go
```diff
@@ -4946,6 +4903,157 @@ resource "test_object" "a" {
...
+	})
+	if !diags.HasErrors() {
+		t.Fatalf("expected error")
+	}
+	if !strings.Contains(diags.Err().Error(), "Cannot import to non-existent resource address") {
+		t.Fatalf("expected error to be \"Cannot import to non-existent resource address\", but it was %s", diags.Err().Error())
+	}
+}
+
+func TestContext2Plan_importIntoNonExistentConfiguration(t *testing.T) {
```
RLRabinowitz, line 4933: For all these test cases - the behaviour should be the same if config generation was defined. I was wondering if I should run these tests both with and without config generation, but have decided to only run them without config generation for now

Would be interested to hear more opinions on that

internal/tofu/context_plan.go
```diff
@@ -557,15 +555,30 @@ func (c *Context) findImportTargets(config *configs.Config, priorState *states.S
...
 
+func (c *Context) validateImportTargets(config *configs.Config, importTargets []*ImportTarget) (diags tfdiags.Diagnostics) {
+	for _, imp := range importTargets {
+		staticAddress := imp.StaticAddr()
+		descendantConfig := config.Descendent(staticAddress.Module)
+		if descendantConfig == nil {
+			diags = diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Cannot import to non-existent resource address",
+				Detail:   fmt.Sprintf("Importing to resource address %s is not possible, because that address does not exist in configuration. Please ensure that the resource key is correct, or remove this import block.", staticAddress),
```
Yantrio, line 573:
```suggestion
				Detail:   fmt.Sprintf("Importing to resource address '%s' is not possible, because that address does not exist in configuration. Please ensure that the resource key is correct, or remove this import block.", staticAddress),
```

Yantrio (approved):
Overall I'm happy with this code. My comments are just minor nitpicks and requests for tiny refactoring.

However I would have to spend some time to really dive in deep to ensure that evaluation is happening correctly.

I'll approve for now and dive in deeper later on to confirm everything is okay. 👍 nice work

internal/tofu/node_resource_plan.go
```diff
@@ -315,25 +338,12 @@ func (n *nodeExpandPlannableResource) resourceInstanceSubgraph(ctx EvalContext,
...
-						// non-null string.
-						panic(fmt.Sprintf("Invalid import id: %s. This is a bug in OpenTofu; please report it!", diags.Err()))
-					}
-
-					return &graphNodeImportState{
-						Addr:             importTarget.Addr,
-						ID:               importId,
-						ResolvedProvider: n.ResolvedProvider,
-					}
+		for _, c := range commandLineImportTargets {
```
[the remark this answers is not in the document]
RLRabinowitz: The comment still matches with the logic:
If there are import targets originating from an `tofu import` call, then these should be address by creating a `graphNodeImportState` node instead of a `NodePlannableResourceInstance`.
And technically, import targets either all originate from `tofu import` (`CommandLineImportTarget`) or all of them originate from `import` blocks (`Config`). The code does already rely on that - [example here](https://github.com/opentofu/opentofu/blob/64f419e33548d4a243b60c14ffe37be70e3eaf34/internal/tofu/node_resource_abstract.go#L302-L309)
Maybe this is something that could be better captured with some different encapsulation

I will change the "legacy import" mode comment to just say "import CLI" instead, as we do not currently consider the `tofu import` command as "legacy"

internal/tofu/node_resource_abstract.go
```diff
@@ -213,7 +213,12 @@ func (n *NodeAbstractResource) RootReferences() []*addrs.Reference {
 	var root []*addrs.Reference
 
 	for _, importTarget := range n.importTargets {
-		refs, _ := lang.ReferencesInExpr(addrs.ParseRef, importTarget.ID)
+		// References are only possible in import targets originating from an import block
+		if importTarget.Config == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
RLRabinowitz: ae4fd5f486fa1e0e5c7ff757fd86d1c539d3a171

internal/tofu/node_resource_plan.go
```diff
@@ -304,6 +300,33 @@ func (n *nodeExpandPlannableResource) expandResourceInstances(globalCtx EvalCont
...
+			evaluatedConfigImportTargets = append(evaluatedConfigImportTargets, EvaluatedConfigImportTarget{
+				Config: importTarget.Config,
+				ID:     importId,
+			})
+
+			resolvedImports := ctx.ResolvedImports().imports
+			resolvedImports[ResolvedConfigImportsKey{
+				AddrStr: importTarget.Config.To.String(),
+				ID:      importId,
+			}] = true
```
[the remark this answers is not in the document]
RLRabinowitz: ae4fd5f486fa1e0e5c7ff757fd86d1c539d3a171

internal/tofu/node_resource_plan.go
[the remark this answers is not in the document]
RLRabinowitz: e8ed3ea2c02d4540de5673882d74d510bf90e439 - Changed the comment a bit, removing the "legacy" terminology from "import CLI"

internal/tofu/context_plan.go
RLRabinowitz: c503f29be56d92f253c22f1cbfcd9b1056177af8

cube2222 (approved):
I really appreciate the amount of informative comments you've added, they made the code much clearer.

Yantrio:
I also appreciate how you broke the changes up into smaller commits :D made it much easier to review. Thanks @RLRabinowitz . This is my stamp of approval, please feel free to merge when happy.

PR #1208, opened 2024-01-29
docs: Adding DNF as  install option for the Redhat based distributions

### LINKED ISSUE #1205: Add DNF alongisde Yum in the lnstall guide
Although ```yum``` commands are automatically "translated" to ```dnf``` and will still work, as of Fedora 22 and RHEL 8 / CentOS 8 ```dnf``` is the default package manager for the RedHat based distributions.

In general, yum has been officially deprecated:

```
    Yum command has been deprecated, use dnf instead.
    See 'man dnf' and 'man yum2dnf' for more information
```

And while that is true, it's highly likely that there are users that will install ```tofu``` on distributions like CentOS 7.

Based on the above, I believe that ```dnf``` should be added as an install option in website/docs/intro/install/rpm.mdx

Relevant links:
1. https://docs.fedoraproject.org/en-US/quick-docs/dnf/
2. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/managing-software-packages_configuring-basic-system-settings
3. https://github.com/opentofu/opentofu/issues/706

I do have a PR ready to go, in case the community is in favour of this change.

nckbnv:
Adding DNF as a installation option in the website installation guide. 
This PR adds a third tab, with DNF instead of replacing Yum completely.

Even though ```dnf``` is the default for the most recent versions of the major RedHat based distributions, yum is still a valid command for RHEL / CentOS 7 and below.

Resolves #1205.

### DIFF
--- a/website/docs/intro/install/install-opentofu.sh
+++ b/website/docs/intro/install/install-opentofu.sh
@@ -558,10 +558,17 @@ EOF
 # This function installs OpenTofu via the yum command line utility. It returns $TOFU_INSTALL_EXIT_CODE_INSTALL_REQUIREMENTS_NOT_MET
 # if yum is not available.
 install_yum() {
-  if ! command_exists "yum"; then
+  RPM_PACKAGE_MANAGER=""
+  if command_exists "dnf"; then
+    RPM_PACKAGE_MANAGER="dnf"
+  elif command_exists "yum"; then
+    RPM_PACKAGE_MANAGER="yum"
+  else
+    log_error "Neither yum nor dnf found. Unable to install OpenTofu."
     return "${TOFU_INSTALL_EXIT_CODE_INSTALL_REQUIREMENTS_NOT_MET}"
   fi
-  log_info "Installing OpenTofu using yum..."
+  
+  log_info "Installing OpenTofu using ${RPM_PACKAGE_MANAGER}..."
   if [ "${SKIP_VERIFY}" -ne "1" ]; then
     GPGCHECK=1
     GPG_URL="${RPM_GPG_URL}"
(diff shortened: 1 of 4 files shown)


janosdebugs (changes requested):
Hey @nckbnv thank you for adding the DNF instructions. Could you also add it to the test scripts located in this directory?

nckbnv:
@janosdebugs I've made some adjustments to install-opentofu.sh and rpm.sh. Are those the scripts you have referred to?

janosdebugs:
Hey @nckbnv thank you, I haven't had time to review it yet, but I will, hopefully tomorrow.

website/docs/intro/install/install-opentofu.sh
```diff
@@ -558,10 +558,17 @@ EOF
 # This function installs OpenTofu via the yum command line utility. It returns $TOFU_INSTALL_EXIT_CODE_INSTALL_REQUIREMENTS_NOT_MET
 # if yum is not available.
 install_yum() {
-  if ! command_exists "yum"; then
+  RPM_PACKAGE_MANAGER=""
```
cam72cam, line 561: This file actually lives in the get.opentofu.org repository and needs to be changed there instead.  It was accidentally committed to the main repository and removed in https://github.com/opentofu/opentofu/pull/1314.

website/docs/intro/install/rpm.sh
```diff
@@ -2,14 +2,22 @@
 
 set -e
 
-if [ -f /usr/bin/zypper ]; then
+if command_exists "zypper"; then
```
cam72cam, line 5: command_exists is not defined within this file.  You are trying to refer to a function defined inside the install-opentofu.sh script that lives in the get.opentofu.org repository.  You can pull a similar function into this script if need be.

cam72cam (changes requested):
See comments, sorry for the delay and confusion!

website/docs/intro/install/install-opentofu.sh
ksemele: What do you think about creating a separate command for yum and dnf?

PR #1220, opened 2024-02-01
Added aliases for `state list`, `state mv`, and `state rm`

### LINKED ISSUE #1136: Improve the consistency of `state` subcommands
### OpenTofu Version

```shell
OpenTofu v1.6.0
on linux_amd64
```

### Use Cases

Below is the list of `state` subcommands:

```raw
Subcommands:
    list                List resources in the state
    mv                  Move an item in the state
    pull                Pull current state and output to stdout
    push                Update remote state from a local state file
    replace-provider    Replace provider in the state
    rm                  Remove instances from the state
    show                Show a resource in the state
```

I think there are some inconsistencies in terms of subcommand names since `list` is a full word whereas `mv` is an abbreviation. Same story for `rm`.

Being used to UNIX commands, I'm tempted to write `terraform state ls` but then, when I realise it is in fact `list`, my reflex is to write `terraform state move` afterwards instead of `terraform state mv` for example.

It is not a big deal of course but I think aliases would be very welcome to improve the overall consistency of these subcommands.

### Attempted Solutions

N/A

### Proposal

Adding the following aliases to the `state` subcommands:

* `ls` to `list`
* `move` to `mv`
* `remove` to `rm`

### References

Issue originally created for Terraform 4 years ago: https://github.com/hashicorp/terraform/issues/20998

lettucemode:
Adds alias commands for `state list`, `state mv`, and `state rm`. Aliases don't show when listing `state`'s subcommands, and are denoted in each subcommand's help text.

Resolves #1136 

## Target Release
1.7.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ BUG FIXES:
 * Fix Global Schema Cache not working in provider acceptance tests ([#1054](https://github.com/opentofu/opentofu/pull/1054))
 * Fix `tofu show` and `tofu state show` not working with state files referencing Terraform registry providers in some instances ([#1141](https://github.com/opentofu/opentofu/pull/1141))
 * Improved stability on 32-bit architectures ([#1154](https://github.com/opentofu/opentofu/pull/1154))
+* Added aliases for `state list`, `state mv`, and `state rm` ([#1220](https://github.com/opentofu/opentofu/pull/1220))
 
 ## Previous Releases
 
--- a/cmd/tofu/commands.go
+++ b/cmd/tofu/commands.go
@@ -370,6 +370,14 @@ func initCommands(
 			}, nil
 		},
 
+		"state ls": func() (cli.Command, error) {
+			return &command.AliasCommand{
+				Command: &command.StateListCommand{
+					Meta: meta,
+				},
+			}, nil
+		},
+
 		"state rm": func() (cli.Command, error) {
 			return &command.StateRmCommand{
 				StateMeta: command.StateMeta{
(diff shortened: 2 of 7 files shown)


ilmax:
>Aliases don't show when listing state's subcommands

Is that done for consistency reasons? Because in general aliases should be displayed in the help text and when listing subcommands of a particular command

cube2222 (approved):
Only nit is that we might want to settle on either the long or short forms being the aliases consistently.

But at the same time I can see how the current approach taken is more backwards-compatible UX-wise.

Either way works for me, the PR as it is looks good to me!

lettucemode:
> > Aliases don't show when listing state's subcommands
> 
> Is that done for consistency reasons? Because in general aliases should be displayed in the help text and when listing subcommands of a particular command

Yea correct, I didn't want to change the output of `tofu state` as compared to before. The `cli` library that's used doesn't have explicit support for aliases, so I didn't see a way to have output lines like `(list|ls)                List resources in the state` when listing subcommands. If preferred, I could append `(alias: xyz)` to the `Synopsis()` for each of the affected commands.

janosdebugs (approved):

serdardalgic:
Thank you very much for the implementation, I believe this showcases the constructive nature of OpenTofu project, compared to its predecessor. 🚀 
One extremely minor nitpick: Why is it `(list|ls)` but `(mv|move)` and `(rm|remove)`? Is it because the latter one is added later? My instinct suggests selecting either  `ls|list, mv|move, rm|remove` or `list|ls, move|mv, remove|rm` for UX consistency.

Yantrio:
I agree with what you're saying there @serdardalgic :) I'll put in a quick fix for this now.

Yantrio:
https://github.com/opentofu/opentofu/pull/1244 Opened a PR here.

Is this what you meant @serdardalgic ?

PR #1223, opened 2024-02-01
Added templatestring function similar to templatefile

### LINKED ISSUE #301: Render templates from strings e.g. `templatestring` function similar to `templatefile`
## Context

- workaround is to use the forked template provider with arm support as a drop in replacement https://github.com/cloudposse/terraform-provider-template
- full context https://gist.github.com/nitrocode/cf40a24054a66afe2b19ca54b7be5d68

---

I'd like to use `template` from a string instead of a file. At the moment, `templatefile()` requires a file to be present on disk prior to plan. If I can instead use a string, then I no longer need a file and my string can be even more dynamic. This is the functionality that we used to have with the original hashicorp/terraform-provider-template. However, that has been deprecated by `templatefile` with the missing string functionality and never had an arm release which is why the original was forked as `cloudposse/terraform-provider-template`. Since this fork provider is a workaround, a new string function would be much more beneficial.

A few attempts have been made to PR the `templatestring` function and those have been closed.

sanskruti-shahu:
## Objective

The objective of this issue is to address a limitation in Terraform, where the `templatefile()` function requires a file to be present on disk, thus restricting the dynamism of template usage. The goal is to introduce a new function named `templatestring()` to enable template usage directly from a string, eliminating the need for a file and enhancing flexibility in generating dynamic templates.

## Changes

- Implemented the `MakeTemplateStringFunc()` function inside `string.go`, containing the execution logic for the `templatestring()` function.
- Created a new file, `render_template.go`, to house the common `renderTemplate()` function used by both `templatestring()` and `templatefile()`.
- Added comprehensive tests in `string_test.go` to validate the implementation of the `templatestring()` function and `renderTemplate()` function in `render_template_test.go`.
- Created `templatestring.mdx`, a documentation file outlining the usage of the `templatestring()` function, to be displayed on the opentofu website.
- Updated `language-nav-data.json` and `description.go` files to include information about the `templatestring()` function.
- Updated `CHANGELOG.md` to reflect the addition of the `templatestring()` function.

Resolves #301 

## Target Release

1.7.0


internal/lang/funcs/filesystem.go
```diff
@@ -69,7 +68,7 @@ func MakeFileFunc(baseDir string, encBase64 bool) function.Function {
 // As a special exception, a referenced template file may not recursively call
 // the templatefile function, since that would risk the same file being
 // included into itself indefinitely.
-func MakeTemplateFileFunc(baseDir string, funcsCb func() map[string]function.Function) function.Function {
+func MakeTemplateFunc(baseDir string, funcsCb func() map[string]function.Function, fileInput bool) function.Function {
```
Yantrio, line 71: I was wondering why we chose to do it this way?

I think overall it may be best in it's own file, even if there is some duplicate logic. This could help us with the maintainability in the long run.

Right now, as an end user who would look for the templatestring function, I would be confused to find it in filesystem.go, and to have file logic in it.

I propose we move it out to it's own function and put this back as it was.

internal/lang/funcs/filesystem_test.go
```diff
@@ -217,6 +217,63 @@ func TestTemplateFile(t *testing.T) {
 	}
 }
 
+func TestTemplateString(t *testing.T) {
+	tests := []struct {
+		Data  cty.Value
+		Vars  cty.Value
+		Want  cty.Value
+		Err   string
+	}{
```
Yantrio, line 226: It would be nice if the tests each had a name describing what it's testing.

internal/lang/funcs/filesystem_test.go
```diff
@@ -217,6 +217,63 @@ func TestTemplateFile(t *testing.T) {
...
+		},
+		{
+			cty.StringVal("Hello, ${name}!"),
+			cty.MapVal(map[string]cty.Value{
+				"name": cty.StringVal("Jodie"),
+			}),
+			cty.StringVal("Hello, Jodie!"),
+			``,
+		},
+		// Add more test cases for templatestring here
```
Yantrio, line 241: Can you add the missing tests please, this seems to only be the happy path.

internal/lang/funcs/filesystem_test.go
```diff
@@ -217,6 +217,63 @@ func TestTemplateFile(t *testing.T) {
...
+					t.Errorf("wrong error\ngot:  %s\nwant: %s", got, want)
+				}
+				return
+			} else if err != nil {
+				t.Fatalf("unexpected error: %s", err)
+			}
+
+			if !got.RawEquals(test.Want) {
+				t.Errorf("wrong result\ngot:  %#v\nwant: %#v", got, test.Want)
+			}
```
Yantrio, line 272: I wonder if we can re-rder some of this logic so we don't require the early `return` on line 265. 

Maybe inverting some of the ifs to handle the shorter path first may be favourable here.

website/docs/language/functions/templatestring.mdx
```diff
@@ -0,0 +1,38 @@
+---
```
Yantrio, line 1: The documentation here seems quite minimal. It may be difficult for new users to consume this.

Yantrio (changes requested):
Hi @sanskruti-shahu , thanks for submitting this. I've left some feedback in comments.

sanskruti-shahu:
@Yantrio I'll work upon the changes that you requested.

internal/lang/funcs/filesystem.go
sanskruti-shahu: For this I was thinking of two options either creating new file named templatestring.go containing all the templatestring function's logic in it or string.go file is already in opentofu so i can add templatestring function in that file as templatefile function is present in filesystem coz it is related to files so similarly templatestring can be in string  file as it is related to string. which one do you suggest?

Yantrio: My personal preference would be to place this alongside other string based methods inside `string.go`.

sanskruti-shahu: commit #[4fb1950](https://github.com/opentofu/opentofu/pull/1223/commits/4fb1950bb377f66c61af3af40afb85f8d11d343a) resolves this

internal/lang/funcs/filesystem_test.go
sanskruti-shahu: commit [4fb1950](https://github.com/opentofu/opentofu/pull/1223/commits/4fb1950bb377f66c61af3af40afb85f8d11d343a) resolves this

internal/lang/funcs/filesystem_test.go
sanskruti-shahu: commit [4fb1950](https://github.com/opentofu/opentofu/pull/1223/commits/4fb1950bb377f66c61af3af40afb85f8d11d343a) resolves this

internal/lang/funcs/filesystem_test.go
sanskruti-shahu: commit [4fb1950](https://github.com/opentofu/opentofu/pull/1223/commits/4fb1950bb377f66c61af3af40afb85f8d11d343a) resolves this

website/docs/language/functions/templatestring.mdx
sanskruti-shahu: commit [4fb1950](https://github.com/opentofu/opentofu/pull/1223/commits/4fb1950bb377f66c61af3af40afb85f8d11d343a) resolves this

website/docs/language/functions/templatestring.mdx
```diff
@@ -0,0 +1,91 @@
...
+output "result" {
+  value = templatestring("Hello, ${name}!", { name = "Alice" })
+}
+```
+```
+result = Hello, Alice!
+```
+## Advanced Usage
+
+Although in principle `templatestring` can be used with an inline template string, we don't recommend this approach because it requires awkward escaping. Instead, iconsider exploring alternative options such as you can use the local block to create a local variable and then pass that variable to the templatestring function or utilizing the file function. For a clearer understanding, refer to the example below.
```
Yantrio, line 43: I don't think we should mention using the file function here, they should use the other template function there

website/docs/language/functions/templatestring.mdx
```diff
@@ -0,0 +1,91 @@
...
+  value = templatestring(file("map.tftpl"), { items = {key1 = "value1", key2 = "value2", key3 = "value3" }})
+}
+```
+```
+result = <<-EOT
+    key1:value1
+    key2:value2 
+    key3:value3
+EOT
+```
```
Yantrio, line 86: Let's keep using templatestring here without files,

website/docs/language/functions/templatestring.mdx
```diff
@@ -0,0 +1,91 @@
+---
```
Yantrio, line 1: Can you somehow link to how string templates work and how they can be used?

Yantrio: https://opentofu.org/docs/language/expressions/strings/#string-templates

website/docs/language/functions/templatestring.mdx
```diff
@@ -0,0 +1,91 @@
...
+  The `templatestring` function takes a string and renders it as a template using a supplied set of template variables.
+---
+
+# `templatestring` Function
+
+The `templatestring` function takes a string and renders it as a template using a supplied set of template variables.
+
+```hcl
+templatestring(str, vars)
+```
```
Yantrio, line 13: This is still missing a lot of documentation. This needs to be informative for the end users. Please see [templatefile.mdx](https://github.com/opentofu/opentofu/blob/199ec898e634b5c2782664cf436bac94eae3fd10/website/docs/language/functions/templatefile.mdx#L8-L43) as an example of this

Yantrio (changes requested):

internal/lang/funcs/filesystem_test.go
```diff
@@ -198,20 +198,20 @@ func TestTemplateFile(t *testing.T) {
...
+				if !got.RawEquals(test.Want) {
+					t.Errorf("wrong result\ngot:  %#v\nwant: %#v", got, test.Want)
 				}
-				return
-			} else if err != nil {
-				t.Fatalf("unexpected error: %s", err)
-			}
-
-			if !got.RawEquals(test.Want) {
-				t.Errorf("wrong result\ngot:  %#v\nwant: %#v", got, test.Want)
```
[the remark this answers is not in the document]
sanskruti-shahu: I incorporated your suggestion to refactor the logic of the if-else statements to avoid early returns in the templatestring test. Similarly, I addressed the early return in this code as well.

internal/lang/funcs/string.go
```diff
@@ -162,3 +164,113 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
...
+			AllowMarked: true,
+		},
+		{
+			Name: "vars",
+			Type: cty.DynamicPseudoType,
+		},
+	}
+	loadTmpl := func(content string, marks cty.ValueMarks) (hcl.Expression, error) {
+		// We re-use File here to ensure the same filename interpretation
+		// as it does, along with its other safety checks.
```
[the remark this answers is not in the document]
sanskruti-shahu: yeah, will remove it

website/docs/language/functions/templatestring.mdx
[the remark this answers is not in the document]
sanskruti-shahu: noted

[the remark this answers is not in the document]
sanskruti-shahu: Here, when I tried running loop or join func as inline string then it was giving me error (screenshot atached). So, I thought it was due to awkward escaping, mentioned in [template_file](https://registry.terraform.io/providers/hashicorp/template/latest/docs/data-sources/file) doc of hashicorp. But, I think this is not due to awkward escaping, it is something which is related to HCL syntax. I'll try to find solution for this.
Screenshot 2024-02-05 at 6 14 51 PM

internal/lang/funcs/string_test.go
```diff
@@ -253,3 +254,105 @@ func TestStartsWith(t *testing.T) {
...
+			`vars map does not contain key "name"`,
+		},
+		"Interpolation of a boolean value": {
+			cty.StringVal("${val}"),
+			cty.ObjectVal(map[string]cty.Value{
+				"val": cty.True,
+			}),
+			cty.True, 
+			``,
+		},
```
Yantrio, line 324: I think it would be helpful here if we added tests for sensitive arguments too

internal/lang/functions.go
```diff
@@ -160,6 +160,12 @@ func (s *Scope) Functions() map[string]function.Function {
 			return s.funcs
 		})
 
+		s.funcs["templatestring"] = funcs.MakeTemplateStringFunc(s.BaseDir, func() map[string]function.Function {
+			// The templatestring function prevents recursive calls to itself
+			// by copying this map and overwriting the "templatestring" entry.
+			return s.funcs
```
Yantrio, line 166: This logic isn't doing what the comment says. We should implement this.

Yantrio (changes requested):

internal/lang/functions.go
sanskruti-shahu: I think the recursive call for templatestring is being already handled (screenshots attached) so I need not to do anything explicitly for that.

sanskruti-shahu: yeah sure will do that

internal/lang/funcs/string.go
```diff
@@ -162,3 +164,110 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
...
+			if !(args[0].IsKnown() && args[1].IsKnown()) {
+				return cty.DynamicPseudoType, nil
+			}
+
+			// We'll render our template now to see what result type it produces.
+			// A template consisting only of a single interpolation an potentially
+			// return any type.
+
+			pathArg, pathMarks := args[0].Unmark()
+			expr, err := loadTmpl(pathArg.AsString(), pathMarks)
```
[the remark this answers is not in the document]
sanskruti-shahu: when a sensitive string is passed to template string, the resultant template will be sensitive as well, so do you want templatestring func to give error when sensitive string is passed or is it fine this way?

Screenshot 2024-02-08 at 4 54 01 PM

Screenshot 2024-02-08 at 4 54 24 PM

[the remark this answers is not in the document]
sanskruti-shahu: I will be adding a check for sensitive string which will give error and will ask user to remove any sensitive content passed as input.

Yantrio (changes requested):
I've discussed this with the core team, and we believe that in situations where you pass in either
- A sensitive template
- OR any senstive variables

The resultant templated string should be sensitive.

This is in line with other functionality in OpenTofu.

Would you mind implementing it this way please instead of having it error? Thanks!

Please reach out with any questions you have before beginning any other implementations to avoid re-doing some work!

internal/lang/funcs/string.go
```diff
@@ -164,3 +166,129 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
...
```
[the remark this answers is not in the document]
sanskruti-shahu: yeah, will do that.

internal/lang/funcs/string.go
```diff
@@ -164,3 +166,129 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
...
+			Name:        "vars",
+			Type:        cty.DynamicPseudoType,
+			AllowMarked: true,
+		},
+	}
+	loadTmpl := func(content string, marks cty.ValueMarks) (hcl.Expression, error) {
+
+		// This condition checks if the provided string to be rendered as a template is marked as sensitive.
+		// If the string is marked as sensitive, it returns an error indicating that sensitive strings cannot be used as template strings.
+		if strings.Contains(marks.GoString(), "Sensitive") {
```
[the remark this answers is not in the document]
sanskruti-shahu: yeah i'll update this.

internal/lang/funcs/string.go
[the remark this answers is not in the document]
sanskruti-shahu: will do that.

internal/lang/funcs/string.go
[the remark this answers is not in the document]
sanskruti-shahu: I dont think I need to explicitly add any check marking the resultant value as sensitive because, this is already implemented in opentofu itself. (screenshots attached)

Screenshot 2024-02-12 at 4 16 45 PM

Screenshot 2024-02-12 at 4 17 31 PM

internal/lang/funcs/string.go
```diff
@@ -164,3 +166,129 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
...
+			return nil, function.NewArgErrorf(0, "Sensitive strings cannot be used as template strings. Please ensure that any sensitive information is removed from your template before using them")
```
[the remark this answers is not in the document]
sanskruti-shahu: Based on the feedback provided earlier, I've removed the error check. Therefore, I believe this comment is no longer valid. Should I proceed with committing my changes?"

sanskruti-shahu:
Hey @Yantrio , I've replied to your comment. Could you please check it once? Also, could you review the code changes again to let me know if there are any further changes needed?

sanskruti-shahu:
Hey @Yantrio , I've replied to your comments. Could you please check it once? Also, could you review the code changes again to let me know if there are any further changes needed?

internal/lang/funcs/string.go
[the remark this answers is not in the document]
Yantrio: I'm not quite sure what you mean here. I'm suggesting two things here:

- checking the marks in a different way, instead you're pulling it out manually and checking against the string, instead of using `marks.Has`

- checking the marks in a different location (This could've happened one level higher) 

however we've now decided to propogate the sensitivity so this check is now redundant, right?

internal/lang/funcs/string.go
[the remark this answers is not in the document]
Yantrio: yes please

Yantrio:
Hi @sanskruti-shahu , we discussed it in one of the conversations that the sensitive value should be propogated throughout instead of erroring.

Would you mind implementing this please? I do not see this in the current code.

internal/lang/funcs/string.go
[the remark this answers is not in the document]
sanskruti-shahu: yeah right.

sanskruti-shahu:
Yeah sure will do that.

sanskruti-shahu:
Hey, @Yantrio, I have implemented the mentioned changes, please have a review.

sanskruti-shahu:
Hello, @Yantrio , I've implemented the changes you mentioned. Could you please review them?

internal/lang/funcs/string.go
```diff
@@ -164,3 +166,110 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
...
+			funcs[name] = fn
+		}
+		ctx.Functions = funcs
+
+		val, diags := expr.Value(ctx)
+		if diags.HasErrors() {
+			return cty.DynamicVal, diags
+		}
+		return val, nil
+	}
```
janosdebugs, line 242: This function seems to be copied verbatim from `filesystem.go`. It serves the same purpose, so there is no reason to duplicate this function.  This needs to be refactored so both the templatefile and templatestring function can use the same function instead of copying, otherwise the code will have to be maintained in two places, which is less than ideal.

janosdebugs (changes requested):
I'm sorry to jump in this late into the PR, but there's a large chunk of copy-pasted code and there's no good reason to maintain two copies of the code.

internal/lang/funcs/string.go
sanskruti-shahu: At start of the PR, I had used the same piece of code for both the functions with the respective modifications needed but then @Yantrio raised concerns regarding potential confusion for end-users who would look for the templatestring function, they would be confused to find it in filesystem.go and as templatestring is related to string so it was suggested that its logic be moved to the string.go file to ensure clarity so that's why I created new function copying all logic into string.go file for templatestring

janosdebugs: @sanskruti-shahu if this is the exact same function, it should definitely not be copied over. Instead, this specific function should be moved out from both here and the templatefile function and called accordingly. @Yantrio was right, the templatefile function should not use the templatestring function, but copying 50-ish lines of code is also not the right way to go.

sanskruti-shahu: @janosdebugs I got your point but I am not clear about what am I supposed to do next? Do I have to make a third function which will contain the logic which is common in both and it will be called separately for templatefile and templatestring and if this is so then where shall I place that function? creating new file or inside any (don't know which) existing file?

janosdebugs: Yes, the renderTmpl function should be extracted into possibly a separate file.

sanskruti-shahu: Ok then I'll create new file having func containing the logic of renderTmpl which will be called by both templatestring and templatefile.

sanskruti-shahu:
@janosdebugs , I have implemented the changes you mentioned. Could you please review them?

internal/lang/funcs/render_template.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// currFilename stores the filename of the template file, if any.
+	currFilename := expr.Range().Filename
+
+	// We'll pre-check references in the template here so we can give a
+	// more specialized error message than HCL would by default, so it's
+	// clearer that this problem is coming from a templatefile/templatestring call.
+	for _, traversal := range expr.Variables() {
+		root := traversal.RootName()
+		referencedPos := func() string {
```
cam72cam, line 48: Why is this a func?  It's a bit of an odd pattern here.

Why not something like:
```
referencedPos = fmt.Sprintf("%q", root)
if currFilename != TemplateStringFilename  {
  referencedPos = fmt.Sprintf("%q, referenced at %s", root, traversal[0].SourceRange())
}
```

internal/lang/funcs/render_template.go
```diff
@@ -0,0 +1,91 @@
...
+						Type:        cty.String,
+						AllowMarked: true,
+					},
+					{
+						Name: "vars",
+						Type: cty.DynamicPseudoType,
+					},
+				},
+				Type: func(args []cty.Value) (cty.Type, error) {
+					return cty.NilType, fmt.Errorf("cannot recursively call templatefile from inside templatefile call")
```
cam72cam, line 77: `from inside templatefile or templatestring`

cam72cam:
Two small changes @sanskruti-shahu requested, but it's looking good !

internal/lang/funcs/render_template.go
sanskruti-shahu: noted

internal/lang/funcs/render_template.go
sanskruti-shahu: yeah will update that

cam72cam (dismissed):

internal/lang/funcs/render_template.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"fmt"
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+func RenderTemplate(expr hcl.Expression, varsVal cty.Value, funcsCb func() map[string]function.Function) (cty.Value, error) {
```
janosdebugs, line 17: I do not think this needs to be exported because it won't be used outside of this package.

internal/lang/funcs/string.go
```diff
@@ -164,3 +166,69 @@ func Replace(str, substr, replace cty.Value) (cty.Value, error) {
 func StrContains(str, substr cty.Value) (cty.Value, error) {
 	return StrContainsFunc.Call([]cty.Value{str, substr})
 }
+
+// This constant provides a placeholder value for filename indicating
+// that no file is needed for templatestring.
+const (
+	TemplateStringFilename = "NoFileNeeded"
```
janosdebugs, line 173: This constant seems to be internal and should not be imported.

janosdebugs (approved):
Thanks @sanskruti-shahu great job! Tiny nitpick, but I'll approve since these don't change much.

internal/lang/funcs/string.go
sanskruti-shahu: noted

internal/lang/funcs/render_template.go
sanskruti-shahu: noted

janosdebugs (approved):

Yantrio (approved):

cam72cam (approved):

PR #1132, opened 2024-01-17
state encryption draft methods and key providers (NOT FOR MERGING)

StephanHCB:
This is a draft PR that is NOT intended for merging.

It is based on https://github.com/opentofu/opentofu/pull/1063, so the first commits include the changes from that PR.

The code changes and additions under `internal/states/encryption` are a working draft of the following subtasks of https://github.com/opentofu/opentofu/issues/1030
  - implement encryption interfaces
  - Implement Encryption Method - Full
  - Implement Key Providers - Direct
  - Implement Key Providers - Passphrase

It also contains a draft of Encryption Method - Partial (I added this to ensure the Method & KeyProvider interfaces work with
partial encryption methods).

The code changes outside `internal/states/encryption` are the minimal integration into the rest of tofu that allow using
the feature. LIMITATIONS:
  - only supports configuration via environment variables
  - uses the same configuration for all remote states

This is a draft in progress, if changes are made to the base PR https://github.com/opentofu/opentofu/pull/1063, I will rebase it.

RFC see https://github.com/opentofu/opentofu/issues/874

PR #1227, opened 2024-02-02
State Encryption Documentation and Partial Implementation

### LINKED ISSUE #1030: Implement State Encryption

Yantrio:
This pull request includes 2 main parts of the implementation.

Firstly it contains a large document (`docs/state_encryption.md`) which we intend to keep up to date as a living design document for the intentions behind the implementation. 

Secondly, it contains an initial draft of the library component and a basic prototype implementation. 

This does **not** change the behavior of OpenTofu.

This draft has been made because we are seeking reviews on this pull request. However we would like to ask people to review this from 3 perspectives.

- Review this as an implementer of a `KeyProvider` or `Method`
- Review this as an end-user consumer of the new `encryption` block in OpenTofu Configuration
- Review this as a consumer of the library, either OpenTofu or another project.

Resolves #1030 

## Target Release

1.7.0


docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Implementation
+
+When implementing the encryption tooling, the implementation should be split in two parts: the library and the OpenTofu implementation. The library should rely on the cty types and hcl as a means to specify schema, but should not be otherwise tied to the OpenTofu codebase. This is necessary to enable encryption capabilities for third party tooling that may need to work with state and plan files.
+
+### Library implementation
+
+The encryption library should create an interface that other projects and OpenTofu itself can use to encrypt and decrypt state. The library should express its schema needs (e.g. for key provider config) using [cty](https://github.com/zclconf/go-cty) and [hcl](https://github.com/hashicorp/hcl), but be otherwise independent of the OpenTofu codebase. Ideally, the OpenTofu project should provide this library as a standalone dependency that does not pull in the entire OpenTofu dependency tree.
```
StephanHCB, line 249: Should the library include the functionality to parse a configuration in HCL and JSON from a string value into cty? This would free users of the library from the (IMHO very complex) interface of go-cty if all they want to do is decrypt some state file.

internal/encryption/keyprovider/keyprovider.go
```diff
@@ -0,0 +1,22 @@
...
+	//
+	// Common errors:
+	// - Returning a struct without a pointer
+	// - Returning a non-struct
+	ConfigStruct() Config
+}
+
+type KeyProvider interface {
+	// Provide provides an encryption key. If the process fails, it returns an error.
+	Provide() ([]byte, error)
```
StephanHCB, line 21: During encryption, key providers may need to place information in the "encryption" JSON key, such as a nonce used to derive the key.

During decryption, key providers may need to read information from the "encryption" JSON key, such as a nonce needed to derive the same key again.

Compare the KeyProvider interface in https://github.com/opentofu/opentofu/pull/1132/files#diff-e7749ff1b789f4930684fd4132a65fe3227b0a2dcbb23fe80794ddd5066cd88b this is why the parameters include an EncryptionInfo block, which later gets encoded into the "encryption" JSON substructure.

internal/encryption/encryption.go
```diff
@@ -0,0 +1,364 @@
...
+			}, "remote_data_source."+target.Name)
+			diags = append(diags, mDiags...)
+			enc.remote[target.Name] = NewState(m)
+		}
+	}
+
+	return diags
+}
+
+func (enc *encryption) setupTarget(cfg *TargetConfig, name string) ([]method.Method, hcl.Diagnostics) {
```
RLRabinowitz, line 310: I am a bit confused as to why `setupTarget` returns an array of `Method`s. Seems like it'll be a 1-2 sized array, the first being the actual encryption method, and the optional second being the fallback method. 

If that's the case then the type here and in other usages (In `NewPlan`, etc.) is a bit confusing, especially since we've explicitly declared we don't want more than 1 fallback

If I understand correctly, then this means the fallback logic is not yet implemented (at least for plan encryption), correct?

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+  encryption {
+    key_provider "static" "my_passphrase" {
+      key = "this is my encryption key"
+    }
+    key_provider "pbkdf2" "my_derivative_key" {
+      key_providers = [key_provider.static.my_passphrase]
+    }
+    method "aes_gcm" "foo" {
+      key_provider = key_provider.pbkdf2.my_derivative_key
+    }
```
RLRabinowitz, line 148: Reading through here initially, I had thought that a `key_provider` supports `key` and `key_providers` fields, and that `method` requires a `key_provider` field, but seems like the schema is dynamic and is per `key_provider`/`method` type. IMO the docs should clarify this a bit more, but that's just me

RLRabinowitz (commented):
Very well done!

I left some comments. Overall I like this approach 👏🏻

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+```hcl2
+terraform {
+  encryption {
+    // Encryption options
+  }
+}
+```
+
+The mere presence of the `encryption` block alone should not enable encryption because the user should explicitly specify what key and method to use. The implementation should warn the user if the encryption block is present but has no configuration.
```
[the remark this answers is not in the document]
Yantrio: This should error, you're correct. I'll update the wording.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+The user can set either of these structures in the `TF_ENCRYPTION` environment variable:
+
+```bash
+export TF_ENCRYPTION='{"key_provider":{...},"method":{...},"statefile":{...}}'
+```
+
+When the user specifies both an environment and a code configuration, `tofu` merges the two configurations. If two values conflict, the environment configuration takes precedence.
+
+To ensure that the encryption cannot be accidentally forgotten or disabled and the data stored unencrypted, the user can specify the `enforced` option in the HCL configuration:
```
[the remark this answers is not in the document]
Yantrio: The main reason for the `enforced` is when someone does the following:

```
    statefile {
      method = method.aes_gcm.abc
    }
```

And they:
- Pass in configuration through environment variables
- Don't specify a planfile block in hcl/json

This means the planfile could possibly not be encrypted if the environment variable wasn't constructed properly.

Adding the `enforced` flag ensures that the overall configuration (merged from hcl, overrides and environment variables) 100% contains a configuration.

On that note: I think that if the top level block defines `enforced = true` then we should try and not allow overrides and env vars to just set `enforced = false`. I'm not sure how to approach that with our current situation though (using gohcl to read in)

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+    terraform_remote_states {
+      default {
+        enforced = true
+      }
+    }
+  }
+}
+```
+
+**Note:** The `enforced` option is also available in the environment configuration and works as intended, but doesn't make much sense because its primary purpose is to guard against environment variable omission.
```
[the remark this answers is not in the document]
Yantrio: See my comment [above](https://github.com/opentofu/opentofu/pull/1227#discussion_r1479601431)

internal/encryption/config.go
```diff
@@ -0,0 +1,54 @@
...
+type MethodConfig struct {
+	Type string   `hcl:"type,label"`
+	Name string   `hcl:"name,label"`
+	Body hcl.Body `hcl:",remain"`
+}
+
+type TargetConfig struct {
+	Enforced bool           `hcl:"enforced,optional"`
+	Method   hcl.Expression `hcl:"method,optional"`
+	Fallback *TargetConfig  `hcl:"fallback,block"`
```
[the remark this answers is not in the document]
cam72cam: Enforced is not relevant, but having multiple fallback is nice (even if it's not recommended)

Yantrio (commented):
.

internal/encryption/registry/registry.go
```diff
@@ -0,0 +1,25 @@
+package registry
+
+import (
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+	"github.com/opentofu/opentofu/internal/encryption/method"
+)
+
+// Registry collects all encryption methods and key providers
+type Registry interface {
+	// RegisterKeyProvider registers a key provider. Use the keyprovider.Any() function to convert your key provider
```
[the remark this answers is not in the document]
Yantrio: Ah, we removed it, I'll update this comment

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Goals
+
+The goal of this feature is to allow OpenTofu users to fully encrypt state files when they are stored on the local disk or transferred to a remote backend. The feature should also allow reading from an encrypted remote backend using the `terraform_remote_state` data source. The encrypted version should still be a valid JSON file, but not necessarily a valid state file.
+
+Furthermore, this feature should allow users to encrypt plan files when they are stored. However, plan files are not JSON, they are ZIP files. If feasible, this feature should produce a structurally correct encrypted ZIP file.
+
+For the encryption key, users should be able to specify a key directly, use a remote key provider (such as AWS KMS, etc.), or create derivative keys (such as pbkdf2) from another key source. The primary encryption method should be AES-GCM, but the implementation should be open to different encryption methods. The user should also have the ability to decrypt a state or plan file with one (older) key and then re-encrypt data with a newer key. Multiple fallbacks should be avoided in the implementation.
```
cube2222, line 15: Could you expand on the use-case for doing derivative keys from another key source? I.e. with pbkdf2 the point is to use a human-readable password that is then expanded to a full key. I don't think there's any point in doing this for automated systems (which should just provide an AES key)?

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+To enable use cases where multiple teams need to collaborate, the user should be able to specify separate encryption methods and keys for individual uses, especially for the `terraform_remote_state` data source. However, to simplify configuration, the user should be able to specify a default configuration for all remote state data sources.
```
cube2222, line 17: > However, to simplify configuration, the user should be able to specify a default configuration for all remote state data sources.

Should the default affect only remote state data sources, or also the main remote state-file, if not otherwise specified?

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+**Note:** the user is responsible for keeping this key safe and follow disaster recovery best practices. OpenTofu is a read-only consumer for the key the user provided and will not perform tasks on the key management system like key rotation.
+
+The user also has to specify at least one encryption method referencing the key provider. This encryption method determines how the encryption takes place. It is the user's responsibility to make sure that they provided a key that is suitable for the encryption method.
+
+```hcl2
+terraform {
+  encryption {
+    //...
+    method "aes_gcm" "bar" {
+      key_provider = key_provider.static.foo
```
cube2222, line 80: I love the new UX, feels much more TF-native <3

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+      }
+      remote_data_source "some_module.remote_data_source.foo" {
+        method = method.aes_gcm.ijk
+      }
+    }
+  }
+}
+```
+
+To facilitate key and method rollover, the user can specify a fallback configuration for state and plan decryption. When the user specifies a `fallback` block, `tofu` will first attempt to decrypt any state or plan it reads with the primary method and then fall back to the `fallback` method. When `tofu` writes a state or plan, it will not use the `fallback` method and always writes with the primary method.
```
cube2222, line 113: I think it's worth noting here that if there *is no* primary, and there *is a* fallback, then tofu *will not* encrypt anything (after successfully decrypting it on read).

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+When a user desires to create a composite key, such as for creating a passphrase-based derivative key or a shared custody key, they may avail themselves of a key provider that supports multiple inputs. The user can chain key providers together:
+
+```hcl2
+terraform {
+  encryption {
+    key_provider "static" "my_passphrase" {
+      key = "this is my encryption key"
+    }
+    key_provider "pbkdf2" "my_derivative_key" {
```
cube2222, line 143: If we foresee adding future encryption methods that aren't based on aes256 keys, then we need a way to indicate to pbkdf2 what the key length is. I don't think we need to implement it yet, but we should plan for it. I suppose a simple key_length field that defaults to 256 would be fine.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+      key_providers = [key_provider.static.my_passphrase]
```
cube2222, line 144: I don't think pbkd2 should allow multiple input keys. I don't think encouraging composite keys constructed this way is a good thing. If any of the "sub-keys" is unavailable, then you have a problem. The fault tolerance is very bad. If we want to support composite keys in a first class way, we should support it via a dedicated method that is "non-experienced encryption user-safe", like https://en.wikipedia.org/wiki/Shamir%27s_secret_sharing. In practice, I would recommend not including this in the v1. Power users who want composite keys can construct them prior to `tofu` invocation and then pass them as a raw key.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+    }
```
cube2222, line 145: With the above comment (where I believe we shouldn't support composite keys via pbkdf2) I don't really see why pbkdf2 should be accepting a separate key provider.

The point of pbkdf2 is to expand a human-readable password to an encryption key. If you're getting the password from automated key management systems, then you should just use an aes key from there, without pbkdf2 in the loop.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+    method "aes_gcm" "foo" {
+      key_provider = key_provider.pbkdf2.my_derivative_key
+    }
+    //...
+  }
+}
```
cube2222, line 151: I believe we've discussed this towards the beginning of this project - I'm not sure if exposing raw encryption details like "aes_gcm" or "pbkdf2" is great UX. I'm worried it could be confusing to users.

TBF, I think it would be better to have a simple "symmetric" method. In practice, I don't expect us to ever have any other methods, frankly. The underlying implementation details might of course change (e.g. from gcm to something else) but ideally we'd handle that transparently to the user (via fallback automatically inserted fallback methods).

In practice, I suppose we could use just "symmetric" for now, and in the future, if we need to, include separate parameters for the "aes" and "gcm" (mode_of_operation) and "256" (key_length) parts.

This is all based on my experience with the Go standard library, which exposes multiple modes of operation, and if you're not at least moderately well-versed with cryptography, you'll have a hard time deciding which one you should be using.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+The mere presence of the `encryption` block alone should not enable encryption because the user should explicitly specify what key and method to use. The implementation should warn the user if the encryption block is present but has no configuration.
+
+The encryption relies on an encryption key, or a composite encryption key, which the user can provide directly or via a key management system. The user must provide at least one `key_provider` block with the settings described below. These key providers serve the purpose of creating or providing the encryption key. For example, the user could hard-code a static key named foo:
+
+```hcl2
+terraform {
+  encryption {
+    key_provider "static" "foo" {
+      key = "6f6f706830656f67686f6834616872756f3751756165686565796f6f72653169"
```
cube2222, line 65: For hex-encoded keys, I would suggest using the "hex" field. Similar to the random_bytes resource.

internal/encryption/keyprovider/static/config.go
```diff
@@ -0,0 +1,19 @@
+package static
+
+import (
+	"encoding/hex"
+	"fmt"
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+)
+
+type Config struct {
+	Key string `hcl:"key"`
```
cube2222, line 10: This will fail the example in the design doc you provided with pbkdf2.

In practice, as mentioned in my design document comment I would suggest both:
1. Using hex as the field name for passing a key in hex format.
2. Passing the passphrase to pbkdf2 directly. I don't think having a shared "resource" for both static keys and passphrases is a good idea.

internal/encryption/keyprovider/keyprovider.go
cube2222: Moreover, I believe this metadata should then be passed as "authenticated data" to the GCM Seal method.

docs/state_encryption.md
cube2222: Actually, when we were originally designing this, I believe we called the method "full" which I believe made sense.

It's obvious that we're doing symmetric encryption, and the fact that we're using aes_gcm should imo not be explicit in the config (or as an optional parameter).

And in the future, we might add a "partial" method.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Notes
+
+This document mentions `HCL` as a short-form for OpenTofu code. Unless otherwise specified, everything written also applies to the [JSON-equivalent of the HCL code](https://opentofu.org/docs/language/syntax/json/).
+
+## Goals
+
+The goal of this feature is to allow OpenTofu users to fully encrypt state files when they are stored on the local disk or transferred to a remote backend. The feature should also allow reading from an encrypted remote backend using the `terraform_remote_state` data source. The encrypted version should still be a valid JSON file, but not necessarily a valid state file.
+
+Furthermore, this feature should allow users to encrypt plan files when they are stored. However, plan files are not JSON, they are ZIP files. If feasible, this feature should produce a structurally correct encrypted ZIP file.
```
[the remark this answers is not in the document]
cube2222: The design suggests that you were leaning towards zipping encrypted files (specifically this part "this feature should produce a structurally correct encrypted ZIP file").

I'm a proponent of encrypting the ZIP file as well.

docs/state_encryption.md
Yantrio: In the case of pbkdf2, using a non-human-readable password has always been best practice as far as I'm aware, to avoid brute-force attacks.

In this case, I think it could also be helpful if we have a 1password or similar key provider to pull the password from there.

docs/state_encryption.md
Yantrio: My personal opinion is that it's better to have the two separate and explicitly defined.

docs/state_encryption.md
cube2222: If you're already doing a non-human-readable password (so don't use the approach of https://xkcd.com/936/) then you can just as well go with a raw AES key.

docs/state_encryption.md
```diff
@@ -0,0 +1,333 @@
...
+
+To enable use cases where multiple teams need to collaborate, the user should be able to specify separate encryption methods and keys for individual uses, especially for the `terraform_remote_state` data source. However, to simplify configuration, the user should be able to specify a default configuration for all remote state data sources.
+
+It is the goal of this feature to let users specify their encryption configuration both in (HCL) code and in environment variables. The latter is necessary to allow users the reuse of code for both encrypted and unencrypted state storage.
+
+Finally, it is the goal of the encryption feature to make available a library that third party tooling can use to encrypt and decrypt state. This may be implemented as a package within the OpenTofu repository, or as a standalone repository.
```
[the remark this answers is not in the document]
Yantrio: It's something we should bear in mind when developing this feature so that it's not more work to do in the future.

There are 2 steps to this in my mind:

1. Decouple the encryption interface and the opentofu implementation in their own packages
	- This helps avoid circular dependencies too!
2. Later on (post-v1 when we're happy with it), we can move the package from internal/ to pkg/ or it's own repo (however we decide!) to expose it publically for consumption.

docs/state_encryption.md
Yantrio: Correct, I'll update the document to reflect this.

docs/state_encryption.md
Yantrio: I agree but would like the opinion of @janosdebugs / @cam72cam on this one.

docs/state_encryption.md
cube2222: So, overall, the UX I would recommend, is the following:
```
terraform {
  encryption {
    key_provider "static" "key1" {
      hex = "6f6f706830656f67686f6834616872756f3751756165686565796f6f72653169"
    }
    key_provider "azure" "key2" {
      // Yada yada
    }
    method "full" "one" {
       key_provider = key_provider.static.key1
       mode = "gcm" // Optional, gcm is default.
       key_length = 256 // Optional.
    }
    method "partial" "two" {
       key_provider = key_provider.azure.key2
    }
    statefile {
      method = method.full.one
    }
    remote_data_sources {
      default {
        method = method.partial.two
      }
    }
  }
}
```

docs/state_encryption.md
cube2222, line 143: Also, pbkdf2 is an implementation detail. I suggest just calling it "passphrase".

cube2222: In practice, it seems like https://pkg.go.dev/golang.org/x/crypto/argon2 is the currently-recommended key derivation method.

internal/encryption/config.go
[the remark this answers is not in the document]
Yantrio: Note: We should document our choice here

We've discussed this between the team and we have decided to allow nested Fallback methods on the backend, but add a validation step to ensure we only have one depth of fallback.

internal/encryption/encryption.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		return append(diags, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Unable to fetch key data",
+			Detail:   fmt.Sprintf("key_provider.%s.%s failed with error: %s", cfg.Type, cfg.Name, err.Error()),
+		})
+	}
+	data, err := keyProvider.Provide()
+	if err != nil {
+		enc.keyValues[cfg.Type][cfg.Name] = cty.UnknownVal(cty.DynamicPseudoType)
```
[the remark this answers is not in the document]
Yantrio: TODO: @cam72cam can you document this in the code as a comment please. Mainly discussing why we are setting it to an unknown val.

docs/state_encryption.md
Yantrio: TODO: We have had a discussion and we will approach this in a separate PR when we decide how to handle things like propagating salts

docs/state_encryption.md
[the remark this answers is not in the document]
Yantrio: Choice made as a team:

- We will zip the file in memory
- Encrypt that in-memory blob
- Dump that encrypted data to disk as a binary file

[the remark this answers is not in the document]
janosdebugs: TODO: update document to describe that there needs to be authentication for the content.

docs/state_encryption.md
Yantrio: Team choice made:

We will discuss how we can implement a 'simple' method in the future and what the UX would look like for that, as apposed to the current setup.

internal/encryption/config/config_load.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/hashicorp/hcl/v2/json"
+)
+
+// LoadConfigFromString loads a configuration from a string. The sourceName is used to identify the source of the
+// configuration in error messages.
+// However! this method should only be used in tests, as we should be using gohcl to parse the configuration.
+// TODO: Discuss if we can remove this now.
```
Yantrio, line 19: We want to keep this here because it's great as a reference implementation. We should update the comment to make this clear.

internal/encryption/encryption.go
```diff
@@ -0,0 +1,74 @@
...
+
+package encryption
+
+import (
+	"github.com/opentofu/opentofu/internal/encryption/config"
+	"github.com/opentofu/opentofu/internal/encryption/registry"
+)
+
+// New creates a new Encryption provider from the given configuration and registry.
+func New(reg registry.Registry, cfg *config.Config) Encryption {
```
Yantrio, line 14: Tiny tiny tiny tiny tiny tiny nitpick: we should move this New method below the interface

internal/encryption/keyprovider.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Lookup the KeyProviderDescriptor from the registry
+	id := keyprovider.ID(cfg.Type)
+	keyProviderDescriptor, err := e.reg.GetKeyProviderDescriptor(id)
+	if err != nil {
+		if errors.Is(err, &registry.KeyProviderNotFoundError{}) {
+			return hcl.Diagnostics{&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Unknown key_provider type",
+				Detail:   fmt.Sprintf("Can not find %q", cfg.Type),
```
Yantrio, line 88:
```suggestion
				Detail:   fmt.Sprintf("Can not find encryption key provider of type %q", cfg.Type),
```

internal/encryption/keyprovider.go
```diff
@@ -0,0 +1,192 @@
...
+			Summary:  "Unable to build encryption key data",
+			Detail:   fmt.Sprintf("%s failed with error: %s", metakey, err.Error()),
+		})
+	}
+
+	meta := e.keyProviderMetadata[metakey]
+
+	data, newmeta, err := keyProvider.Provide(meta)
+	if err != nil {
+		e.keyValues[cfg.Type][cfg.Name] = cty.UnknownVal(cty.DynamicPseudoType)
```
Yantrio, line 174: @cam72cam can you add a comment here explaining why we add the dynamic type when things are unknown please?

internal/encryption/method/aesgcm/config.go
```diff
@@ -0,0 +1,24 @@
...
+
+package aesgcm
+
+import (
+	"fmt"
+	"github.com/opentofu/opentofu/internal/encryption/method"
+)
+
+type Config struct {
+	Key []byte `hcl:"cipher"`
```
Yantrio, line 14: tiny nitpick: naming

internal/encryption/keyprovider/keyprovider.go
Yantrio: We now handle this by passing through metadata between keyproviders.

internal/encryption/config.go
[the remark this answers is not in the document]
Yantrio: https://github.com/opentofu/opentofu/pull/1227/files#diff-b48996a7c3f29955173abf2555b30d81dc28c06b4b29adb5e8d068a6792eb275R73
This has now been implemented here.

cam72cam (approved):

PR #1235, opened 2024-02-06
1078 prerelease vs release

### LINKED ISSUE #1078: Separate release flow into one for stable releases and one for pre-releases.
The stable release flow should only be runnable from `v1.x` branches and should be the one we have today.

The pre-release flow (for alphas, betas, and release candidates) should be runnable from main and v1.x branches, and should **not** push linux archives (neither packagecloud nor snapcraft). Snapcraft does support separate edge/beta/candidate/stable channels, so eventually we could support this, but that's not important in this issue. Important is not to push the "stable" channel for pre-releases anymore.

siddharthasonker95:
Objective: 
The stable release flow should only be runnable from v1.x branches and should be the one we have today.

The pre-release flow (for alphas, betas, and release candidates) should be runnable from main and v1.x branches, and should not push linux archives (neither packagecloud nor snapcraft). Snapcraft does support separate edge/beta/candidate/stable channels, so eventually we could support that too.

Changes:

- Added new input variable `prerelease` in `workflow_dispatch` 
- Updated an existing variable in `workflow_dispatch` from `tag` to `ref` based on it's use case and for better code readability.
- Added a new variable `TRIGGER_PRERELEASE_FLOW`  and a conditional statement that stable release should be runnable for `v1.x` branch and pre-release flow should be runnable for `main` and `v1.x` branches.
- Added a new variable `SNAPCRAFT_CHANNEL` which is used to specify `grade` in **Snapcraft**. SNAPCRAFT_CHANNEL can be either **stable** or **edge**.

Resolves #1078 

## Target Release

1.7.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,6 @@ on:
       tag:
         description: "Git tag (leave empty for dry run)"
         required: false
-    inputs:
       prerelease:
         description: "Is this a pre-release?"
         required: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,8 @@ name: release
 on:
   workflow_dispatch:
     inputs:
-      ref:
-        description: "Git tag or branch (leave empty for dry run)"
+      tag:
+        description: "Git tag (leave empty for dry run)"
         required: false
       prerelease:
         description: "Is this a pre-release?"
@@ -16,8 +16,6 @@ jobs:
     name: Release
     runs-on: ubuntu-latest
     environment: gpg
-    env:
-      TRIGGER_PRERELEASE_FLOW: ${{ inputs.ref == 'main' || inputs.prerelease }}
     permissions:
       contents: write
       id-token: write
@@ -31,7 +29,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
-        if: startsWith(inputs.ref, 'v') || env.TRIGGER_PRERELEASE_FLOW
+        if: startsWith(inputs.tag, 'v')
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
@@ -41,16 +39,30 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          ref: ${{ inputs.ref }}
+          ref: ${{ inputs.tag }}
 
       - name: Fetch tags
         run: git fetch --force --tags
 
       - name: Compare versions
-        if: startsWith(inputs.ref, 'v')
+        if: startsWith(inputs.tag, 'v')
         run: ./.github/scripts/compare-release-version.sh
         env:
-          TARGET_VERSION: ${{inputs.ref}}
+          TARGET_VERSION: ${{inputs.tag}}
+
+      - name: Check if tag is on main branch
+        id: validate_tag
+        run: |
+          IS_TAG_ON_MAIN=$(git branch --contains ${{inputs.tag}} | grep -q "main" && echo true || echo false)
+          echo "IS_TAG_ON_MAIN=${IS_TAG_ON_MAIN}" >> $GITHUB_OUTPUT
+
+      - name: Check if release is allowed or not
+        id: validate_release
+        run: |
+          if [[ !inputs.prerelease && steps.validate_tag.outputs.IS_TAG_ON_MAIN == true ]]; then
+            echo "ERROR: Creating stable release from a tag on main is not allowed."
+            exit 1
+          fi
 
       - name: Determine Go version
         id: go
@@ -80,7 +92,7 @@ jobs:
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_TOKEN }}
 
       - name: Import GPG key
-        if: startsWith(inputs.ref, 'v') || env.TRIGGER_PRERELEASE_FLOW
+        if: startsWith(inputs.tag, 'v')
         run: |
           GPG_KEY_FILE=/tmp/signing-key.gpg
           echo "${{ secrets.GPG_PRIVATE_KEY }}" | base64 --decode > "${GPG_KEY_FILE}"
@@ -97,7 +109,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v4
         with:
           version: v1.21.2
-          args: release --clean --timeout=60m --snapshot=${{ !startsWith(inputs.ref, 'v') }}
+          args: release --clean --timeout=60m --snapshot=${{ !startsWith(inputs.tag, 'v') }}
         env:
           # Note: the GPG_FINGERPRINT and GPG_KEY_FILE are defined in the task above. If they are not set,
           # goreleaser won't sign the packages.
@@ -120,7 +132,7 @@ jobs:
           path: dist
 
       - name: Upload Debian packages to PackageCloud
-        if: startsWith(inputs.ref, 'v') && !inputs.prerelease
+        if: startsWith(inputs.tag, 'v') && !inputs.prerelease
         uses: computology/packagecloud-github-action@v0.6
         with:
                 PACKAGE-NAME: dist/*.deb
@@ -129,7 +141,7 @@ jobs:
                 PACKAGECLOUD-DISTRO: any/any
                 PACKAGECLOUD-TOKEN: ${{ secrets.PACKAGECLOUD_TOKEN }}
       - name: Upload RPM packages to PackageCloud
-        if: startsWith(inputs.ref, 'v') && !inputs.prerelease
+        if: startsWith(inputs.tag, 'v') && !inputs.prerelease
         uses: computology/packagecloud-github-action@v0.6
         with:
                 PACKAGE-NAME: dist/*.rpm
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
-        if: startsWith(inputs.ref, 'v') || env.ALLOW_MAIN_PRERELEASE
+        if: startsWith(inputs.ref, 'v') || env.TRIGGER_PRERELEASE_FLOW
         with:
           registry: ghcr.io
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -25,7 +29,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
-        if: startsWith(inputs.tag, 'v')
+        if: startsWith(inputs.tag, 'v') || (inputs.tag == 'main' && inputs.prerelease)
```
Yantrio, line 32: It may be worth pulling out `(inputs.tag == 'main' && inputs.prerelease)` to it's own variable. This way you can use it throughout the workflow easier and it improves readability.

right now I'm noticing there's areas where we miss the && `inputs.tag == 'main'

.github/workflows/release.yml
```diff
@@ -6,6 +6,10 @@ on:
       tag:
```
Yantrio, line 6: We should rename this variable to `ref` maybe, that way we can pass tags or branches. It's a bit strange to be checking if the `tag == main`

.github/workflows/release.yml
```diff
@@ -97,6 +101,8 @@ jobs:
           # goreleaser won't sign the packages.
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_TOKEN }}
+          SNAPCRAFT_CHANNEL: ${{ !inputs.prerelease &&  'stable' || 'edge' }}
```
Yantrio, line 104: It may be nicer here to invert the ternary.

.goreleaser.yaml
```diff
@@ -189,14 +189,14 @@ nfpms:
 
 snapcrafts:
   - name: opentofu
-    publish: true
+    publish: "{{ .Env.STABLE_RELEASE }}"
```
Yantrio, line 192: I have a feeling we need to keep this as a boolean, not move it to a string. But I could be wrong. I'm not sure how goreleaser handles this

Yantrio (changes requested):
Hi, Thanks for submitting this. I do have a few bits of feedback.

Regarding Testing:
Overall I think this is going to be difficult to test, so we may have to merge and test it live. Because we don't have a nicely documented way about how you can run our releases on your own fork.

Also, in the PR description you said
> . Snapcraft does support separate edge/beta/candidate/stable channels, so eventually we could support this, but that's not important in this issue.

But I see that this PR is making that change. What made you decide to do that?

DCO:
Also please sign the DCO, Thank you 🙏  You can find information about how to do this in our contributing guide or in the details of the failed check on the PR.

siddharthasonker95:
> port separate edge/beta

There were some initial changes in the branch by @cam72cam, which include modifications related to Snapcraft, so I followed those accordingly.

.goreleaser.yaml
siddharthasonker95: In my opinion, it should work as a boolean variable, as `STABLE_RELEASE` is being assigned the negation of inputs.prerelease in `release.yml`. However, we do need to fully test this first.

.github/workflows/release.yml
siddharthasonker95: Makes sense. Updated the code accordingly.

.github/workflows/release.yml
siddharthasonker95: Done.

.github/workflows/release.yml
```diff
@@ -25,7 +31,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
-        if: startsWith(inputs.tag, 'v')
+        if: startsWith(inputs.ref, 'v') || env.ALLOW_MAIN_PRERELEASE
```
Yantrio, line 34: should this be `TRIGGER_PRERELEASE_FLOW` ?

.goreleaser.yaml
Yantrio: But you're wrapping it in quotes, that is changing the type here from boolean to string.

Yantrio (changes requested):
I've raised 2 small nitpicks. Also, would you be able to update your PR description to explain what you're doing now with the snapcraft package channel. Thanks!

.goreleaser.yaml
Yantrio: We discussed this in a call, we think that we can drop the `publish` property completely here.

.github/workflows/release.yml
```diff
@@ -91,12 +97,14 @@ jobs:
...
         with:
           version: v1.21.2
-          args: release --clean --timeout=60m --snapshot=${{ !startsWith(inputs.tag, 'v') }}
+          args: release --clean --timeout=60m --snapshot=${{ !startsWith(inputs.ref, 'v') }}
         env:
           # Note: the GPG_FINGERPRINT and GPG_KEY_FILE are defined in the task above. If they are not set,
           # goreleaser won't sign the packages.
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_TOKEN }}
+          SNAPCRAFT_CHANNEL: ${{ inputs.prerelease &&  'edge' || 'stable' }}
```
Yantrio, line 106: @cube2222 as the owner of the snapcraft setup, are you happy with deploying to `edge`?

cube2222: Yes

.goreleaser.yaml
siddharthasonker95: removed the publish changes

Yantrio (dismissed):

.github/workflows/release.yml
```diff
@@ -3,15 +3,21 @@ name: release
...
+        required: true
+        type: boolean
 
 jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
     environment: gpg
+    env:
+      TRIGGER_PRERELEASE_FLOW: ${{ inputs.ref == 'main' && inputs.prerelease }}
```
cube2222, line 20: I would change the logic here slightly. It should be hard to do a non-prerelease by accident. Thus, I would suggest that it's a prerelase if it's on the main branch **or** input.prerelease is true.

One should not be able to make a non-prerelease release on the main branch. And if someone specifies inputs.prerelease as true, then it should also definitely not be a full release.

While someone specifying inputs.prerelease on a branch like v1.6 **should** create a **prerelease**.

I think this is just a case of changing this to `${{ inputs.ref == 'main' || inputs.prerelease }}`

siddharthasonker95: Done.

cube2222 (changes requested):
I'm sorry, this is actually my own confusion with what this was supposed to be, and I should've caught this earlier.

The idea is that all releases are triggered from a tag, but we want to make sure that people don't try to make a stable release using a tag that's pushed to the main branch.

In other words, the parts of this PR that check whether ref is a branch are unnecessary (left comments in all relevant places).

Now, a tag can be part of many branches (as a tag is really a commit pointer, and a commit can be on many branches). However, for stable release tags, I belive there is **no way** for them to be on the main branch, if the process is followed.

Thus, what I think would be the right thing to do as a validation is to use `git branch --contains` to check if the tag we're running with is contained by the main branch or not. Here's a relevant stackoverflow question: https://stackoverflow.com/questions/15806448/git-how-to-find-out-on-which-branch-a-tag-is

Basically, if `!inputs.prerelease` **and** `git branch --contains my_tag` contains main, then we can fail, saying that the maintainer is trying to create a stable release from a tag on main.

In the future we could actually replace the input.prerelease with just checking whether the tag is a prerelease tag or not (so whether it's of the form v1.2.3-xyz, or just v1.2.3). But I don't think that is necessary for now.

Again, apologies, I should've caught that earlier.

siddharthasonker95:
> I'm sorry, this is actually my own confusion with what this was supposed to be, and I should've caught this earlier.
> 
> The idea is that all releases are triggered from a tag, but we want to make sure that people don't try to make a stable release using a tag that's pushed to the main branch.
> 
> In other words, the parts of this PR that check whether ref is a branch are unnecessary (left comments in all relevant places).
> 
> Now, a tag can be part of many branches (as a tag is really a commit pointer, and a commit can be on many branches). However, for stable release tags, I belive there is **no way** for them to be on the main branch, if the process is followed.
> 
> Thus, what I think would be the right thing to do as a validation is to use `git branch --contains` to check if the tag we're running with is contained by the main branch or not. Here's a relevant stackoverflow question: https://stackoverflow.com/questions/15806448/git-how-to-find-out-on-which-branch-a-tag-is
> 
> Basically, if `!inputs.prerelease` **and** `git branch --contains my_tag` contains main, then we can fail, saying that the maintainer is trying to create a stable release from a tag on main.
> 
> In the future we could actually replace the input.prerelease with just checking whether the tag is a prerelease tag or not (so whether it's of the form v1.2.3-xyz, or just v1.2.3). But I don't think that is necessary for now.
> 
> Again, apologies for not catching this earlier.

@cube2222 
As per my understanding, the release should not proceed further, and it should log an error and exit when the following condition is met: `!inputs.prerelease and git branch --contains my_tag` contains main.
On the contrary, the issue specifies that in case of pre release skip pushing linux archives instead of logging an error.
Please correct me if I am wrong.

cube2222:
@siddharthasonker95 

Yeah
- if it's a prerelease (and is thus running with a tag that's either on main or on a version branch) it should skip pushing archives
- if it's not a prerelease and the tag is on main, then it should error

cube2222 (changes requested):
Actually, it seems to me like this will still push the :latest tag of the docker image on prereleases? That is definitely not something we want to happen.

I suppose we could do it similar to snapcraft - push latest for stable releases, and edge for prereleases.

EDIT: Also, we definitely don't want to push the major and major.minor docker images on prerelease.

siddharthasonker95:
I have tested the changes in the `release.yml` file (mentioned in this PR) in this [forked repository](https://github.com/opentofu-ci-test/opentofu/actions). 
Please refer to this repository for checking and validating any edge cases.

cam72cam (approved):

janosdebugs (approved):

PR #1237, opened 2024-02-07
Post weekly update for 20230207.

cube2222:


WEEKLY_UPDATES.md
```diff
@@ -1,5 +1,27 @@
 # Weekly Updates
 
+## 2023-02-07
+
+The last two weeks we've been focusing on working towards the 1.7.0 release, and on going through our issue backlog.
+
+This weekend some of us have been to FOSDEM and had a stand there - it was amazing to hear in person how many people have already migrated to OpenTofu! We've also distributed
```
Yantrio, line 7: "We've also distributed" what?

Yantrio (changes requested):

WEEKLY_UPDATES.md
```diff
@@ -1,5 +1,27 @@
...
+
+Additionally, most of the core team is off on winter holidays this week, so it's fairly quiet.
+
+- Current Status and Up Next
+  - We've been iterating on the skeleton of the client-side state encryption project, in order to get that right. [A PR for this has been opened recently and is pending reviews](https://github.com/opentofu/opentofu/pull/1227).
+  - The removed block project also [has a PR up](https://github.com/opentofu/opentofu/pull/1158), and is pending reviews.
+  - As part of adding for_each to import blocks, we've refactored the import code to be unified across the block implementation and the cli command implementation. [This is posted as a PR](https://github.com/opentofu/opentofu/pull/1207), reviewed, and will be merged soon.
```
RLRabinowitz, line 14: > we've refactored the import code to be unified across the block implementation and the cli command implementation

We haven't really unified the implementation here, but actually went the other way - making the difference between a CommandLine triggered import and an ImportBlock triggered import more obvious (For example, import targets from CLI already have addresses resolved, but for the import block that's no longer gonna be the case)

WEEKLY_UPDATES.md
cube2222: Looks like a copilot completion I thought I've applied.

WEEKLY_UPDATES.md
cube2222: Thanks! Adjusted the messaging.

RLRabinowitz (approved):

Yantrio (approved):

IS2511:
Wrong year?

Issue #1201, opened 2024-01-26
Add `windows_amd64` to our Pull Request checks unit tests matrix


ghost:
Currently, not all tests pass when run on Windows. This prevents us from running the tests in the CI pipeline, see #1135. If you have a Windows box, please help us by fixing a few unit tests.

sshiv5768:
Hey, @janosdebugs I would like to contribute to this issue.

janosdebugs:
Hey @sshiv5768 thank you very nuch for your help, I've assigned you. There are a number of issues, so to make reviewing easier I'd suggest to go one package at a time and submit multiple PRs.

lettucemode:
@janosdebugs I'm interested in being a co-assignee on this if that's doable.

Yantrio:
I think because we do have a number of issues it may be best to break this up and do a PR for each as Janos suggested above.

@sshiv5768 would you be able to share any progress you've made and if @lettucemode could also take some of the work? 

If not, we do have other issues under the `help wanted` label you could look at @lettucemode :)

sshiv5768:
I also think that would be really good. I have not started yet, actually I was not able to setup on windows.

lettucemode:
Happy to help out then. @Yantrio did you want to split this issue up into multiple issues and PR each, or just multiple PRs all `part of`ing this issue?

If the latter, maybe people could use this issue to claim which suite they are working on. For example I have a PR ready for the `/cmd/tofu` suite, so I'll claim that. However you want to organize it is fine with me, though.

For reference, here's the current output of `go clean -testcache && TF_ACC=1 gotestsum ./...` on my Windows machine.
<details>
<summary>Output</summary>

```
✖  cmd/tofu (114ms)
✓  internal/backend/init (136ms)
✖  internal/backend/remote-state/s3 (268ms)
✖  internal/backend/remote-state/kubernetes (344ms)
✖  internal/addrs (1.503s)
✖  internal/backend/remote-state/consul (1.733s)
✖  internal/backend/remote-state/azure (1.892s)
✖  internal/backend/remote-state/pg (1.805s)
✓  internal/backend/remote-state/cos (1.998s)
✓  internal/backend/remote-state/gcs (2.013s)
✓  internal/builtin/provisioners/file (1.216s)
✖  internal/backend (2.499s)
✓  internal/builtin/providers/tf (1.294s)
✓  internal/backend/remote-state/inmem (2.125s)
[...]

PR #1242, opened 2024-02-09
fix: `cmd/tofu` unit tests on windows

lettucemode:
Fixes the `cmd/tofu` unit test suite on Windows.

Part of #1201

### DIFF
--- a/cmd/tofu/main_test.go
+++ b/cmd/tofu/main_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"testing"
 
 	"github.com/mitchellh/cli"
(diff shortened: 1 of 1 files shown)


cmd/tofu/main_test.go
```diff
@@ -322,6 +323,9 @@ func TestMkConfigDir_new(t *testing.T) {
 
 	mode := int(info.Mode().Perm())
 	expectedMode := 0755
+	if runtime.GOOS == "windows" {
```
Yantrio, line 326: I think for our future selves it may be best if we could document why this is how it is.

Would you mind adding a small comment here explaining why its 777 and not 755 please?

lettucemode: No problem - comment added, lemme know if you were looking for something different.

Yantrio (approved):

cam72cam (approved):

PR #1243, opened 2024-02-09
Issue 248: left-over resources after tofu test should be written to a file

### LINKED ISSUE #248: When `tofu test` has left-over resources in the state file, it should write it out to a file for investigation purposes.
### OpenTF Version

```shell
`OpenTF v1.6.0-dev`
```

### Use Cases

As `opentf test` runs `opentf apply` and `opentf destroy` if for some reason OpenTF fails to destroy the resources it only prints the OpenTF resource ID, which makes the cleaning process hard as you don't know the actual Cloud Resource Id.

`main.tf`
```
resource "null_resource" "failing" {
  depends_on = [null_resource.failing_will_depend_on_me]
  provisioner "local-exec" {
    when    = destroy
    command = "exit 1"
  }
}

resource "null_resource" "failing_will_depend_on_me" {

}

check "check" {
  assert {
    condition = length(null_resource.failing.id) > 0
    error_message = "invalid value"
  }
}
```

`main.tftest.hcl`
```
run "test" {
}
```

```
> opentf test
main.tftest.hcl... pass
  run "test"... pass
OpenTF encountered an error destroying resources created while executing main.tftest.hcl/test.
╷
│ Error: local-exec provisioner error
│ 
│   with null_resource.failing,
│   on main.tf line 3, in resource "null_resource" "failing":
│    3:   provisioner "local-exec" {
│ 
│ Error running command 'exit 1': exit status 1. Output: 
╵

OpenTF left the following resources in state after executing main.tftest.hcl/test, and they need to be cleaned up manually:
  - null_resource.failing
  - null_resource.failing_will_depend_on_me

Success! 1 passed, 0 failed.
```

And as `opentf test` manages the state in memory, I can't print the resource id by using `opentf state show ...`

### Attempted Solutions

_No response_

### Proposal

Print the actual cloud resource ids after OpenTF resource id

### References

_No response_

Kazzhar:
## When tofu test has left-over resources in the state file, it should write it out to a file for investigation purposes.

When `tofu test` is run, in case certain resources fail to destroy, they are logged out to the console by `DestroySummary()` implemented in `internal/command/views/test.go`. 

### I have made the following changes:
- Since the issue asks to dump the state file after these logs, I make a call to a new function I declared and implemented, `SaveErroredTestStateFile()`, in `internal/command/views/test.go`. This function stores the statefile in a file named `errored_test.tfstate`. The implementation of it is simple, it calls a pre-existing function named `WriteStateForMigration()` used throughout the OpenTofu codebase wherever there was a need to store a file to the disk. The function is invoked in `internal/command/test.go`, after `DestroySummary()` is called.

- Updated the log message to let the user know that a statefile is being created in case of an error during `tofu test`

- Corresponding to the aforementioned log message change, I have changed the associated tests for `DestroySummary` in `internal/command/views/test_test.go` .

- Added tests for `SaveErroredTestStateFile()`, required data is fed, `SaveErroredTestStateFile` is invoked to create the statefile, the existence of the statefile created is being checked, and later on deleted.

Tested locally, `errored_test.tfstate` is created with no problems. 

Resolves #248 

## Target Release

1.7.0


internal/states/statemgr/filesystem.go
```diff
@@ -92,6 +92,16 @@ func NewFilesystemBetweenPaths(readPath, writePath string) *Filesystem {
 	}
 }
 
+// SaveErroredTestStateFile saves the state present in memory as a stateFile in disk
+// and is invoked when "tofu test" fails to destroy resources.
+func SaveErroredTestStateFile(stateFilePath string, state *states.State) error {
+	log.Printf("Writing state to file: errored_test.tfstate")
+	fs := NewFilesystem(stateFilePath)
+	schema := &tofu.Schemas{}
+	err := WriteAndPersist(fs, state, schema)
```
Yantrio, line 101: I'm concerned here that WriteState ends up being called here and this will attempt to refresh the state. https://github.com/opentffoundation/opentf/blob/b66111f8bf80d0bdb92684f9287e0751487902f6/internal/states/statemgr/filesystem.go#L139-L144

I have a feeling we should be calling `WriteStateForMigration` similar to how the errored state writing works.

Is there any chance we could re-work this PR to use code similar to [this](https://github.com/opentffoundation/opentf/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/backend/local/backend_apply.go#L284-L335) (or extract this code out to a re-usable method where you can pass different filenames?

Yantrio (changes requested):
Thank you for your submission. I've reviewed this and left some comments.

Would you also be able to add some tests to ensure that this is working as expected please? Thanks!

internal/states/statemgr/filesystem.go
```diff
@@ -92,6 +92,16 @@ func NewFilesystemBetweenPaths(readPath, writePath string) *Filesystem {
 	}
 }
 
+// SaveErroredTestStateFile saves the state present in memory as a stateFile in disk
+// and is invoked when "tofu test" fails to destroy resources.
+func SaveErroredTestStateFile(stateFilePath string, state *states.State) error {
+	log.Printf("Writing state to file: errored_test.tfstate")
```
Yantrio, line 98: I don't think this log message should be the responsibility of the filesystem logic. we only seem to log inside filesystem for trace logs.

I would recommend moving this log up a level, and using the stream to write direct to the view.

internal/states/statemgr/filesystem.go
```diff
@@ -92,6 +92,16 @@ func NewFilesystemBetweenPaths(readPath, writePath string) *Filesystem {
 	}
 }
 
+// SaveErroredTestStateFile saves the state present in memory as a stateFile in disk
+// and is invoked when "tofu test" fails to destroy resources.
+func SaveErroredTestStateFile(stateFilePath string, state *states.State) error {
```
Yantrio, line 97: I don't think we need to pass in the stateFilePath as it never changes. Hardcoding the value as it's passed into the NewFilesystem call is fine.

internal/states/statemgr/filesystem.go
Kazzhar: > I have a feeling we should be calling `WriteStateForMigration` similar to how the errored state writing works.
> 
> Is there any chance we could re-work this PR to use code similar to [this](https://github.com/opentffoundation/opentf/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/backend/local/backend_apply.go#L284-L335) (or extract this code out to a re-usable method where you can pass different filenames?

I am a little unsure of how do you want me to change this part. Do you want me to write a function similar to `backupStateForError`, where it attempts to write a backup for `errored.tfstate`, and even that fails it logs out the state file to the console as a last ditch attempt. Or should I only make a call to `WriteStateForMigration` where the write operation occurs without an attempt to refresh. Please do let me know,  I have made the other changes requested and will push changes after this. Thanks!

Yantrio: You're right, and I missed that WriteStateForMigration calls refresh too. Maybe this isn't so bad, but I've not tracked down yet what is passed around without debugging this.

In my original comment I asked

> Is there any chance we could re-work this PR to use code similar to [this](https://github.com/opentffoundation/opentf/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/backend/local/backend_apply.go#L284-L335) (or extract this code out to a re-usable method where you can pass different filenames?

And I think this still stands. Mainly for the following reasons:

1. The code linked has great diagnostic logging and error handling
2. The code linked just has a simple call out to `local.WriteStateForMigration` which your code can also use.
3. I don't see the point in introducing a new method that has a cut down version of the logic when the code already exists in some form (that we could either reuse, abstract out or base new work on top of)

Kazzhar: > Is there any chance we could re-work this PR to use code similar to [this](https://github.com/opentffoundation/opentf/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/backend/local/backend_apply.go#L284-L335) (or extract this code out to a re-usable method where you can pass different filenames?

Hi, here is what I have tried:
- I went ahead and tried extracting the code to a method, to which I would pass a filename and it would handle storing the state to the file and associated diagnostics and error handling. 

- I was unsure of where to place this method, I initially placed it in `filesystem.go` [https://github.com/opentofu/opentofu/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/states/statemgr/filesystem.go](url), however that gave me a cyclic dependency with package imports. Finding a suitable place to write this method would solve this problem, I would like your insights on this.

- I am going ahead and replicating the contents of 
`backupStateForError` https://github.com/opentofu/opentofu/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/backend/local/backend_apply.go#L290-L334
in `DestorySummary` https://github.com/opentofu/opentofu/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/command/views/test.go#L233

I personally don't like the idea of having similar code in two places, I would much rather have a method that can accept the filename as a parameter. Please do let me know of your opinions on how to tackle this.

internal/states/statemgr/filesystem.go
Kazzhar: done

internal/states/statemgr/filesystem.go
Kazzhar: done

internal/command/views/test.go
```diff
@@ -220,7 +221,6 @@ func (t *TestHuman) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Ru
 	if diags.HasErrors() {
 		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("OpenTofu encountered an error destroying resources created while executing %s.\n", identifier), t.view.errorColumns()))
 	}
-	t.Diagnostics(run, file, diags)
 
 	if state.HasManagedResourceInstanceObjects() {
 		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, and they need to be cleaned up manually:\n", identifier), t.view.errorColumns()))
```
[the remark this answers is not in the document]
Kazzhar: done

internal/command/views/test.go
```diff
@@ -220,17 +221,22 @@ func (t *TestHuman) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Ru
 	if diags.HasErrors() {
 		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("OpenTofu encountered an error destroying resources created while executing %s.\n", identifier), t.view.errorColumns()))
 	}
+
 	t.Diagnostics(run, file, diags)
 
 	if state.HasManagedResourceInstanceObjects() {
-		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, and they need to be cleaned up manually:\n", identifier), t.view.errorColumns()))
+		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, the left-over resources can be cleaned using the statefile written to disk(errored_test.tfstate) or can be done manually:\n", identifier), t.view.errorColumns()))
```
Yantrio, line 228: I think the wording of this is a bit unclear, making me end users should be running some commands against the errored_test.tfstate file to remove things.

When in reality this feature is to just dump the statefile to disk so that people have a reference. Or am I incorrect in this assumption?

internal/command/views/test.go
```diff
@@ -570,3 +579,45 @@ func testStatus(status moduletest.Status) string {
...
+// SaveErroredTestStateFile is a helper function to invoked in DestorySummary
+// to store the state to errored_test.tfstate and handle associated diagnostics and errors with this operation
+func SaveErroredTestStateFile(state *states.State, op Operation) tfdiags.Diagnostics {
+	var diags tfdiags.Diagnostics
+	localFileSystem := statemgr.NewFilesystem("errored_test.tfstate")
+	stateFile := statemgr.NewStateFile()
+	stateFile.State = state
+	writeErr := localFileSystem.WriteStateForMigration(stateFile, true)
+	if writeErr != nil {
+		if dumpErr := op.EmergencyDumpState(stateFile); dumpErr != nil {
```
Yantrio, line 592: It may be nice to have some comments here explaining why we dump the state here like this so nobody comes by and cleans this up in the future.

internal/command/views/test.go
```diff
@@ -570,3 +579,45 @@ func testStatus(status moduletest.Status) string {
...
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Failed to persist state",
+			stateWriteConsoleFallbackError,
+		))
+	}
+	return diags
+}
+
+const stateWriteFatalErrorFmt = `Failed to save state after apply.
```
Yantrio, line 608:
```suggestion
const stateWriteFatalErrorFmt = `Failed to save state after an errored test run.
```

internal/command/views/test.go
```diff
@@ -570,3 +579,45 @@ func testStatus(status moduletest.Status) string {
...
+
+Error serializing state: %s
+
+A catastrophic error has prevented OpenTofu from persisting the state file or creating a backup. Unfortunately this means that the record of any resources created during this tes has been lost, and such resources may exist outside of OpenTofu's management.
+
+For resources that support import, it is possible to recover by manually importing each resource using its id from the target system.
```
Yantrio, line 614: This line doesn't apply to tests.

Yantrio (changes requested):
Just some tiny nitpicks about some error wording.

Would you mind adding some test cases for this too please? The tests will help us identify regressions in the future, act as documentation for the expected behavior and also help give us some security in the knowledge that this works.

internal/command/views/test.go
Kazzhar: https://github.com/opentofu/opentofu/issues/248#issuecomment-1837497775

Here is a reference to what was discussed in the issue, I assumed that if we stored a state file then there was a way by which we could source that particular state file and then run `tofu destroy` or other tofu commands. Correct me if I am wrong please.

internal/command/views/test.go
Kazzhar: Understood, will add a few comments explaining what is being done.

internal/command/views/test.go
```diff
@@ -223,14 +224,18 @@ func (t *TestHuman) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Ru
 	t.Diagnostics(run, file, diags)
 
 	if state.HasManagedResourceInstanceObjects() {
-		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, and they need to be cleaned up manually:\n", identifier), t.view.errorColumns()))
+		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, the left-over resources can be cleaned using the statefile written to disk(errored_test.tfstate) or can be done manually:\n", identifier), t.view.errorColumns()))
```
Yantrio, line 227: tiny tiny tiny nitpick:

```suggestion
		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, the left-over resources can be managed using the statefile written to disk(errored_test.tfstate) or can be done manually:\n", identifier), t.view.errorColumns()))
```

Yantrio: (same for the other instances of this error message)

internal/command/views/test.go
janosdebugs: I think in this case the user needs to clean up by hand, I wouldn't encourage them to run commands against something that's broken, there's a chance to make an even bigger mess.

Evi1Pumpkin: I agree with @janosdebugs. Let's keep it simple. If a need arises, we can add this as a separate feature.

internal/command/views/test.go
Kazzhar: done 👍🏼

internal/command/views/test.go
```diff
@@ -223,14 +224,18 @@ func (t *TestHuman) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Ru
 	t.Diagnostics(run, file, diags)
 
 	if state.HasManagedResourceInstanceObjects() {
-		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, and they need to be cleaned up manually:\n", identifier), t.view.errorColumns()))
+		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, the left-over resources can be managed using the statefile written to disk(errored_test.tfstate) or can be done manually:\n", identifier), t.view.errorColumns()))
```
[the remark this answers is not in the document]
Kazzhar: Yes, that makes sense to me. The error message that I had written makes sense only if other tofu commands could be run on the state file. Since we aren't doing that anymore, the message will have to be changed.

Yantrio (dismissed):
LGTM 👍 thanks for this.

internal/command/views/test.go
```diff
@@ -223,14 +224,18 @@ func (t *TestHuman) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Ru
 	t.Diagnostics(run, file, diags)
 
 	if state.HasManagedResourceInstanceObjects() {
-		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, and they need to be cleaned up manually:\n", identifier), t.view.errorColumns()))
+		t.view.streams.Eprint(format.WordWrap(fmt.Sprintf("\nOpenTofu left the following resources in state after executing %s, these left-over resources can be viewed by reading the statefile written to disk(errored_test.tfstate).\n", identifier), t.view.errorColumns()))
```
RLRabinowitz, line 227: Personally - I'd still say here that "they need to be cleaned up manually", as I'm not sure it's now clear that the user should do that (Same for all other places you've changed the text)

internal/command/views/test.go
```diff
@@ -223,14 +224,18 @@ func (t *TestHuman) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Ru
...
 		for _, resource := range state.AllResourceInstanceObjectAddrs() {
 			if resource.DeposedKey != states.NotDeposed {
 				t.view.streams.Eprintf("  - %s (%s)\n", resource.Instance, resource.DeposedKey)
 				continue
 			}
 			t.view.streams.Eprintf("  - %s\n", resource.Instance)
 		}
+
+		t.view.streams.Eprint(format.WordWrap("\nWriting state to file: errored_test.tfstate\n", t.view.errorColumns()))
+		op := NewOperation('H', false, t.view)
```
RLRabinowitz, line 237: nit

Prefer `arguments.ViewHuman` over `'H'`

internal/command/views/test.go
```diff
@@ -570,3 +578,45 @@ func testStatus(status moduletest.Status) string {
 		panic("unrecognized status: " + status.String())
 	}
 }
+
+// SaveErroredTestStateFile is a helper function to invoked in DestorySummary
+// to store the state to errored_test.tfstate and handle associated diagnostics and errors with this operation
+func SaveErroredTestStateFile(state *states.State, op Operation) tfdiags.Diagnostics {
```
RLRabinowitz, line 584: It's technically not `SaveErroredTestStateFile`'s place to be inside `DestroySummary`.

1. `DestroySummary` is part of the view, and is only a view action. 
2. There's another usage of `DestroySummary`, at the start of `cleanup`, in cases of inconsistent state (only caused by a bug in OpenTofu). Not sure we'd like to save the state in this case

Personally I'd use `SaveErroredTestStateFile` outside of `DestroySummary`

internal/command/views/test.go
```diff
@@ -460,21 +465,24 @@ func (t *TestJSON) DestroySummary(diags tfdiags.Diagnostics, run *moduletest.Run
...
 				"@testfile", file.Name)
 		}
-
+		t.view.log.Info("Writing state to file: errored_test.tfstate")
 	}
-
+	op := OperationJSON{
+		view: t.view,
+	}
+	diags = diags.Append(SaveErroredTestStateFile(state, &op))
```
RLRabinowitz, line 485: You are always writing to the errored-state file (even if nothing errored), but you only write the log line `"Writing state to file"` when it errored

Please make sure the state file is only written when there's an error in destroy

Also, maybe adapt the tests to check the existence of the state file

RLRabinowitz (changes requested):
Thank you very much for this submission!

I've left some comments. 

Do note that the errored state is always being written (even if the logs appear only when there's an error during destroy)
We should probably adapt the tests to not miss this in the future (actually attempt to find the file, and maybe even check the state)

internal/command/views/test.go
Kazzhar: I am in agreement, the message does not clearly convey that the user has to manually clean up the resources after viewing the statefile. 

This is what I propose:
`OpenTofu left the following resources in state after executing %s, these left-over resources can be viewed by reading the statefile written to disk(errored_test.tfstate) and they need to be cleaned up manually:
`

WDYT @Yantrio ?

internal/command/views/test.go
Kazzhar: agreed, and done 👍🏼

internal/command/views/test.go
Kazzhar: I think this is a fair point. This is what I have in mind:

- The check for remaining resources is done by `state.HasManagedResourceInstanceObjects()`[here](https://github.com/opentofu/opentofu/blob/96b7fce5d1da52563260f6c445be14bfa0303357/internal/command/views/test.go#L225).

- Let's leave DestroySummary to perform logging/view related functionalities, writing to the file can be done in [here](url) instead, after invoking `state.HasManagedResourceInstanceObjects()` and checking if there are remnant resources.

WDYT?

internal/command/views/test.go
Kazzhar: This was an oversight on my end, fixed it 👍🏼.  

As for the tests, I am not clear on how you'd like the tests to be changed. Could you please elaborate on this?

internal/command/views/test.go
```diff
@@ -570,3 +569,61 @@ func testStatus(status moduletest.Status) string {
...
+
+A catastrophic error has prevented OpenTofu from persisting the state during an errored test run. 
+
+This is a serious bug in OpenTofu and should be reported.
+`
+
+const stateWriteConsoleFallbackError = `The errors shown above prevented OpenTofu from writing the state to
+the errored_test.tfstate. As a fallback, the raw state data is printed above as a JSON object.
+
+To retry writing this state, copy the state data (from the first { to the last } inclusive) and save it into a local file called errored_test.tfstate, and then run tofu commands on this state file.
```
Yantrio, line 628: We should re-word this to not promote the idea of running commands on the statefile.

internal/command/views/test.go
RLRabinowitz: Right now the tests only check the logs, checking whether or not the logs 
write `"Writing state to file"`, but they are not actually testing whether the file was created or not (and whether or not the state in the file is correct).

Also, right now, the test is testing out the `SaveErroredTestStateFile` function, and not the actual test command, like the other tests in the file. I personally think that `SaveErroredTestStateFile` didn't really have to be a public function.

Ideally, the tests should actually run the test command, and later check whether or not the state file exists, and what it contains. Doing this would have made the tests more precise, and would have detected this issue early on.

However, I would say that doing this is optional, in my opinion, as setting up such a test would not be too simple (would require mocking failure when destroying), so feel free to resolve this comment

internal/command/views/test.go
RLRabinowitz: Sounds good.

internal/command/views/test.go
```diff
@@ -570,3 +569,61 @@ func testStatus(status moduletest.Status) string {
...
+	var op Operation
+	switch v := view.(type) {
+	case *TestHuman:
+		op = NewOperation(arguments.ViewHuman, false, v.view)
+		v.view.streams.Eprint(format.WordWrap("\nWriting state to file: errored_test.tfstate\n", v.view.errorColumns()))
+	case *TestJSON:
+		op = &OperationJSON{
+			view: v.view,
+		}
+		v.view.log.Info("Writing state to file: errored_test.tfstate")
```
Kazzhar, line 591: Here's why I think that `SaveErroredTestStateFile` belongs inside `internal/command/views/test.go` (as a part of the views package):
- The creation of `OperationJson`/`OperationHuman` requires the `view` field of `TestJson`/`TestHuman`,  whose scope is limited to this file.

- The function mostly performs view operations, by writing to the view using a stream. `EmergencyDumpState` is also a view operation, it dumps the statefile to the terminal. The actual file write is done by a call to `WriteStateForMigration`, which  is a part of the filesystem logic.

internal/command/views/test.go
Kazzhar: I understand how testing doing that would make the tests more precise and inclusive overall, but in my opinion, I feel like this would mean unnecessary additional complexities (like mocking failure when destroying). 

Essentially we would be testing out the same function, but with the workflow of "tofu test" running from the start, instead of how `SaveErroredTestStateFile` is currently being tested. I took inspiration from how tests for DestroySummary has been written, I believe that the current testing should suffice.

internal/command/test.go
```diff
@@ -976,6 +976,9 @@ func (runner *TestFileRunner) Cleanup(file *moduletest.File) {
 		}
 		runner.Suite.View.DestroySummary(diags, state.Run, file, updated)
 
+		if updated.HasManagedResourceInstanceObjects() {
+			views.SaveErroredTestStateFile(updated, state.Run, file, runner.Suite.View)
```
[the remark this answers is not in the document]
RLRabinowitz: I see your point. But technically the `views` package should only do view related things.
Ideally, `views` package should implement a way to print out "I'm going to save the error state file", and to dump the error state file to screen / JSON.

Then, it would be this package's responsibility to: 1. Call `views` to print out "I'm going to save the error state file". 2. Attempt to write the error state file. 3. If write failed, call `views` to dump the error state file

However, this is a nit. Feel free to keep it as is, and resolve the comment

RLRabinowitz (approved):

Yantrio (dismissed):

cube2222:
It seems like a lot of tests are failing after the rebase due to changed interfaces after the addition of encryption. It seems like you need to add `encryption.StateEncryptionDisabled()` in a few places @Kazzhar

Yantrio (approved):

internal/command/views/test_test.go
```diff
@@ -3078,6 +3255,320 @@ func TestTestJSON_FatalInterruptSummary(t *testing.T) {
...
+	state  *states.State
+	run    *moduletest.Run
+	file   *moduletest.File
+	stderr string
+	want   interface{}
+}, viewType arguments.ViewType) {
+	for name, data := range tc {
+		t.Run(name, func(t *testing.T) {
+			// Create a temporary directory
+			tempDir := t.TempDir()
```
cube2222, line 3518: You're creating this temporary directory, but I don't see any place where this is cleaned up.

That makes sense for the actual running (obviously, we want people to be able to inspect the statefile, that's the point), but for unit tests we definitely need to clean up, or each unit test run will create more files on the users system.

cube2222 (changes requested):
@Kazzhar one nit left as far as I'm concerned

internal/command/views/test_test.go
cube2222: This is actually good, it's a testing-specific TempDir and is cleaned up automatically.

cube2222 (approved):

cube2222:
Thanks for the contribution!

PR #1250, opened 2024-02-12
Allow for templatefile recursion (up to 1024 depth default)

### LINKED ISSUE #402: Accept recursive calls to templatefile()
### OpenTF Version

```shell
1
```

### Use Cases

2

### Attempted Solutions

4

### Proposal

5

### References

6

cam72cam:
This changes templatefile to track call depth when building templatefile functions.  It works fine, even when nested inside other function calls.

As mentioned in #402, due to the lack of tail recursion it can be memory intensive during evaluation.  In my opinion this is a reasonable trade off and the default max depth should prevent most OOM crashes during reasonable use.

Example:
```hcl
main.tf 
locals {
        x = templatefile("rec.tmpl", {})
}

rec.tmpl 
${ base64encode(templatefile("other.tmpl", {})) }

other.tmpl 
${ templatefile("rec.tmpl", {}) }

```

image

With `TF_LOG=debug`
image

TODO:
* Documentation changes to templatefile
* Changelog entry

Resolves #402

## Target Release

1.7.0

### DIFF
--- a/internal/lang/funcs/filesystem.go
+++ b/internal/lang/funcs/filesystem.go
@@ -9,8 +9,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"unicode/utf8"
 
 	"github.com/bmatcuk/doublestar/v4"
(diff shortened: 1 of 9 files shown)


internal/lang/funcs/filesystem.go
```diff
@@ -58,6 +61,50 @@ func MakeFileFunc(baseDir string, encBase64 bool) function.Function {
 	})
 }
 
+func templateMaxRecursionDepth() (int, error) {
+	envkey := "TF_TEMPLATE_RECURSION_DEPTH"
```
janosdebugs, line 65: I wonder if this shouldn't be hard coded. It seems like a footgun to change this because it would make modules break under certain situations.

internal/lang/funcs/filesystem.go
```diff
@@ -58,6 +61,50 @@ func MakeFileFunc(baseDir string, encBase64 bool) function.Function {
...
+
+type ErrorTemplateRecursionLimit struct {
+	sources []string
+}
+
+func (err ErrorTemplateRecursionLimit) Error() string {
+	trace := make([]string, 0)
+	maxTrace := 16
+
+	// Look for repitition in the first N sources
```
janosdebugs, line 85:
```suggestion
	// Look for repetition in the first N sources
```

internal/lang/funcs/filesystem.go
cam72cam: The default limit (1024) should work for all of the proposed scenarios so far.  However, I usually like to provide escape hatches for users where possible (even if they are not recommended).  Would documentation around using this environment variable as a last resort assuage your concerns?

cube2222 (approved):

cube2222 (approved):

janosdebugs (approved):

PR #1253, opened 2024-02-13
Added a User-Agent header for remote-state http backend client

NetMoose:
Added a User-Agent header for the http client to determine which client call to the http backend for change state.

Sometimes need to be able to send specially formatted state in different ways (for example, hiding sensitive data that will be shown in Swagger output but not hidden for terraform cli). In this case, would like to be able to identify which client sends the request for change state.

## Target release 

1.7.0

### DIFF
--- a/internal/backend/remote-state/http/client.go
+++ b/internal/backend/remote-state/http/client.go
@@ -61,6 +61,7 @@ func (c *httpClient) httpRequest(method string, url *url.URL, data *[]byte, what
 	// Work with data/body
 	if data != nil {
 		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Opentofu-http-client/1.1")
 		req.ContentLength = int64(len(*data))
 
 		// Generate the MD5


internal/backend/remote-state/http/client.go
```diff
@@ -61,6 +61,7 @@ func (c *httpClient) httpRequest(method string, url *url.URL, data *[]byte, what
 	// Work with data/body
 	if data != nil {
 		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Opentofu-http-client/1.1")
```
janosdebugs, line 64: I'm not sure about the capitalization, it should either be all lower case or correctly capitalized (`OpenTofu`). 

```suggestion
		req.Header.Set("OpenTofu-HTTP-client/1.1")
```

```suggestion
		req.Header.Set("opentofu-http-client/1.1")
```

Yantrio, line 64: I would prefer if we could use existing mechanisms to create a user agent.

For reference, you can see this code here: https://github.com/opentofu/opentofu/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/httpclient/useragent.go#L35-L51

This is used throughout the codebase to set the useragent on clients.

cam72cam:
I'm closing this PR for now.  Feel free to re-open @NetMoose if you have time to address the review comments.

PR #1254, opened 2024-02-13
#1212 Problem referencing output from test run in local variables block (tofu test)

### LINKED ISSUE #1212: Problem referencing output from test run in local variables block (tofu test)
### OpenTofu Version

```shell
OpenTofu v1.6.1
on darwin_arm64
+ provider registry.opentofu.org/hashicorp/aws v5.34.0
+ provider registry.opentofu.org/hashicorp/http v3.4.0
```

Extract from  https://github.com/crayfishx/tofu-test-demo/blob/main/tests/website.tftest.hcl

```hcl
run "deploy_html" {
  command = apply
  
  assert {
    condition = aws_s3_object.index.etag == filemd5("./site/index.html")
    error_message = "index.html MD5 hash does not match source"
  }
}

run "resolve_endpoint" {
  command = plan

  module {
    source = "./tests/e2e"
  }

  variables {
    endpoint_url = run.deploy_html.endpoint_url
  }

  assert {
    condition = data.http.request.status_code == 200
    error_message = "HTTP code ${data.http.request.status_code} returned from endpoint"
  }

}
```

### Debug Output

```
│ Error: Variables not allowed
│ 
│   on tests/website.tftest.hcl line 52, in run "resolve_endpoint":
│   52:     endpoint_url = run.deploy_html.endpoint_url
│ 
│ Variables may not be used here.
╵
```

### Expected Behavior

The variable should be assigned with the value of the endpoint_url output generated in the `deploy_html` run

### Actual Behavior

It seems that when trying to reference the output of the apply run deploy_html OpenTofu doesn't want to use a variable to assign it and gives me the above error that I cannot interpolate a variable here.

### Steps to Reproduce

1. `tofu init`
2. `tofu test`

### Additional Context

I'm trying to get this use case working based on existing tutorials I've seen - my understanding is that the module output endpoint_url should be able to be resolved from another run block as `run.deploy_html.endpoint_url`  - I understand this feature should be compatible with the test feature in Terraform, but I don't get this error when running under Terraform.

The repo used for this example is at https://github.com/crayfishx/tofu-test-demo

The offending block is: https://github.com/crayfishx/tofu-test-demo/blob/main/tests/website.tftest.hcl#L51-L53

### References

_No response_

siddharthasonker95:
**Objective**:

Fix the issue of Problem referencing output from test run in local variables block (tofu test).
Example: 
In the code snippet below, the expected behavior is that the variable should be assigned the value of the endpoint_url output generated in the deploy_html run.

```
run "deploy_html" {
  command = apply
  
  assert {
    condition = aws_s3_object.index.etag == filemd5("./site/index.html")
    error_message = "index.html MD5 hash does not match source"
  }
}

run "resolve_endpoint" {
  command = plan

  module {
    source = "./tests/e2e"
  }

  variables {
    endpoint_url = run.deploy_html.endpoint_url
  }

  assert {
    condition = data.http.request.status_code == 200
    error_message = "HTTP code ${data.http.request.status_code} returned from endpoint"
  }
}
```

**Changes**:

- Added `getEvalContextFromStates()` in `test.go` to extracts the relevant information from the states to create a context suitable for HCL evaluation.
- Provided `hcl.EvalContext` to `v.expr.Value()` which returns the value resulting from evaluating the expression in the given evaluation context. 

Resolves #1212

## Target Release

1.7.0


internal/command/meta_vars.go
```diff
@@ -231,11 +231,17 @@ func (m *Meta) addVarsFromFile(filename string, sourceType tofu.ValueSourceType,
...
 
 func (v unparsedVariableValueExpression) ParseVariableValue(mode configs.VariableParsingMode) (*tofu.InputValue, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
-	val, hclDiags := v.expr.Value(nil) // nil because no function calls or variable references are allowed here
+	var ctx *hcl.EvalContext // By default, the hcl.EvalContext will be nil.
+	if v.ctx != nil {
+		ctx = v.ctx
+	}
+
+	val, hclDiags := v.expr.Value(ctx)
```
Yantrio, line 244: This could be shortened down to just 
```suggestion
	val, hclDiags := v.expr.Value(v.ctx)
```

internal/command/meta_vars.go
```diff
@@ -231,11 +231,17 @@ func (m *Meta) addVarsFromFile(filename string, sourceType tofu.ValueSourceType,
 type unparsedVariableValueExpression struct {
 	expr       hcl.Expression
 	sourceType tofu.ValueSourceType
+	ctx        *hcl.EvalContext
```
Yantrio, line 234: I feel like this context shouldn't be attached to the unparsedVaraibleValueExpression, especially as we already have a similar type introduced in #1129 here: https://github.com/opentofu/opentofu/pull/1129/files#diff-16f7ccd0b34c619fda941531b649e919dcc79d87306854db9d212c3050296305R1026-R1044

Would you be able to talk through why you chose to attach the context here to the unparsedVariableValueExpression instead of re-using the existing type or introducing a new abstracted type?

internal/command/test.go
```diff
@@ -1028,6 +1029,34 @@ func buildInputVariablesForTest(run *moduletest.Run, file *moduletest.File, conf
...
+		mod := state.State.Modules[""] // Empty string is what is used by the module in the test runner
+		for outName, out := range mod.OutputValues {
+			outputs[outName] = out.Value
+		}
+		runCtx[state.Run.Name] = cty.ObjectVal(outputs)
+	}
+	ctx := &hcl.EvalContext{Variables: map[string]cty.Value{"run": cty.ObjectVal(runCtx)}}
+
+	return ctx
+}
```
Yantrio, line 1058: Am I correct in thinking that this logic is copied from [here](https://github.com/opentofu/opentofu/pull/1129/files#diff-16f7ccd0b34c619fda941531b649e919dcc79d87306854db9d212c3050296305R1062-R1076) ?

If so I think we could look at abstracting that logic out to a common method and updating the logic in command/test.go

internal/command/test.go
```diff
@@ -997,6 +997,7 @@ func buildInputVariablesForTest(run *moduletest.Run, file *moduletest.File, conf
 				variables[name] = unparsedVariableValueExpression{
 					expr:       expr,
 					sourceType: tofu.ValueFromConfig,
+					ctx:        getEvalContextFromStates(states),
```
Yantrio, line 1000: Is there a reason to keep reconstructing this context inside the loop instead of constructing it once and passing the same context to all the variables here?

internal/command/test.go
```diff
@@ -988,7 +988,7 @@ func (runner *TestFileRunner) Cleanup(file *moduletest.File) {
 // Crucially, it differs from prepareInputVariablesForAssertions in that it only
 // includes variables that are reference by the config and not everything that
 // is defined within the test run block and test file.
-func buildInputVariablesForTest(run *moduletest.Run, file *moduletest.File, config *configs.Config, globals map[string]backend.UnparsedVariableValue) (tofu.InputValues, tfdiags.Diagnostics) {
+func buildInputVariablesForTest(run *moduletest.Run, file *moduletest.File, config *configs.Config, globals map[string]backend.UnparsedVariableValue, states map[string]*TestFileState) (tofu.InputValues, tfdiags.Diagnostics) {
```
Yantrio, line 991: I talked this through with @cube2222, 

And we have a couple ideas about how to best approach this.

1. Construct the context at the top of this method, and evaluate each variable as we come across it. Then store a literal reference (evaluated) instead of an unknown value.

2. re-use the type for testVariableValueExpression in pr #1129 instead of tacking the evaluation context onto this type.

Which do you think is the better approach?

Yantrio (changes requested):
Thanks for submitting this @siddharthasonker95 !

I've left some comments in line in the code, but I also wanted to mention that this is lacking a lot of tests.

For making sure that we do not regress in the future, but also to assert that we've thought of and caught the edge cases would you please be able to add tests for all of the functionality introduced here please?

If you have any questions, please feel free to reach out!

internal/command/meta_vars.go
siddharthasonker95: I referred to the approach implemented in #1129 only. Since it was my initial task, before confirming the approach, I wanted to deep dive into the code myself and gain an understanding of the code ethics and approach we follow. 
I will go through the suggestions you've provided, and after that, I will resolve the comments.

siddharthasonker95:
> Thanks for submitting this @siddharthasonker95 !
> 
> I've left some comments in line in the code, but I also wanted to mention that this is lacking a lot of tests.
> 
> For making sure that we do not regress in the future, but also to assert that we've thought of and caught the edge cases would you please be able to add tests for all of the functionality introduced here please?
> 
> If you have any questions, please feel free to reach out!

Sure, I will add the test cases for the above code changes. I was waiting to confirm the approach first before writing the tests.

internal/command/test.go
cube2222: Yes, I think that makes sense.

Though I'd still recommend creating the evaluation context before the for loop and reusing it across testVariableValueExpression's

siddharthasonker95: Done.

internal/command/test.go
siddharthasonker95: Done

cube2222 (approved):
Looks great! I'd like to get one more pair of eyes on this, though. Seems like James requested changes, so let's wait for his approval too.

cube2222:
Actually, please also add a CHANGELOG entry in this PR @siddharthasonker95 

Also, after this is merged, we'll want to backport it to the 1.6 branch as well.

Yantrio (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ ENHANCEMENTS:
 * Added "cidrcontains" function. ([$366](https://github.com/opentofu/opentofu/issues/366))
 * Allow test run blocks to reference previous run block's module outputs ([#1129](https://github.com/opentofu/opentofu/pull/1129))
 * Support the XDG Base Directory Specification ([#1200](https://github.com/opentofu/opentofu/pull/1200))
+* Added support to resolve the issue of referencing output from a test run in the local variables block (tofu test). ([#1254](https://github.com/opentofu/opentofu/pull/1254))
```
Yantrio, line 21: I think this reads a little better: WDYT?

```suggestion
* Allow referencing the output from a test run in the local variables block of another run (tofu test). ([#1254](https://github.com/opentofu/opentofu/pull/1254))
```

siddharthasonker95: Done. Thanks

Yantrio:
@cam72cam as you've been working in this area recently, would you mind giving this one last sanity check/approval please?

cam72cam (approved):

cam72cam:
LGTM

PR #1262, opened 2024-02-13
fix s3 backup documentation - workspace_key_prefix

### LINKED ISSUE #1263: s3 backend: missing info
### Summary

You should add the info about the parameter in the s3 backend documentation: the path indicated by the sample in the documentation is wrong (you should add env:WORKSPACE)

A more appropriater sample should be

```hlc
 terraform {
  backend "s3" {
    bucket = "mybucket"
    key    = "default.tfstate"
    region = "eu-west-1"
    workspace_key_prefix = "env:"
  }
}
```

and the target s3 path file will be s3://mybucket/env:/WORKSPACE/default.tfstate

### Problem Statement

Missing info in the official documentation: the real s3 path is not the indicated path

### User-facing description

more clear doc

### Technical Description

web site doc

better doc

### Downsides

_No response_

### Unresolved Questions

_No response_

### Related Issues

_No response_

### Proof of Concept

_No response_

matteoredaelli:
Resolves #1263

## Target Release

1.7.0

### DIFF
--- a/website/docs/language/settings/backends/s3.mdx
+++ b/website/docs/language/settings/backends/s3.mdx
@@ -23,15 +23,16 @@ on the S3 bucket to allow for state recovery in the case of accidental deletions
 ```hcl
 terraform {
   backend "s3" {
-    bucket = "mybucket"
-    key    = "path/to/my/key"
-    region = "us-east-1"
+    bucket               = "mybucket"
+    workspace_key_prefix = "path/to"
+    key                  = "my/key" # default.tfstate
+    region               = "us-east-1"
   }
 }
 ```
 
 This assumes we have a bucket created called `mybucket`. The
-OpenTofu state is written to the key `path/to/my/key`.
+OpenTofu state is written to the key `path/to/WORKSPACE/my/key`.
 
 Note that for the access credentials we recommend using a
 [partial configuration](/docs/language/settings/backends/configuration#partial-configuration).


website/docs/language/settings/backends/s3.mdx
```diff
@@ -23,15 +23,16 @@ on the S3 bucket to allow for state recovery in the case of accidental deletions
...
+    workspace_key_prefix = "path/to"
+    key                  = "my/key" # default.tfstate
+    region               = "us-east-1"
   }
 }
 ```
 
 This assumes we have a bucket created called `mybucket`. The
-OpenTofu state is written to the key `path/to/my/key`.
+OpenTofu state is written to the key `path/to/WORKSPACE/my/key`.
```
RLRabinowitz, line 35: Actually, this is not correct if the workspace is the `default` workspace

RLRabinowitz (changes requested):
Hey @matteoredaelli thank you for your contribution!

The example at the top of the S3 backend doc page is intended to show to most basic example of usage. I believe that using `workspace_key_prefix` is not a basic need that every user of the S3 backend would need to set.

Specifically here, the docs are not wrong, they are just confusing.
- If the workspace is the `default` workspace, then the `workspace_key_prefix` does not matter at all, and the S3 bucket key would simply be the `key`
- If the workspace is any other workspace, then the S3 bucket key would be `/workspace_key_prefix/workspace/key`

I think that this part should need clarification.
As mentioned in [my comment on the issue](https://github.com/opentofu/opentofu/issues/1263#issuecomment-1943332452), the best course of action would be to change the explanation to something along the lines of:

```
This assumes we have a bucket created called `mybucket`. The OpenTofu state is written to the key `path/to/my/key`, assuming that the workspace used is the default workspace.
See https://opentofu.org/docs/language/settings/backends/s3/#s3-state-storage for more info
```

PR #1270, opened 2024-02-15
Make `import` block's `to` possibly more dynamic

RLRabinowitz:
Related to https://github.com/opentofu/opentofu/issues/1033

Current plan for implementing `for_each` for the `import` block:
1. Prepare `ImportTarget` types and codebase for more dynamic addresses (addresses that are not fully known from the get-go) **DONE**
2. Allow for having a more dynamic address in the `to` field of an `import` block **<--- You are here**
3. Add `for_each` field for `import` block, and adapt the code for resolving multiple import targets from the same import configuration

### Background - What this PR does
In order to implement `for_each` for the `import` block, the `to` address of it would need to be able to be more dynamic than a "static" reference

**Before this PR**

Import target addresses can only be of the following syntax: `module.my_module1["static key"].aws_s3_bucket.bucket[0]`. The address must be a representation of an address of a single resource instance.

**After this PR**

Import target addresses can be more dynamic then that. They still need to be in the format of `module.my_module1[expression1].aws_s3_bucket.bucket[expression2]`, but the expressions in the indexes can be more dynamic than that. 

You would be able to reference variables/locals/resources/datas, and OpenTofu would resolve these expressions before performing the import action

The following would be allowed:
- `module.my_module1[locals.my_local].aws_s3_bucket.bucket[locals.my_bool ? var.var1 : var.var2]`
- `module.my_module1[data.my_data_source.my_data].aws_s3_bucket.bucket[my_resource.res]`

### In detail

- When decoding the `import` block, decode the "static" part of the address, to be able to get an address for the resource (without yet knowing the instance keys, if there are any). Also, attempt to fully resolve the address, if possible, in order to keep some early validations
- When evaluating an import target, fully resolve the address
- Resolve imports before the expansion of resources, to avoid import target resolution happening multiple times
- Adapt validations/checks
  - Detect duplicate import blocks, when resolving the addresses of the import targets
  - Since we can't always tell if the state for the resource instance exists early on, move this check to the node

## Differences compared to how `import` block behaves in Terraform 1.7

- Terraform has harsher restrictions on import target addresses:
  - Only allows the expressions inside an index key that are literal, if `for_each` isn't used
  - Only allows the expressions inside an index key to contain `each.key`/`each.value`, if `for_each` is used
- Up until 1.6 Terraform did not error when attempting to run `terraform import` on a resource that is already in state. However, in 1.7 the behaviour changed to error "Resource already managed by Terraform". I did not change the behaviour of `tofu import` to do the same, as it is _technically_ a breaking change (albeit erroring here makes more sense)
  - I could change the behaviour of `tofu import` to error as well when state already exists, but I wanted to get more feedback on the beforehand
- Terraform only makes sure that no two `import` blocks target the same resource **if the resource addresses are static**. If two different `import` blocks have a dynamic address, and end up resolving to the same address, then the last resolved import block "wins" the import. In OpenTofu, this would error as a duplicate import block
```hcl
import {
  for_each = {
    one = "one"
  }
  to = random_integer.randomy[true ? "staging" : "production"]
  id = "2,1,10"
}

import {
  for_each = {
    one = "one"
  }
  to = random_integer.randomy[true ? "staging" : "production"]
  id = "5,1,10"
}
```
Would not fail in Terraform, and would result in importing `5,1,10` into the resource. However, in OpenTofu (currently, without the `for_each` field) it would error on a duplicate import block

## Target Release

1.7.0


cube2222:
> Terraform has harsher restrictions on import target addresses

I think your approach of being more liberal in expressions we accept makes sense, esp. since we move validations to a later stage.

> I could change the behaviour of tofu import to error as well when state already exists, but I wanted to get more feedback on the beforehand

I think it makes sense to error in this situation.

> Terraform only makes sure that no two import blocks target the same resource if the resource addresses are static. If two different import blocks have a dynamic address, and end up resolving to the same address, then the last resolved import block "wins" the import. In OpenTofu, this would error as a duplicate import block

I think erroring here makes sense. Frankly, I'm surprised that Terraform allows it, even though they error on imports to resources whose state already exists.

internal/configs/module.go
```diff
@@ -459,30 +459,14 @@ func (m *Module) appendFile(file *File) hcl.Diagnostics {
...
-			if mb.From.String() == i.To.String() {
-				diags = append(diags, &hcl.Diagnostic{
-					Severity: hcl.DiagError,
-					Summary:  "Cannot import to a move source",
-					Detail:   fmt.Sprintf("An import block for ID %q targets resource address %s, but this address appears in the \"from\" argument of a moved block, which is invalid. Please change the import target to a different address, such as the move target.", i.ID, i.To),
-					Subject:  &i.DeclRange,
-				})
-			}
-		}
-
```
Evi1Pumpkin, line 485: Removed unneeded logic. 
Without this logic, we still get proper errors: 
1. That the config resource in Import "to" is missing (as the Moved block requires removing the "From" resource).
2. Or, if we use the `-generate-config-out flag`, we'll get an error that the Moved "from" object still exists because it's recreated.

internal/tofu/transform_config.go
```diff
@@ -164,35 +161,18 @@ func (t *ConfigTransformer) transformSingle(g *Graph, config *configs.Config, ge
...
-		}
-
-		// In case of config generation - We can error early here in two cases:
-		// 1. When attempting to import a resource with a key (Config generation for count / for_each resources)
-		// 2. When attempting to import a resource inside a module.
 		if len(generateConfigPath) > 0 {
-			if address.Resource.Key != addrs.NoKey {
-				return fmt.Errorf("Config generation for count and for_each resources not supported.\n\nYour configuration contains an import block with a \"to\" address of %s. This resource instance does not exist in configuration.\n\nIf you intended to target a resource that exists in configuration, please double-check the address. Otherwise, please remove this import block or re-run the plan without the -generate-config-out flag to ignore the import block.", address)
-			}
-
```
Evi1Pumpkin, line 192: Moved this logic to internal/tofu/context_plan.go `validateImportTargets()`

internal/tofu/context_import.go
```diff
@@ -71,76 +72,80 @@ func (i *ImportTarget) IsFromImportCommandLine() bool {
...
+
+	importId, evalDiags := evaluateImportIdExpression(importTarget.Config.ID, rootCtx)
 	if evalDiags.HasErrors() {
-		return evalDiags.Err()
+		return evalDiags
 	}
 
-	// TODO - Change once an import target's address is more dynamic
-	importAddress := importTarget.Config.To
+	importAddress, addressDiags := rootCtx.EvaluateImportAddress(importTarget.Config.To)
```
cam72cam, line 126: This should be merging the evalDiags with addressDiags.  The code as written will skip warnings.  This comment also applies to the for loop

cam72cam (dismissed):

internal/tofu/context_import.go
Evi1Pumpkin: Fixed [Ensure ResolveImport code will support passing warnings](https://github.com/opentofu/opentofu/pull/1270/commits/78417b862bdc20cd4fbdbd359b35d40b751a55c0)

cam72cam (approved):

internal/tofu/context_plan.go
```diff
@@ -568,32 +563,100 @@ func (c *Context) postPlanValidateImports(importResolver *ImportResolver, allIns
 func (c *Context) findImportTargets(config *configs.Config, priorState *states.State) []*ImportTarget {
 	var importTargets []*ImportTarget
 	for _, ic := range config.Module.Import {
-		if priorState.ResourceInstance(ic.To) == nil {
-			importTargets = append(importTargets, &ImportTarget{
-				Config: ic,
-			})
-		}
+		importTargets = append(importTargets, &ImportTarget{
```
siddharthasonker95, line 566: If `priorState` is an unused variable, then we can remove it.

internal/configs/import.go
```diff
@@ -7,12 +7,41 @@ package configs
...
+	// To is the address HCL expression given in the `import` block configuration.
+	// It supports the following address formats:
+	// 1. aws_s3_bucket.my_bucket
+	// 2. module.my_module.aws_s3_bucket.my_bucket
+	// 3. aws_s3_bucket.my_bucket["static_key"]
+	// 4. module.my_module[0].aws_s3_bucket.my_buckets["static_key"]
+	// 3. aws_s3_bucket.my_bucket[expression]
+	// 4. module.my_module[expression].aws_s3_bucket.my_buckets[expression]
+	// A dynamic instance key supports a dynamic expression like - a variable, a local, a condition (for example,
+	// ternary), a resource block attribute, a data block attribute, etc.
```
Yantrio, line 27:
```suggestion
	// - aws_s3_bucket.my_bucket
	// - module.my_module.aws_s3_bucket.my_bucket
	// - aws_s3_bucket.my_bucket["static_key"]
	// - module.my_module[0].aws_s3_bucket.my_buckets["static_key"]
	// - aws_s3_bucket.my_bucket[expression]
	// - module.my_module[expression].aws_s3_bucket.my_buckets[expression]
	// - A dynamic instance key supports a dynamic expression like - a variable, a local, a condition (for example,
	//  ternary), a resource block attribute, a data block attribute, etc.
```

internal/tofu/context_import.go
```diff
@@ -71,80 +72,87 @@ func (i *ImportTarget) IsFromImportCommandLine() bool {
...
-	resolvedImportKey := ResolvedConfigImportsKey{
-		AddrStr: importAddress.String(),
-		ID:      importId,
+	resolvedImportKey := importAddress.String()
+
+	if importTarget, exists := ri.imports[resolvedImportKey]; exists {
+		return diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  fmt.Sprintf("Duplicate import configuration for %q", importAddress),
+			Detail:   fmt.Sprintf("An import block for the resource %q was already declared at %s. A resource can have only one import block.", importAddress, importTarget.Config.DeclRange),
```
Yantrio, line 144: Just for my curiousity, can someone declare an import block AND an import through the CLI somehow?

I wonder if we need to handle it with a more generic error message

internal/tofu/context_import.go
```diff
@@ -71,80 +72,87 @@ func (i *ImportTarget) IsFromImportCommandLine() bool {
...
 
 	ri.mu.Lock()
 	defer ri.mu.Unlock()
 
-	resolvedImportKey := ResolvedConfigImportsKey{
-		AddrStr: importAddress.String(),
-		ID:      importId,
+	resolvedImportKey := importAddress.String()
+
+	if importTarget, exists := ri.imports[resolvedImportKey]; exists {
```
Yantrio, line 140: do we need to be locking the ri.mu like we do in GetImport here?

internal/tofu/context_plan.go
```diff
@@ -568,32 +563,100 @@ func (c *Context) postPlanValidateImports(importResolver *ImportResolver, allIns
...
+			Config: ic,
+		})
 	}
 	return importTargets
 }
 
-func (c *Context) validateImportTargets(config *configs.Config, importTargets []*ImportTarget) (diags tfdiags.Diagnostics) {
+// validateImportTargets makes sure all import targets are not breaking the following rules:
+//  1. Imports are attempted to resources that do not exist (if config generation is not enabled).
```
Yantrio, line 574: tiny tiny tiny nitpick
```suggestion
//  1. Imports are attempted in to resources that do not exist (if config generation is not enabled).
```

internal/tofu/context_plan.go
```diff
@@ -568,32 +563,100 @@ func (c *Context) postPlanValidateImports(importResolver *ImportResolver, allIns
...
+			} else {
+				diags = diags.Append(importResourceWithoutConfigDiags(staticAddress.String(), imp.Config))
+				continue
+			}
 		}
 	}
 	return
 }
 
+func importConfigGenerationInModuleDiags(addressStr string, config *configs.Import) *hcl.Diagnostic {
```
Yantrio, line 618: I like that we pulled out the diag creations here, it makes it a bit more readable 👍

internal/tofu/eval_context_builtin.go
```diff
@@ -11,6 +11,9 @@ import (
 	"log"
 	"sync"
 
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/opentofu/opentofu/internal/lang/marks"
+
 	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
```
Yantrio, line 18: I think these imports need ordering

internal/tofu/eval_context_builtin.go
```diff
@@ -408,6 +411,108 @@ func (ctx *BuiltinEvalContext) EvaluateReplaceTriggeredBy(expr hcl.Expression, r
...
+
+	val, diags := ctx.EvaluateExpr(expr, cty.DynamicPseudoType, nil)
+	if diags.HasErrors() {
+		return idx, diags
+	}
+
+	if !val.IsKnown() {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Import block 'to' address contains invalid key",
```
Yantrio, line 475:
```suggestion
			Summary:  "Import block 'to' address contains an invalid key",
```

internal/tofu/eval_context_builtin.go
```diff
@@ -408,6 +411,108 @@ func (ctx *BuiltinEvalContext) EvaluateReplaceTriggeredBy(expr hcl.Expression, r
...
+			Detail:   "Import block contained a resource address using an index will only be known after apply. Please make sure to use expressions that are known at plan time for the index of an import target address",
```
Yantrio, line 476:
```suggestion
			Detail:   "Import block contained a resource address using an index that will only be known after apply. Please ensure to use expressions that are known at plan time for the index of an import target address",
```

internal/tofu/eval_context_builtin.go
```diff
@@ -408,6 +411,108 @@ func (ctx *BuiltinEvalContext) EvaluateReplaceTriggeredBy(expr hcl.Expression, r
...
+			Subject:  expr.Range().Ptr(),
+		})
+		return idx, diags
+	}
+
+	if val.IsNull() {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Import block 'to' address contains invalid key",
+			Detail:   "Import block contained a resource address using an index which is null. Please make sure the expression for the index is not null",
```
Yantrio, line 486:
```suggestion
			Summary:  "Import block 'to' address contains an invalid key",
			Detail:   "Import block contained a resource address using an index which is null. Please make sure the expression for the index is not null",
```

internal/tofu/eval_context_builtin.go
```diff
@@ -408,6 +411,108 @@ func (ctx *BuiltinEvalContext) EvaluateReplaceTriggeredBy(expr hcl.Expression, r
...
+			Subject:  expr.Range().Ptr(),
+		})
+		return idx, diags
+	}
+
+	if val.Type() != cty.String && val.Type() != cty.Number {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Import block 'to' address contains invalid key",
+			Detail:   "Import block contained a resource address using an index which is not valid for a resource instance (not a string or a number). Please make sure the expression for the index is correct, and returns either a string or a number",
```
Yantrio, line 496:
```suggestion
			Summary:  "Import block 'to' address contains an invalid key",
			Detail:   "Import block contained a resource address using an index which is not valid for a resource instance (not a string or a number). Please make sure the expression for the index is correct, and returns either a string or a number",
```

internal/tofu/eval_context_builtin.go
```diff
@@ -408,6 +411,108 @@ func (ctx *BuiltinEvalContext) EvaluateReplaceTriggeredBy(expr hcl.Expression, r
...
+		})
+		return idx, diags
+	}
+
+	unmarkedVal, valMarks := val.Unmark()
+	if _, sensitive := valMarks[marks.Sensitive]; sensitive {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Import block 'to' address contains invalid key",
+			Detail:   "Import block contained a resource address using an index which is sensitive. Please make sure indexes used in the resource address of an import target are not sensitive",
```
Yantrio, line 507:
```suggestion
			Summary:  "Import block 'to' address contains an invalid key",
			Detail:   "Import block contained a resource address using an index which is sensitive. Please make sure indexes used in the resource address of an import target are not sensitive",
```

internal/tofu/node_resource_abstract.go
```diff
@@ -220,7 +257,12 @@ func (n *NodeAbstractResource) RootReferences() []*addrs.Reference {
 			continue
 		}
 
-		refs, _ := lang.ReferencesInExpr(addrs.ParseRef, importTarget.Config.ID)
+		refs, _ := referencesInImportAddress(importTarget.Config.To)
+
+		// TODO - Add RootReferences of ForEach here later one, once for_each is added
```
Yantrio, line 262: Is this coming in another PR?

Yantrio (changes requested):
Overall it looks good to me. I don't have many qualms but im not super familiar with this area of importing.

The comments were super helpful and descriptive code was really helpful in reviewing this. Thanks!

I've left a few small comments

internal/tofu/context_plan.go
Evi1Pumpkin: Thanks. Fixed [here](https://github.com/opentofu/opentofu/pull/1270/commits/d06b19f31fd33929e4ce754a4dcc99d78a5a242a)

internal/tofu/context_import.go
Evi1Pumpkin: It cannot be done. When you run an import using the CLI, all other `import` blocks in the configuration are ignored.

internal/configs/import.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/ba46e21344ef6380bb31d023efc8f73ee2ccf1a3

internal/tofu/context_plan.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/a02afd9341273a015d3f1c4eba0ecbee0ddbeb41

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: I agree it looks weird.
But when I run `go run golang.org/x/tools/cmd/goimports -w -l internal/tofu/eval_context_builtin.go` it changes nothing (plus, all checks, including the imports ordering, passed for the commit).

internal/tofu/node_resource_abstract.go
Evi1Pumpkin: Yes, it'll be included in the next PR, supporting the ForEach syntax for the import block.

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: Thanks for catching these https://github.com/opentofu/opentofu/pull/1270/commits/ac25cf3ebf6457893dfc4412ff594d38093042a6

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/ac25cf3ebf6457893dfc4412ff594d38093042a6

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/ac25cf3ebf6457893dfc4412ff594d38093042a6

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/ac25cf3ebf6457893dfc4412ff594d38093042a6

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/ac25cf3ebf6457893dfc4412ff594d38093042a6

internal/tofu/eval_context_builtin.go
ollevche: Hey @Evi1Pumpkin! You need to delete an empty line and rerun sorting. `goimports` treats newlines as a separator for a "group" and thus it sorts imports inside each "group". This is the order I would expect with two "groups":

```
import (
	"context"
	"fmt"
	"log"
	"sync"

	"github.com/hashicorp/hcl/v2"
	"github.com/hashicorp/hcl/v2/hclsyntax"
	"github.com/opentofu/opentofu/internal/addrs"
	"github.com/opentofu/opentofu/internal/checks"
	"github.com/opentofu/opentofu/internal/configs/configschema"
	"github.com/opentofu/opentofu/internal/encryption"
	"github.com/opentofu/opentofu/internal/instances"
	"github.com/opentofu/opentofu/internal/lang"
	"github.com/opentofu/opentofu/internal/lang/marks"
	"github.com/opentofu/opentofu/internal/plans"
	"github.com/opentofu/opentofu/internal/providers"
	"github.com/opentofu/opentofu/internal/provisioners"
	"github.com/opentofu/opentofu/internal/refactoring"
	"github.com/opentofu/opentofu/internal/states"
	"github.com/opentofu/opentofu/version"
	"github.com/zclconf/go-cty/cty"
)
```

These "groups" could be configured in IDEs. Here is an example for builtin, local and external "groups":

```
import (
	// builtin
        "context"
	"fmt"
	"log"
	"sync"
	
	// local
	"github.com/opentofu/opentofu/internal/addrs"
	"github.com/opentofu/opentofu/internal/checks"
	"github.com/opentofu/opentofu/internal/configs/configschema"
	"github.com/opentofu/opentofu/internal/encryption"
	"github.com/opentofu/opentofu/internal/instances"
	"github.com/opentofu/opentofu/internal/lang"
	"github.com/opentofu/opentofu/internal/lang/marks"
	"github.com/opentofu/opentofu/internal/plans"
	"github.com/opentofu/opentofu/internal/providers"
	"github.com/opentofu/opentofu/internal/provisioners"
	"github.com/opentofu/opentofu/internal/refactoring"
	"github.com/opentofu/opentofu/internal/states"
	"github.com/opentofu/opentofu/version"
	
	// external
	"github.com/hashicorp/hcl/v2"
	"github.com/hashicorp/hcl/v2/hclsyntax"
	"github.com/zclconf/go-cty/cty"
)
```

internal/tofu/context_plan2_test.go
```diff
@@ -4286,6 +4286,407 @@ import {
...
+				}
+				if got, want := instPlan.Action, plans.NoOp; got != want {
+					t.Errorf("wrong planned action\ngot:  %s\nwant: %s", got, want)
+				}
+				if got, want := instPlan.ActionReason, plans.ResourceInstanceChangeNoReason; got != want {
+					t.Errorf("wrong action reason\ngot:  %s\nwant: %s", got, want)
+				}
+				if instPlan.Importing.ID != "123" {
+					t.Errorf("expected import change from \"123\", got non-import change")
+				}
```
Yantrio, line 4505: tiny nitpick: I'm aware that all the test cases right now use an id of `123`, but if they ever changed it could be annoying to have to edit this code too. 

How do you feel about defining the expected id of `123` in the testcases?

internal/tofu/eval_import.go
```diff
@@ -28,10 +27,6 @@ func evaluateImportIdExpression(expr hcl.Expression, ctx EvalContext) (string, t
 		})
 	}
 
-	// The import expression is declared within the root module
-	// We need to explicitly use that context
-	ctx = ctx.WithPath(addrs.RootModuleInstance)
-
```
Yantrio, line 34: Would you mind explaining why this was removed here? It's a bit unclear

Evi1Pumpkin: Yes, of course.
We moved it to the function that is calling `evaluateImportIdExpression()` [here](https://github.com/opentofu/opentofu/pull/1270/files#diff-b57ebc9c33d78d38d14349f826cce5d6b06348b1ce868061075d79c03e827d57R121). So now we get this value in the parameter `ctx EvalContext`.

internal/tofu/context_plan2_test.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1270/commits/c2b98f67597a15b23401363faa2f2c9440cbfecb

internal/tofu/eval_context_builtin.go
Evi1Pumpkin: Oh! thanks for the tip :)
fixed it here https://github.com/opentofu/opentofu/pull/1270/commits#:~:text=c2b98f6-,fix%20imports%20order,-%E2%80%A6

Yantrio (dismissed):

ollevche (approved):

Yantrio (approved):

PR #1277, opened 2024-02-18
Fix race when using `ResolvedImports`

RLRabinowitz:
Fix a race introduced in https://github.com/opentofu/opentofu/pull/1207 (example run [here](https://github.com/opentofu/opentofu/actions/runs/7950028478/job/21701859808?pr=1158))

Access to `imports` of `ResolvedImports` was not _technically_ safe, causing failures in the race tests (A race here shouldn't really have affected the code much for now, as the `ResolvedImports` was only used for validation that all import actions have been satisfied)

Here I fix the race by:
- Introducing a lock for accessing the `imports`
- Changed `ResolvedImports` to be `ImportResolver`, which has its own access functions for reading and writing the resolved imports (in preparation for https://github.com/opentofu/opentofu/pull/1270)

## Target Release

1.7.0

### DIFF
--- a/internal/tofu/context_import.go
+++ b/internal/tofu/context_import.go
@@ -7,6 +7,7 @@ package tofu
 
 import (
 	"log"
+	"sync"
 
 	"github.com/hashicorp/hcl/v2"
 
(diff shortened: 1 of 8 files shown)


janosdebugs (approved):

Evi1Pumpkin (approved):
LGTM!

PR #1279, opened 2024-02-19
issue 966: different user agent for unit tests

### LINKED ISSUE #966: Unit tests should use a different useragent to help differentiate in logging
Right now, our unit tests are indistinguishable in our registry statistics, and we would like to try and omit any unit test hits to our registry by using a different user agent so we can determine what is real usage, and what is unit testing.

SypherSP:
Resolves #966

## Target Release

1.7.0

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -37,8 +37,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      TF_APPEND_USER_AGENT: TEST
-      OPENTOFU_USER_AGENT: TEST
+      TF_APPEND_USER_AGENT: INTEGRATION-TEST
+      OPENTOFU_USER_AGENT: OPENTOFU
 
     steps:
       - name: "Fetch source code"
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -36,6 +36,10 @@ jobs:
     name: "Unit Tests"
     runs-on: ubuntu-latest
 
+    env:
+      TF_APPEND_USER_AGENT: TEST
+      OPENTOFU_USER_AGENT: TEST
+
     steps:
       - name: "Fetch source code"
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -9,7 +9,7 @@
 # The build.yml workflow includes some additional checks we run only for
 # already-merged changes to release branches and tags, as a compromise to
 # keep the PR feedback relatively fast. The intent is that checks.yml should
-# catch most problems but that build.yml might occasionally by the one to catch
+# catch most problems but that build.yml might occasionally be the one to catch
 # more esoteric situations, such as architecture-specific or OS-specific
 # misbehavior.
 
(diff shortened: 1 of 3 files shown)


SypherSP:
Have made this PR to confirm if this is the expected change. 
Also need suggestions for what the environment variables should be set to.

Also, currently looking for other locations where this needs to be done.

.github/workflows/checks.yml
```diff
@@ -36,6 +36,10 @@ jobs:
     name: "Unit Tests"
     runs-on: ubuntu-latest
 
+    env:
+      TF_APPEND_USER_AGENT: TEST
+      OPENTOFU_USER_AGENT: TEST
```
janosdebugs, line 41: I think this user agent might be a bit too nondescriptive if this is the entire user agent?

SypherSP: the full user agent is OPENTOFU_USER_AGENT+TF_APPEND_USER_AGENT. We can change the APPEND_USER_AGENT env var depending on where the test is being run from. This is just a placeholder for now.

janosdebugs: In that case the user agent should be a bit more verbose, such as `OpenTofu-integration-test`

SypherSP: Done

janosdebugs (approved):
Assuming these environment variables work.

SypherSP:
Could you check by running the workflow on your end. Im trying to figure out the following in order to test it locally as mentioned by @Yantrio in the issue
> To be able to test this you should be able to set up a private registry (or reverse proxy) and check the requests sent to that. Or add some logging in the code to dump the http requests.

SypherSP:
Yeah they're failing because the test cases have not been updated. But the logging is as expected.
Is there a way to get the version of OpenTofu in the yaml file for the workflow? because the tests compare the env var with the version appended to "OpenTofu".

RLRabinowitz:
> Is there a way to get the version of OpenTofu in the yaml file for the workflow? because the tests compare the env var with the version appended to "OpenTofu".

Sorry I'm not entirely sure what you mean by that. The workflow simply checks out the code and runs `go test ./...`, it's not using any release of OpenTofu. You can run it locally as well in order to find the source of failures in the unit tests

Yantrio:
I have a feeling we should probably unset those env vars for the user agent tests.

SypherSP:
@Yantrio I dont know how that can be done, unsetting those vars just for tests in internal/httpclient package. What I can do is update the test so that it passes on these exact values.

It would help knowing where else tests are being run on the registry.

.github/workflows/checks.yml
```diff
@@ -36,6 +36,10 @@ jobs:
     name: "Unit Tests"
     runs-on: ubuntu-latest
 
+    env:
+      TF_APPEND_USER_AGENT: INTEGRATION-TEST
+      OPENTOFU_USER_AGENT: OPENTOFU
```
Yantrio, line 41: I think if we're appending, we dont need to override the OPENTOFU_USER_AGENT too

SypherSP: Test cases would still need to be changed because TF_APPEND_USER_AGENT is also being tested here 

Also, does unsetting the environment variable not defeat the purpose of having tests for them?

image

Yantrio (approved):

cam72cam (approved):

PR #1283, opened 2024-02-20
Feat: urldecode function #1234

### LINKED ISSUE #1234: urldecode function
### OpenTofu Version

```shell
OpenTofu v1.6.1
on linux_amd64
+ provider registry.opentofu.org/hashicorp/azuread v2.47.0
+ provider registry.opentofu.org/hashicorp/azurerm v3.87.0
+ provider registry.opentofu.org/hashicorp/local v2.4.1
+ provider registry.opentofu.org/hashicorp/random v3.6.0
```

### Use Cases

There are some scenarios where we need to urldecode strings. My scenario is extracting the token from the uri of `azurerm_automation_webhook`, urldecoding the token and then passing it to a `azurerm_logic_app_action_http`.

### Attempted Solutions

Right now, I have to do something like `replace(replace(replace(local.webhook_token, "%3d", "="), "%2f", "/"), "%2b", "+")` - but would be really nice to have the function.

### Proposal

Add `urldecode()` function.

### References

`urlencode` already exists: https://developer.hashicorp.com/terraform/language/functions/urlencode

pooriaghaedi:
Resolves #1234

## Target Release

1.7.0

### DIFF
--- a/internal/lang/funcs/encoding.go
+++ b/internal/lang/funcs/encoding.go
@@ -227,6 +227,26 @@ var URLEncodeFunc = function.New(&function.Spec{
 	},
 })
 
+// URLDecodeFunc constructs a function that applies URL decoding to a given string.
+var URLDecodeFunc = function.New(&function.Spec{
+	Params: []function.Parameter{
+		{
+			Name: "str",
+			Type: cty.String,
+		},
+	},
+	Type:         function.StaticReturnType(cty.String),
+	RefineResult: refineNotNull,
+	Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+		query, err := url.QueryUnescape(args[0].AsString())
+		if err != nil {
+			return cty.UnknownVal(cty.String), fmt.Errorf("failed to decode URL '%s': %v", query, err)
+		}
+
+		return cty.StringVal(query), nil
+	},
+})
+
 // Base64Decode decodes a string containing a base64 sequence.
 //
 // OpenTofu uses the "standard" Base64 alphabet as defined in RFC 4648 section 4.
@@ -279,6 +299,16 @@ func Base64Gunzip(str cty.Value) (cty.Value, error) {
 // UTF-8 and then percent encoding is applied separately to each UTF-8 byte.
 func URLEncode(str cty.Value) (cty.Value, error) {
 	return URLEncodeFunc.Call([]cty.Value{str})
+}
+
+// URLDecode decodes a URL-encoded string.
+//
+// This function decodes the given string that has been encoded.
+//
+// If the given string contains non-ASCII characters, these are first encoded as
+// UTF-8 and then percent decoding is applied separately to each UTF-8 byte.
+func URLDecode(str cty.Value) (cty.Value, error) {
+	return URLDecodeFunc.Call([]cty.Value{str})
 }
 
 // TextEncodeBase64 applies Base64 encoding to a string that was encoded before with a target encoding.
--- a/internal/lang/funcs/encoding.go
+++ b/internal/lang/funcs/encoding.go
@@ -227,7 +227,7 @@ var URLEncodeFunc = function.New(&function.Spec{
 	},
 })
 
-// URLDecodeFunc constructs a function that applies URL decoding to a given string.
+// URLDecodeFunc constructs a function that applies URL decoding to a given encoded string.
 var URLDecodeFunc = function.New(&function.Spec{
 	Params: []function.Parameter{
 		{
@@ -301,7 +301,7 @@ func URLEncode(str cty.Value) (cty.Value, error) {
 	return URLEncodeFunc.Call([]cty.Value{str})
 }
 
-// URLDecode decodes a URL-encoded string.
+// URLDecode decodes a URL encoded string.
 //
 // This function decodes the given string that has been encoded.
 //
--- a/internal/lang/funcs/encoding_test.go
+++ b/internal/lang/funcs/encoding_test.go
@@ -243,6 +243,157 @@ func TestURLEncode(t *testing.T) {
 	}
 }
 
+func TestURLDecode(t *testing.T) {
+	tests := []struct {
+		String cty.Value
+		Want   cty.Value
+		Err    bool
+	}{
+		{
+			cty.StringVal("abc123-_"),
+			cty.StringVal("abc123-_"),
+			false,
+		},
+		{
+			cty.StringVal("foo%3Abar%40localhost%3Ffoo%3Dbar%26bar%3Dbaz"),
+			cty.StringVal("foo:bar@localhost?foo=bar&bar=baz"),
+			false,
+		},
+		{
+			cty.StringVal("mailto%3Aemail%3Fsubject%3Dthis%2Bis%2Bmy%2Bsubject"),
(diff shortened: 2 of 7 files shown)


internal/lang/funcs/encoding.go
```diff
@@ -227,6 +227,47 @@ var URLEncodeFunc = function.New(&function.Spec{
...
+					reconstructedURL += "?"
+					first = false
+				} else {
+					reconstructedURL += "&"
+				}
+				reconstructedURL += key + "=" + values[0]
+			}
+		}
+
+		return cty.StringVal(reconstructedURL), nil
```
Yantrio, line 267: Is there a reason you're not using url.QueryUnescape here? https://pkg.go.dev/net/url#QueryUnescape

Doing this manually feels a little hacky. I expected this to be much smaller, similar to the UrlEncode function.

internal/lang/funcs/encoding_test.go
```diff
@@ -243,6 +243,54 @@ func TestURLEncode(t *testing.T) {
...
+			cty.StringVal("mailto%3Aemail%3Fsubject%3Dthis%2Bis%2Bmy%2Bsubject"),
+			cty.StringVal("mailto:email?subject=this+is+my+subject"),
+			false,
+		},
+		{
+			cty.StringVal("foo/bar"),
+			cty.StringVal("foo%2Fbar"),
+			false,
+		},
+	}
```
Yantrio, line 272: I think we have to be careful here when it comes to edge cases. We cant just take the tests from urlencode and assume that the inverse covers everything.

Would you also mind adding tests for the following

- Incomplete percent encodings (ie: just `%` or `%2` and not `%20`
- Invalid percent encodings, (ie: `%GG`)
- Unicode character support (ensure we only support utf8?)
- Null bytes ie %00 should not be allowed


It may also be cool to write some tests that test the full circle of UrlEncode + UrlDecode

Yantrio (changes requested):

Yantrio:
I also noticed that the DCO has not been signed for this set of commits. Please see [this page ](https://github.com/opentofu/opentofu/pull/1283/checks?check_run_id=21760292704)for more information. Or click `Details` next to the failing check.

internal/lang/funcs/encoding_test.go
pooriaghaedi: Done, and I appreciate the suggestion!

internal/lang/funcs/encoding.go
```diff
@@ -227,6 +228,33 @@ var URLEncodeFunc = function.New(&function.Spec{
...
+		rawQuery := args[0].AsString()
+
+		// Check for null bytes
+		if strings.Contains(rawQuery, "%00") {
+			return cty.UnknownVal(cty.String), fmt.Errorf("null bytes are not be allowed")
+		}
+
+		query, err := url.QueryUnescape(rawQuery)
+		if err != nil {
+			return cty.UnknownVal(cty.String), fmt.Errorf("failed to parse URL '%s': %v", query, err)
```
Yantrio, line 251: tiny tiny nitpick:
```suggestion
			return cty.UnknownVal(cty.String), fmt.Errorf("failed to decode URL '%s': %v", query, err)
```

internal/lang/funcs/encoding.go
```diff
@@ -227,6 +228,33 @@ var URLEncodeFunc = function.New(&function.Spec{
...
+		},
+	},
+	Type:         function.StaticReturnType(cty.String),
+	RefineResult: refineNotNull,
+	Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+		rawQuery := args[0].AsString()
+
+		// Check for null bytes
+		if strings.Contains(rawQuery, "%00") {
+			return cty.UnknownVal(cty.String), fmt.Errorf("null bytes are not be allowed")
```
Yantrio, line 246: tiny tiny nitpick:

```suggestion
			return cty.UnknownVal(cty.String), fmt.Errorf("null bytes (%00) are not supported in the input to urldecode")
```

website/docs/language/functions/urldecode.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+`urldecode` targets encoded characters within a string. 
+
+The function is capable of decoding a comprehensive range of characters, 
+including those outside the ASCII range. Non-ASCII characters are first treated as UTF-8 bytes,
+followed by the application of percent decoding to each byte, 
+facilitating the accurate decoding of multibyte characters.
+
+Ideal for sanitizing and interpreting query string parameters, `urldecode` ensures
+that values embedded in URLs are correctly understood when extracted for processing or display.
```
Yantrio, line 16: I think these 2 lines still are talking about encoding, and not decoding.

website/docs/language/functions/urldecode.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+## Implementation Details
+
+The decoding process includes a safety check against null bytes (`%00`),
+which are prohibited due to their potential to introduce security vulnerabilities.
+Upon encountering such sequences, the function halts and returns an error,
+signaling the presence of invalid input.
+
+The core decoding operation leverages the `url.QueryUnescape` method to perform the conversion from encoded to decoded form.
+This method is robust, handling a wide array of encoding scenarios in compliance with the relevant standards for URL encoding.
```
Yantrio, line 26: I'm not sure if we need to mention what it uses under the hood.

internal/lang/funcs/encoding.go
janosdebugs: Is there any specific reason why null bytes should not be supported? Are we 100% certain nobody wants to decode binary data with this?

janosdebugs (approved):

internal/lang/funcs/encoding.go
cube2222: I agree with Janos, Go doesn't use null-terminated strings, I don't think there's any reason to explicitly check for them.

pooriaghaedi: Please let me know the decision so I can delete it.

website/docs/language/functions/urldecode.mdx
```diff
@@ -0,0 +1,32 @@
...
+signaling the presence of invalid input.
+
+## Examples
+
+```
+> urldecode("Hello+World%21")
+Hello World!
+> urldecode("%E2%98%83")
+☃
+> "foo%3Abar%40localhost%3Ffoo%3Dbar%26bar%3Dbaz"
```
Evi1Pumpkin, line 30: I believe that in this example, we forgot to include the `urldecode()` function.

internal/lang/funcs/encoding.go
Evi1Pumpkin: We discussed it internally with @Yantrio and he agreed that we should support null bytes because of the above reasons.
Please remove it.

janosdebugs (dismissed):

cam72cam (dismissed):

Yantrio (approved):

Evi1Pumpkin (approved):

cam72cam (approved):

Yantrio (approved):

Yantrio:
Thanks for this contribution @pooriaghaedi 🎉

PR #1288, opened 2024-02-20
Integrate Encryption into State Backends

cam72cam:
This PR updates the statefile interface to require an `encryption.StateEncryption` for read/write operations.  It then propagates through the backends which are now constructed with an `encryption.StateEncryption` for all StateMgr operations.

`encryption.StateEncryptionDisabled()` was added as an easy path to opt out of state encryption, `tofu show` for example.

`encryption.StateEncryptionTODO()` was added as a alias for Disabled and to mark locations where the config layer needs to interface with the new encryption code paths.  It will be removed before #1030 is closed.

~TODO: Consider how errors returned from state encryption are actually diags and could be added to tfdiags in different code paths.~ #1294

~TODO: Think about how the cloud backend should work.  Encryption is disabled in this pr, but team consensus is that it should be allowed.~ [054fc5b](https://github.com/opentofu/opentofu/pull/1288/commits/054fc5bf3ac5ba4c3ce41344067d6bd92660e41c)

Part of #1030 

## Target Release

1.7.0


internal/backend/local/backend.go
```diff
@@ -89,20 +90,28 @@ type Local struct {
...
+		encryption: enc,
+	}
 }
 
 // NewWithBackend returns a new local backend initialized with a
 // dedicated backend for non-enhanced behavior.
-func NewWithBackend(backend backend.Backend) *Local {
+func NewWithBackend(backend backend.Backend, enc encryption.StateEncryption) *Local {
+	if backend == nil && enc == nil {
+		panic("either backend or encryption required for backend.Local initialization")
```
Yantrio, line 110: tiny tiny nitpick: I think this error would be more comprehensible if it was talking about what has happened, rather than what is required.

I would recommend we do something like:
```suggestion
		panic("backend.Local NewWithBackend called with nil backend and encryption, atleast one is required.")
```

internal/encryption/state.go.orig
```diff
@@ -0,0 +1,39 @@
+package encryption
```
Yantrio, line 1: i think we can remove this .orig file

Yantrio (dismissed):
Overall it LGTM, I have some reservations about merging this until I've tested it manually though.


But I really do like this approach to injecting the encryption into each of the backends.

internal/command/show.go
```diff
@@ -338,7 +339,7 @@ func getStateFromPath(path string) (*statefile.File, error) {
 	defer file.Close()
 
 	var stateFile *statefile.File
-	stateFile, err = statefile.Read(file)
+	stateFile, err = statefile.Read(file, encryption.StateEncryptionTODO()) // Should we use encryption -> statefile config here?
```
cube2222, line 342: Yes, I believe so.

internal/command/state_push.go
```diff
@@ -68,7 +69,7 @@ func (c *StatePushCommand) Run(args []string) int {
 	}
 
 	// Read the state
-	srcStateFile, err := statefile.Read(r)
+	srcStateFile, err := statefile.Read(r, encryption.StateEncryptionTODO()) // Should we use encryption -> statefile config here?
```
cube2222, line 72: Yep, definitely.

internal/configs/static.go
```diff
@@ -0,0 +1,338 @@
...
+	if len(ref.Type) != 0 {
+		val = ref.Type + "." + val
+	}
+	if len(ref.Module) != 0 {
+		val = ref.Module + "." + val
+	}
+	return val
+}
+
+type StaticReference struct {
```
cube2222, line 35: Please add comments (ideally with short examples embedded) of what each of these structs are.

internal/configs/static.go
```diff
@@ -0,0 +1,338 @@
...
+	Identifier StaticIdentifier
+	Value      func(stack []StaticIdentifier) (cty.Value, hcl.Diagnostics)
```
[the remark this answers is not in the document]
cam72cam: Removed

internal/configs/static.go
cam72cam: Removed

cube2222 (approved):
LGTM!

Yantrio (approved):

cube2222 (approved):

PR #1291, opened 2024-02-21
Fixes #1169: AES-GCM implementation

### LINKED ISSUE #1169: [State Encryption] Methods (full)
**Parent:** #1030
**Depends on:** #1166
**Blueprint PR:** #1132

## Acceptance criteria

1. Implement full encryption for both state files and for plan files.
2. The encrypted format should use the `EncryptionTopLevelJsonKey` const key to identify the produced JSON as an encrypted payload.

janosdebugs:
Resolves #1169

This PR adds the AES-GCM encryption method for state encryption (see the dedicated [README](https://github.com/opentofu/opentofu/blob/features/1169-aes-gcm/internal/encryption/method/aesgcm/README.md))

## TODO:

- [x] Key rotation with AES-GCM is mandatory. Decide how to best explain this to users.
- [x] The methods in `internal/collections` may be useful to others, decide if we keep them there or move them into the `aesgcm` package.
- [x] Add tests and documentation for `internal/collections`.
- [x] The README is incorrect, you can't pass strings as byte slices in HCL.
- [x] User-facing documentation.

## Target Release

1.7.0


Yantrio (changes requested):
Sorry for pulling this apart so much @janosdebugs !

This is great work though, especially handling the panics from the GCM implementation. Everything I'm mentioning here are small items

internal/errorhandling/safe.go
```diff
@@ -0,0 +1,36 @@
...
+	}()
+	return f()
+}
+
+// Safe2 runs the specified function and returns its result value or returned error. If a panic occurs, it returns the
+// panic as an error. Any errors returned are passed through the passed wrap function.
+//
+// Note: this is equivalent to a try-catch and you should probably not use it. Only use if you need to handle
+// panics from third party libraries or from Golang itself.
+func Safe2[TValue any](f func() (TValue, error), wrap func(err error) error) (result TValue, err error) {
```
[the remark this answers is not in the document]
janosdebugs: My main issue is that Go doesn't have method / function overrides, which would be really helpful here. Some functions may not have a return value, which would then need a function called `Safe1`. On the flip side, some functions may return two values, or some function may not return an error but only panic. The naming should be such that it's extensible to that end. (I have removed these functions at the request of the review in the engineering meeting because we don't need them now.)

internal/collections/set.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Set is a container that can hold each item only once and has a fast lookup time.
+//
+// You can define a new set like this:
+//
+//	var validKeyLengths = golang.Set[int]{
+//	    16: {},
+//	    24: {},
+//	    32: {},
+//	}
```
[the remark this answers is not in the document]
janosdebugs: I would be tempted too, but performance-wise using the `NewSet` is less than optimal since it needs to iterate over all items, whereas directly setting the values doesn't have any performance cost.

internal/errorhandling/safe.go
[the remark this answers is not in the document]
janosdebugs: Regarding the callsites: we can, but it makes the code a bit harder to read.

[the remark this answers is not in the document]
cam72cam: Is there a reason to make this it's own function at this point?  Should we wait until there are multiple instances to refactor it out?

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,49 @@
...
+|--------------------|---------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `key` (*required*) |         | Your encryption key in binary format, likely referencing a key provider. Must be 16, 24, or 32 bytes long.                                                                                       |
+| `aad`              |         | Additional Authenticated Data. This data is stored along the encrypted form and authenticated. The AAD value of the encrypted form must match the configuration, otherwise the decryption fails. |
+| `nonce_size`       | 12      | Size of the Initialization Vector for AES-GCM. This field should not be changed and is provided for future compatibility.                                                                        |
+| `tag_size`         | 16      | Size of the authentication tag (12-16 bytes). This field should not be changed and is provided for future compatibility.                                                                         |
```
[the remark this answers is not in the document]
Yantrio: That makes sense 👍

internal/errorhandling/safe.go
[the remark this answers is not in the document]
janosdebugs: I tried implementing it inline and it became very disgusting. I could make this a package-internal function, but that feels wrong to me because it has no business being in the `aesgcm` package. We should also encourage code reuse rather than discourage it.

internal/encryption/method/aesgcm/aesgcm.go
```diff
@@ -9,53 +9,125 @@ import (
...
+				return nil, &method.ErrEncryptionFailed{Cause: err}
+			}
 
-	nonce := make([]byte, gcm.NonceSize())
-	if _, err = rand.Read(nonce); err != nil {
-		return nil, err
-	}
+			nonce := make([]byte, gcm.NonceSize())
+			if _, err := rand.Read(nonce); err != nil {
+				return nil, &method.ErrEncryptionFailed{Cause: &method.ErrCryptoFailure{Message: "could not generate nonce", Cause: err}}
```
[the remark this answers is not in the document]
janosdebugs: Not sure it's worth it for 3 lines of code that will never be reused and the compiler will inline anyway.

serdardalgic (commented):
Just a few bits and pieces, ideas and 2 cents to the ongoing discussions..
Other than that, I'm not so much familiar with the AES-GCM implementation. I just took a quick look and overall, the implementation looks quite exciting. Elegant work @janosdebugs 👏 Kudos 👍 

The tests are self explanatory and AFAICS, they cover all the cases I can think of, but again, I took a quick look and didn't dive deeper.

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Key rotation
+
+AES-GCM keys have a limited lifetime of `2^32` blocks, equaling roughly 64 GB of data that can be encrypted before the keys should be considered compromised. Users should rotate keys well before this limit is reached.
+
+## Encryption vs. Authentication
+
+The AES-GCM implementation protects data at rest from being accessed. It does not, however, protect against malicious actors reusing old data (replay attacks) to compromise the integrity of the system. Users with the need for payload authentication should rotate their key and/or AAD frequently to ensure that old data cannot be used in this manner.
```
cipherboy, line 38: Perhaps too late, but if there's no requirements for FIPS compliance, there's two alternatives to pure AES-GCM I'd suggest considering:

 - If you really want AES, choose AES-GCM-SIV, which prevents the nonce reuse problem entirely. You still have a limited number of encryptions and still have to rotate keys, but it is a much safer construction overall. Sadly NIST hasn't standardized this mode for whatever reason. This hasn't [landed in the stdlib yet](https://github.com/golang/go/issues/54364) so you're stuck using a third-party library.
 - If you're fine swapping out AES, go for XChaCha20-Poly1305. The "X" for extended allows longer nonces, so you don't have the key rotation problem at all. NIST hasn't seriously considered standardizing ChaCha20 at all to my knowledge. This is available in `x/crypto`. 

Ultimately, you still have to deal with key rotation in the event of compromise, it is just you're no longer under a risky scenario (having to count encryptions and rotate keys well ahead of hitting the limit or otherwise using time as a proxy metric).

internal/errorhandling/safe.go
```diff
@@ -0,0 +1,36 @@
+package errorhandling
+
+import "fmt"
+
+// safe2 runs the specified function and returns its result value or returned error. If a panic occurs, it returns the
```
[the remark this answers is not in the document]
cipherboy: I don't quite understand the value in this either --- all the error conditions that AES-GCM panics on, could be handled by us, the caller, and thus assert they won't panic.

E.g,. https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/crypto/cipher/gcm.go;l=205 can be caught [by checking against `AEAD.NonceSize()`](https://pkg.go.dev/crypto/cipher#AEAD) and the inexact buffer overlaps can be proven not to occur if we're decrypting to a different buffer (guaranteed by passing `nil` as `dst` like we're doing). Message length checks should also probably be done here, as I don't think we want to be encrypting 4GB of data, generally speaking, and probably want a much lower threshold.

Aside: I suspect any panics observed were during development and shouldn't apply to productions (and even then, should be caught by explicit safeguards in the code here and not passing unchecked values straight to the stdlib).

My 2c., but I think this can be reworked to avoid the need for this entirely, in this code.

internal/encryption/method/aesgcm/aesgcm.go
```diff
@@ -9,53 +9,129 @@ import (
...
+
+	"github.com/opentofu/opentofu/internal/errorhandling"
+
+	"github.com/opentofu/opentofu/internal/encryption/method"
+	// This unsafe is required for go:linkname
+	_ "unsafe"
 )
 
+// Note: the linking below is a workaround for Go issue #42470. We want to support setting both the nonce and the tag
+// size specifically because if the defaults for any of these parameters change due to new cryptography research,
```
cipherboy, line 22: I don't think this is likely to happen, fwiw. 

Tag size is limited at most at 16 bytes, which is the default IIRC. There's no reason to shorten the tag for this use case.

Nonce size also shouldn't change as I discuss elsewhere. I'd simplify this to the default `NewGCM` call. 

If you do anything, I'd consider https://github.com/opentofu/opentofu/pull/1291/files#r1499974976 instead.

internal/encryption/method/aesgcm/README.md
Yantrio: I think FIPS is something that's always at the back of our mind, but it is worth noting that with the way that state encryption is set up, we can have multiple different methods available for people to use.

When I was discussing this with Janos earlier, I think we discussed the fac that we want to explain some of these shortcomings with AES-GCM and help make the end users make informed choices based on their requirements. This may be another method, or more frequent key rotation.

internal/collections/set.go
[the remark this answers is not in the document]
Yantrio: Do we really have to worry about this level of performance of a small Set package here?

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+## Implementation notes
+
+### Additional Authenticated Data (AAD)
+
+
```
Yantrio, line 45:
```suggestion


```

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,49 @@
...
+| `key` (*required*) |         | Your encryption key in binary format, likely referencing a key provider. Must be 16, 24, or 32 bytes long.                                                                                       |
+| `aad`              |         | Additional Authenticated Data. This data is stored along the encrypted form and authenticated. The AAD value of the encrypted form must match the configuration, otherwise the decryption fails. |
+| `nonce_size`       | 12      | Size of the Initialization Vector for AES-GCM. This field should not be changed and is provided for future compatibility.                                                                        |
+| `tag_size`         | 16      | Size of the authentication tag (12-16 bytes). This field should not be changed and is provided for future compatibility.                                                                         |
+
+## Key rotation
+
+AES-GCM keys have a limited lifetime of `2^32` blocks, equaling roughly 64 GB of data that can be encrypted before the keys should be considered compromised. Users should rotate keys well before this limit is reached.
```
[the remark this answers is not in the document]
cipherboy: Go's default, and NIST's recommendation, is that safe balance. There is no better value than 12 bytes given a 32-bit counter.

I think [NIST SP800-38D](https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-38d.pdf) Section 8.3 Constraints on the Number of Invocations is the guidance you're looking for.

Practically though, you don't get any benefit in changing your nonce size with Go; either way you're still limited to an (implicit) 2^32 counter size (which 16-byte nonces don't really help: now you can have two nonces `A`, `B` with `A != B` such that `A+j == B+K` and you break the cryptosystem, just in a more convoluted way. A least with 12-byte nonces you can ensure that as long as `A` & `B` are distinct, you'll never have `A+j == B+k`).

All this to say, you _do_ have to worry about rotation -- and you're not getting out of it with AES-GCM.

internal/encryption/method/aesgcm/config.go
```diff
@@ -8,18 +8,124 @@ package aesgcm
...
+)
+
+// validKeyLengths holds the valid key lengths supported by this method.
+var validKeyLengths = collections.Set[int]{
+	16: {},
+	24: {},
+	32: {},
+}
+
+// validTagSizes contains the valid tag sizes according to NIST SP 800-38D 5.2.1.2
```
cipherboy, line 30: I would not make this configurable as there is no reason to use anything other than the default. 

12 byte tags give a 2^48 probability of collision, and no benefit in performance or storage relative to a state file. The entire tag is still computed and the remainder is truncated before comparison. This is mostly beneficial to resource constrained network protocols, which this is not. 

My 2c.

internal/encryption/method/aesgcm/config.go
```diff
@@ -8,18 +8,124 @@ package aesgcm
...
+	// 256, respectively.
+	Key []byte `hcl:"key" json:"key" yaml:"key"`
+
+	// AAD is the Additional Authenticated Data that is authenticated, but not encrypted. In the Go implementation, this
+	// data serves as a canary value against replay attacks. The AAD value on decryption must match this setting,
+	// otherwise the decryption will fail. (Note: this is Go-specific and differs from the NIST SP 800-38D description
+	// of the AAD.)
+	AAD []byte `hcl:"aad,optional" json:"aad,omitempty" yaml:"aad,omitempty"`
+
+	// NonceSize describes the length of the nonce. The default (and minimum) value is 12 bytes as per the NIST
```
cipherboy, line 55: If you were to make anything configurable, I would opt for different named schemes: call this one `aes-gcm`  with default 12-bit nonce and 16-bit tag and introduce others later (`xchacha20-poly1305` or `aes-gcm-siv`). This gives you extensibility in the direction that matters, and ensures users don't shoot themselves in the foot.

internal/errorhandling/safe.go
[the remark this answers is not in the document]
janosdebugs: Thank you @cipherboy  and @serdardalgic ! My thinking when implementing this was that the authors of the crypto package seem to think that it's a perfectly fine thing to panic in the face of an invalid input. Panics should, according to a lot of opinions, only be used for unrecoverable errors, which an invalid input clearly isn't. I don't trust the authors of the crypto package to not add panics in the future when they want to handle another case. I am not opposed to preemptively checking the input, but I don't want to run into a situation when in a future Go version, this starts crashing for some reason.

internal/errorhandling/must.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	// Handle normal errors:
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Must2 converts an error into a panic, returning a value if no error happened.
+func Must2[T any](value T, err error) T {
```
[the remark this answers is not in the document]
janosdebugs: Unfortunately, `require.Error` needs a `testing.T`, which is not available in runnable examples. :(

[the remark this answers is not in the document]
cipherboy: Ah, that's too bad.

internal/encryption/method/aesgcm/config.go
cipherboy, line 1: This AES-GCM config looks a lot better @janosdebugs! :tada:

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+You can configure the encryption by specifying the following method block:
+
+```hcl2
+terraform {
+  encryption {
+    method "aes_gcm" "mymethod" {
+      key        = "key here" # Pass your 16, 24, or 32 byte encryption key here.
+      aad        = ""         # Leave empty unless needed
+      nonce_size = 12         # Minimum: 1, do not change unless needed
```
cipherboy, line 18: Given the changes to config, I think this can be removed?

cipherboy (approved):
From a cipher usage perspective, this looks a lot better :-) Thanks for addressing my comments!

internal/encryption/method/aesgcm/README.md
janosdebugs: Yes, I still need to work on the docs, thank you!

internal/encryption/method/aesgcm/README.md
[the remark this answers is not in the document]
janosdebugs: Our current understanding is that if we use key-derivation functions, such as PBKDF2 or Argon2 in conjunction with AES-GCM, this is no longer a problem. Is this correct @cipherboy ?

internal/collections/set.go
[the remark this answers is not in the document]
janosdebugs: @Yantrio would you mind adding this into the branch as a commit?

docs/state_encryption.md
```diff
@@ -36,6 +36,8 @@ In this section we describe the features that are out of scope for state and pla
 
 The primary goal of this feature is to protect state and plan files **at rest**. It is not the goal of this feature to protect other channels secrets may be accessed through, such as the JSON output. As such, it is not a goal of this feature to encrypt any output on the standard output, or file output that is not a state or plan file.
 
+Furthermore, it is not a goal of this feature to *authenticate* the plan file. It does not protect against, among others, replay attacks where a malicious actor replaces a current plan or state file with an old one.
```
cube2222, line 39: Wait, the encryption **does** do authentication (that is, that a message hasn't been tampered with). It doesn't protect against replay attacks though.

janosdebugs: It authenticates that the encryption is authenticated, but it does not authenticate that the plan file is the plan file you want to run. This may be in need of better wording.

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,51 @@
...
+      # Leave the AAD empty unless needed. Pass as a list of bytes if needed:  
+      aad  = [1,2,3,4,...]
+    }
+  }
+}
+```
+
+| Field               | Description                                                                                                                                                                                      |
+|---------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `keys` (*required*) | Encryption and decryption key, as well as key provider metadata.                                                                                                                                 |
```
cube2222, line 28: Why is this plural "keys"? This is very confusing to me. Esp. since it's defined as singular "Encryption and decryption key".

internal/encryption/method/method.go
```diff
@@ -5,8 +5,18 @@
...
+// the .tf code into it. For example:
+//
+//	type MyConfig struct {
+//	    Key string `hcl:"key"`
+//	}
+//
+//	func (m MyConfig) Build() (Method, error) { ... }
 type Config interface {
 	// Build takes the configuration and builds an encryption method.
+	// TODO this may be better changed to return hcl.Diagnostics so warnings can be issued?
```
cube2222, line 19: +1

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+However, encryption does not protect against data loss (your state file getting damaged) and it also does not protect against replay attack (an attacker using an older state or plan file and tricking you into running it). Additionally, OpenTofu does not and cannot protect the sensitive values in the state file from the person running the `tofu` command.
+
+### What precautions do I need to take?
+
+When you enable encryption, consider who needs access to your state file directly. If you have more than a very small number of people with access needs, you may want to consider running your production `plan` and `apply` runs from a continuous integration system to protect both the encryption key and the sensitive values in your state.
+
+You will also need to decide what kind of key you would like to use. You can either opt for a static passphrase which you manually rotate regularly, or you can choose a key management system. If feasible, we recommend using a key management system that can rotate keys automatically to give you the maximum amount of protection.
```
cube2222, line 35: I don't think there's any proper reason to regularly rotate passphrases, is there?

With key derivation we'll get a new encryption key every time anyway.

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+If you decide for a key management system, please make sure to configure regular key rotation. This is especially important if the chosen encryption method can exhaust the key. (This is not a problem with passphrase-based static keys.) You can find more information in the [encryption methods](#encryption-methods) section below.
```
Yantrio, line 37:
```suggestion
If you opt for a key management system, it is imperative to configure automatic key rotation. This is particularly crucial if the encryption algorithm you choose has the potential to reach a point of 'key saturation', where the maximum safe usage limit of the key is approached. You can find more information about this in the [encryption methods](#encryption-methods) section below.
```

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+Finally, before enable encryption, please exercise your disaster recovery plan and make a temporary backup of your unencrypted state file. Once you enable encryption, OpenTofu cannot read your state file without the correct key.
```
cube2222, line 39: before enable => before enabling

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+        <CodeBlock language={"shell"}>{ConfigurationSH}</CodeBlock>
+    </TabItem>
+    <TabItem value="env-ps1" label="Environment (Powershell)">
+        <CodeBlock language={"powershell"}>{ConfigurationPS1}</CodeBlock>
+    </TabItem>
+</Tabs>
+
+:::warning
+
+Once your data is encrypted, do not rename key providers and methods in your configuration! The encrypted data contains metadata related to their specific names. Instead, use a [fallback block](#key-and-method-rollover).
```
Yantrio, line 61:
```suggestion
Once your data is encrypted, you should not rename key providers and methods in your configuration! The encrypted data stored in the backend contains metadata related to their specific names. Instead you should use a [fallback block](#key-and-method-rollover) to handled changes to key providers.

```

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### AES-GCM
+
+The only currently supported encryption method is AES-GCM. You can configure it in the following way:
+
+<CodeBlock language="hcl">{AESGCM}</CodeBlock>
+
+:::warning
+
+When using a key-management system as a key provider, always configure automatic key rotation in conjunction with AES-GCM, otherwise your security may be compromised! (This does not affect passphrase-based local keys.)
```
cube2222, line 93: Frankly, with our official implementations and no plugins, I don't think this will affect **any** key providers.

website/docs/language/state/encryption.mdx
cipherboy: See https://github.com/opentofu/opentofu/pull/1291#discussion_r1512927813 -- there's the same reason to rotate passwords as for phrases, unless additional data is mixed in to the password to make it unique.

Likely there isn't, otherwise you could use a KBKDF in place of a PBKDF and get the same properties, IMHO.

website/docs/language/state/encryption.mdx
Yantrio, line 93: this is a repeat of L37

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+:::
+
+## Key and method rollover
+
+In some cases, you may want to change your encryption configuration. This can include renaming a key provider or method, changing a passphrase for a key provider, or switching key-management systems. OpenTofu supports an automatic rollover of your encryption configuration if you provide your old configuration in a `fallback` block:
+
+<CodeBlock language="hcl">{Fallback}</CodeBlock>
+
+If OpenTofu fails to read your state or plan file with the new method, it automatically tries the fallback method. When OpenTofu saves your state or plan files, it will always use the new method.
```
Yantrio, line 103: Maybe we should embolden read and saves here to highlight the difference in actions being the driving factor behind the differences.

website/docs/language/state/encryption.mdx
cube2222: Since they will be generating a new data encryption key every time.

internal/encryption/method/aesgcm/README.md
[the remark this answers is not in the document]
cipherboy: Now, I will say, some things sometimes called PBKDFs aren't suitable for _key_ derivation and only for password storage, e.g., bcrypt and scrypt incorporate entropy  (randomness) into each password hash and thus generate different "keys"/outputs each time and thus won't work for this. You need a deterministic scheme.

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+## Remote state data sources
+
+For projects using the `terraform_remote_state` data source, you can also configure encryption. This can be the same encryption setup as your main configuration, but you can also define a separate set of keys and methods. The configuration syntax looks as follows:
```
Yantrio, line 107:
```suggestion
You can also configure an encryption setup for projects using the `terraform_remote_state` data source. This can be the same encryption setup as your main configuration, but you can also define a separate set of keys and methods. The configuration syntax looks as follows:

```

Yantrio (changes requested):

internal/encryption/method/aesgcm/README.md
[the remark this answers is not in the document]
cube2222: > Every time a user (whether intentionally or unintentionally, e.g., everyone using password1234) uses the same password for encryption, they get the same AES key as output and thus runs into this problem. They'd then have to rotate passwords, or better, choose a strong enough password that roughly looks like a key and use it only once. However, now you don't know if someone else choose the same password, whereas you could be guaranteed nobody else chose the same 128-to-256-bit AES key. :-)

If I understand correctly, you'd be generating a new salt every time you're generating an encryption key, which means a new encryption key whenever you're encrypting.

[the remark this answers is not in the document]
cube2222: There's some notes about it here: https://crypto.stackexchange.com/questions/84439/is-it-dangerous-to-encrypt-lots-of-small-files-with-the-same-key

But basically, both pbkdf2 and argon2 accept a salt. If we use an 8-byte or 16-byte salt, then I believe we're safe, and a passphrase can be reused without worrying about key reuse.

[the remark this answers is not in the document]
cube2222: (we're planning to store the salt as part of the statefile metadata field, which is stored unencrypted)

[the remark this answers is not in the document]
cipherboy: Ah, ok, thanks!

This still ends up being a trade off: you trade a known 128-to-256-bits of security for a questionable security margin in the password itself. Password management and injection is no easier or harder than key management, and you (usually) cannot take advantage of things like HSM storage.

I don't think its a silver bullet, even with public salt, as any weak passwords can still be cracked relatively easily. It is why things like PKCS#12 aren't widely adopted outside of niche communities/use cases and why unencrypted PEM ultimately prevails. If you're just putting the password along side the file itself (so a service can read it), its not different than having an unencrypted blob, IMO.

---

Why have externally specified keys outside of a PKCS#11 interface? Why not have OpenTofu generate a random key on encryption (when migrating from unencrypted->encrypted; otherwise reuse the same specified key, perhaps with an internal counter) and have the user manage that as a password (a la encrypted backups of things like Signal or Matrix)?

My 2c., but I suppose there's no easy design here, otherwise it would've already been done :-)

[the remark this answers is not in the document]
cube2222: > This still ends up being a trade off: you trade a known 128-to-256-bits of security for a questionable security margin in the password itself. Password management and injection is no easier or harder than key management, and you (usually) cannot take advantage of things like HSM storage.

However, assuming a good 32-byte password with a (randomly generated on each encryption) 16-byte salt, you do at least get protection from reusing IV's for an AES key in GCM, unless I'm missing something.

Of course, a user could still just write a script to generate a new encryption key, store it in a key management system, fetch the previous one, and then pass them as the encryption/decryption keys to opentofu, every time it's ran.

In practice, we do expect most users to utilize key providers like AWS KMS, which mostly handle everything for you with their managed keys.

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### What precautions do I need to take?
+
+When you enable encryption, consider who needs access to your state file directly. If you have more than a very small number of people with access needs, you may want to consider running your production `plan` and `apply` runs from a continuous integration system to protect both the encryption key and the sensitive values in your state.
+
+You will also need to decide what kind of key you would like to use. You can either opt for a static passphrase which you manually rotate regularly, or you can choose a key management system. If feasible, we recommend using a key management system that can rotate keys automatically to give you the maximum amount of protection.
+
+If you decide for a key management system, please make sure to configure regular key rotation. This is especially important if the chosen encryption method can exhaust the key. (This is not a problem with passphrase-based static keys.) You can find more information in the [encryption methods](#encryption-methods) section below.
+
```
StephanHCB, line 38: I'd suggest adding paragraph about backing up keys and the configuration used for encryption. If either of those are lost, the state data may be unrecoverable. Maybe something like:

"Ensure you will be able to recover both the keys (passphrase, kms data, ...) and the relevant configuration (code, environment variable values) used to encrypt your state."

internal/encryption/method/aesgcm/aesgcm.go
```diff
@@ -9,53 +9,114 @@ import (
...
+	"errors"
+
+	"github.com/opentofu/opentofu/internal/encryption/method"
 )
 
+// aesgcm contains the encryption/decryption methods according to AES-GCM (NIST SP 800-38D).
 type aesgcm struct {
-	key []byte
+	encryptionKey []byte
+	decryptionKey []byte
```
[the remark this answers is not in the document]
cube2222: The decryption key is deterministically generated based on the config and metadata, as it's used to decrypt the previously-encrypted statefile. The encryption key will be a key generated in a random way, based on randomized metadata that will be added to the newly-stored and newly-encrypted statefile.

internal/encryption/method/aesgcm/README.md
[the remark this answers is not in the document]
cipherboy: You could get that same protection with a regular KDF (HKDF or an SP 800-108 KBKDF) with published salt though. :-)

The goal of adding a PB KDF shouldn't be to avoid nonce reuse. It should be because it's a workflow you want to encourage (use a password to unlock/share X).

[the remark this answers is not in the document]
cipherboy: Let me know if you want me to build a PKCS#11 compatible SP800-108 KDF library. I've written them a few times and I wouldn't mind trying to upstream it to x/crypto, though I note they've not taken many PRs from me that quickly :-D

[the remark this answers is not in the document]
cube2222: Thanks for that.

That said, I suppose there's no big reason **not** to use pbkdf2 here though, right? Mainly, I see that hkdf, pbkdf2 and argon2 **are** in the golang.org/x/crypto package, while the only two Go implementations of `SP800-108 Feedback mode KDF` I found are 1. [an unmaintained repo with one star](https://pkg.go.dev/github.com/chrisccoulson/go-sp800.108-kdf) 2. [HashiCorp's implementation in the Vault sdk](https://pkg.go.dev/github.com/hashicorp/vault/sdk/helper/kdf) (which, fwiw, is still MPL lincensed).

But I agree with you that we might want to revise the messaging here. I don't think we expect or encourage people to use memorable passwords, and would rather encourage generating a long password and keeping it in something like aws secrets manager, or 1password.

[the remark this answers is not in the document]
cipherboy: That is a counter mode KDF, I was thinking of a pipeline mode KDF as that mirrors HKDF. PKCS#11 and SP800-108 has additional flexibility that the linked code doesn't (also present in OpenBao fwiw) for parameter ordering that let's you fenangle pipeline mode to behave exactly as the second stage of HKDF (Expand).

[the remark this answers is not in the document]
cipherboy: I wonder if you could get away with the opposite argument, tbh, and just call HKDF_Expand and pretend it is SP800-108 Feedback if anyone asks. :D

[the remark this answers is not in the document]
cube2222: Just to clarify, is there any reason I'm not seeing (other than users possibly getting the wrong message about expected passphrase complexity, which we can handle via documentation) not to just go with pbkdf2 here?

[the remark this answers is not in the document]
cipherboy: Nah, if it's the UX you want, go for it. But if you're passing in "real" keys to it, I think NIST would frown on that for FIPS.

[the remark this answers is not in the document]
cube2222: That makes sense, we can just recommend sufficiently complex "passphrases".

Thanks a lot for this discussion @cipherboy, I really appreciate you joining the discussion here!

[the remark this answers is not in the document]
cipherboy: No problem, glad the feedback was useful!

internal/encryption/method/aesgcm/README.md
janosdebugs: The key provider returns an object containing an `encryption_key` and a `decryption_key` field. There is more than one key involved. Alternatively, we can name it `key_provider`, but that's inaccurate on a technical level, the user could (if they were into causing themselves pain) pass a list of individual bytes as keys.

internal/collections/set.go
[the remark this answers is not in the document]
cam72cam: I can take this

cam72cam (dismissed):

internal/encryption/method/aesgcm/example_test.go
```diff
@@ -56,14 +56,13 @@ func Example_config() {
 
 	// Obtain a modifiable, buildable config. Alternatively, you can also use ConfigStruct() method to obtain a
 	// struct you can fill with HCL or JSON tags.
-	config := descriptor.TypedConfig()
+	config := descriptor.ConfigStruct().(*aesgcm.Config)
```
Yantrio, line 59: I discussed this with Janos thismorning and the reason we had the TypedConfig() method here was to avoid down-casting and we should add this `TypedConfig()` method back in.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/dylanmei/winrmtest v0.0.0-20210303004826-fbc9ae56efb6
 	github.com/go-test/deep v1.0.3
+	github.com/go-viper/mapstructure/v2 v2.0.0-alpha.1
```
Yantrio, line 31: We should not bring in alpha here, we should use a stable release.

Yantrio: Correction: this is fine: https://github.com/go-viper/mapstructure/compare/v1.6.0...v2.0.0-alpha.1

2.0.0-alpha is just a rename of the module from 1.6.0 with

internal/encryption/method/aesgcm/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Implementation notes
+
+### Additional Authenticated Data (AAD)
+
+The AAD in AES-GCM is a general-purpose authenticated, but not encrypted field in the encrypted payload. The Go implementation only supports using this field as a canary value, rejecting decryption if the value mismatches. AES-GCM would support using this field as a means to store data. Since Go does not support it, neither do we.
+
+### Future compatibility
+
+The current nonce and tag size recommendations may change in the future. The configuration fields here are allowed to permit users to decrypt their old state files.
+
```
Yantrio, line 48:
```suggestion
```

internal/encryption/method/aesgcm/aesgcm.go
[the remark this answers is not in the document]
Yantrio: We should add a quick comment to the code on the output struct to explain this.

internal/encryption/method/method.go
Yantrio: Right now its not clear about where the boundary is between this package and opentofu code code.

When we pull out to an external package, I would like to strongly define a "boundary" where we say everything on one side is `error` and the other side is `hcl.Diagnostics` or `diags.Diagnostics`.

website/docs/language/state/encryption.mdx
janosdebugs:
```suggestion
### What precautions do I need to take?

When you enable encryption, consider who needs access to your state file directly. If you have more than a very small number of people with access needs, you may want to consider running your production `plan` and `apply` runs from a continuous integration system to protect both the encryption key and the sensitive values in your state.

You will also need to decide what kind of key you would like to use based on your security requirements. You can either opt for a static passphrase or you can choose a key management system. If you opt for a key management system, it is imperative to configure automatic key rotation for some encryption methods. This is particularly crucial if the encryption algorithm you choose has the potential to reach a point of 'key saturation', where the maximum safe usage limit of the key is approached, such as AES-GCM. You can find more information about this in the [encryption methods](#encryption-methods) section below.

Finally, before enable encryption, please exercise your disaster recovery plan and make a temporary backup of your unencrypted state file. You should also have backups of your keys. Once you enable encryption, OpenTofu cannot read your state file without the correct key.
```

website/docs/language/state/encryption.mdx
janosdebugs:
```suggestion
AES-GCM is a secure, industry-standard encryption algorithm, but suffers from "key saturation". In order to configure a secure setup, you should either use a key-derivation key provider (such as PBKDF2) with a long and complex passphrase, or use a key management system that automatically rotates keys regularly. Using short, static keys will degrade your encryption.
```

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,117 @@
...
+:::
+
+## Key and method rollover
+
+In some cases, you may want to change your encryption configuration. This can include renaming a key provider or method, changing a passphrase for a key provider, or switching key-management systems. OpenTofu supports an automatic rollover of your encryption configuration if you provide your old configuration in a `fallback` block:
+
+<CodeBlock language="hcl">{Fallback}</CodeBlock>
+
+If OpenTofu fails to read your state or plan file with the new method, it automatically tries the fallback method. When OpenTofu saves your state or plan files, it will always use the new method.
```
Yantrio, line 104:
```suggestion
If OpenTofu fails to **read** your state or plan file with the new method, it will automatically try the fallback method. When OpenTofu **saves** your state or plan file, it will always use the new method and not the fallback.

```

internal/encryption/method/method.go
Yantrio: Overall, we will leave this TODO in the code, and revisit later.

internal/encryption/keyprovider/output.go
```diff
@@ -0,0 +1,24 @@
+package keyprovider
+
+import "github.com/zclconf/go-cty/cty"
+
+type Output struct {
+	EncryptionKey []byte `hcl:"encryption_key" cty:"encryption_key" json:"encryption_key" yaml:"encryption_key"`
+	DecryptionKey []byte `hcl:"decryption_key" cty:"decryption_key" json:"decryption_key" yaml:"decryption_key"`
+	Metadata      any
+}
```
janosdebugs, line 9:
```suggestion
// Output is the standardized structure a key provider must return when providing a key.
// It contains two keys because some key providers may prefer include random data (e.g. salt)
// in the generated keys. 
// Additionally, the Metadata should contain a struct with `meta` tags that OpenTofu can serialize
// into the encrypted form. OpenTofu will inject the `meta` tagged items back into the
// key provider configuration when an item needs to be decrypted.
type Output struct {
	EncryptionKey []byte `hcl:"encryption_key" cty:"encryption_key" json:"encryption_key" yaml:"encryption_key"`
	DecryptionKey []byte `hcl:"decryption_key" cty:"decryption_key" json:"decryption_key" yaml:"decryption_key"`
	Metadata      any
}
```

website/docs/language/state/encryption.mdx
janosdebugs:
```suggestion
### What precautions do I need to take?

When you enable encryption, consider who needs access to your state file directly. If you have more than a very small number of people with access needs, you may want to consider running your production `plan` and `apply` runs from a continuous integration system to protect both the encryption key and the sensitive values in your state.

You will also need to decide what kind of key you would like to use based on your security requirements. You can either opt for a static passphrase or you can choose a key management system. If you opt for a key management system, it is imperative to configure automatic key rotation for some encryption methods. This is particularly crucial if the encryption algorithm you choose has the potential to reach a point of 'key saturation', where the maximum safe usage limit of the key is approached, such as AES-GCM. You can find more information about this in the [encryption methods](#encryption-methods) section below.

Finally, before enabling encryption, please exercise your disaster recovery plan and make a temporary backup of your unencrypted state file. You should also have backups of your keys. Once you enable encryption, OpenTofu cannot read your state file without the correct key.

```

janosdebugs:
```suggestion
### What precautions do I need to take?

When you enable encryption, consider who needs access to your state file directly. If you have more than a very small number of people with access needs, you may want to consider running your production `plan` and `apply` runs from a continuous integration system to protect both the encryption key and the sensitive values in your state.

You will also need to decide what kind of key you would like to use based on your security requirements. You can either opt for a static passphrase or you can choose a key management system. If you opt for a key management system, it is imperative to configure automatic key rotation for some encryption methods. This is particularly crucial if the encryption algorithm you choose has the potential to reach a point of 'key saturation', where the maximum safe usage limit of the key is approached, such as AES-GCM. You can find more information about this in the [encryption methods](#encryption-methods) section below.

Finally, before enabling encryption, please exercise your disaster recovery plan and make a temporary backup of your unencrypted state file. You should also have backups of your keys. Once you enable encryption, OpenTofu cannot read your state file without the correct key.
```

website/docs/language/state/encryption.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+You will also need to decide what kind of key you would like to use. You can either opt for a static passphrase which you manually rotate regularly, or you can choose a key management system. If feasible, we recommend using a key management system that can rotate keys automatically to give you the maximum amount of protection.
+
+If you opt for a key management system, it is imperative to configure automatic key rotation. This is particularly crucial if the encryption algorithm you choose has the potential to reach a point of 'key saturation', where the maximum safe usage limit of the key is approached. You can find more information about this in the [encryption methods](#encryption-methods) section below.
+
+Finally, before enable encryption, please exercise your disaster recovery plan and make a temporary backup of your unencrypted state file. Once you enable encryption, OpenTofu cannot read your state file without the correct key.
```
janosdebugs, line 39:
```suggestion
### What precautions do I need to take?

When you enable encryption, consider who needs access to your state file directly. If you have more than a very small number of people with access needs, you may want to consider running your production `plan` and `apply` runs from a continuous integration system to protect both the encryption key and the sensitive values in your state.

You will also need to decide what kind of key you would like to use based on your security requirements. You can either opt for a static passphrase or you can choose a key management system. If you opt for a key management system, it is imperative to configure automatic key rotation for some encryption methods. This is particularly crucial if the encryption algorithm you choose has the potential to reach a point of 'key saturation', where the maximum safe usage limit of the key is approached, such as AES-GCM. You can find more information about this in the [encryption methods](#encryption-methods) section below.

Finally, before enabling encryption, please exercise your disaster recovery plan and make a temporary backup of your unencrypted state file. You should also have backups of your keys. Once you enable encryption, OpenTofu cannot read your state file without the correct key.
```

cube2222 (approved):

Yantrio (approved):

Yantrio:
Thank you so much for your effort and time reviewing this @StephanHCB + @cipherboy <3

Without you two we wouldn't have the great implementation we're merging today

PR #1292, opened 2024-02-21
Integrate encryption into plan serialization

cam72cam:
This PR updates the planfile package to support encryption.PlanEncryption.  All call sites currently use PlanEncryptionDisabled or PlanEncryptionTODO depending on if it will be integrated with the config layer loading.

Part of #1030

This is the compliment of #1288

## Target Release

1.7.0


internal/plans/planfile/reader.go
```diff
@@ -50,15 +51,27 @@ func (e *ErrUnusableLocalPlan) Unwrap() error {
...
-	r, err := zip.OpenReader(filename)
+func Open(filename string, enc encryption.PlanEncryption) (*Reader, error) {
+	raw, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+	println("A")
+	println(raw)
+
+	decrypted, diags := enc.DecryptPlan(raw)
```
cam72cam, line 69: This is diags for now as go has wierd edge cases about casting a nil object to an error.

Yantrio (approved):

internal/e2e/e2e.go
```diff
@@ -202,11 +203,10 @@ func (b *binary) StateFromFile(filename string) (*states.State, error) {
 // Plan is a helper for easily reading a plan file from the working directory.
 func (b *binary) Plan(path string) (*plans.Plan, error) {
 	path = b.Path(path)
-	pr, err := planfile.Open(path)
```
cube2222, line 205: Actually, we might think about also running some e2e tests with encryption disabled. But I suppose it's not worth adding to this PR.

cube2222 (dismissed):

Yantrio (dismissed):

Yantrio (approved):

PR #1294, opened 2024-02-21
State Encryption Error Handling / Diagnostics

cam72cam:
A lot of the call sites seen in #1288 and #1292 only deal with errors, especially deep in the state serialization package.  By returning hcl.Diagnostics, we complicate each of those call sites, and the entire call tree thereof.  Although technically hcl.Diagnostics implements the error interface, it must be used carefuly.  In the case of `var err error; err = hcl.Diagnostics(nil)`, `err != nil` is a false statement as the type information does not match.

hcl.Diagnostics is primarily useful when decoding configuration or evaluating expressions.  We want to preserve the ability to return those from the encryption library where it makes sense, but simplify all other call sites as much as possible to the standard error interface.

Other than errors in the metadata, all of the hcl.Diagnostics should be known ahead of time when dealing with building encryption methods.   We can call that when the encryptor is constructed and return those diagnostics early.  This will primarily be called from packages that already are heavily dealing with hcl.Diagnostics and tfdiags.Diagnostics.

Once we know that the configuration/registry pair is valid for the given target, we can be confident that plain errors will suffice when calling the encrypt/decrypt actions.

This is a prerequisite of #1288 and #1292, though the code in those PRs won't change at all (other than 1 cleanup in planfile/reader.go)

Part of #1030 

## Target Release

1.7.0

### DIFF
--- a/internal/encryption/encryption.go
+++ b/internal/encryption/encryption.go
@@ -6,6 +6,7 @@
 package encryption
 
 import (
+	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/encryption/config"
 	"github.com/opentofu/opentofu/internal/encryption/registry"
 )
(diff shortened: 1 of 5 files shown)


Yantrio (approved):

internal/encryption/encryption.go
```diff
@@ -41,34 +42,27 @@ func New(reg registry.Registry, cfg *config.Config) Encryption {
...
-func (e *encryption) RemoteState(name string) ReadOnlyStateEncryption {
+func (e *encryption) RemoteState(name string) (ReadOnlyStateEncryption, hcl.Diagnostics) {
 	for _, remoteTarget := range e.cfg.Remote.Targets {
 		if remoteTarget.Name == name {
-			return &stateEncryption{
-				// TODO the addr here should be generated in one place.
-				base: newBaseEncryption(e, remoteTarget.AsTargetConfig(), false, "remote.remote_state_datasource."+remoteTarget.Name),
-			}
+			// TODO the addr here should be generated in one place.
+			addr := "remote.remote_state_datasource." + remoteTarget.Name
```
cube2222, line 61: This address is weird to me. Isn't it called `datasource.terraform_remote_state`?

Or is this meant to be the address of the encryption method?

I'm a bit confused here.

cube2222 (approved):
One nit, overall LGTM.

internal/encryption/encryption.go
cam72cam: This is the address of the remote encryption configuration for use in errors
`terraform { encryption{ remote { remote_data_source "foo" {} } } } `

We may want to rename that in the future.

Yantrio (approved):

PR #1295, opened 2024-02-21
Integrate encryption config into configs package

cam72cam:
Due to cyclic dependencies, part of the configs package needed to be moved into hcl2shim.  After this change, the root module will contain a single Encryption config.  That will need to be merged with the environment in the command package and used to build an encryption.Encryption with a configured registry.

Part of #1030 

## Target Release

1.7.0

### DIFF
--- a/internal/configs/hcl2shim/merge_body.go
+++ b/internal/configs/hcl2shim/merge_body.go
@@ -3,7 +3,7 @@
 // Copyright (c) 2023 HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
-package configs
+package hcl2shim
 
 import (
 	"github.com/hashicorp/hcl/v2"
@@ -46,7 +46,7 @@ var _ hcl.Body = mergeBody{}
 func (b mergeBody) Content(schema *hcl.BodySchema) (*hcl.BodyContent, hcl.Diagnostics) {
 	var diags hcl.Diagnostics
 	baseSchema := schemaWithDynamic(schema)
-	overrideSchema := schemaWithDynamic(schemaForOverrides(schema))
+	overrideSchema := schemaWithDynamic(SchemaForOverrides(schema))
 
 	baseContent, _, cDiags := b.Base.PartialContent(baseSchema)
 	diags = append(diags, cDiags...)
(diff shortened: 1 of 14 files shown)


internal/configs/shim.go
```diff
@@ -0,0 +1,24 @@
...
+	return hcl2shim.ExprIsNativeQuotedString(expr)
+}
+
+func schemaForOverrides(schema *hcl.BodySchema) *hcl.BodySchema {
+	return hcl2shim.SchemaForOverrides(schema)
+}
+
+func SynthBody(filename string, values map[string]cty.Value) hcl.Body {
+	return hcl2shim.SynthBody(filename, values)
+}
```
Yantrio, line 24: should we add a TODO: to try and get rid of this shim?

Yantrio (changes requested):
one small nitpick

internal/configs/shim.go
cam72cam: Done

Yantrio (dismissed):

janosdebugs:
Note: this change will actively prevent the encryption package from ever being its own library and will always be tied to the OpenTofu codebase, pulling in its massive amounts of dependencies. (Also, go modules are terrible.)

I personally really don't like having a centralized `config` package, especially since we don't do centralized config parsing, so there is really no reason to have one, but I will also not object if this is merged because the code base is what it is.

cam72cam:
The code that @Yantrio already introduced for merge called into the configs package.  This is actually a step in the direction of making it more modular.  hcl2shim could easily be made into it's own library that's used by both the standalone encryption library and OpenTofu

internal/encryption/config/config_merge.go
```diff
@@ -7,12 +7,18 @@ package config
 
 import (
 	"github.com/hashicorp/hcl/v2"
-	"github.com/opentofu/opentofu/internal/configs"
+	"github.com/opentofu/opentofu/internal/configs/hcl2shim"
```
janosdebugs, line 10: I'm a bit unhappy about having hcl2shim referenced in this way. I would prefer the encryption library to be standalone OR if not, then we should name hcl2shim something more... descriptive and open it up with this library.

internal/configs/hcl2shim/merge_body.go
```diff
@@ -3,7 +3,7 @@
 // Copyright (c) 2023 HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
-package configs
+package hcl2shim
```
janosdebugs, line 6: This package desperately needs documentation. There should be a README.md in it as we will likely open it up the same time we move the encryption package into the public. Exported functions also all need documentation please.

janosdebugs (dismissed):
The PR is good, but the hcl2shim package needs work.

cube2222 (approved):

internal/configs/hcl2shim/merge_body.go
cam72cam: All of the exported structures/functions have documentation.

internal/encryption/config/config_merge.go
Yantrio: We discussed this in one of our syncs today. We'll be merging this PR and I will visit this renaming and documentation as a follow up PR 👍

Yantrio (approved):

PR #1304, opened 2024-02-24
Only run check/tests workflow when relevant file paths are modified

### LINKED ISSUE #1303: Only run check/tests workflow when relevant file paths are modified
### Current behaviour

For any given PR on this repo, a whole host of tests are executed by the checks.yml workflow, including: unit, race, end-to-end.

Given the scope of the repo includes the website as well, it seems a bit wasteful of the GitHub Action runner minutes _and_ adds friction to the PR review process to require these tests for website/docs-only changes.

### Expected behaviour

Perhaps we can add `paths` or `paths-ignore` ([see docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths)) so that the end-to-end/unit test workflows are only required on PRs which actually touch code, instead of everything including website/docs-only changes?

RDhar:
Currently, for any given PR on this repo, a whole host of tests are required by the checks.yml workflow.

When it comes to website/docs-only changes, this workflow requirement is particularly wasteful of GitHub Action runner minutes _and_ adds unnecessary friction to the PR review process.

[Per docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths), we can `paths` and `paths-ignore` to filter when the workflow is run depending on the file changes introduced by the PR.

While I've drafted a "broad stroke" approach of either include or excluding the `'website/**'` directory and its content, it's perfectly open/adaptable to both positive and negative matching of both files paths and (nested) directories.

Resolves #1303

## Target Release

1.6.1

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -17,16 +17,19 @@ name: Quick Checks
 
 on:
   pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'website/**'
   push:
     branches:
       - main
       - 'v[0-9]+.[0-9]+'
       - checks-workflow-dev/*
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+*'
     paths-ignore:
       - 'docs/**'
       - 'website/**'
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
 
 # This workflow runs for not-yet-reviewed external contributions and so it
 # intentionally has no write access and only limited read access to the
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -24,6 +24,8 @@ on:
       - checks-workflow-dev/*
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+*'
+    paths-ignore:
+      - 'website/**'
 
 # This workflow runs for not-yet-reviewed external contributions and so it
 # intentionally has no write access and only limited read access to the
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -24,6 +24,8 @@ on:
       - checks-workflow-dev/*
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+*'
+    paths-ignore:
+      - 'website/**'
```
Yantrio, line 28: I think we should also ignore the `docs` folder

rdhar: Agreed, will append it to the list.

cam72cam (approved):

Yantrio:
@opentofu/core-engineers if we do bring this in, we may have to re-think how we do required checks.

cube2222:
Hey! It seems like this wouldn't play well with branch protection (that is, if required checks get skipped this way, the PR's will be unmergable). See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

What I believe we'd need is that the workflow starts, but one of its first steps checks the diff, or something like that.

Anyway, I'm not sure if this is worth spending too much time on?

RDhar:
Ah, looking at the docs, I reckon you're right, @cube2222. If either the `paths-ignore` or `paths` filter is not met, then the workflow will remain at "Pending" instead of "Skipped", which is a bit of a bummer for required  checks.

Given this PR was to minimise redundant GitHub Actions runner usage--and if that's not really all that much worth worrying--then I can go ahead and close this PR along with the associated #1303. Thanks for responding!

cube2222:
Sure thing, thanks for raising it!

PR #1305, opened 2024-02-24
feat: Add support for `tofu.workspace` which will be resolved in the same way as `terraform.workspace`

### LINKED ISSUE #1156: terraform.workspace should be tofu.workspace
### OpenTofu Version

```shell
OpenTofu v1.6.0
on darwin_arm64
+ provider registry.opentofu.org/hashicorp/aws v5.33.0
```

```hcl
locals {
  workspace = local.workspaces[tofu.workspace]
  workspaces = {
    prod = {
       ...
    }
  }
}
```

Versus

```hcl
locals {
  workspace = local.workspaces[terraform.workspace]
  workspaces = {
    prod = {
       ...
    }
  }
}
```

### Debug Output

Not needed

### Expected Behavior

"tofu.workspace" version works, due to the power of trademark law.

### Actual Behavior

A managed resource "tofu" "workspace" has not been declared in the root module.

It's required to use `terraform.workspace` instead.

### Steps to Reproduce

Have code that uses `terraform.workspace` interpolation thingy.

### Additional Context

tofu fmt should replace any instnaces of terraform.workspace w/ opentofu compatible interpolation sequences.

### References

_No response_

Ericwww:
Resolves #1156 

This PR reuse the `TerraformAttr` struct, and add a `alias` property in it. 

Also, adding the logic for parsing `tofu.workspace` on ParseRef and evalContext function.

Here are the details on how `tofu.workspace` or `terraform.workspace` be evaluated.

## Details

I don't have a lot of background on tf syntax parsing.

But it seems that a graph will be built and walk validate in every tofu operation (like `plan`,`apply`,`validate` etc.)

So I checked the trace log and found `(*BuiltinEvalContext)EvaluateExpr` will be called when tofu want to evaluate the value of `terraform.workspace`.

Therefore, I implemented the parsing logic of `tofu.workspace` by referring to the parsing logic of `terraform.workspace`.

### Unit Test

I noticed some unit test about `terraform.workspace` is named with tofu.

I do some refactor on them. 

Now, the test about `terraform.workspace` will be named `tf-workspace` or `tfWorkspace`. (alias `terraform` to `tf` to avoid some legal risk)
And the test about `tofu.workspace` will be named `tofu-workspace` or `tofuWorkspace`

### Cloud E2E Test

Also I noticed some e2e test at [internal/cloud/e2e](https://github.com/opentofu/opentofu/tree/main/internal/cloud/e2e) is using `terraform.workspace`. But it seems that all test will be skipped. So I did't add some test about `tofu.workspace` here.

**If you have a different opinion about this, please let me know.**

## TODO

* Change the word on document. (Janos will change it)
* <del>Throw some warning diagnostic when user are using `terraform.workspace` **(I'm not sure. Should be discuss)**</del> (not accept, throw warning will make for a bad user experience for people who just migrated.)

## Target Release

1.7.0

### DIFF
--- a/internal/addrs/parse_ref.go
+++ b/internal/addrs/parse_ref.go
@@ -328,10 +328,18 @@ func parseRef(traversal hcl.Traversal) (*Reference, tfdiags.Diagnostics) {
 	case "terraform":
 		name, rng, remain, diags := parseSingleAttrRef(traversal)
 		return &Reference{
-			Subject:     TerraformAttr{Name: name},
+			Subject:     NewTerraformAttr(IdentTerraform, name),
 			SourceRange: tfdiags.SourceRangeFromHCL(rng),
 			Remaining:   remain,
 		}, diags
+
+	case "tofu":
+		name, rng, remain, parsedDiags := parseSingleAttrRef(traversal)
+		return &Reference{
+			Subject:     NewTerraformAttr(IdentTofu, name),
+			SourceRange: tfdiags.SourceRangeFromHCL(rng),
+			Remaining:   remain,
+		}, parsedDiags
 
 	case "var":
 		name, rng, remain, diags := parseSingleAttrRef(traversal)
--- a/internal/tofu/evaluate.go
+++ b/internal/tofu/evaluate.go
@@ -916,7 +916,6 @@ func (d *evaluationStateData) getResourceSchema(addr addrs.Resource, providerAdd
 func (d *evaluationStateData) GetTerraformAttr(addr addrs.TerraformAttr, rng tfdiags.SourceRange) (cty.Value, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 	switch addr.Name {
-
 	case "workspace":
 		workspaceName := d.Evaluator.Meta.Env
 		return cty.StringVal(workspaceName), diags
@@ -927,8 +926,8 @@ func (d *evaluationStateData) GetTerraformAttr(addr addrs.TerraformAttr, rng tfd
 		// removed.
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
-			Summary:  `Invalid "terraform" attribute`,
-			Detail:   `The terraform.env attribute was deprecated in v0.10 and removed in v0.12. The "state environment" concept was renamed to "workspace" in v0.12, and so the workspace name can now be accessed using the terraform.workspace attribute.`,
+			Summary:  fmt.Sprintf("Invalid %q attribute", addr.Alias),
+			Detail:   fmt.Sprintf(`The %s.env attribute was deprecated in v0.10 and removed in v0.12. The "state environment" concept was renamed to "workspace" in v0.12, and so the workspace name can now be accessed using the %s.workspace attribute.`, addr.Alias, addr.Alias),
 			Subject:  rng.ToHCL().Ptr(),
 		})
 		return cty.DynamicVal, diags
@@ -936,8 +935,8 @@ func (d *evaluationStateData) GetTerraformAttr(addr addrs.TerraformAttr, rng tfd
 	default:
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
-			Summary:  `Invalid "terraform" attribute`,
-			Detail:   fmt.Sprintf(`The "terraform" object does not have an attribute named %q. The only supported attribute is terraform.workspace, the name of the currently-selected workspace.`, addr.Name),
+			Summary:  fmt.Sprintf("Invalid %q attribute", addr.Alias),
+			Detail:   fmt.Sprintf(`The %q object does not have an attribute named %q. The only supported attribute is %s.workspace, the name of the currently-selected workspace.`, addr.Alias, addr.Name, addr.Alias),
 			Subject:  rng.ToHCL().Ptr(),
 		})
 		return cty.DynamicVal, diags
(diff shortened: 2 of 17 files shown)


Ericwww:
I noticed some unit test or e2e test about `terraform.workspace` is named with `tofu`.

* [internal/command/testdata/apply-tofu-workspace/main.tf](https://github.com/opentofu/opentofu/blob/74a479e94bcd5f4ee9a3a5ce552e1f98a4395d33/internal/command/testdata/apply-tofu-workspace/main.tf) used by [internal/command/apply_test.go:1737 TestApply_tofuEnv](https://github.com/opentofu/opentofu/blob/main/internal/command/apply_test.go#L1737) and [internal/command/apply_test.go:1774 TestApply_tofuEnvNonDefault](https://github.com/opentofu/opentofu/blob/main/internal/command/apply_test.go#L1774)

And some test reference `terraform.workspace` but the aim of test is other. Shall I change it to `tofu.workspace`?

janosdebugs:
@Ericwww that's a good question, I don't know. These tests seem to be fairly short, so I'd go for duplication, but if you find more and it gets out of hand let's quickly discuss.

internal/tofu/evaluate.go
```diff
@@ -943,6 +943,37 @@ func (d *evaluationStateData) GetTerraformAttr(addr addrs.TerraformAttr, rng tfd
 	}
 }
 
+func (d *evaluationStateData) GetTofuAttr(addr addrs.TofuAttr, rng tfdiags.SourceRange) (cty.Value, tfdiags.Diagnostics) {
```
Yantrio, line 946: I think it may be best to have a common extracted function between this and GetTerraofmrAttr so that if we make any changes to one, the other does not regress.

Yantrio (changes requested):
Thanks for contributing this,

I've had a quick look and it seems okay to me. I've left one comment and I would also like to ask that you add the TofuAttr to the TestUniqueKeyer test here: https://github.com/opentofu/opentofu/blob/cb2e9119aa75eeb8e1fa175e2b7a205a4fef129f/internal/addrs/unique_key_test.go#L18-L63

Thanks!

internal/tofu/evaluate.go
Ericwww: OK. 

extract a common method `getTerraformOrTofuAttr`

In order to ensure the accuracy of the information in diagnostic, I defined an additional ObjectName method on TerraformAttr and TofuAttr.

cam72cam:
I've got a question on the approach we should take here.  Creating a TofuAttr that mirrors TerraformAttr seems straight forward enough, but it might be worth using TerraformAttr in both scenarios, with a constructor param for the variant https://github.com/opentofu/opentofu/pull/1305/files#diff-6cf74ea30982b698346b82c1bd12acb0b051fbe0fe4ad9e10ef14e45ef799a02R25

IMO it would make the changes in this PR much smaller, but may make it easier to miss cases where we need to handle both "tofu" and "terraform".  We could rename TerraformAttr to ProductAttr to make it more clear? 

Thoughts @opentofu/core-engineers ?

janosdebugs (dismissed):

Ericwww:
OK, it seems you prefer to change `TerraformAttr` to `ProductAttr` or other name and with a constructor. So that we can re using the structure? If so, I will commit new code later.

 @cam72cam @Yantrio

RLRabinowitz (commented):
One thing that was probably left out here - Referencing `tofu.workspace` would not be allowed in destroy-time provisioners, but `terraform.workspace` would be allowed

See https://github.com/opentofu/opentofu/blob/5125db58410c9a67bab0bd53ccd3f063b9743b1e/internal/configs/provisioner.go#L154-L198

cam72cam:
That sounds good @Ericwww!  Also, per the community slack discussion we will keep you posted on a decision from the Technical Steering Committee on the `tofu` or `meta` name for this attribute.  Sorry for not validating that before you started work on the code!

cam72cam:
This is looking great @Ericwww !

cam72cam:
@Ericwww I think once https://github.com/opentofu/opentofu/pull/1699/ is merged, we can get back to this PR.  Apologies again for the delay!

cam72cam:
@Ericwww now that we have support for the .tofu extension, would you be willing to rebase this PR?  Apologies again for the delay!

Ericwww:
@cam72cam Sure. I will rebase it this week.

Ericwww:
fix the failure of `Code Consistency Checks`. PTAL again. 😃

cam72cam:
@Ericwww I've duplicated some code since you first worked on this branch: https://github.com/opentofu/opentofu/blob/main/internal/configs/static_scope.go#L208

I should have added a note in the corresponding location, but forgot to.  I should really move a lot of the common attribute handling to a shared location.

internal/addrs/parse_ref.go
```diff
@@ -328,11 +328,19 @@ func parseRef(traversal hcl.Traversal) (*Reference, tfdiags.Diagnostics) {
 	case "terraform":
 		name, rng, remain, diags := parseSingleAttrRef(traversal)
 		return &Reference{
-			Subject:     TerraformAttr{Name: name},
+			Subject:     NewTerraformAttr("terraform", name),
```
cam72cam, line 331: We should make "terraform" and "tofu" constants in the addrs package.  Perhaps:
* `addrs.IdentTerraform`
* `addrs.IdentTofu`
or something similar?

Ericwww: OK. Extracted them as constant.

cam72cam (approved):

Yantrio (approved):

PR #1313, opened 2024-02-27
Fixes: show insensitive plan details when resource in set with sensitive value

### LINKED ISSUE #708: Plan detail for a resource's TypeSet block is hidden due to sensitive values
### OpenTofu Version

```shell
OpenTofu v1.6.0-alpha2
on darwin_arm64
```

```hcl
terraform {
  required_providers {
    fastly = {
      source = "fastly/fastly"
      version = "5.5.0"
    }
  }
}

resource "fastly_service_vcl" "example" {
  name = "opentofu_example_block_sensitive"

  domain {
    name    = "opentofu_example_block_sensitive.integralist.co.uk"
  }

  backend {
    address = "127.0.0.1"
    name    = "localhost"
    port    = 80
  }

  force_destroy = true
}
```

### Debug Output

https://gist.github.com/Integralist/2820e49513ef22f0fd083c5ef316b742

### Expected Behavior

We should see a complete diff for the `backend` block, something like...

```
+ backend {
  + address               = "127.0.0.1"
  + auto_loadbalance      = false
  + between_bytes_timeout = 10000
  + connect_timeout       = 1000
  + error_threshold       = 0
  + first_byte_timeout    = 15000
  + healthcheck           = ""
  + max_conn              = 200
  + name                  = "localhost"
  + override_host         = ""
  + port                  = 80
  + request_condition     = ""
  + ssl_cert_hostname     = ""
  + ssl_check_cert        = true
  + ssl_sni_hostname      = ""
  + use_ssl               = false
  + weight                = 100
}
```

### Actual Behavior

Instead, we see the following...

```
+ backend {
  # At least one attribute in this block is (or was) sensitive,
  # so its contents will not be displayed.
}
```

### Steps to Reproduce

1. `tofu init`
2. `tofu plan`

### Additional Context

This was [originally reported](https://github.com/hashicorp/terraform/issues/33112) in the Terraform repository and is currently still open (the change in behaviour was introduced in the `terraform` CLI version 1.4.0, and OpenTofu forked from 1.5.0).

Now, although I appreciate the rationale of hiding sensitive data from the plan, there are two problems here...

1. We're hiding the _entire_ `backend` block's plan, and this makes it very dangerous for anyone to apply this configuration change (as we can't see what's changing). Yes, it's likely the author of the change is the one running the `tofu plan` but that's not always the case, and even still mistakes happen and it's important to be able to verify what is about to change in your infrastructure. It would be nice if just the _values_ for each sensitive attribute was hidden rather than hiding everything.
2. In this particular case, I'm not actually setting any of the 'sensitive' attributes defined in the `backend` block. So I shouldn't really have the block's entire plan hidden from me as there actually is no sensitive value being set.

### References

- https://github.com/hashicorp/terraform/issues/33112

chenzejun:
### Solution: 
MarkWithPaths(marks) will move marks to the whole set if the mark was applied to value in the set, and it makes the whole set to be hide when render plan details
To avoid this, use SensitiveAsBoolWithPathValueMarks to get more accurate sensitive as bool value and keep more detail.

### test:
use tfcoremock provider with dynamic_resources.json: 
```
{
  "tfcoremock_dynamic_resource": {
    "attributes": {
      "description": {
        "type": "string",
        "required": true
      }
    },
    "blocks": {
      "backend": {
        "mode": "set",
        "attributes": {
          "name": {
            "type": "string",
            "required": true
          },
          "password": {
            "type": "string",
            "required": true,
            "sensitive": true
          }
        }
      }
    }
  }
}
```

and plan on template:
```
terraform {
  required_providers {
    tfcoremock = {
      source = "hashicorp/tfcoremock"
    }
  }
}

resource "tfcoremock_dynamic_resource" "example" {
  id          = "id"
  description = "whatever"
  backend {
    name = "foo"
    password = "bar"
  }
}

```

the plan result:
```
  + create

  # tfcoremock_dynamic_resource.example will be created
  + resource "tfcoremock_dynamic_resource" "example" {
      + description = "whatever"
      + id          = "id"

      + backend {
          + name     = "foo"
          + password = (sensitive value)
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

 

Resolves #708

## Target Release

1.7.0


chenzejun:
Hi @janosdebugs I found some case need to consider, such as value with sensitive data source reference. For now just leave this pr WIP.
When I think my pr is ready, I will ping you again.

chenzejun:
Hi, @janosdebugs I have a new approch to fix the issue, can you PTAL when you are avaliable?

Besides, in case of dynamic sensitive value scenario, I found it hard to fix, so this is not included in my pr yet.
This it my test case and some investigation:
 dynamic_resources.json:
```
{
  "tfcoremock_dynamic_resource": {
    "attributes": {
      "description": {
        "type": "string",
        "required": true
      }
    },
    "blocks": {
      "backend": {
        "mode": "set",
        "attributes": {
          "name": {
            "type": "string",
            "required": true
          },
          "password": {
            "type": "string",
            "required": true,
            "sensitive": true
          },
          "data": {
            "type": "string",
            "optional": true,
            "sensitive": false
          }
        }
      }
    }
  },
  "tfcoremock_secret_resource": {
    "attributes": {
      "data": {
        "type": "string",
        "required": true,
        "sensitive": true
      }
    }
  }
}
```
And I use a template when tfcoremock_dynamic_resource.example.backend.data use datasource tfcoremock_secret_resource.source.data,
```
terraform {
  required_providers {
    tfcoremock = {
      source = "hashicorp/tfcoremock"
    }
  }
}

data "tfcoremock_secret_resource" "source" {
  id = "secretid"
}

resource "tfcoremock_dynamic_resource" "example" {
  id          = "id"
  description = "whatever"
  backend {
    name     = "foo"
    password = "pass"
  }
  backend {
    name     = "bar"
    password = "pass"
    data = data.tfcoremock_secret_resource.source.data
  }
}
```

The planed value will have sensitive mark on tfcoremock_dynamic_resource.backend(top level of the set) when evaluate config block
https://github.com/opentofu/opentofu/blob/b5889b10ebdba23061e8da3b14c61f2a12fdb61d/internal/tofu/node_resource_abstract_instance.go#L744C1-L745C1

because the set decoder also use cty.SetVal to construct a set form list in lib hcl, which means we lost tracking of sensitive mark from the beginning in dynamic sensitive value scenario.

https://github.com/hashicorp/hcl/blob/57f8bbf184a628a6dedb36520db1c90bfab7fb06/hcldec/spec.go#L832C1-L833C1

janosdebugs (commented):
@opentofu/core-engineers I don't know this part of the code well enough, could you please take a look?

cam72cam:
Apologies again for letting this idle, our team has been quite scattered over the past few weeks between KubeCon and PTO.  Thank you for your patience!

cam72cam:
@chenzejun I've done a fairly thorough review of this PR and it looks great!  We will need some significant test coverage of this feature, but the approach is definitely solid.

Looks like internal/command/jsonstate/state_test.go already has some basic cases you could expand upon.

Apologies again for the delay in review!

cam72cam:
Assigning @Yantrio as he is also making some related changes and has some experience in this area.

chenzejun:
> Assigning @Yantrio as he is also making some related changes and has some experience in this area.

Hi @Yantrio , I haven't found a solution to this problem yet https://github.com/opentofu/opentofu/pull/1313#issuecomment-1976088224, do you have any ideas?

chenzejun:
Added some tests I was missing before,  would you like ptal again? @cam72cam @Yantrio

internal/command/jsonstate/state.go
```diff
@@ -518,16 +538,19 @@ func SensitiveAsBool(val cty.Value) cty.Value {
 		vals := make([]cty.Value, 0, length)
 		it := val.ElementIterator()
 		for it.Next() {
-			_, v := it.Element()
-			vals = append(vals, SensitiveAsBool(v))
+			kv, ev := it.Element()
+			path := append(path, cty.IndexStep{
+				Key: kv,
+			})
```
ollevche, line 544: Does `IndexStep` work fine with `Tuple`? From the `IndexStep` doc:

> IndexStep is a Step implementation representing applying the index operation to a value, which must be of either a list, map, or set type.

internal/command/jsonstate/state.go
```diff
@@ -496,14 +496,34 @@ func marshalResources(resources map[string]*states.Resource, module addrs.Module
...
+
+func SensitiveAsBoolWithPathValueMarks(val cty.Value, pvms []cty.PathValueMarks) cty.Value {
+	var path cty.Path
+	sensitiveMarks := make([]cty.PathValueMarks, 0)
+	for _, pvm := range pvms {
+		if _, ok := pvm.Marks[marks.Sensitive]; ok {
+			sensitiveMarks = append(sensitiveMarks, pvm)
+		}
+	}
+	return sensitiveAsBoolWithPathValueMarks(val, path, sensitiveMarks)
```
ollevche, line 510: tiny style suggestion: it may be more obvious to the reader if we pass an empty `cty.Path` here (with no variable)

internal/command/jsonstate/state.go
```diff
@@ -518,16 +538,19 @@ func SensitiveAsBool(val cty.Value) cty.Value {
...
+			vals = append(vals, sensitiveAsBoolWithPathValueMarks(ev, path, pvms))
 		}
 		// The above transform may have changed the types of some of the
 		// elements, so we'll always use a tuple here in case we've now made
 		// different elements have different types. Our ultimate goal is to
 		// marshal to JSON anyway, and all of these sequence types are
 		// indistinguishable in JSON.
 		return cty.TupleVal(vals)
-	case ty.IsMapType() || ty.IsObjectType():
+	case ty.IsMapType():
```
ollevche, line 553: Since it is certainly a map, we can simplify the length check starting on line 559. However, I'm curious if that's necessary to separate maps and objects processing here (see comment below)

internal/command/jsonstate/state.go
```diff
@@ -561,8 +587,27 @@ func SensitiveAsBool(val cty.Value) cty.Value {
 		// marshal to JSON anyway, and all of these mapping types are
 		// indistinguishable in JSON.
 		return cty.ObjectVal(vals)
+	case ty.IsObjectType():
+		if !val.IsKnown() || ty.Equals(cty.EmptyObject) {
```
ollevche, line 591: Why is the obj check moved separately? As far as I understand, `IndexStep` supports objects as well as maps, and also objects could be iterated via `ElementIterator` so it could be the same code for both maps and objects. WDYT?

ollevche (commented):
Thank you for your contribution!

I left some questions/comments to understand the change better, however it looks nice overall

internal/command/jsonstate/state.go
chenzejun: `IndexStep` for map and `GetAttrStep` for object

internal/command/jsonstate/state.go
chenzejun: two difference between obj and map
1. length or empty check
2. different path step type

internal/command/jsonstate/state.go
chenzejun: updated

internal/command/jsonstate/state.go
chenzejun: Yes, see https://github.com/zclconf/go-cty/blob/15a9d85f4808a6d2a34baf813f3806c8e3d334c7/cty/path.go#L216C1-L219C4
and I also add a test for tuple

chenzejun:
hi @ollevche, I update my PR for codecheck, can you help terigle the workflows again?

ollevche:
Hey @chenzejun, I rerun the check, however it still fails. You can also run golint locally for a quicker check. Ping me if you have any questions

chenzejun:
Hi @ollevche I have some trouble to run golint  locally on Windows, so I trigger these checks it in my own fork as a workaround and now all the tests have passed. can you PTAL again?

ollevche:
Hey @chenzejun, thank you for the fixes!

The code looks nice to me, however, I'm not deeply familiar with the sensitivity implementation so it would be safer to have one more review from @Yantrio and/or @janosdebugs.

ollevche (dismissed):

chenzejun:
Hi @janosdebugs @Yantrio, would you mind take a look?

Yantrio:
hi @chenzejun I want to sincerly apologise for taking so long to review this. I've been meaning to comment on here for a while.

I've been busy working on #1640 and spending a lot of time understanding how sensitivity marks are propogated throughout tofu. I'm sure you're aware that we take security really seriously and sensitivity in tofu is a big deal, so I wanted to make sure that I had the full context before I really started reviewing this deeply.

Now I have the big picture I think that this PR should be okay to accept. 

@cam72cam are you happy with the testing here as you requested some? If so I think we can get this in.

Yantrio (approved):
Approving assuming @cam72cam is happy with the tests

cam72cam:
Thanks @Yantrio I'll take a final pass later today :+1:

cam72cam (approved):

cam72cam:
Another read through of the code looked good, and test coverage is complete for all added/modified code.

cam72cam:
Thank you again @chenzejun!

PR #1316, opened 2024-02-27
Initial wiring of encryption through the command package

cam72cam:
This PR fully wires the state encryption package into the command structure.

TODO:
* [x] Make sure Backend/StateFile are used in the correct locations
* [x] Take a closer look at changes made to the provider interface for encrypted remote state access
* [ ] Add a boat load of tests

Related to #1030 

## Target Release

1.7.0

### DIFF
--- a/internal/backend/local/backend.go
+++ b/internal/backend/local/backend.go
@@ -106,9 +106,6 @@ func New(enc encryption.StateEncryption) *Local {
 // NewWithBackend returns a new local backend initialized with a
 // dedicated backend for non-enhanced behavior.
 func NewWithBackend(backend backend.Backend, enc encryption.StateEncryption) *Local {
-	if backend == nil && enc == nil {
-		panic("either backend or encryption required for backend.Local initialization")
-	}
 	return &Local{
 		Backend:    backend,
 		encryption: enc,
--- a/internal/command/meta_encryption.go
+++ b/internal/command/meta_encryption.go
@@ -13,6 +13,8 @@ import (
 	"github.com/opentofu/opentofu/internal/tfdiags"
 )
 
+const encryptionConfigEnvName = "TF_ENCRYPTION"
+
 func (m *Meta) Encryption() (encryption.Encryption, tfdiags.Diagnostics) {
 	path, err := os.Getwd()
 	if err != nil {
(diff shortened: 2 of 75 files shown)


internal/backend/local/backend.go
```diff
@@ -106,9 +106,6 @@ func New(enc encryption.StateEncryption) *Local {
 // NewWithBackend returns a new local backend initialized with a
 // dedicated backend for non-enhanced behavior.
 func NewWithBackend(backend backend.Backend, enc encryption.StateEncryption) *Local {
-	if backend == nil && enc == nil {
```
cam72cam, line 109: This is actually valid for some commands that do not use encryption, like get schema

internal/grpcwrap/provider6.go
```diff
@@ -373,6 +374,7 @@ func (p *provider6) ReadDataSource(_ context.Context, req *tfplugin6.ReadDataSou
 		TypeName:     req.TypeName,
 		Config:       configVal,
 		ProviderMeta: metaVal,
+		Encryption:   encryption.StateEncryptionDisabled(), // Not supported via GRCP
```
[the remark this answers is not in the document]
cam72cam: Thanks for catching that!

internal/command/meta_encryption.go
```diff
@@ -0,0 +1,64 @@
...
+		panic(err)
+	}
+	if err := reg.RegisterMethod(aesgcm.New()); err != nil {
+		panic(err)
+	}
+
+	cfg := module.Encryption
+	var diags tfdiags.Diagnostics
+
+	envName := "TF_ENCRYPTION"
```
Yantrio, line 49: Make this a constant with a comment explaining what it does. Also please add this to the docs here: https://opentofu.org/docs/cli/config/environment-variables

internal/command/state_meta.go
```diff
@@ -28,17 +28,17 @@ type StateMeta struct {
...
-func (c *StateMeta) State() (statemgr.Full, error) {
+func (c *StateMeta) State(enc encryption.Encryption) (statemgr.Full, error) {
 	var realState statemgr.Full
 	backupPath := c.backupPath
 	stateOutPath := c.statePath
 
 	// use the specified state
 	if c.statePath != "" {
-		realState = statemgr.NewFilesystem(c.statePath, encryption.StateEncryptionTODO())
+		realState = statemgr.NewFilesystem(c.statePath, encryption.StateEncryptionDisabled()) // User specified state file should not be encrypted
```
Yantrio, line 38: We should check manually that this works with something like `tofu output`

internal/encryption/base.go
```diff
@@ -100,16 +111,20 @@ func (s *baseEncryption) decrypt(data []byte, validator func([]byte) error) ([]b
...
-		if err != nil {
+		verr := validator(data)
+		if verr != nil {
 			// Nope, just bad input
-			return nil, fmt.Errorf("unable to determine data structure during decryption: %w", err)
+
+			// Return the outer json error if we have one
+			if err != nil {
+				return nil, fmt.Errorf("invalid data format for decryption: %w, %w", err, verr)
+			}
```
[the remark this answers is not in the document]
cam72cam: 002b5d5c7f

PR #1332, opened 2024-02-29
Adding `removed` block docs

Evi1Pumpkin:
Adding the documentation for the `removed` block introduced in https://github.com/opentofu/opentofu/pull/1158

**Possible sections to add explanations about the `removed` block:**
1. [Language -> Resources -> Resource Blocks](https://opentofu.org/docs/language/resources/syntax/) - added here the main explanation. 
2. [Language -> Resources -> Resource Behavior](https://opentofu.org/docs/language/resources/behavior/) - added an explanation about the new `forget` action.
3. [Language -> Modules -> Refactoring](https://opentofu.org/docs/language/modules/develop/refactoring/) - I initially considered putting the main explanation here, as this section mainly discusses the `moved` block. But the `removed` block forgets resources and doesn't specifically belong to modules. So, I decided not to include it here. However, I am still considering putting a reference to 1.
4. [Language -> State](https://opentofu.org/docs/language/state/) - We have the '[Importing Existing Resources](https://opentofu.org/docs/language/state/import/)' page, so I considered adding a similar page about removing existing resources. Ultimately, it would not benefit this location, as people will most likely search for how to remove resources in the 'Resources' section and not in the 'State' section.

**More considerations:**
1. Since we still don't have docs per OpenTofu version, should we mention that this feature is supported from the 1.7.0 version onwards?
2. Should we also mention the `tofu state rm` command here, as it is a similar way to do the same thing?

## Target Release

1.7.0

### DIFF
--- a/website/docs/language/resources/syntax.mdx
+++ b/website/docs/language/resources/syntax.mdx
@@ -119,8 +119,7 @@ configuration, see
 
 ## Removing Resources
 
-By default, deleting a resource block from your configuration will result in a plan to destroy it, effectively removing
-the infrastructure object described by the resource.
+If you remove a resource block from your configuration, OpenTofu will destroy it as a default behavior.
 
 However, there are instances when you want to remove a resource from your configuration without destroying the
 corresponding infrastructure object. In such cases, you can remove it from the [OpenTofu state](/docs/language/state)
@@ -137,6 +136,11 @@ removed {
   from = aws_instance.web
 }
 ```
+
+:::note
+The address in the `from` attribute cannot include instance keys (for example, "aws_instance.web[0]").
+:::
+
 
 Upon executing `tofu plan`, OpenTofu will indicate that the resource is slated for removal from the state but will not
 be destroyed.
--- a/website/docs/language/resources/syntax.mdx
+++ b/website/docs/language/resources/syntax.mdx
@@ -137,7 +137,9 @@ removed {
 ```
 
 :::note
+
 The address in the `from` attribute cannot include instance keys (for example, "aws_instance.web[0]").
+
 :::
 
 Upon executing `tofu plan`, OpenTofu will indicate that the resource is slated for removal from the state but will not
--- a/website/docs/language/resources/syntax.mdx
+++ b/website/docs/language/resources/syntax.mdx
@@ -140,7 +140,6 @@ removed {
 The address in the `from` attribute cannot include instance keys (for example, "aws_instance.web[0]").
 :::
 
-
 Upon executing `tofu plan`, OpenTofu will indicate that the resource is slated for removal from the state but will not
 be destroyed.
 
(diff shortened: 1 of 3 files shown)


website/docs/language/resources/syntax.mdx
```diff
@@ -117,6 +117,37 @@ For more information about how OpenTofu manages resources when applying a
 configuration, see
 [Resource Behavior](/docs/language/resources/behavior).
 
+## Removing Resources
+
+By default, deleting a resource block from your configuration will result in a plan to destroy it, effectively removing
+the infrastructure object described by the resource.
```
janosdebugs, line 123:
```suggestion
If you remove a resource block from your configuration, OpenTofu will destroy it as a default behavior.
```

website/docs/language/resources/syntax.mdx
```diff
@@ -117,6 +117,37 @@ For more information about how OpenTofu manages resources when applying a
...
+
+However, there are instances when you want to remove a resource from your configuration without destroying the
+corresponding infrastructure object. In such cases, you can remove it from the [OpenTofu state](/docs/language/state)
+while allowing it to persist in the remote system.
+
+To achieve this, follow these steps:
+1. Delete the resource from your configuration.
+2. Add a removed block instead, specifying the resource address you want to "forget" in the `from` attribute. Note that
+the address cannot include instance keys (for example, "aws_instance.web[0]").
```
janosdebugs, line 132: The note block could be moved into an [admonition](https://docusaurus.io/docs/markdown-features/admonitions).

janosdebugs (dismissed):
Very good description, just a few minor nitpicks.

janosdebugs:
> Since we still don't have docs per OpenTofu version, should we mention that this feature is supported from the 1.7.0 version onwards?
No, we just shouldn't backport the docs change. When 1.7 comes out, we will need to make versioned docs.

> Should we also mention the tofu state rm command here, as it is a similar way to do the same thing?
It may be good to link together the imperative and declarative versions of these commands, such as state rm and the removed block, import and the import block, etc. but that could be done in a separate PR.

website/docs/language/resources/syntax.mdx
Evi1Pumpkin: ✅

website/docs/language/resources/syntax.mdx
```diff
@@ -138,6 +136,10 @@ removed {
 }
 ```
 
+:::note
+The address in the `from` attribute cannot include instance keys (for example, "aws_instance.web[0]").
+:::
```
janosdebugs, line 141:
```suggestion
:::note

The address in the `from` attribute cannot include instance keys (for example, "aws_instance.web[0]").

:::
```

janosdebugs (approved):
Tiny nitpick Docusaurus recommended writing it this way.

janosdebugs (approved):

cam72cam (approved):

Issue #1339, opened 2024-03-04
Adding AWS sample for Amazon S3 remote backend to documentation 


LumberjackUsingMath:
### OpenTofu Version

```shell
any/latest
```

### Use Cases

The link to the AWS sample would be a great reference in the documentation regarding the usage of an Amazon S3 remote backend for OpenTofu. It provides an example to kickstart a OpenTofu project following best practices

### Attempted Solutions

If someone let's me know where to put it, I can prepare the pull request 

### Proposal

Referencing the https://github.com/aws-samples/bootstrap-amazon-s3-remote-backend-for-open-tofu

Yantrio:
Hi, I'm aware we dont have tutorial-style documentation right now but there is a vast amount online already. For example: https://spacelift.io/blog/terraform-s3-backend

Yantrio:
@LumberjackUsingMath Firstly, thank you for the contribution.

In the future, we do recommend that people follow our [Contributing guide](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md) as we would've loved to have discussed this a bit more before accepting the issue

Firstly I'm not too sure about pitching to users that they should use CloudFormation to manage their infra, when OpenTofu would work well for this. This could increase complexity overall for users new to IaC.

I am also a little concerned by the license in the repo you are pitching: https://github.com/aws-samples/bootstrap-amazon-s3-remote-backend-for-open-tofu/blob/main/Mozilla%20Public%20License%202.0 I'm mainly a bit confused by the hashicorp license being there. Would you be able to explain what's going on here with the licensing of this repo?

I've also left a comment on the pull request but our intention in the long run with the s3 backend with OpenTofu is to support multiple S3-compatible services. This means that where possible we are not looking to tie things too closely to AWS specifically. We want to provide options to the users.

In my personal opinion, we should probably have some sort of guide directly in the OpenTofu documentation that end users can consume rather than an external resource that could disappear or become unmaintained over time.

LumberjackUsingMath:
Hello @Yantrio, 

thanks for the great feedback. I released that repository as I could not find any IaC example which solves the creation of the remote backend in an automated way and following best practices. For instance, https://spacelift.io/blog/terraform-s3-backend gives a great head start but leaves the reader all the work to provision the resource and has the following three weakness. 1/ it creates the backend using a local state 2/ it does not recommend a KMS key 3/ the bucket policy could be more restrictive. 
[...]

PR #1340, opened 2024-03-04
Add AWS sample to bootstrap the remote backend for OpenTofu to documentation

LumberjackUsingMath:
Adding a reference to the AWS sample which provides an example to bootstrap the Amazon S3 backend for OpenTofu following best practices. 

closes  https://github.com/opentofu/opentofu/issues/1339

Resolves #

## Target Release

1.7.0

### DIFF
--- a/website/docs/language/settings/backends/s3.mdx
+++ b/website/docs/language/settings/backends/s3.mdx
@@ -1,9 +1,9 @@
 ---
-sidebar_label: s3
+sidebar_label: Amazon S3
 description: OpenTofu can store state remotely in S3 and lock that state with DynamoDB.
 ---
 
-# Backend Type: s3
+# Backend Type: Amazon S3
 
 Stores the state as a given key in a given bucket on
 [Amazon S3](https://aws.amazon.com/s3/).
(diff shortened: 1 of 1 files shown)


website/docs/language/settings/backends/s3.mdx
```diff
@@ -1,9 +1,9 @@
 ---
-sidebar_label: s3
+sidebar_label: Amazon S3
```
Yantrio, line 2: I would prefer to keep this as just "s3" if possible as we may in the future try and support other s3-compatible backends officially.

Yantrio (changes requested):

website/docs/language/settings/backends/s3.mdx
```diff
@@ -18,6 +18,12 @@ It is highly recommended that you enable
 on the S3 bucket to allow for state recovery in the case of accidental deletions and human error.
 :::
 
+## Bootstrap the backend using Amazon CloudFormation
+
+Amazon provides a sample repository for bootstrapping the backend with Amazon CloudFormation, accessible at the following link: [bootstrap-amazon-s3-remote-backend-for-open-tofu
+](https://github.com/aws-samples/bootstrap-amazon-s3-remote-backend-for-open-tofu).
```
virtualroot, line 24: This repo is gone (404)

janosdebugs:
Hi @LumberjackUsingMath thank you for the PR and the discussion. Personally, I'm not a fan of including vendor-specific links in the official documentation, as you can see we didn't include the Spacelift link in the documentation either, even though it is one of the companies sponsoring the OpenTofu development.

If this were to be included, I would either like to see a comprehensive list of resources for cloud providers such as Amazon, DigitalOcean, IONOS, etc, as well as on-prem solutions such as Minio or Synology C2. OpenTofu is a vendor-neutral tool and the documentation should reflect that.As far as the naming is concerned, S3 (or S3-compatible object storage) is a pretty well-known term in the industry, I don't think it needs more explanation.

website/docs/language/settings/backends/s3.mdx
janosdebugs: Did you copy the `)`? It works for me.

virtualroot: Weird, it works now. 
I did search for tofu in the `aws-samples` org with no results, but I see it now.

PR #1342, opened 2024-03-05
gcs backend: fix race condition on locking

### LINKED ISSUE #1341: In concurrent scenarios lock with gcs backend fails with "Error acquiring the state lock"
### OpenTofu Version

```shell
OpenTofu v1.6.2
on linux_amd64
+ provider registry.opentofu.org/hashicorp/google v5.18.0
+ provider registry.opentofu.org/hashicorp/google-beta v5.18.0
```

```hcl
terraform {
  backend "gcs" {
    bucket = "some-bucket"
    prefix = "some-prefix"
  }
}
```

### Debug Output

 none

### Expected Behavior

Lock obtained

### Actual Behavior

```
Acquiring state lock. This may take a few moments...

Error: Error acquiring the state lock

Error message: 2 errors occurred:
	* writing
"gs://some-bucket/some-prefix/default.tflock"
failed: googleapi: Error 412: At least one of the pre-conditions you
specified did not hold., conditionNotMet
	* storage: object doesn't exist
	
```

### Steps to Reproduce

tofu apply -auto-approve -lock-timeout=30m -no-color

### Additional Context

The scenario is next - we have many root modules and created 'shared' module and use it via `data "terraform_remote_state"` 
as result up to 20 processes run apply with the same root module/lock file/gcs backend

I am moving this issue from other tool because source seems the same and other tool don't fix this issue for 2 years 

from what I discovered the problem is next
- when [Lock](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L91) can't  [write lock file](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L104) it tries to [read exists lock file](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L149) via [lockError()](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L111) and [lockInfo()](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L137)
- there is some probability that there already no file (deleted by other process) and lockInfo() and lockError() returns error without filled Info field and info.ID 
- but [LockWithContext()](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/states/statemgr/locker.go#L92) expects either no error or filled err.Info.ID  

### References

_No response_

oprudkyi:
- handles situation when Lock() fails because lock file exists, and then lockInfo() fails because file was deleted by concurrent process
- try to fix #1341

Details: 
- when [Lock](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L91) can't  [write lock file](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L104) it tries to [read exists lock file](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L149) via [lockError()](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L111) and [lockInfo()](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L137)
- there is some probability that there already no file (deleted by other process) and lockInfo() and lockError() returns error without filled Info field and info.ID 
- but [LockWithContext()](https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/states/statemgr/locker.go#L92) expects either no error or filled err.Info.ID  
- for such a case the patch returns dummy err.Info.ID so waiting for real lock continues

Resolves #1341

## Target Release

1.7.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ BUG FIXES:
 * `tofu test` now supports accessing module outputs when the module has no resources. ([#1409](https://github.com/opentofu/opentofu/pull/1409))
 * Fixed support for provider functions in tests ([#1603](https://github.com/opentofu/opentofu/pull/1603))
 * Added a better error message on `for_each` block with sensitive value of unsuitable type. ([#1485](https://github.com/opentofu/opentofu/pull/1485))
+* Fix race condition on locking in gcs backend ([#1342](https://github.com/opentofu/opentofu/pull/1342))
 
 ## Previous Releases
 
--- a/internal/backend/remote-state/gcs/client.go
+++ b/internal/backend/remote-state/gcs/client.go
@@ -7,6 +7,7 @@ package gcs
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
@@ -135,9 +136,13 @@ func (c *remoteClient) lockError(err error) *statemgr.LockError {
 	}
 
 	info, infoErr := c.lockInfo()
-	if infoErr != nil {
+	switch {
+	case errors.Is(infoErr, storage.ErrObjectNotExist):
+		// Race condition - file exists initially but then has been deleted by other process
+		lockErr.InconsistentRead = true
+	case infoErr != nil:
 		lockErr.Err = multierror.Append(lockErr.Err, infoErr)
-	} else {
+	default:
 		lockErr.Info = info
 	}
 	return lockErr
--- a/internal/states/statemgr/locker.go
+++ b/internal/states/statemgr/locker.go
@@ -89,14 +89,17 @@ func LockWithContext(ctx context.Context, s Locker, info *LockInfo) (string, err
 			return "", err
 		}
 
-		if le == nil || le.Info == nil || le.Info.ID == "" {
-			// If we don't have a complete LockError then there's something
-			// wrong with the lock.
+		if !le.Retriable() {
 			return "", err
 		}
 
 		if postLockHook != nil {
 			postLockHook()
+		}
+
+		// Lock() can be repeated without sleep
+		if le.RetriableWithoutDelay() {
+			continue
 		}
 
 		// there's an existing lock, wait and try again
@@ -214,6 +217,10 @@ func (l *LockInfo) String() string {
 type LockError struct {
 	Info *LockInfo
 	Err  error
+
+	// Set when writing of lock file fails because of conflict and
+	// then reading fails because file doesn't exist (removed by other process)
+	InconsistentRead bool
 }
 
 func (e *LockError) Error() string {
@@ -226,4 +233,20 @@ func (e *LockError) Error() string {
 		out = append(out, e.Info.String())
 	}
 	return strings.Join(out, "\n")
+}
+
+// Retriable returns true when locking should be retried
+func (e *LockError) Retriable() bool {
+	// If we don't have a complete LockError then there's something
+	// wrong with the lock.
+	if e == nil {
+		return false
+	}
+
+	return e.InconsistentRead || (e.Info != nil && e.Info.ID != "")
+}
+
+// RetriableWithoutDelay returns true when delaying can be avoided
+func (e *LockError) RetriableWithoutDelay() bool {
+	return e != nil && e.InconsistentRead
 }


internal/backend/remote-state/gcs/client.go
```diff
@@ -147,6 +148,15 @@ func (c *remoteClient) lockError(err error) *statemgr.LockError {
 // LockInfo struct.
 func (c *remoteClient) lockInfo() (*statemgr.LockInfo, error) {
 	r, err := c.lockFile().NewReader(c.storageContext)
+
+	// When lock file do exist in Lock() but writing fails because file removed by other process,
```
marcinwyszynski, line 152: > writing fails

Not sure I understand. Did you mean "reading fails"?

oprudkyi: Hi, 
writing fails there https://github.com/opentofu/opentofu/blob/36eb93f95862c9f1e8a270f6a6e31f8818b0bfa5/internal/backend/remote-state/gcs/client.go#L104 in Lock() (file exists) then it calls lockInfo() and then reading fails in lockInfo() 
race conditions are hard to explain clearly 
What text should I use there ?

marcinwyszynski: So it's "writing fails because of conflict", then "reading fails because the file is no longer there", correct?

marcinwyszynski (commented):
First, thanks for contributing!

Having read the code back and forth, I can see that the problem is real.

But I'm having some thoughts about the solution.

The most superficial concern that I have is that `lockInfo` is indirectly used by the `Unlock` function, too, so this new logic will pass there, too, with unknown consequences.

More importantly, I'm not sure if we're solving this problem on the right level. The way I understand your approach is that you're trying to make the error look like `LockError` (with bogus data) so it can be handled by the `LockWithContext` function.

But isn't it a square peg in a round hole? Would it not be more graceful if a new error type was introduced, and then handled by the `LockWithContext` function, representing just this scenario - which BTW may be universal for all the locking backends?

So my three main high-level concerns here are:
- knock-on effect on other functions within the GCS backend;
- readability of the code for future contributors - by using what I'd call a clever gimmick we could be making the logic more difficult to follow (and it is already difficult to follow as-is, at least to me);
- making the fix as universal as possible;

Looking forward to your thoughts on the matter @oprudkyi !

internal/backend/remote-state/gcs/client.go
oprudkyi: the sequence is something like 
- lock() checks storage and there is file already (so w.Write fails by condition DoesNotExist:true) 
```
w := lockFile.If(storage.Conditions{DoesNotExist: true}).NewWriter(c.storageContext)
err = func() error {
		if _, err := w.Write(infoJson); err != nil {
			return err
		}
		return w.Close()
	}()
```
- then as error handling, lockInfo() tries to read file 
```
r, err := c.lockFile().NewReader(c.storageContext)
```
and fails too because other process already deleted file

marcinwyszynski: Gotcha, I think we're generally on the same page.

oprudkyi:
Hi @marcinwyszynski yes, probably universal solution will be more useful. I  just created ugly workaround for issue. 
Sadly, I do not have enough experience with golang and code base to be sure I can make it universal and avoid introducing new bugs 
I could try though. but not before weekend

marcinwyszynski:
@oprudkyi Please let me know if you need any help. I think there's a cheaper way - I discussed it with the core team:

Cursor_and_dev__Channel__-_opentf_-_Slack

...then you'd be calling `if le.Retriable()` in the caller (`LockWithContext`) to encapsulate this logic. A small optimization would be to not sleep at all if you get an `InconsistentRead`.

internal/states/statemgr/locker.go
```diff
@@ -227,3 +233,25 @@ func (e *LockError) Error() string {
...
+	if e == nil || e.Info == nil {
+		return false
+	}
+
+	return e.Info.InconsistentRead || e.Info.ID != ""
+}
+
+// RetriableWithoutDelay returns true when delaying can be avoided
+func (e *LockError) RetriableWithoutDelay() bool {
+	// If we don't have a complete LockError then there's something
```
marcinwyszynski, line 250: Shorter version?

```golang
func (e *LockError) RetriableWithoutDelay() bool {
  return e.Retriable() && e.Info.ID == ""
}
```

internal/backend/remote-state/gcs/client.go
```diff
@@ -147,6 +147,14 @@ func (c *remoteClient) lockError(err error) *statemgr.LockError {
 // LockInfo struct.
 func (c *remoteClient) lockInfo() (*statemgr.LockInfo, error) {
 	r, err := c.lockFile().NewReader(c.storageContext)
+
+	// Race condition - file exists initially but then has been deleted by other process
+	if err == storage.ErrObjectNotExist {
+		info := &statemgr.LockInfo{}
```
marcinwyszynski, line 153: Style nit: why not build and return immediately?

```golang
return &statemgr.LockInfo{InconsistentRead: true}, nil
```

marcinwyszynski (dismissed):
See nits. The logic however addresses my concerns, so handing it over to the core team for the review.

internal/states/statemgr/locker.go
```diff
@@ -140,6 +143,9 @@ type LockInfo struct {
 
 	// Path to the state file when applicable. Set by the Lock implementation.
 	Path string
+
+	// Set when there is race conditions on locking in non-locking backends (GCS)
```
marcinwyszynski, line 147: What do you mean by "non-locking backends"? GCS is a locking backend.
I think it would also make sense to explain what the exact race condition is (write fails because of conflict, read returns nothing).

oprudkyi: updated with 
```
	// Set when writing of lock file fails because of conflict and
	// then reading fails because file doesn't exist (removed by other process)
```

internal/states/statemgr/locker.go
oprudkyi: updated

internal/backend/remote-state/gcs/client.go
```diff
@@ -147,6 +147,11 @@ func (c *remoteClient) lockError(err error) *statemgr.LockError {
 // LockInfo struct.
 func (c *remoteClient) lockInfo() (*statemgr.LockInfo, error) {
 	r, err := c.lockFile().NewReader(c.storageContext)
+
+	// Race condition - file exists initially but then has been deleted by other process
+	if err == storage.ErrObjectNotExist {
```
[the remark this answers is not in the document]
oprudkyi: updated

internal/states/statemgr/locker.go
```diff
@@ -140,6 +143,10 @@ type LockInfo struct {
 
 	// Path to the state file when applicable. Set by the Lock implementation.
 	Path string
+
+	// Set when writing of lock file fails because of conflict and
+	// then reading fails because file doesn't exist (removed by other process)
+	InconsistentRead bool `json:"-"`
```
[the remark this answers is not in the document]
oprudkyi: Hi @ollevche 
I tried to do so , but it seems isn't convenient as for me 
so I placed it as another branch https://github.com/opentofu/opentofu/compare/main...oprudkyi:opentofu:issue/gcs-lock-fails-v2
if it is fine , I will merge it there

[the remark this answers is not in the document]
ollevche: Hey @oprudkyi, thanks for the changes - it looks pretty fine to me!

I would only simplify a bit `Retriable` and `RetriableWithoutDelay` functions, but it's only a suggestion, not a blocker.

cam72cam (dismissed):

oprudkyi:
Hi @ollevche 

I've pushed version with  `InconsistentRead` flag under the `LockError` structure. Please review

internal/states/statemgr/locker.go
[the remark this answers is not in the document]
oprudkyi: done

ollevche (commented):
Thank you for the changes! I'm happy to approve once the following is addressed:
* existing changelog entries shouldn't be removed (see PR comment)
* golint should pass for all new code (see failing checks)

CHANGELOG.md
```diff
@@ -12,9 +12,7 @@ BUG FIXES:
 * Fixed crash in gcs backend when using certain commands ([#1618](https://github.com/opentofu/opentofu/pull/1618))
 * Fix inmem backend crash due to missing struct field ([#1619](https://github.com/opentofu/opentofu/pull/1619))
 * Added a check in the `tofu test` to validate that the names of test run blocks do not contain spaces. ([#1489](https://github.com/opentofu/opentofu/pull/1489))
-* `tofu test` now supports accessing module outputs when the module has no resources. ([#1409](https://github.com/opentofu/opentofu/pull/1409))
-* Fixed support for provider functions in tests ([#1603](https://github.com/opentofu/opentofu/pull/1603))
-* Added a better error message on `for_each` block with sensitive value of unsuitable type. ([#1485](https://github.com/opentofu/opentofu/pull/1485))
```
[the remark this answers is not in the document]
oprudkyi: done

oprudkyi:
@ollevche  
pr has been fixed

ollevche (approved):
Thank you for your contribution!

Yantrio (approved):

cam72cam (approved):

oprudkyi:
Today the first time I was able to update 50+ k8 clusters without need to re-run failed jobs
Thank you very much everyone for help with fix

PR #1347, opened 2024-03-07
Added support to allow provider blocks in tests to access variables and outputs

### LINKED ISSUE #1185: `tofu test` - allow provider blocks in tests to access variables and run outputs
### OpenTofu Version

```shell
1.6.0
```

### Use Cases

Like in the main tofu configuration, there are cases where we'd like to initialize providers based on dynamic configuration in the tests. 

One example is referring to variables that have been defined in the test file. That would allow to test the module with different regions for a provider, for example
Another example is referring to run outputs - Running some setup configuration, which might get some secret AWS credentials, and then you could initialize the provider with these credentials

### Attempted Solutions

Currently that's not possible, and the provider configuration in test suites is pretty static

### Proposal

Allow `provider` blocks in test files to refer to variables defined in the test file, and also allow them to refer to `run` block outputs:

```
variables {
  aws_access_key_id = "aa"
  aws_secret_access_key = "bb"
}

provider "aws" {
  access_key = var.aws_access_key_id
  secret_key = var.aws_secret_access_key
}
```

and:
```
provider "aws" {
  access_key = run.setup.aws_access_key_id
  secret_key = run.setup.aws_secret_access_key
}

run "setup" {
  module {
    source = "./testing/setup"
  }
}
```

### References

Related issue in the Terraform Repository - https://github.com/hashicorp/terraform/issues/34007

siddharthasonker95:
Objective:
Add support to allow `provider` blocks in tests to access `variables` and `outputs`

Use Case:
Like in the main tofu configuration, there are cases where we'd like to initialize providers based on dynamic configuration in the tests.

One example is referring to variables that have been defined in the test file. That would allow to test the module with different regions for a provider, for example
Another example is referring to run outputs - Running some setup configuration, which might get some secret AWS credentials, and then you could initialize the provider with these credentials

Changes:

- Added support for `run` block in parse_ref.go to incorporate changes from run block output.
- Added `run` key in `eval.go` to store `cty.Value` for run block outputs.
- Added test cases for the run outputs.

Resolves #1185

## Target Release

1.7.0


internal/command/testdata/test/pass_with_run_block_variable/main.tf
chenzejun, line 1: I think this test file should be added to internal\command\test_test.go for automation.
Btw, can we have some failed case here?

internal/command/testdata/test/pass_with_run_block_variable/main.tf
```diff
@@ -0,0 +1,31 @@
...
+
+output "aws_access_key_id" {
+  sensitive = false
+  value = var.aws_access_key_id
+}
+
+terraform {
+  required_providers {
+    docker = {
+      source  = "kreuzwerker/docker"
```
chenzejun, line 16: can we simply use 'provider "test"' like other tests？

Yantrio: I agree, using a real provider here is a little strange.

internal/command/testdata/test/pass_with_run_block_variable/main.tf
```diff
@@ -0,0 +1,31 @@
+# main.tf
```
Yantrio, line 1: This isn't needed

internal/lang/eval.go
```diff
@@ -453,6 +453,12 @@ func (s *Scope) evalContext(refs []*addrs.Reference, selfAddr addrs.Referenceabl
 
 	if len(outputValues) > 0 {
 		vals["output"] = cty.ObjectVal(outputValues)
+
+		// "run" block will have output values when size of ref.Remaining is greater than 0.
+		if len(refs) > 0 && len(refs[0].Remaining) > 0 {
```
Yantrio, line 458: I may not be fully understanding the code in this area, but what if `refs[0].Remaining` was 0 and `refs[1].Remaining` was > 0, shouldn't we also handle that?

I think we shouldn't just check the 0th element in the slice.

internal/lang/eval.go
```diff
@@ -462,6 +468,21 @@ func (s *Scope) evalContext(refs []*addrs.Reference, selfAddr addrs.Referenceabl
...
 }
 
+func getRunBlockValue(refs []*addrs.Reference, outputValues map[string]cty.Value) map[string]cty.Value {
+	var block, outputVariable string
+	if attrTrav, ok := refs[0].Remaining[0].(hcl.TraverseAttr); ok {
+		block = attrTrav.Name
+	}
+	if attrTrav, ok := refs[0].Remaining[1].(hcl.TraverseAttr); ok {
+		outputVariable = attrTrav.Name
+	}
```
Yantrio, line 478: why are we only ever pulling from `refs[0]` ?

Yantrio (changes requested):

internal/addrs/parse_ref.go
```diff
@@ -333,6 +333,25 @@ func parseRef(traversal hcl.Traversal) (*Reference, tfdiags.Diagnostics) {
 			Remaining:   remain,
 		}, diags
 
+	case "run":
```
Evi1Pumpkin, line 336: I don't think this logic should be a part of the general `parseRef` function, as it would leak into the regular TF operations.
From looking at the code, I see there is the [ParseRefFromTestingScope](https://github.com/opentofu/opentofu/blob/8ce1b42085b36e5a2f8a3fd536623df2386873ba/internal/addrs/parse_ref.go#L93) function., and I believe this logic should fit there.

internal/lang/eval.go
```diff
@@ -462,6 +468,21 @@ func (s *Scope) evalContext(refs []*addrs.Reference, selfAddr addrs.Referenceabl
 	return ctx, diags
 }
 
+func getRunBlockValue(refs []*addrs.Reference, outputValues map[string]cty.Value) map[string]cty.Value {
```
ronnyorot, line 471: Like my previous comment, this logic shouldn't be here, as it leaks into the general tofu logic. The `evalContext` is the general context for the plan/apply, and you are populating it with something strictly related to the test's feature. It shouldn't be here.
In this [PR](https://github.com/opentofu/opentofu/pull/1129/files#diff-16f7ccd0b34c619fda941531b649e919dcc79d87306854db9d212c3050296305R1063-R1075) @cam72cam  already implemented what you are trying to do in this function - save the run block outputs. As part of his PR, he passed the `run` outputs so other `run` blocks would be able to access them.
So IMO, you need to use his existing logic and figure out how to use the run block outputs in the `provider` block inside the test file.

internal/command/testdata/test/pass_with_run_block_variable/tests/main.tftest.hcl
```diff
@@ -0,0 +1,16 @@
+# tests/main.tftest.hcl
+
+provider "docker" {
+  host = run.setup.aws_access_key_id
```
ronnyorot, line 4: This is a test for the scenario of allowing `provider` blocks to access `run` outputs. What about the other scenario -  allowing `provider` blocks to access variables?
I'm asking because I didn't see any relevant code change or relevant test.

internal/lang/eval.go
siddharthasonker95: Thanks @ronnyorot , I will check the code again and figure out the solution.

internal/command/testdata/test/pass_with_run_block_variable/tests/main.tftest.hcl
siddharthasonker95: I had a fix for accessing variables in the provider block too, but after the latest pull from the main branch, the variables were working fine in the provider block without any explicit code changes. However, I will add some test cases for reference.

siddharthasonker95:
I have been working on this issue and need some input from the team. With respect to this issue, I have added a 'run' block in the [parseRef](https://github.com/opentofu/opentofu/blob/4117fc281a0a65bfa75714bceed8847a2dd3e489/internal/addrs/parse_ref.go#L174)() method, which is incorrect. The correct implementation would be to add it in [ParseRefFromTestingScope](https://github.com/opentofu/opentofu/blob/4117fc281a0a65bfa75714bceed8847a2dd3e489/internal/addrs/parse_ref.go#L93)(). ParseRefFromTestingScope() is not being called from any place in the tofu test command. To implement that, we need a new ProviderTestConfig struct having hcl.body interface, which will trigger custom PartialContent() based on the ProviderTestConfig hcl.body (PartialContent() is triggered from [graph_builder.go](https://github.com/opentofu/opentofu/blob/4117fc281a0a65bfa75714bceed8847a2dd3e489/internal/tofu/graph_builder.go#L45)), and in that, we can trigger ParseRefFromTestingScope(). Along with that, the changes in EvalContext() in [eval.go](https://github.com/opentofu/opentofu/blob/60b262dcdfbbc250c419ea4d0076be5b77945269/internal/lang/eval.go#L458) with respect to the run block remain the same. Please let me know what you think of the approach and suggest some alternative approaches.

The changes mentioned in this PR indicate that it is being called after the Validate() function. Therefore, I'm unsure of how to reuse this 'run' block parser in the implementation mentioned above.

internal/addrs/parse_ref.go
```diff
@@ -113,6 +113,24 @@ func ParseRefFromTestingScope(traversal hcl.Traversal) (*Reference, tfdiags.Diag
...
+				Detail:   `The "run" block output values must be followed by two attribute names: the resource type and the property. For eg: run.resource_type.property.`,
+				Subject:  traversal.SourceRange().Ptr(),
+			})
+			return nil, diags
+		}
+
+		remainingRefs := traversal[1:]
+		outputVariable := remainingRefs[len(remainingRefs)-1]
+		return &Reference{
+			Subject:     OutputValue{Name: outputVariable.(hcl.TraverseAttr).Name},
```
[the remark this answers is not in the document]
siddharthasonker95: Here we are trying to access the outputVariable from the previous "run" block only, and `outputVariable` is initialized with `remainingRefs`, which contains the output variable path (escaped first occurrence from traversal, i.e., "run")
But still I will add a check to avoid panic.

internal/lang/eval.go
```diff
@@ -455,13 +459,31 @@ func (s *Scope) evalContext(refs []*addrs.Reference, selfAddr addrs.Referenceabl
...
+	}
+
 	if self != cty.NilVal {
 		vals["self"] = self
 	}
 
 	return ctx, diags
 }
 
+func getRunBlockValue(ref *addrs.Reference, value cty.Value, runBlockValue map[string]map[string]cty.Value) {
```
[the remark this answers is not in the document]
siddharthasonker95: `getRunBlockValue()` is used to store the output value with respect to the run block nested structure. for eg: if provider is trying to access the value of `run.test_module.test_value` then we need to store the value following the same nested structure in the [hcl.EvalContext](https://github.com/opentofu/opentofu/blob/17bc1acfc2678db707e005d5ecf32e5cc881ddc6/internal/lang/eval.go#L247) so that in [hcldec.Decode()](https://github.com/opentofu/opentofu/blob/17bc1acfc2678db707e005d5ecf32e5cc881ddc6/internal/lang/eval.go#L73) can evaluate it’s value based on the provided [hcl.body](https://github.com/opentofu/opentofu/blob/17bc1acfc2678db707e005d5ecf32e5cc881ddc6/internal/lang/eval.go#L52).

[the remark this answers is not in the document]
ollevche: Thanks for the explanation! Also, there are some confusing parts:

* Why do we check only the first and second elements of `ref.Remaining` slice? Do we need to add a length check at least?

* In the case of type cast failure we will just write the value to empty keys: `runBlockValue[block][outputVariable] = value`, where `block` and/or `outputVariable` are empty . Is it expected?

* The function is writing its return value into the argument passed. Can we refactor it a bit to follow more common practices (read-only arguments with return values)? The signature could be similar to the following: `func getRunBlockKeys(ref *addrs.Reference) (block, outputVar string)`

internal/tofu/evaluate.go
```diff
@@ -76,10 +76,10 @@ type Evaluator struct {
 // If the "self" argument is nil then the "self" object is not available
 // in evaluated expressions. Otherwise, it behaves as an alias for the given
 // address.
-func (e *Evaluator) Scope(data lang.Data, self addrs.Referenceable, source addrs.Referenceable) *lang.Scope {
+func (e *Evaluator) Scope(data lang.Data, self addrs.Referenceable, source addrs.Referenceable, parseRef lang.ParseRef) *lang.Scope {
 	return &lang.Scope{
 		Data:          data,
-		ParseRef:      addrs.ParseRef,
+		ParseRef:      parseRef,
```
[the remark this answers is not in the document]
siddharthasonker95: `parseRef` cannot be part of `Evaluator` as we need parseRef to be part of the block we are evaluating. The block can be of run, output, module, etc. type as mentioned in [parseRef()](https://github.com/opentofu/opentofu/blob/17bc1acfc2678db707e005d5ecf32e5cc881ddc6/internal/addrs/parse_ref.go#L192) method. 
I will try to figure out more ways where we can put the parseRef field.

internal/lang/eval.go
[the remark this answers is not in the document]
siddharthasonker95: We check first two elements because the we are expecting the output value referred from run block is not going to be nested further. For eg: run.setup.sample_test_value is being used in this [test file.](https://github.com/opentofu/opentofu/blob/17bc1acfc2678db707e005d5ecf32e5cc881ddc6/internal/command/testdata/test/pass_with_run_block_output/tests/main.tftest.hcl#L2)

I have added the length before calling getRunBlockValue() method [here](https://github.com/opentofu/opentofu/blob/17bc1acfc2678db707e005d5ecf32e5cc881ddc6/internal/lang/eval.go#L418).
I will update this if block to check for exact two elements as per the use case.

internal/addrs/parse_ref.go
```diff
@@ -478,3 +494,13 @@ func parseSingleAttrRef(traversal hcl.Traversal) (string, hcl.Range, hcl.Travers
 	})
 	return "", hcl.Range{}, nil, diags
 }
+
+func appendRunDiag(subject *hcl.Range) tfdiags.Diagnostics {
+	var diags tfdiags.Diagnostics
+	return diags.Append(&hcl.Diagnostic{
+		Severity: hcl.DiagError,
+		Summary:  "Invalid reference",
+		Detail:   `The "run" block output values must be followed by two attribute names: the resource type and the property. For example: run.resource_type.property.`,
```
Evi1Pumpkin, line 503: I don't believe this error is correct. We are not referencing a resource + property of the `run` block. We are referencing and output in the format of `run.aws_access_key_id`, which means that after the `run` we have the run name and the name of the output.

Evi1Pumpkin (changes requested):
This is going in the right direction, but please notice my comments.
I have two main concerns:
1. For some reason, while testing locally, the said functionality doesn't seem to work.
2. We are still integrating `test` logic inside the general `plan`/`apply` logic. The `test` logic should be separated and not affect the `plan`/`apply` operations.

internal/lang/eval.go
```diff
@@ -290,6 +290,7 @@ func (s *Scope) evalContext(refs []*addrs.Reference, selfAddr addrs.Referenceabl
 	countAttrs := map[string]cty.Value{}
 	forEachAttrs := map[string]cty.Value{}
 	checkBlocks := map[string]cty.Value{}
+	runBlockValues := map[string]map[string]cty.Value{}
```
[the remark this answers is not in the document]
Evi1Pumpkin: Another hint that might help to figure it out:
Today, `provider` blocks in the regular operations (not `test`)[ already support variables](https://opentofu.org/docs/language/providers/configuration/#:~:text=You%20can%20use%20expressions%20in%20the%20values%20of%20these%20configuration%20arguments%2C%20but%20can%20only%20reference%20values%20that%20are%20known%20before%20the%20configuration%20is%20applied.) in their config. For example:
```
# main.tf

variable "region" {
	default = "us-east-1"
}

 provider "aws" {
   region = var.region
}
```

It'll be good to investigate how it works so we can use a similar logic regarding tests.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 UPGRADE NOTES:
 * Backend/S3: The default of `use_legacy_workflow` changed to `false` and is now deprecated. The S3 backend will follow the same behavior as AWS CLI and SDKs for credential search, preferring backend configuration over environment variables. To support the legacy credential search workflow, you can set this option as `true`. It'll be completely removed in a future minor version.
+* Allow provider blocks in `tofu tests` to access variables and run outputs. ([#1347](https://github.com/opentofu/opentofu/pull/1347))
```
Yantrio, line 5: this feels like an enhancement, not a UPGRADE NOTE

internal/command/test.go
```diff
@@ -1065,11 +1090,58 @@ func getEvalContextFromStates(states map[string]*TestFileState) *hcl.EvalContext
...
+		}
+		providerSchema := schemas.ProviderSchema(config.ProviderForConfigAddr(provider.Addr()))
+		spec := providerSchema.Provider.Block.DecoderSpec()
+
+		// If attributes are nil, then it iterates to the next ProviderConfig.
+		attributes, _ := provider.Config.JustAttributes()
+		for _, attribute := range attributes {
+			if len(attribute.Expr.Variables()) > 0 {
+				// Check if the variable root name exists in the eval context
+				if _, ok := evalCtx.Variables[attribute.Expr.Variables()[0].RootName()]; ok {
```
Yantrio, line 1127: to reduce nesting and improve readability it may be best to invert these if statements.

internal/command/test_test.go
```diff
@@ -815,6 +815,14 @@ func TestTest_Modules(t *testing.T) {
 			expected: "main.tftest.hcl... pass\n  run \"setup\"... pass\n  run \"test\"... pass\n\nSuccess! 2 passed, 0 failed.\n",
 			code:     0,
 		},
+		"pass_with_run_block_variable": {
```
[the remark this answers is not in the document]
Yantrio: @siddharthasonker95 ping

internal/configs/test_file.go
```diff
@@ -41,6 +42,15 @@ const (
 	RefreshOnlyTestMode TestMode = 'R'
 )
 
+// TestProviderConfig represents the config for test providers.
```
Yantrio, line 45:
```suggestion
// TestProviderConfig represents the config for providers used in tests.
```

internal/configs/config.go
```diff
@@ -950,13 +951,42 @@ func (c *Config) TransformForTest(run *TestRun, file *TestFile) (func(), hcl.Dia
...
 		// the test file itself.
 		for key, provider := range file.Providers {
-			next[key] = provider
+			next[key] = &Provider{
+				Name:       provider.Name,
+				NameRange:  provider.NameRange,
+				Alias:      provider.Alias,
+				AliasRange: provider.AliasRange,
+				Version:    provider.Version,
+				Config:     &TestProviderConfig{Body: provider.Config, Value: cty.NilVal},
```
Yantrio, line 960: I'm a little concerned we may be mixing up real nil values and cty.NilVal here.

Is there a reason the Value field isn't a *cty.Value so we can just pass nil instead of cty.NilVal?

I would rather check if the value is unset, rather than cty.NilVal

internal/tofu/node_provider.go
```diff
@@ -56,6 +57,12 @@ func (n *NodeApplyableProvider) ValidateProvider(ctx EvalContext, provider provi
 
 	configBody := buildProviderConfig(ctx, n.Addr, n.ProviderConfig())
 
+	// If a provider configuration is of type TestProviderConfig, then skip the validation,
+	// and its value will be evaluated before the Plan operation.
```
Yantrio, line 61:
```suggestion
	// and its value should be evaluated before the Plan operation.
```

internal/tofu/node_provider.go
```diff
@@ -56,6 +57,12 @@ func (n *NodeApplyableProvider) ValidateProvider(ctx EvalContext, provider provi
...
+	if _, ok := configBody.(*configs.TestProviderConfig); ok {
```
Yantrio, line 62: can you please add a unit test for this case please? This method is pretty well tested and it helps us avoid regressions in the future. Thanks!

internal/tofu/node_provider.go
```diff
@@ -113,10 +120,21 @@ func (n *NodeApplyableProvider) ConfigureProvider(ctx EvalContext, provider prov
...
+	var evalDiags tfdiags.Diagnostics
+
+	// Check if the config body is of type TestProviderConfig.
+	// Use the value directly if available
+	testProviderConfig, isTestProviderConfig := configBody.(*configs.TestProviderConfig)
+	if isTestProviderConfig && testProviderConfig.Value != cty.NilVal {
+		configVal = testProviderConfig.Value
+	} else {
+		configVal, configBody, evalDiags = ctx.EvaluateBlock(configBody, configSchema, nil, EvalDataForNoInstanceKey)
+		diags = diags.Append(evalDiags)
```
Yantrio, line 134: this diag should probably be wrapped in a more user-friendly one. similar to below.

Yantrio (changes requested):

internal/command/test_test.go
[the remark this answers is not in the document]
siddharthasonker95: The tests are working fine with revised approach in PR. Thanks!

internal/tofu/node_provider.go
siddharthasonker95: added [here.](https://github.com/opentofu/opentofu/pull/1347/commits/b15cb4962966b088eebc00b34e6dac0470d81eaa)

simonostendorf:
Whats the current status here?

RLRabinowitz (changes requested):
I tested this locally, and there are some cases that do not work properly. I will list them here:

1. When having a provider reference a `run` output, for a run that had not yet been executed, the error is very unclear and pops up multiple times

```hcl
provider "aws" {
  region = run.test.my_region
}

// Assuming the configuration uses the AWS provider, we will fail here as the provider is needed here, but the run which the provider relies on has not yet executed
// The error is unclear, and appears multiple times, which is the problem
run "test" {
  command = plan
}

run "test2" {
  command = plan
}
```

2. Same for providers that reference nonexistent run blocks. Error is unclear
3. The error with referencing a non-existent variable is not so clear (Unsupported attribute. This object does not have an attribute named "x")
4. Using both variables and run outputs inside a provider definition in a test block somehow breaks the run output reference

```hcl
variables {
  region = "us-east-1"
}

provider "aws" {
  access_key = run.setup.aws_access_key_id // For some reason having "region = var.region" breaks this line. Running the test command say that "run.setup" cannot be resolved
  region = var.region
}

run "setup" {
  module {
    source = "./testing/setup"
  }
}
```

PR #1349, opened 2024-03-07
Initial implementation of aws_kms encryption.key_provider

### LINKED ISSUE #1172: [State Encryption] Key providers (AWS KMS)
**Parent:** #1030
**Depends on:** #1166
**Blueprint PR:** #1132

## Acceptance criteria

1. Implement a key provider that loads an encryption key from the AWS KMS.

cam72cam:
This PR implements the AWS KMS key provider using identical configuration to the s3 backend.  We will be able to copy over most of that documentation once we have a dedicated place on the website for encryption docs.

I've run this through a myriad of testing locally with no issues.

Resolves #1172 

## Target Release

1.7.0


internal/encryption/base.go
```diff
@@ -36,14 +36,14 @@ func newBaseEncryption(enc *encryption, target *config.TargetConfig, enforced bo
...
-	_, diags := base.buildTargetMethods(make(map[keyprovider.Addr]any))
+	_, diags := base.buildTargetMethods(make(map[keyprovider.Addr][]byte))
 	return base, diags
 }
 
 type basedata struct {
-	Meta    map[keyprovider.Addr]any `json:"meta"`
-	Data    []byte                   `json:"encrypted_data"`
-	Version string                   `json:"encryption_version"` // This is both a sigil for a valid encrypted payload and a future compatability field
+	Meta    map[keyprovider.Addr][]byte `json:"meta"`
```
cube2222, line 44: I feel like I've already seen the move in the opposite direction in a previous PR.

Would appreciate to hear the reasoning behind it.

internal/encryption/keyprovider/aws_kms/provider.go
```diff
@@ -0,0 +1,98 @@
...
+			KeySpec: spec,
+		})
+
+		if err != nil {
+			return out, outMeta, err
+		}
+
+		// Set inital outputs
+		out.EncryptionKey = generatedKeyData.Plaintext
+		out.DecryptionKey = generatedKeyData.Plaintext
```
cube2222, line 53: I'd add a comment here that the decryption key will be overriden later, if the statefile is already encrypted. Just because I was surprised when reading the code, and only later saw the end of the function.

internal/encryption/keyprovider/aws_kms/provider.go
```diff
@@ -0,0 +1,98 @@
...
+		outMeta.CiphertextBlob = generatedKeyData.CiphertextBlob
+	} else {
+		var spec types.DataKeyPairSpec
```
cube2222, line 56: I would propose not supporting assymetric kms keys for now. Imo people should just use symmetric ones for this.

internal/encryption/keyprovider/aws_kms/provider.go
```diff
@@ -0,0 +1,98 @@
...
+			KeyPairSpec: spec,
+		})
+
+		if err != nil {
+			return out, outMeta, err
+		}
+
+		// Set inital outputs
+		out.EncryptionKey = generatedKeyData.PublicKey
+		out.DecryptionKey = generatedKeyData.PrivateKeyPlaintext
```
cube2222, line 78: **If** we want to do assymetric crypto (and as indicated in the other comment, I'm against) then this is not the way to do it. Asymmetric keys must be used with an encryption method dedicated to them - RSA in this case most likely (not sure if KMS supports elliptic). RSA also has a very strong limit on the message size (I believe length of the key).

So the way this needs to be done is as follows:
- On encryption
  - Generate a new AES key.
  - Encrypt the AES key using the public key and put it into the metadata.
  - Put the encrypted private key into the metadata.
  - Pass the plaintext AES key out as the encryption key.
- On decryption
  - Decrypt the encrypted private key using kms.
  - Decrypt the AES key using the private key.
  - Pass the decrypted AES key as the decryption key.

But overall, as I said, I'd suggest we leave this out for now, and possibly add it in the future, if someone requests it.

internal/encryption/keyprovider/keyprovider.go
```diff
@@ -24,5 +27,5 @@ type Descriptor interface {
 
 type KeyProvider interface {
 	// Provide provides an encryption and decryption keys. If the process fails, it returns an error.
-	Provide() (Output, error)
+	Provide(KeyMeta) (Output, KeyMeta, error)
```
cube2222, line 30: I think @janosdebugs's idea was that the metadata is passed in via constructor?

No strong opinion on my side, but please make sure to discuss.

internal/encryption/keyprovider/aws_kms/provider.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/kms"
+	"github.com/aws/aws-sdk-go-v2/service/kms/types"
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+)
+
+type keyMeta struct {
+	CiphertextBlob []byte `json:"ciphertext_blob"`
```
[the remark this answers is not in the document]
cam72cam: I was just matching the AWS convention, don't mind changing it

internal/encryption/keyprovider/keyprovider.go
cam72cam: @janosdebugs and I spent an hour going over different designs and settled on this one for now.

internal/encryption/keyprovider/aws_kms/provider.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	// Set inital outputs
+	out.EncryptionKey = generatedKeyData.Plaintext
+	out.DecryptionKey = generatedKeyData.Plaintext
+	outMeta.CiphertextBlob = generatedKeyData.CiphertextBlob
+
+	if len(inMeta.CiphertextBlob) != 0 {
+		// We have an existing decryption key to decode
+		decodedKeyData, err := p.svc.Decrypt(p.ctx, &kms.DecryptInput{
```
cube2222, line 57: Total nit: decryptedKeyData

cube2222 (dismissed):
LGTM!

internal/encryption/keyprovider/aws_kms/provider.go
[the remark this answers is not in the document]
cube2222: Fair, no strong opinion.

cube2222 (approved):

internal/encryption/keyprovider/aws_kms/provider_test.go
```diff
@@ -0,0 +1,160 @@
...
+	output, meta, err := provider.Provide(metaIn)
+	if err != nil {
+		t.Fatalf("Error providing keys: %s", err)
+	}
+
+	if len(output.EncryptionKey) == 0 {
+		t.Fatalf("No encryption key provided")
+	}
+
+	if len(output.DecryptionKey) == 0 {
```
janosdebugs, line 79: This behavior is actually not correct. Since you are not providing an input meta, the key provider does not need to provide a decryption key.

cam72cam: That's true, though I don't think it hurts to provide the decryption key for testing?  I imagine we will refactor this pattern at some point.

janosdebugs: I would be cautious because it could lead to confusing error messages. The error message on incorrect/incomplete metadata should be that there is no decryption kry, not that the decryption failed.

janosdebugs (dismissed):

cube2222 (approved):

cube2222 (approved):

PR #1350, opened 2024-03-08
Insert configuration marks to state value when import

### LINKED ISSUE #1231: Planable import get unexpected update action when import resource with sensitive path
### OpenTofu Version

```shell
-> % tofu -version 
OpenTofu v1.7.0-dev
```

```
data "kubernetes_namespace" "ns" {
  metadata {
    name = "namespace-test"
  }
}

data "kubernetes_secret" "src" {
  metadata {
    name      = "secret-src"
    namespace = data.kubernetes_namespace.ns.id
  }
}

resource "kubernetes_secret" "dst" {
  metadata {
    name      = "secret-dst"
    namespace = data.kubernetes_namespace.ns.id
  }
  wait_for_service_account_token = false
  type                           = "kubernetes.io/basic-auth"
  data = data.kubernetes_secret.src.data
}

import {
  to = kubernetes_secret.dst
  id = "namespace-test/secret-dst"
}

```

### Expected Behavior

Plan: 1 to import, 0 to add, 0 to change, 0 to destroy.

### Actual Behavior

image

### Steps to Reproduce

tofu init
prepare two secret with same value
tofu plan

### Additional Context

the bug may be caused here:
https://github.com/opentofu/opentofu/blob/199ec898e634b5c2782664cf436bac94eae3fd10/internal/tofu/node_resource_abstract_instance.go#L1087C1-L1096C1

### References

_No response_

chenzejun:
For resource import scenario, the instance refresh state should contains all the marks(include marks form configuration ) to avoid a update action caused by difference marks between planned value and refresh state(for example: write or delete sensitive paths is an update action).

Resolves #1231

## Target Release

1.7.0

### DIFF
--- a/internal/tofu/node_resource_import.go
+++ b/internal/tofu/node_resource_import.go
@@ -268,9 +268,8 @@ func (n *graphNodeImportStateSub) Execute(ctx EvalContext, op walkOperation) (di
 	// Insert marks from configuration
 	if n.Config != nil {
 		// Since the import command allow import resource with incomplete configuration, we ignore diagnostics here
-		valueWithSchemaMarks, _, _ := ctx.EvaluateBlock(n.Config.Config, n.Schema, nil, EvalDataForNoInstanceKey)
-		_, pvm := valueWithSchemaMarks.UnmarkDeepWithPaths()
-		state.Value = state.Value.MarkWithPaths(pvm)
+		valueWithConfigurationSchemaMarks, _, _ := ctx.EvaluateBlock(n.Config.Config, n.Schema, nil, EvalDataForNoInstanceKey)
+		state.Value = copyMarksFromValue(state.Value, valueWithConfigurationSchemaMarks)
 	}
 
 	diags = diags.Append(riNode.writeResourceInstanceState(ctx, state, workingState))
--- a/internal/tofu/node_resource_import.go
+++ b/internal/tofu/node_resource_import.go
@@ -10,6 +10,8 @@ import (
 	"log"
 
 	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/configs"
+	"github.com/opentofu/opentofu/internal/configs/configschema"
 	"github.com/opentofu/opentofu/internal/providers"
 	"github.com/opentofu/opentofu/internal/states"
 	"github.com/opentofu/opentofu/internal/tfdiags"
@@ -20,6 +22,10 @@ type graphNodeImportState struct {
 	ID               string                    // ID is the ID to import as
 	ProviderAddr     addrs.AbsProviderConfig   // Provider address given by the user, or implied by the resource type
 	ResolvedProvider addrs.AbsProviderConfig   // provider node address after resolution
+
+	Schema        *configschema.Block // Schema for processing the configuration body
+	SchemaVersion uint64              // Schema version of "Schema", as decided by the provider
+	Config        *configs.Resource   // Config is the resource in the config
 
 	states []providers.ImportedResource
 }
@@ -178,6 +184,9 @@ func (n *graphNodeImportState) DynamicExpand(ctx EvalContext) (*Graph, error) {
 			TargetAddr:       addrs[i],
 			State:            state,
 			ResolvedProvider: n.ResolvedProvider,
+			Schema:           n.Schema,
+			SchemaVersion:    n.SchemaVersion,
+			Config:           n.Config,
 		})
 	}
 
@@ -194,6 +203,11 @@ type graphNodeImportStateSub struct {
 	TargetAddr       addrs.AbsResourceInstance
 	State            providers.ImportedResource
 	ResolvedProvider addrs.AbsProviderConfig
+
+	Schema        *configschema.Block // Schema for processing the configuration body
+	SchemaVersion uint64              // Schema version of "Schema", as decided by the provider
+	Config        *configs.Resource   // Config is the resource in the config
+
 }
 
 var (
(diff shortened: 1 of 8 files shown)


cam72cam (approved):

cam72cam:
Thanks @chenzejun, would you mind adding a changelog entry?

chenzejun:
> Thanks @chenzejun, would you mind adding a changelog entry?
sure, np

cam72cam (approved):

chenzejun:
@RLRabinowitz @janosdebugs @Yantrio  Hi folks, will someone ship it? 😊

internal/tofu/node_resource_plan_instance.go
```diff
@@ -554,6 +554,17 @@ func (n *NodePlannableResourceInstance) importState(ctx EvalContext, addr addrs.
...
+	// insert marks from schema
+	if n.Config != nil {
+		valueWithSchemaMarks, _, configDiags := ctx.EvaluateBlock(n.Config.Config, n.Schema, nil, EvalDataForNoInstanceKey)
+		diags = diags.Append(configDiags)
+		if configDiags.HasErrors() {
+			return instanceRefreshState, diags
+		}
+		_, pvm := valueWithSchemaMarks.UnmarkDeepWithPaths()
+		instanceRefreshState.Value = instanceRefreshState.Value.MarkWithPaths(pvm)
+	}
```
RLRabinowitz, line 566: This fix is specific to the import block, and does not affect the `tofu import` command, which have the same results.

Also, feels a bit weird to me that we need to perform HCL block evaluation here in order to get the schema marks. You could probably simply rely on

RLRabinowitz (changes requested):
Thank you for your contribution @chenzejun !

I can see that the reproducible case you've included in the issue is indeed resolved here, but for the `import` command this case still happens (try to remove the `import` block, and to run `tofu import` prior to the plan).

I think that ideally the `refresh` function should return the value including the schema marks. We would need to see, though, how this affects other usages of it (regular plan, deposed, orphan), but it does make sense for the refreshed state to include the schema marks.

Either way, this fix should work for `tofu import` as well

internal/tofu/node_resource_plan_instance.go
chenzejun: Thanks @RLRabinowitz,  actually, the refresh state contains schema marks defined in provider.
But in this case, the sensitive mark is from configuration(sensitive datasouce), so here I do a evaluation to avoid the difference between planned val and current state. And I edited my PR description to make it clear.

chenzejun: As for the `tofu import`, I will continue work on it

RLRabinowitz: @chenzejun sorry if I wasn't clear here, but the following does not work with the current code, and should behave the same:

Let's say you have the following configuration fresh (same as with the issue, just without the import block):
```hcl
data "kubernetes_namespace" "ns" {
  metadata {
    name = "namespace-test"
  }
}

data "kubernetes_secret" "src" {
  metadata {
    name      = "secret-src"
    namespace = data.kubernetes_namespace.ns.id
  }
}

resource "kubernetes_secret" "dst" {
  metadata {
    name      = "secret-dst"
    namespace = data.kubernetes_namespace.ns.id
  }
  wait_for_service_account_token = false
  type                           = "kubernetes.io/basic-auth"
  data = data.kubernetes_secret.src.data
}
```

And then you:
1. `tofu init`
2. `tofu import "kubernetes_secret.dst" "namespace-test/secret-dst"`
3. `tofu plan`

You would get the same result of an unnecessary "will-update-in-place" plan

chenzejun: Hi, @RLRabinowitz I have also fixed the import command, PTAL again

internal/tofu/node_resource_import.go
```diff
@@ -251,6 +265,14 @@ func (n *graphNodeImportStateSub) Execute(ctx EvalContext, op walkOperation) (di
 		return diags
 	}
 
+	// Insert marks from configuration
+	if n.Config != nil {
+		// Since the import command allow import resource with incomplete configuration, we ignore diagnostics here
```
RLRabinowitz, line 270: Shouldn't this technically be the same for plannable import as well? We might need to ignore diagnostics there as well

internal/tofu/node_resource_plan_instance.go
```diff
@@ -554,6 +554,17 @@ func (n *NodePlannableResourceInstance) importState(ctx EvalContext, addr addrs.
 		return instanceRefreshState, diags
 	}
 
+	// Insert marks from configuration
```
RLRabinowitz, line 557: nit

This and the above parts (refreshing and failure in case of importing a non-existent object) are shared with the implementation of the import command as well.
Consider extracting them to some common function `refreshImportedState`, which would refresh + error on non-existent remote object + add schema marks.

Alternatively, just extract the "add schema marks" part to a common function

RLRabinowitz (commented):
Thank you @chenzejun !

I took a deeper look into how the marks work, and it seems like importing doesn't act the same as other planning behaviour (like a regular plan / orphan), as the state that is attempted to refresh does not exist already, with the necessary schema marks.

Initially I thought that this is an issue with how the `refresh` works, but it doesn't seem like it.

I've left a couple of comments, but overall LGTM

RLRabinowitz (approved):

internal/tofu/node_resource_import.go
chenzejun: In fact, before plannable import, the validate walk will perform this evaluate to make sure the config is valid, so it will never fail when plannable import. 
Ignore or not are both acceptable, but I prefer to check it to make the code looks clean. 

https://github.com/opentofu/opentofu/blob/739a66a86c4d775adf60d67cdc6c066059e46c12/internal/tofu/node_resource_validate.go#L383

internal/tofu/node_resource_plan_instance.go
chenzejun: Good suggestion, and I prefer the latter. 
I am not sure how will the plannable part differ to the import command, with we add more feature to plannable import(such as support for_each). 
So extract the "add schema marks" part to a common function is good enough for now.

RLRabinowitz (approved):

chenzejun:
Hi @cam72cam, can you help merge it?

PR #1377, opened 2024-03-12
[State Encryption] Compliance tests

janosdebugs:
This test suite is a comprehensive check of all key providers, the registry, and the methods to ensure compliance as far as error types, metadata, and behavior are concerned.

## Target Release

1.7.0


internal/encryption/keyprovider/id.go
```diff
@@ -5,11 +5,18 @@
 
 package keyprovider
 
+import "fmt"
+
 // ID is a type alias to make passing the wrong ID into a key provider harder.
 type ID string
 
 // Validate validates the key provider ID for correctness.
 func (i ID) Validate() error {
```
cube2222, line 14: Total nit: Could we use id for the method receiver here? When I see i in the code, I automatically think "loop index".

cube2222 (dismissed):
Impressive pile of reflection

janosdebugs:
Back to draft, this needs another do-over to eliminate the duplications with the method compliance tests.

cam72cam (approved):

PR #1381, opened 2024-03-12
feat: implement issensitive function

### LINKED ISSUE #1370: Add an `issensitive` function, checking whether a value is sensitive or not
### OpenTofu Version

```shell
1.6.2
```

### Use Cases

It would be useful if we could tell whether or not a value is sensitive programatically. A use-case for that is if you'd like to filter all sensitive data for an output of a module

### Attempted Solutions

Before [this change](https://github.com/opentofu/opentofu/issues/394), `nonsensitive` used to error if a value is already not sensitive. So, you could use `!can(nonsensitive(value))` to tell you if a value is sensitive. However, that no longer works in 1.7.0

### Proposal

Add a function `issensitive` that does that explicitly - return a boolean saying whether or not a value is sensitive

### References

Related issue in the Terraform registry - https://github.com/hashicorp/terraform/issues/34572,

JonasBak:
This PR implements a `issensitive` function that returns whether or not a value is sensitive as described in #1370.

Resolves #1370

## Target Release

1.7.0

### DIFF
--- a/internal/lang/funcs/sensitive.go
+++ b/internal/lang/funcs/sensitive.go
@@ -60,10 +60,34 @@ var NonsensitiveFunc = function.New(&function.Spec{
 	},
 })
 
+// IsSensitiveFunc returns whether or not the value is sensitive.
+var IsSensitiveFunc = function.New(&function.Spec{
+	Params: []function.Parameter{
+		{
+			Name:             "value",
+			Type:             cty.DynamicPseudoType,
+			AllowUnknown:     true,
+			AllowNull:        true,
+			AllowMarked:      true,
+			AllowDynamicType: true,
+		},
+	},
+	Type: func(args []cty.Value) (cty.Type, error) {
+		return cty.Bool, nil
+	},
+	Impl: func(args []cty.Value, retType cty.Type) (ret cty.Value, err error) {
+		return cty.BoolVal(args[0].HasMark(marks.Sensitive)), nil
+	},
+})
+
 func Sensitive(v cty.Value) (cty.Value, error) {
 	return SensitiveFunc.Call([]cty.Value{v})
 }
 
 func Nonsensitive(v cty.Value) (cty.Value, error) {
 	return NonsensitiveFunc.Call([]cty.Value{v})
+}
+
+func IsSensitive(v cty.Value) (cty.Value, error) {
+	return IsSensitiveFunc.Call([]cty.Value{v})
 }
--- a/website/docs/language/functions/issensitive.mdx
+++ b/website/docs/language/functions/issensitive.mdx
@@ -1,0 +1,40 @@
+---
+sidebar_label: issensitive
+description: >-
+  The issensitive function returns a boolean saying whether or not a value is
+  marked as sensitive.
+---
+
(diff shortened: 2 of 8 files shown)


cube2222 (approved):
LGTM, thanks for working on this!

janosdebugs (approved):

website/docs/language/functions/issensitive.mdx
```diff
@@ -0,0 +1,40 @@
...
+  description = "A list of environment variables that may contain both sensitive and non-sensitive values"
+  type        = map(string)
+  sensitive   = true
+}
+
+locals {
+  sensitive_variables = [for key, value in var.environment_variables: key if issensitive(value)]
+  nonsensitive_variables = [for key, value in var.environment_variables: key if !issensitive(value)]
+}
+```
```
chenzejun, line 29: This example seems can not construct a list with both sensitive and nonsensitive variables.I would recommend using this
```
locals {
  mixed_variables = tolist(["a", sensitive("b"), "c"])
}

locals {
  sensitive_variables = [for key, value in local.mixed_variables: key if issensitive(value)]
  nonsensitive_variables = [for key, value in local.mixed_variables: key if !issensitive(value)]
}

```

internal/lang/funcs/sensitive.go
```diff
@@ -60,10 +60,34 @@ var NonsensitiveFunc = function.New(&function.Spec{
...
+			AllowNull:        true,
+			AllowMarked:      true,
+			AllowDynamicType: true,
+		},
+	},
+	Type: func(args []cty.Value) (cty.Type, error) {
+		return cty.Bool, nil
+	},
+	Impl: func(args []cty.Value, retType cty.Type) (ret cty.Value, err error) {
+		return cty.BoolVal(args[0].HasMark(marks.Sensitive)), nil
```
chenzejun, line 79: HasMark can only check if the Value itself has mark
I am not sure if the Value is a list that contains some sensitive value, what should it return?
@cube2222 @janosdebugs which does the core team prefer?

website/docs/language/functions/issensitive.mdx
JonasBak: I tested the example in a module and used the module like this:

```
module "test" {
  source = "./test"
  environment_variables = {
    "TEST_SENSITIVE" = sensitive("A")
    "TEST_NONSENSITIVE" = "B"
  }
}

output "sensitive" {
  value = module.test.sensitive_variables
}

output "nonsensitive" {
  value = module.test.nonsensitive_variables
}
```

Which produced the following output:

```
Changes to Outputs:
  + nonsensitive = [
      + "TEST_NONSENSITIVE",
    ]
  + sensitive    = [
      + "TEST_SENSITIVE",
    ]
```

But I can change the example if it's more clear that way 😄

internal/lang/funcs/sensitive.go
cube2222: I think checking for the mark is fine, and what is expected here. It's how it's handled everywhere else, too.

cube2222:
Thanks a lot for the contribution @JonasBak!

PR #1397, opened 2024-03-14
Version bump to 1.7.0-alpha1

cube2222:


janosdebugs:
#1377 should still make it into the alpha!

cube2222:
@janosdebugs the version bump doesn't have to be the last commit to go in, we just need the version to be there for once we push the tag

janosdebugs (approved):

PR #1421, opened 2024-03-21
Generate all encryption keys during encryption setup

### LINKED ISSUE #1404: [State Encryption] Consider unlikely failure scenarios when writing state
### Configuration
```hcl 
encryption {
                key_provider "aws_kms" "basic" {
                        // configured
                }
                method "aes_gcm" "example" {
                        keys = key_provider.aws_kms.basic
                }
                state {
                        method = method.aes_gcm.example
                }

}
```
### Current Code Path `tofu apply`
* init apply command
  * load root module
  * construct encryption interface
  * construct state backend w/ state encryption
  * read state
    * call aws_kms key provider with given state metadata
    * call aes_gcm with keys
    * decrypt state
  * plan graph
  * apply graph
    * partial apply infrastructure changes
    * checkpoint write state
      * call aws_kms key provider
         * ERROR: US-EAST-1 is DOWN!
         * bubble error state all the way back to command package
  *  Call error handler `backend_apply:backupStateForError`
    * **Write state to "errored.tfstate"**
       * Use backend state encryption
          * aws_kms fails again!
    * **Write state to stderr**
       * Use backend state encryption
          * aws_kms fails again!
    * **State with partial changes is lost!!!!**
    
#### Don't encrypt either errored.tfstate or stderr state

This gives the user an easy to understand state file that they can read, manipulate, and react to.  The main downside is that the state will now be accessible to anyone who has access to that machine or even potentially backed up as part of a CI workflow.

#### Only try to encrypt errored.tfstate and if that fails write state to stderr unencrypted

This is similar to the previous solution, and still has the issue where if stderr is captured as part of CI it can expose secrets in the state.  However, it provides a file that can be manually put into the backend state store and potentially reduces the overhead of recovery.

#### Only generate the new encryption keys for a provider once per tofu invocation

We always need to be dynamic with how we Decrypt state files, the configuration depends on the provided state's metadata. However, we don't necessarily need to generate a new encryption key for a given key provider during the same invocation of tofu.  It would write variations of the state file using the same encryption key that is generated during the encryption interface init.  This would write more data with the same key which may have safety implications that I am personally not qualified to evaluate.

cam72cam:
See large multiline comment and #1404 for rationale on this change.

Resolves #1404

## Target Release

1.7.0

### DIFF
--- a/internal/encryption/base.go
+++ b/internal/encryption/base.go
@@ -21,10 +21,12 @@ const (
 )
 
 type baseEncryption struct {
-	enc      *encryption
-	target   *config.TargetConfig
-	enforced bool
-	name     string
+	enc        *encryption
+	target     *config.TargetConfig
+	enforced   bool
+	name       string
+	encMethods []method.Method
+	encMeta    map[keyprovider.Addr][]byte
 }
 
 func newBaseEncryption(enc *encryption, target *config.TargetConfig, enforced bool, name string) (*baseEncryption, hcl.Diagnostics) {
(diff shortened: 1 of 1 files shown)


internal/encryption/base.go
```diff
@@ -147,6 +169,7 @@ func (s *baseEncryption) decrypt(data []byte, validator func([]byte) error) ([]b
 		return nil, fmt.Errorf("invalid encrypted payload version: %s != %s", es.Version, encryptionVersion)
 	}
 
+	// TODO We could potentially cache this based on a json-encoded version of es.Meta and reduce overhead dramatically
```
Yantrio, line 172:
```suggestion
	// TODO Discuss if we should potentially cache this based on a json-encoded version of es.Meta and reduce overhead dramatically
```

Yantrio (approved):

PR #1422, opened 2024-03-21
Change migration path to and from encrypted state/plans

### LINKED ISSUE #1365: [State Encryption] Syntax for migrating from unencrypted state/plan
Dear everyone, we need your input on syntax for migrating from unencrypted to encrypted state and plan files. Please let us know which solution would work best for you and why. This is a companion issue to #1366.

## Background

Under normal operations, OpenTofu will refuse to read unencrypted state and plan files when encryption is configured. This is protecting users from smuggled-in state and plan files that could cause damage.

However, when migrating from unencrypted state and plan files to encrypted ones, the users need to have the ability to perform this migration. Furthermore, we want to enable users of CI systems to do this declaratively.

For more details on the state encryption function please see [the design document](https://github.com/opentofu/opentofu/blob/main/docs/state_encryption.md), [the user documentation](https://github.com/opentofu/opentofu/blob/main/website/docs/language/state/encryption.mdx), or [this video](https://www.youtube.com/watch?v=rR4IbhlRSkI).

## Solution 1

This solution infers the users' intention to read unencrypted data by the presence of an empty `fallback{}` block. This is somewhat unusual because users may not expect empty blocks to have such an implication.

```hcl
terraform {
  encryption {
    # Methods and key providers here

    statefile {
      method = method.aes_gcm.mymethod
      fallback{} # The empty fallback block signals that reading unencrypted state is OK.
    }
  }
}
```

Note that this functionality should be disabled if the `enforced` option is set to true.

## Solution 2

This solution uses an explicit option to allow reading unencrypted data.

```hcl
terraform {
  encryption {
    # Methods and key providers here

    statefile {
      method = method.aes_gcm.mymethod
      fallback{
        allow_unencrypted = true # Allows reading unencrypted by default
      }
    }
  }
}
```

Note that this functionality will be overridden by the `enforced` option, which will not allow unencrypted state to be read under any circumstances.

## Solution 3

This option changes solution 2 by moving the migration option up from the fallback block. This option would also actively refuse to work if it encounters an encrypted state file, forcing the user to immediately remove this option once the migration is complete. As before, the `enforced` flag would prevent unencrypted data manipulation.

```hcl
terraform {
  encryption {
    # Methods and key providers here

    statefile {
      method = method.aes_gcm.mymethod
      migrate_from_unencrypted = true
    }
  }
}
```

### LINKED ISSUE #1366: [State Encryption] Syntax for migrating to unencrypted state/plan
This issue is a companion to #1365 and deals with migrating from an encrypted to an unencrypted state/plan. Your input is needed on the syntax.

## Background

Under normal operations, OpenTofu will refuse to read unencrypted state and plan files when encryption is configured. This is protecting users from smuggled-in state and plan files that could cause damage.

However, when migrating from encrypted state and plan files to unencrypted ones, the users need to have the ability to perform this migration. Furthermore, we want to enable users of CI systems to do this declaratively.

For more details on the state encryption function please see [the design document](https://github.com/opentofu/opentofu/blob/main/docs/state_encryption.md), [the user documentation](https://github.com/opentofu/opentofu/blob/main/website/docs/language/state/encryption.mdx), or [this video](https://www.youtube.com/watch?v=rR4IbhlRSkI).

## Solution 1

This solution implicitly migrates from encrypted to unencrypted by not specifying a method. To prevent this behavior, users will need to use the `enforced` option.

```hcl
terraform {
  encryption {
    # Methods and key providers here

    statefile {
      # Note the lack of a method here.
      fallback {
        method = method.aes_gcm.mymethod
      }
    }
  }
}
```

## Solution 2

This solution will add an explicit option to allow decryption:

```hcl
terraform {
  encryption {
    # Methods and key providers here

    statefile {
      allow_unencrypted = true
      fallback {
        method = method.aes_gcm.mymethod
      }
    }
  }
}
```

cam72cam:
Added migrate_to_[un]encrypted fields and remove empty fallback logic.

This implements #1365 and #1366, with a few minor changes.

~~Additionally, it removes the need for enforced.  If a state{} or plan{} block is defined, it is automatically enforced.  We will probably remove the field.~~ See conversation below.

Migration enabled:
image

Trying to use encryption on an existing project without migration:
image

Resolves #1365 
Resolves #1366

## Target Release

1.7.0

### DIFF
--- a/internal/encryption/config/config.go
+++ b/internal/encryption/config/config.go
@@ -73,9 +73,12 @@ type TargetConfig struct {
 //
 // Note: This struct is copied because gohcl does not support embedding.
 type EnforcableTargetConfig struct {
-	Enforced bool           `hcl:"enforced,optional"`
-	Method   hcl.Expression `hcl:"method,optional"`
-	Fallback *TargetConfig  `hcl:"fallback,block"`
+	//TODO if a state{} or plan{} block is defined it implies enforced
+	Enforced             bool           `hcl:"enforced,optional"`
+	Method               hcl.Expression `hcl:"method"`
+	Fallback             *TargetConfig  `hcl:"fallback,block"`
+	MigrateToUnencrypted bool           `hcl:"migrate_to_unencrypted,optional"`
+	MigrateToEncrypted   bool           `hcl:"migrate_to_encrypted,optional"`
 }
 
 // AsTargetConfig converts the struct into its parent TargetConfig.
--- a/internal/encryption/config/config.go
+++ b/internal/encryption/config/config.go
@@ -59,8 +59,8 @@ func (m MethodConfig) Addr() (method.Addr, hcl.Diagnostics) {
 // RemoteConfig describes the terraform.encryption.remote block you can use to declare encryption for remote state data
 // sources.
 type RemoteConfig struct {
-	Default *TargetConfig       `hcl:"default,block"`
-	Targets []NamedTargetConfig `hcl:"remote_state_data_source,block"`
+	Default *RemoteDefaultTargetConfig `hcl:"default,block"`
+	Targets []NamedTargetConfig        `hcl:"remote_state_data_source,block"`
 }
 
 // TargetConfig describes the target.encryption.state, target.encryption.plan, etc blocks.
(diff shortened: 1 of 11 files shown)


internal/encryption/config/config.go
```diff
@@ -73,9 +73,12 @@ type TargetConfig struct {
 //
 // Note: This struct is copied because gohcl does not support embedding.
 type EnforcableTargetConfig struct {
-	Enforced bool           `hcl:"enforced,optional"`
-	Method   hcl.Expression `hcl:"method,optional"`
-	Fallback *TargetConfig  `hcl:"fallback,block"`
+	//TODO if a state{} or plan{} block is defined it implies enforced
```
janosdebugs, line 76: No, enforced has a different purpose. Enforced needs to be there to make sure the state is encrypted even if there is no available method. (e.g. in an environment variable scenario.) This should be explicit to the user, empty blocks can be confusing. An empty state block should not be allowed with this change.

janosdebugs (approved):
I give thee my blessing.

internal/encryption/config/config.go
cam72cam: We chatted a bit and I'm convinced.  Going to re-work the PR to revert this.

cam72cam:
Converting back to a draft as this breaks a lot of assumptions around remote state access

PR #1425, opened 2024-03-22
Improve panic handling within go-routines

### LINKED ISSUE #1424: Improve crash logs when handling a panic within a go-routine
Many of the operations and graph functions are run from within go-routines and when something very goes wrong they panic.

We should improve the logging panic handler to include as much of a detailed stack trace as possible.  Right now we only include the stack trace within the go-routine that panic'd.

Recently we were debugging a crash I introduced via some of the state encryption work.  The stack trace was deep within an apply operation and was missing the context for how the operation was constructed and called.  It took a bunch of guess-work to determine the code path which originally spawned that operation go-routine.  When tofu was augmented with an initial prototype of this work, confirmed the initial guesswork and showed exactly where we can configured a backend incorrectly.

cam72cam:
I've introduced a new panic handler which captures the calling stack trace.  It is a fairly small modification to the codebase which provides a much better view into the stack during a panic.  It does make a additional call to debug.Stack() before go-routines are spawned, but this is not deeply nested in any loops and should not incur any significant performance penalty.

This only is a partial solution to the problem of panics within deeply nested go-routines.  It only works between the go-routine being called and the calling go-routine.  If you have multiple nested go-rotuines, it will only preserve the
calling stack and the called panic stack.  Idealy we would be able to use context or a similar construct to build a more comprehensive panic handler, but this is a significant step in the right direction that will dramatically improve crash
debugging.

One of the scenarios which is not completely solved by this PR is `main() -spawn> operation() -spawn> graph_walk()`.  If `graph_walk` panics, it will only include the stack trace from `graph_walk` and `operation` and will not include the main stack trace.  This is still a big improvement, but as before mentioned there is still room for improvement.

Resolves #1424, may be worth creating a followup issue to discuss the deeper improvements mentioned above.

## Target Release

1.7.0

### DIFF
--- a/internal/logging/panic.go
+++ b/internal/logging/panic.go
@@ -53,10 +53,13 @@ func PanicHandler() {
 
 // PanicHandlerWithTraceFn returns a function similar to PanicHandler which is
 // called to recover from an internal panic in OpenTofu, and augments the
-// standard stack trace with a more user friendly error message.
+// standard stack trace with a more complete stack trace.
 // The calling stack trace is captured before returing the augmented panicHandler
 // The returned panicHandler must be called as a defered function, and must be the
 // first defer called at the start of a new goroutine.
+//
+// Callers of this function should create the panicHandler before any tight looping
+// as there may be a performance impact if called excessively.
 //
 // This only is a partial solution to the problem of panics within deeply nested
 // go-routines.  It only works between the go-routine being called and the calling
--- a/internal/logging/panic.go
+++ b/internal/logging/panic.go
@@ -40,6 +40,10 @@ var panicMutex sync.Mutex
 // PanicHandler must be called as a defered function, and must be the first
 // defer called at the start of a new goroutine.
 func PanicHandler() {
+	PanicHandlerWithTrace(nil)
+}
+
+func PanicHandlerWithTrace(trace []byte) {
 	// Have all managed goroutines checkin here, and prevent them from exiting
 	// if there's a panic in progress. While this can't lock the entire runtime
 	// to block progress, we can prevent some cases where OpenTofu may return
@@ -58,6 +62,10 @@ func PanicHandler() {
 	// When called from a deferred function, debug.PrintStack will include the
 	// full stack from the point of the pending panic.
 	debug.PrintStack()
+	if trace != nil {
+		fmt.Fprint(os.Stderr, "With go-routine called from:\n")
+		os.Stderr.Write(trace)
+	}
 
 	// An exit code of 11 keeps us out of the way of the detailed exitcodes
 	// from plan, and also happens to be the same code as SIGSEGV which is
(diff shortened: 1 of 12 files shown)


internal/tofu/graph.go
```diff
@@ -44,11 +44,13 @@ func (g *Graph) walk(walker GraphWalker) tfdiags.Diagnostics {
 	// The callbacks for enter/exiting a graph
 	ctx := walker.EvalContext()
 
+	panicHandler := logging.PanicHandlerWithTraceFn()
```
cam72cam, line 47: Note: This panicHandler (and call to debug.Stack()) only happens once, and not for each vertex within the graph.  This means the performance impact should be negligible.

internal/logging/panic.go
```diff
@@ -48,6 +48,39 @@ func PanicHandler() {
...
+//
+// This only is a partial solution to the problem of panics within deeply nested
+// go-routines.  It only works between the go-routine being called and the calling
+// go-routine.  If you have multiple nested go-rotuines, it will only preserve the
+// calling stack and the called panic stack.  Idealy we would be able to use context
+// or a similar construct to build a more comprehensive panic handler, but this
+// is a significant step in the right direction that will dramatically improve crash
+// debugging
+func PanicHandlerWithTraceFn() func() {
+	trace := debug.Stack()
```
janosdebugs, line 69: What's the performance impact of this? This would run on every goroutine and collect the stack trace regardless if a panic happened or not.

janosdebugs (dismissed):
Looks good in theory, performance impact may be a problem.

internal/logging/panic.go
cam72cam: It is called from every "calling" go routine.  The only place that we spawn lots of routines is in the graph walker.  That captures the calling trace before iteration and correctly re-uses the same calling stack.

internal/logging/panic.go
```diff
@@ -48,6 +48,39 @@ func PanicHandler() {
 	defer panicMutex.Unlock()
 
 	recovered := recover()
+	panicHandler(recovered, nil)
+}
+
+// PanicHandlerWithTraceFn returns a function similar to PanicHandler which is
+// called to recover from an internal panic in OpenTofu, and augments the
+// standard stack trace with a more user friendly error message.
```
Yantrio, line 56: I dont thin this is more "user friendly" if the user is the "end user" here,

internal/logging/panic.go
Yantrio: Should we add a comment that callers of this function should be aware that calling this could add a performance impact?

cam72cam: Added a comment here and the one place where we are close to a tight loop

Yantrio (approved):

Issue #1174, opened 2024-01-22
[State Encryption] Key providers (OpenBAO)


janosdebugs:
**Parent:** #1030
**Depends on:** #1166
**Blueprint PR:** #1132

## Acceptance criteria

1. Implement a key provider that loads an encryption key from OpenBAO.
2. Implement a test integration that spins up an OpenBAO instance for testing purposes on demand.

cipherboy:
@ollevche do you have a design doc for this? Are you planning on using Transit and offloading encryption? Or are you using k/v and fetching the key locally? IMO the former gives better security properties.

cipherboy:
@ollevche That's a good use for data keys! Thank you!

ollevche:
Closing since we implemented it in https://github.com/opentofu/opentofu/pull/1436

PR #1436, opened 2024-03-26
add OpenBao as key provider for state encryption

ollevche:
WIP

Part of https://github.com/opentofu/opentofu/issues/1174.`

## Target Release

1.7.0


internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("serializing decryption request to openbao: %w", err)
+	}
+
+	secret, err := s.c.WriteBytesWithContext(ctx, path, reqBody)
+	if err != nil {
+		return nil, fmt.Errorf("sending decryption request to openbao: %w", err)
+	}
+
+	return retrievePlaintext(secret)
```
siddharthasonker95, line 72: Can we please handle the error before directly returning the plaintext?

siddharthasonker95 (dismissed):

internal/encryption/keyprovider/openbao/client.go
cam72cam: That already happens in the implementation of retrievePlaintext.  As this is a private function, I don't believe that is necessary here.

cam72cam: As this is already handled by retrievePlaintext and is a private function, I don't see the need for that here.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+	openbao "github.com/openbao/openbao/api"
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+)
+
+type Config struct {
+	Address string `hcl:"address,optional"`
+	Token   string `hcl:"token,optional"`
+
+	KeyName        string `hcl:"key_name"`
+	DataKeyBitSize int    `hcl:"data_key_bit_size,optional"`
```
cam72cam, line 17: Let's keep this consistent as key_length to match the other key providers.

ollevche: Thanks for the comment! Originally, I aimed to match OpenBao docs, but it could confuse the end user. Changed `data_key_bit_size` (bits) to `key_length` (bytes) in 4d09d69182154da558bd7aa116ea4dd6feade7f3.

cam72cam (dismissed):

siddharthasonker95 (dismissed):
LGTM

cube2222 (dismissed):
Nice one!

I'd like for Janos to have a final look at this before merging, please.

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+	c client
+}
+
+type dataKey struct {
+	Plaintext  []byte
+	Ciphertext []byte
+}
+
+func (s service) generateDataKey(ctx context.Context, keyName string, bitSize int) (dataKey, error) {
+	path := fmt.Sprintf("/transit/datakey/plaintext/%s", keyName)
```
janosdebugs, line 28: This looks like a URL. Do we need to urlencode the key name?

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+	reqBody, err := json.Marshal(map[string]interface{}{
+		"bits": bitSize,
+	})
+	if err != nil {
+		return dataKey{}, fmt.Errorf("serializing datakey request to openbao: %w", err)
+	}
+
+	secret, err := s.c.WriteBytesWithContext(ctx, path, reqBody)
+	if err != nil {
+		return dataKey{}, fmt.Errorf("sending datakey request to openbao: %w", err)
```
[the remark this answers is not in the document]
janosdebugs: I would consider under which circumstances this can happen and advise the user what to do with it in the error message.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, errNoOpenBaoTokenFound
+	}
+
+	return c.Logical(), nil
+}
+
+const defaultKeyLength = 32
+
+func validateKeyLength(keyLength int) error {
+	if keyLength != 16 && keyLength != 32 && keyLength != 64 {
```
janosdebugs, line 92: Is this an OpenBao restriction? If not, then we shouldn't validate it here, it's the method's job to validate that.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type clientConstructor func(config *openbao.Config, token string) (client, error)
+
+var errNoOpenBaoTokenFound = errors.New("no openbao token found")
+
+// newClient variable allows to inject different client implementations.
+// In order to keep client interface simple, token setting is in this function as well.
+// It's not possible to pass token in config.
+var newClient clientConstructor = func(config *openbao.Config, token string) (client, error) {
```
janosdebugs, line 72: This is a globally scoped injection, which makes parallelization of tests impossible and cleanup is needed after each test which is easy to mess up. I would prefer if we injected the constructor function directly, e.g. using a constructor function and provide a second constructor function that injects the defaults.

internal/encryption/keyprovider/openbao/mock_test.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+type mockClientFunc func(ctx context.Context, path string, data []byte) (*openbao.Secret, error)
+
+func (f mockClientFunc) WriteBytesWithContext(ctx context.Context, path string, data []byte) (*openbao.Secret, error) {
+	return f(ctx, path, data)
+}
+
+func injectMock(m mockClientFunc) {
+	newClient = func(_ *openbao.Config, _ string) (client, error) {
```
janosdebugs, line 16: Where does this get cleaned up? A `t.Cleanup` function would be in order here.

janosdebugs (dismissed):
Great work, just lots of tiny nit-picks! Thanks for this!

internal/encryption/keyprovider/openbao/client.go
ollevche: Yes, it's a URL, but we don't need to additionally escape it. `openbao/api` uses `net.URL` under the hood to escape the whole path: https://github.com/openbao/openbao/blob/main/api/client.go#L1193

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+	plaintext, err := base64.StdEncoding.DecodeString(base64Plaintext)
+	if err != nil {
+		return nil, fmt.Errorf("base64 decoding 'plaintext' from openbao: %w", err)
+	}
+
+	return plaintext, nil
+}
+
+func retrieveCiphertext(s *openbao.Secret) ([]byte, error) {
+	ciphertext, ok := s.Data["ciphertext"].(string)
```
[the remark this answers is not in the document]
ollevche: It doesn't use separate error messages but catches both cases. Please, see an example: https://go.dev/play/p/LBr8z6D8Ngc

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+	secret, err := s.c.WriteBytesWithContext(ctx, path, reqBody)
+	if err != nil {
+		return nil, fmt.Errorf("sending decryption request to openbao: %w", err)
+	}
+
+	return retrievePlaintext(secret)
+}
+
+func retrievePlaintext(s *openbao.Secret) ([]byte, error) {
+	base64Plaintext, ok := s.Data["plaintext"].(string)
```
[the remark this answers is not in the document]
ollevche: It doesn't use separate error messages but catches both cases. Please, see an example: https://go.dev/play/p/LBr8z6D8Ngc

internal/encryption/keyprovider/openbao/compliance_test.go
```diff
@@ -0,0 +1,229 @@
...
+			},
+		},
+	)
+}
+
+// Mocking is a bit complicated due to how openbao/api package is structured,
+// but in order to test cover as much as we can, it has to have some logic in here.
+
+func prepareClientMockForKeyProviderTest(t *testing.T, testKeyName string) mockClientFunc {
+	generateDataKeyPath := fmt.Sprintf("/transit/datakey/plaintext/%s", testKeyName)
```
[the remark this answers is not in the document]
ollevche: No, because the mock will get raw value (since `openbao/api` handles encoding internally).

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+type Config struct {
+	Address string `hcl:"address,optional"`
+	Token   string `hcl:"token,optional"`
+
+	KeyName   string `hcl:"key_name"`
+	KeyLength int    `hcl:"key_length,optional"`
+}
+
+func (c Config) Build() (keyprovider.KeyProvider, keyprovider.KeyMeta, error) {
+	if c.KeyName == "" {
```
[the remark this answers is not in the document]
ollevche: Unfortunately, there are no formatting rules in public docs. I suggest relying on server-side validation since these rules could differ slightly for multiple OpenBao-compatible implementations.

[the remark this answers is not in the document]
janosdebugs: In that case, please note the urlencode comments below.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+	// DefaultConfig reads BAO_ADDR and some other optional env variables.
+	config := openbao.DefaultConfig()
+	if config.Error != nil {
+		return nil, nil, &keyprovider.ErrInvalidConfiguration{
+			Cause: config.Error,
+		}
+	}
+
+	// Address from HCL supersedes BAO_ADDR.
+	if c.Address != "" {
```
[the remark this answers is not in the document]
ollevche: The client uses the default value - `https://127.0.0.1:8200`. I'm not sure if that's a useful one for our users, but it's possible to not set an address at all.

Do you also mean we should ping OpenBao to validate if it's real?

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+		config.Address = c.Address
+	}
+
+	client, err := newClient(config, c.Token)
+	if err != nil {
+		return nil, nil, &keyprovider.ErrInvalidConfiguration{
```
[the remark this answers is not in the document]
ollevche: I think so. It fails if there's a problem parsing env variables so the error message looks like "failed to parse ENV_VAR_NAME".

However, we may fail to read env vars we don't use. Example: user sets valid address in config and invalid in BAO_ADDR -> tofu will fail because BAO_ADDR is invalid. I would not fix it because I believe it's not practically useful. WDYT?

internal/encryption/keyprovider/openbao/config.go
ollevche: Yes, from [the OpenBao docs](https://github.com/openbao/openbao/blob/main/website/content/api-docs/secret/transit.mdx#generate-data-key):

> `bits` `(int: 256)` – Specifies the number of bits in the desired key. Can be 128, 256, or 512.

internal/encryption/keyprovider/openbao/config.go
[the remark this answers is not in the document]
janosdebugs: I think we may want to add a prefix, such as "failed to initialize the OpenBao key provider: whatever OpenBao said".

internal/encryption/keyprovider/openbao/provider.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (m keyMeta) isPresent() bool {
+	return len(m.Ciphertext) != 0
+}
+
+type keyProvider struct {
+	svc       service
+	keyName   string
+	keyLength int
+	ctx       context.Context
```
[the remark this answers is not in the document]
ollevche: Agree, I removed the ctx from the provider struct: [238499b](https://github.com/opentofu/opentofu/pull/1436/commits/238499b74c9ccaefe30551aa984b3490203ecf47). I would like to change Provide function signature to include context in a separate PR.

internal/encryption/keyprovider/openbao/mock_test.go
ollevche: Makes sense. I added a cleanup here: [7021c8d](https://github.com/opentofu/opentofu/pull/1436/commits/7021c8d304c5def03b23d6f9d4f29718e926e4f6).

internal/encryption/keyprovider/openbao/config.go
ollevche: Not sure that's what you expect, but I made it possible to reuse a default client constructor: [f460dfa](https://github.com/opentofu/opentofu/pull/1436/commits/f460dfa21c2affb99f90b4e8df97f1972e17d4fe). I think we still have a problem with tests parallelization in cases when we need multiple test suits run in parallel with different mocks / implementations. It is true for AWS and GCP providers as well. WDYT?

internal/encryption/keyprovider/openbao/compliance_test.go
```diff
@@ -0,0 +1,229 @@
...
+	openbao "github.com/openbao/openbao/api"
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider/compliancetest"
+)
+
+func getBaoKeyName() string {
+	// Acceptance tests are disabled, running with mock.
+	if os.Getenv("TF_ACC") == "" {
+		return ""
+	}
+	return os.Getenv("TF_BAO_KEY_NAME")
```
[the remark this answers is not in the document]
ollevche: Makes sense: [3ec55f2](https://github.com/opentofu/opentofu/pull/1436/commits/3ec55f243d319b39818bbf58897b0e38d8053f07). I would also like to change it for other providers in another PR

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (s service) generateDataKey(ctx context.Context, keyName string, bitSize int) (dataKey, error) {
+	path := fmt.Sprintf("/transit/datakey/plaintext/%s", keyName)
+
+	reqBody, err := json.Marshal(map[string]interface{}{
+		"bits": bitSize,
+	})
+	if err != nil {
+		return dataKey{}, fmt.Errorf("serializing datakey request to openbao: %w", err)
```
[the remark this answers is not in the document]
ollevche: Removed this block completely in [d628ce6](https://github.com/opentofu/opentofu/pull/1436/commits/d628ce681871fb8f1742ae3daa10c725e75da3c4). Client has a simpler option to use: `WriteWithContext` with internal serialization

internal/encryption/keyprovider/openbao/config.go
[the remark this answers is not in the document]
ollevche: It's already checked via openbao client internally so we don't need to additionally handle this. But it doesn't validate if OpenBao server accessible by the address so we can still get an error later even if the address valid syntactically valid.

internal/encryption/keyprovider/openbao/config.go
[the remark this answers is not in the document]
ollevche: `newClient` is our wrapper over openbao client constructor so the prefix is already present.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+			Cause: err,
+		}
+	}
+
+	return &keyProvider{
+		svc:       service{client},
+		keyName:   c.KeyName,
+		keyLength: c.KeyLength,
+		ctx:       context.Background(),
```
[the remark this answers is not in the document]
ollevche: Agree, I removed the ctx from the provider struct: [238499b](https://github.com/opentofu/opentofu/pull/1436/commits/238499b74c9ccaefe30551aa984b3490203ecf47). I would like to change Provide function signature to include context in a separate PR.

internal/encryption/keyprovider/openbao/client.go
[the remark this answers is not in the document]
ollevche: Added more context to error messages here: [bc2c668](https://github.com/opentofu/opentofu/pull/1436/commits/bc2c668ac9855ab9682b17faee1a023aba6a0c3a)

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+
+func retrievePlaintext(s *openbao.Secret) ([]byte, error) {
+	base64Plaintext, ok := s.Data["plaintext"].(string)
+	if !ok {
+		return nil, errors.New("failed to deserialize 'plaintext' from openbao")
+	}
+
+	plaintext, err := base64.StdEncoding.DecodeString(base64Plaintext)
+	if err != nil {
+		return nil, fmt.Errorf("base64 decoding 'plaintext' from openbao: %w", err)
```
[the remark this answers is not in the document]
ollevche: Fixed openbao to OpenBao here: [bc2c668](https://github.com/opentofu/opentofu/pull/1436/commits/bc2c668ac9855ab9682b17faee1a023aba6a0c3a)

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return plaintext, nil
+}
+
+func retrieveCiphertext(s *openbao.Secret) ([]byte, error) {
+	ciphertext, ok := s.Data["ciphertext"].(string)
+	if !ok {
+		return nil, errors.New("failed to deserialize 'ciphertext' from openbao")
```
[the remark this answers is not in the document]
ollevche: Added more context to error messages here: [bc2c668](https://github.com/opentofu/opentofu/pull/1436/commits/bc2c668ac9855ab9682b17faee1a023aba6a0c3a)

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,85 @@
...
+	key.Plaintext, err = retrievePlaintext(secret)
+	if err != nil {
+		return dataKey{}, err
+	}
+
+	return key, nil
+}
+
+func (s service) decryptData(ctx context.Context, keyName string, ciphertext []byte) ([]byte, error) {
+	path := fmt.Sprintf("/transit/decrypt/%s", keyName)
```
janosdebugs, line 52: This may need urlencoding though. :)

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,97 @@
...
+	openbao "github.com/openbao/openbao/api"
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+)
+
+type Config struct {
+	Address string `hcl:"address,optional"`
+	Token   string `hcl:"token,optional"`
+
+	KeyName   string `hcl:"key_name"`
+	KeyLength int    `hcl:"key_length,optional"`
```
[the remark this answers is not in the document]
ollevche: A separate type would be helpful for such cases, agree. I don't think `Set` will be beneficial here since it's only 3 possible values: [033bddb](https://github.com/opentofu/opentofu/pull/1436/commits/033bddbbeba08f084049882aa2fab9818a654a9a)

internal/encryption/keyprovider/openbao/client.go
ollevche: I believe it's the same as `generateDataKey`. `openbao/api`'s `WriteWithContext` method uses `url.URL`'s `Path` field to store the path and it then becomes encoded in `url.RequestURI`. Here is `openbao/api`'s [NewRequest function](https://github.com/openbao/openbao/blob/main/api/client.go#L1193).

ollevche: @janosdebugs thanks for the additional explanations! With internal escape, it could be the valid URL, but not the one we intend it to be. So I agree that's a problem for certain scenarios so I added escaping here: [6bc13d5](https://github.com/opentofu/opentofu/pull/1436/commits/6bc13d5a9efcf0a884e677f1fc62e325f0ffda5d)

internal/encryption/keyprovider/openbao/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	inMeta, ok := rawMeta.(*keyMeta)
+	if !ok {
+		return keyprovider.Output{}, nil, &keyprovider.ErrInvalidMetadata{
+			Message: "bug: invalid metadata struct type",
+		}
+	}
+
+	ctx := context.Background()
```
janosdebugs, line 37: It might be a good idea add a sensible default timeout here so tofu can't hang indefinitely.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+type clientConstructor func(config *openbao.Config, token string) (client, error)
+
+var errNoOpenBaoTokenFound = errors.New("no OpenBao token found")
+
+// newClient variable allows to inject different client implementations.
+// In order to keep client interface simple, token setting is in this function as well.
+// It's not possible to pass token in config.
+var newClient clientConstructor = newOpenBaoClient
```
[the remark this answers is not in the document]
ollevche: Yes, I want to fix it for OpenBao, GCP, and AWS (we use the same mock strategy for all of them) so I will prepare another PR after the merge.

internal/encryption/keyprovider/openbao/provider.go
ollevche: Agree, I also want to extend the `keyprovider.Provide` method to pass/receive `context.Context` so timeout could be set earlier. I want to prepare this as another PR (kind of tiny refactoring of mentioned tests and context).

cam72cam: Nope.  Contexts in OpenTofu are not well managed.  It's on the TODO list.

internal/encryption/keyprovider/openbao/compliance_test.go
```diff
@@ -0,0 +1,229 @@
...
+						return fmt.Errorf("ciphertext is empty")
+					}
+					return nil
+				},
+			},
+		},
+	)
+}
+
+// Mocking is a bit complicated due to how openbao/api package is structured,
```
cipherboy, line 175: If you want, you can [use `openbao/sdk/helper/testcluster/docker.NewTestDockerCluster(...)`](https://pkg.go.dev/github.com/openbao/openbao/sdk@main/helper/testcluster/docker#NewTestDockerCluster) to get a full OpenBao cluster instance. 

At the moment, this will actually load HashiCorp Vault, but this will be updated once we publish actual Docker images.

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,104 @@
...
+	"errors"
+	"fmt"
+
+	openbao "github.com/openbao/openbao/api"
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+)
+
+type Config struct {
+	Address string `hcl:"address,optional"`
+	Token   string `hcl:"token,optional"`
```
cipherboy, line 13: I kinda wonder how `token` should be managed? Ideally we'd like not to check it in to wherever the configuration lives, but the other fields could be. The [`api/auth` package](https://pkg.go.dev/github.com/openbao/openbao/api@main#section-directories) has various auth providers that perhaps could be detected and configured automatically to fetch & renew the token (e.g., AppRole, LDAP, ...) But I suppose Tofu would be a short-running process and so would be unlikely to need to renew the token in the first place?

Perhaps it should be read from the environment instead? Or something like [OpenBao Proxy](https://janma.github.io/openbao/docs/agent-and-proxy/proxy/apiproxy/) could be recommended instead, where Tofu would not maintain a token and instead contact a local auto-authing transparent proxy?

Some thoughts!

internal/encryption/keyprovider/openbao/config.go
```diff
@@ -0,0 +1,104 @@
...
+const defaultDataKeyLength DataKeyLength = 32
+
+type DataKeyLength int
+
+func (l DataKeyLength) Validate() error {
+	switch l {
+	case 16, 32, 64:
+		return nil
+	default:
+		return fmt.Errorf("data key length should one of 16, 32 or 64 bytes: got %v", l)
```
cipherboy, line 72: I'm confused here :-) While these are acceptable values for _Transit_, are these all acceptable values for OpenTofu? I'd imaging nothing in OpenTofu's encryption mechanisms use a 512-bit key and thus 64 should be excluded?

website/docs/language/state/examples/encryption/openbao.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  encryption {
+    key_provider "openbao" "my_bao" {
+      token = "s.Fg8wA4nDrP08TirpjEXkrTmt"
```
cipherboy, line 4: :see_no_evil:

internal/encryption/keyprovider/openbao/config.go
cipherboy: Tbh, I think it you just make it optional, you can probably use `{BAO,VAULT}_{TOKEN,ADDR}` transparently.

cipherboy: Ah, I see, you're doing that but overriding with the config. I see down below intbhe docs that you're marking token and address both as required, so I thought you weren't using the environment variables. :D

internal/encryption/keyprovider/openbao/config.go
ollevche: We limit OpenBao key provider to the limits of OpenBao itself. Currently, there is no practical sense to use a 64-byte data key, but we plan to extend the list of supported encryption methods.

janosdebugs: The key provider should still support everything OpenBao supports. The introduction of a new method shouldn't necessitate a change in the key provider because we'll forget to add it.

internal/encryption/keyprovider/openbao/config.go
ollevche: Yes, token could be set via `BAO_TOKEN` environment variable and in such case we don't require it in config. However, it was not possible to omit it entirely so I made it completely optional here: [e710e8f](https://github.com/opentofu/opentofu/pull/1436/commits/e710e8f6646771eac7b328c1543c9caad759fb9b). It will allow OpenTofu to be used with OpenBao proxy configured.

Also, we could implement more auth options in the future and limit it to token / proxy from now on.

Thanks for the suggestion!

internal/encryption/keyprovider/openbao/config.go
cipherboy: > The key provider should still support everything OpenBao supports. The introduction of a new method shouldn't necessitate a change in the key provider because we'll forget to add it.

Hmm, I'm still confused what happens here. You generate a 512-bit data key and then... Truncate it? Use it verbatim as if it was an AES key (only to get rejected later at AES key construction time)?

website/docs/language/state/examples/encryption/openbao.tf
cipherboy: Make sure you don't length check the token, many places have run into issues with that when upstream switched to the new format. Since we don't have upstream's replication modes, I ended up reverting to the simpler token construction.

internal/encryption/keyprovider/openbao/client.go
```diff
@@ -0,0 +1,88 @@
...
+	transitPath string
+}
+
+type dataKey struct {
+	Plaintext  []byte
+	Ciphertext []byte
+}
+
+func (s service) generateDataKey(ctx context.Context, keyName string, bitSize int) (dataKey, error) {
+	path := path.Join(s.transitPath, "datakey/plaintext", url.PathEscape(keyName))
```
cipherboy, line 30: I didn't notice this before, but key name must already be a path component:

```
    ^keys/(?P<name>\w(([\w-.]+)?\w)?)$
        Managed named encryption keys
```

I don't think path escaping it does anything; anything that'd need to be URL escaped wouldn't be a valid path, and URL escaping (to add `%`s) would prevent it from matching the path component too.

internal/encryption/keyprovider/openbao/provider.go
```diff
@@ -0,0 +1,66 @@
...
+
+	outMeta := &keyMeta{
+		Ciphertext: dataKey.Ciphertext,
+	}
+
+	out := keyprovider.Output{
+		EncryptionKey: dataKey.Plaintext,
+	}
+
+	if inMeta.isPresent() {
```
cipherboy, line 55: Just making sure I understand the scheme correctly: every encryption, we'll use a fresh data key? I think this makes sense to me.

internal/encryption/keyprovider/openbao/config.go
ollevche: Yes, it will lead to an error from the AES method. I agree we shouldn't let users make a mistake, but for this exact case, it's more about exposing different building blocks (and yes, some of them could be incompatible) to allow agile configuration.

internal/encryption/keyprovider/openbao/provider.go
ollevche: Yes, this is exactly right. For every encryption, we use a fresh data key.

internal/encryption/keyprovider/openbao/client.go
cipherboy: I guess the one counterargument could be, if a future hypothetical update were to be created wherein either `datakey/<plaintext>/<key>` or `decrypt/<key>` had alternative mode with path suffix (e.g., `datakey/<plaintext>/<key>/<mode>` or `decrypt/<key>/<mode>`), URL encoding wouldn't make any invalid-url-component-key-names magically work, but would ensure that it gets restricted to only a single path segment. :shrug:  I don't think this type of URL scheme is likely to exist though.

cipherboy (approved):
I don't think my comments about key and paths are worth holding up this PR over, looks good @ollevche! Looking forward to this feature and thanks for the collaboration :-D

internal/encryption/keyprovider/openbao/client.go
ollevche: Yes, also a future hypothetical update could change the regexp of the keys potentially introducing a bug in our system so we just *a bit safer* that way.

internal/encryption/keyprovider/openbao/config.go
janosdebugs: This is meant to be a pluggable architecture. The OpenBao provider shouldn't try to do validation for the AES-GCM method.

cipherboy: @janosdebugs Not sure that I or others believe in fully plugable architectures when it comes to cryptography. Often there's subtle issues that can arise. For a modern example, consider the [removed SHA-3 invocation](https://words.filippo.io/dispatches/mlkem768/#bonus-track-using-a-ml-kem-implementation-as-kyber-v3) in the final steps of ML-KEM (no longer providing ciphertext commitment in the shared secret compared to round-3 Kyber). If a hypothetical scheme/protocol relied implicitly on that ciphertext commitment, it would no longer have that with ML-KEM (see: lengthy discussion on IETF list about CMS and whether it needs this -- luckily many protocols don't, but it is why [XWing](https://eprint.iacr.org/2024/039.pdf) was carefully designed).

More concretely to this discussion, consider if a provider only allowed 16 byte keys, but the caller separately had requested AES-256 or ChaCha20 which require 32 bytes? If you zero-pad your key, you end up with an unexpectedly working result, with potentially [weird security properties](https://en.wikipedia.org/wiki/Weak_key). Or, while less unexpected, would you truncate a 64-byte key down to 16 or 32 for AES? Consider too some error cases (e.g., in the future, a [KEM](https://en.wikipedia.org/wiki/Key_encapsulation_mechanism) could be used to effectively generate local data keys without server input, which often has a failure mode of an all-zero shared secret -- [sometimes probabilistically even with fully-valid inputs](https://www.ietf.org/archive/id/draft-connolly-tls-mlkem-key-agreement-01.html#name-discussion)). Would you pass along this all zero key, or err out appropriately? Which layer is responsible for that, the do-AES layer or the get-a-key-from-a-provider layer? &c.

Lastly, I'm not aware of any safe ciphers in widespread use (AES, ChaCha20, or Ascon) that have a 512-bit input key.

I guess, this all to say, its a nice ideal, but tying provider closely to the symmetric algorithm (as a "bundle" perhaps), isn't the worst idea, even if you use common code between them. In that sense, removing the 512-bit recommendation would be considered safer cryptographic design.

Better, removing this option altogether and letting the choice of algorithm dictate the data key size rather than make it explicitly configurable to the user, because invalid states shouldn't be representable. My 2c.

janosdebugs (approved):

internal/encryption/keyprovider/openbao/provider.go
ollevche: Just to be clear - I plan to add context passing to the `keyprovider.Provide` call and add the timeout for all the providers so it's planned for a separate PR (to simplify merging since it affects not only Bao).

PR #1440, opened 2024-03-27
allow GCP KMS key to be set as env only

ollevche:
This PR allows GCP KMS key to be set as env only. Previously, it was required in HCL configuration.

## Target Release

1.7.0


cam72cam (approved):

Evi1Pumpkin (approved):

ollevche:
After a quick chat with @cam72cam, we decided to keep similar logic for all kms providers and not expose keys as env variables. So to keep it similar to AWS and OpenBao, I removed optional tag and reading from env.

Now GCP KMS key should be provided in HCL only.

cam72cam (approved):

cam72cam:
Good catch @ollevche !

PR #1439, opened 2024-03-27
Integrate provider functions

cam72cam:
This PR actually wires up the provider functions into the EvaluationContext and Scope.  It lazily loads providers as they are used for functions, which are terminated when the program exits.  Documentation, custom error handling, and rudimentary tests are included.

Example:
```
terraform {
        required_providers {
                myprovider = {
                        source = "registry.opentofu.org/yantrio/helpers"
                }
        }
}

output "fncall" {
        value = provider::myprovider::echo("Hello Functions!")
}
```

Resolves #1326 

## Target Release

1.7.0

Issue #1326, opened 2024-02-29
Functions in Providers


cube2222:
The terraform plugin sdk has added support for implementing functions in providers. We'll want to support these in OpenTofu.

cam72cam:
Are there any existing providers that include support for functions that can be used for testing?

janosdebugs:
Curious how this will work? [Here's the video!](https://youtube.com/shorts/4aHZjDz2VWg)

ErwinSalas:
Is someone working on this right now?

ErwinSalas:
I saw a draft PR but I'm curious if is already assigned.

cam72cam:
I've done some initial prototyping in https://github.com/opentofu/opentofu/pull/1334, being careful to only look at the terraform-plugin-go repository and @Yantrio's example provider.

The next steps for this issue are laying out what the actual technical requirements are.

Here's a summary of some of the initial thoughts I've had when chatting with @Yantrio.

* Should a provider's configuration be able to impact it's function internals?
  - The easiest path is to use a unconfigured provider instance for grpc function calls
  - It already uses pre-cached schemas in the EvalContext
  - There is new Metadata available in the grpc protocol which may allow future optimization without having the schema pre-loaded.
    - If len(metadata.functions) > 0, call getfunctions() and inject them into the context
    - If len(metadata.resources,data) > 0, call getschema()
    - We only have some vauge comments in the grpc definition to work off of and are unable to look at how HashiCorp is currently using and planning to use some of these new capabilities.
  - It is technically possible to use the above to override the provider pruning (from the graph) if the provider has functions.
  - If the providers that are not used by resorces/data but supply functions are still in the graph, they can theoretically be configured.
  - There are some weird implications around providers config blocks depending on functions that other providers define.  That is technically possible to work around, but our conclusion is to just focus on unconfigured instances until that requirement is presented as actually useful/nessesary.
* Do all of the function features exposed in the grpc calls need to be wired in for the initial version?
  - There are items like DescriptionKind (text,markdown) that are not currently useful in OpenTofu directly
  - These are comments in the above prototype
* Should the provider functions show up in `tofu metadata functions`?
* Do we think we should try to get this in to OpenTofu 1.7 and not wait for 1.8?
  - It looks like HashiCorp has stabilized the provider interface for this feature.  We don't have any insight however on if they are planning on making additional changes before they release 1.8.
[...]

PR #1444, opened 2024-03-28
Remove some unused provider schema abstractions

cam72cam:
In working on #1439, I discovered a few areas where old provider schema caches were no longer referenced or used.

Related to #1326 

## Target Release

1.7.0

### DIFF
--- a/internal/tofu/eval_context_builtin.go
+++ b/internal/tofu/eval_context_builtin.go
@@ -155,21 +155,6 @@ func (ctx *BuiltinEvalContext) Provider(addr addrs.AbsProviderConfig) providers.
 }
 
 func (ctx *BuiltinEvalContext) ProviderSchema(addr addrs.AbsProviderConfig) (providers.ProviderSchema, error) {
-	// first see if we have already have an initialized provider to avoid
-	// re-loading it only for the schema
-	p := ctx.Provider(addr)
-	if p != nil {
-		resp := p.GetProviderSchema()
-		// convert any diagnostics here in case this is the first call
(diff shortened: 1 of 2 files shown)


internal/tofu/eval_context_builtin.go
```diff
@@ -155,21 +155,6 @@ func (ctx *BuiltinEvalContext) Provider(addr addrs.AbsProviderConfig) providers.
...
-		resp := p.GetProviderSchema()
-		// convert any diagnostics here in case this is the first call
-		// FIXME: better control provider instantiation so we can be sure this
-		// won't be the first call to ProviderSchema
-		var err error
-		if resp.Diagnostics.HasErrors() {
-			err = resp.Diagnostics.ErrWithWarnings()
-		}
-		return resp, err
-	}
```
cube2222, line 171: Could you expand on why we're removing this?

Isn't this an optimization not to start new provider instances if ones already exist?

cube2222: Without this, won't we always create a new instance?

cam72cam: ctx.Plugins already interfaces with the provider schema cache.  That cache is "global" and populated any time any context starts an instance of the corresponding type.  Additionally, ctx.Plugins provides additional validation that this may short circuit.

This is an older abstraction which is made obsolete by some of the fixes made to provider schema caching.

cube2222: Alright

cube2222 (approved):

PR #1450, opened 2024-03-31
Bug Fix #1403 - Set grade and channel correctly

### LINKED ISSUE #1403: [Release Issue] GoReleaser SnapCraft Grade/Channel not correctly configured
In the 1.7.0-alpha1 release, we ran into an issue with setting the grade / channel correctly:
https://github.com/opentofu/opentofu/actions/runs/8282888828/job/22664796648

We should either use the GoReleaser includes, or go template if statements to correctly set the snapcraft channel and grade.  This is a follow up to the work done in https://github.com/opentofu/opentofu/pull/1235.

bunniseng:
Set snapcraft channel and grade based on pre-release release or stable release.

Changes
- Updated release workflow action to support release flags (prerelease and latest)
- Updated goreleaser to construct grade and channels based on workflow release flags enabled.

Resolves #1403 

## Target Release

1.7.0

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -208,8 +208,9 @@ snapcrafts:
       providers as well as custom in-house solutions.
     base: core22
     channel_templates:
-      - '{{ if eq .Env.PRERELEASE_RELEASE "true" }}edge{{ else }}stable{{ end }}'
-    grade: '{{ if eq .Env.PRERELEASE_RELEASE "true" }}devel{{ else }}stable{{ end }}'
+      - '{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}edge{{ else }}stable{{ end }}'
+      - '{{ .Major }}.{{ .Minor }}/{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}edge{{ else }}stable{{ end }}'
+    grade: '{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}devel{{ else }}stable{{ end }}'
     confinement: classic
     license: MPL-2.0
     apps:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -207,6 +207,7 @@ snapcrafts:
       safely and efficiently. OpenTofu can manage existing and popular service
       providers as well as custom in-house solutions.
     base: core22
+    grade: '{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}devel{{ else }}stable{{ end }}'
     channel_templates:
       - '{{ .Major }}.{{ .Minor }}/{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}edge{{ else }}stable{{ end }}'
       - >-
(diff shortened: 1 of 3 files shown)


bunniseng:
I was unable to perform complete snapcraft publish testing due to an issue with not being able to activate a developer account on snapcraft.io (internal server error when agreeing with T&C).

I have attached screenshots showing the different grades / channels that are being published to.

Screenshot of a prerelease publish

prerelease-publishing

Acreenshot of a stable publish

stable-publishing

cam72cam:
From @cube2222 in the original issue:
> Also, we need a separate input variable for latest, because right now any stable release gets marked as latest, even if it's a backport release.

So if we release 1.6.4 after 1.7.0 goes out, I believe this will still mark 1.6.4 as stable.

We might be able to use a channel prefix if the build is not "latest"
https://goreleaser.com/customization/snapcraft/
```
      - "{{ .Major }}.{{ .Minor }}/stable"
```

cam72cam:
This is a really good start though :+1:

bunniseng:
I will make appropriate adjustments tomorrow evening and come back to you @cam72cam, you mentioned on the issue that this is time sensitive for this week, do you have a specific day I need to work towards ?

Thanks

cam72cam:
This is looking great @bunniseng!  We have been delayed on a few tasks and will be aiming for early next week for the alpha.

bunniseng:
Great, do we want to make any changes to the docker image tags while we have time @cam72cam ?

.goreleaser.yaml
```diff
@@ -207,7 +207,14 @@ snapcrafts:
       safely and efficiently. OpenTofu can manage existing and popular service
       providers as well as custom in-house solutions.
     base: core22
-    grade: "{{ .Env.SNAPCRAFT_CHANNEL }}"
+    grade: '{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}devel{{ else }}stable{{ end }}'
+    channel_templates:
+      - '{{ .Major }}.{{ .Minor }}/{{ if eq .Env.RELEASE_FLAG_PRERELEASE "true" }}edge{{ else }}stable{{ end }}'
```
cube2222, line 212: Right now we only have the latest track, so I believe we should only be referencing stable / edge here, and that's it (which will translate to latest/stable and latest/edge).

It seems like requesting new tracks (like 1.6) is a moderately involved process, and requires a forum thread on the snapcraft forum for each.

cube2222: @bunniseng

cube2222:
Thanks a lot for working on this! Left one comment, but looks good overall!

.goreleaser.yaml
bunniseng: Okay, can you confirm with me the requirements as I understand it in your comments.

We want to release to the `latest/edge` channel when the "Release as prerelease" option has been selected otherwise release to the `latest/stable` channel.

This also means we will no longer have the need for the additional input "Release as latest" for now and it can be removed.

@cube2222

cube2222: I think in both cases we also need "Release as latest" to be true, so I'd keep it.

E.g. if someone releases 1.6.3 when 1.7.0 is out already, it's not latest, so it doesn't go to snapcraft at all.

bunniseng: Okay I think I understand, can you look over the scenarios below and confirm.

Given: "release as prerelease" and "release as latest" is selected
When: the release pipeline runs
Then: the snap is published to the `latest/edge` channel

Given: "release as latest" is selected
When: the release pipeline runs
Then: the snap is published to the `latest/stable` channel

Given: "release as latest" is not selected
When the release pipeline runs
Then: the snap is not published

cam72cam:
@bunniseng it looks like you may have a bad rebase or merge in your branch?

bunniseng:
@cam72cam : Yep looked like it, fixed.

cube2222 (approved):
This looks excellent, thanks a lot for working on this!

cam72cam (approved):

cam72cam:
Thank you for your patience @bunniseng, we have had a very scattered week!

PR #1455, opened 2024-04-02
HTTP backend client cleanup

ash2k:
1. Don't need to close response body twice
2. Pass data as `[]byte`, no need to wrap in a reader

This is a copy of https://github.com/hashicorp/terraform/pull/24639.

## Target Release

1.7.0

### DIFF
--- a/internal/backend/remote-state/http/client.go
+++ b/internal/backend/remote-state/http/client.go
@@ -42,14 +42,13 @@ type httpClient struct {
 }
 
 func (c *httpClient) httpRequest(method string, url *url.URL, data *[]byte, what string) (*http.Response, error) {
-	// If we have data we need a reader
-	var reader io.Reader = nil
+	var body interface{}
 	if data != nil {
-		reader = bytes.NewReader(*data)
+		body = *data
 	}
 
 	// Create the request
-	req, err := retryablehttp.NewRequest(method, url.String(), reader)
+	req, err := retryablehttp.NewRequest(method, url.String(), body)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to make %s HTTP request: %w", what, err)
 	}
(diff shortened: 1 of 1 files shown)


cam72cam:
Thanks for the PR @ash2k, we'll take a look over the next day or so!

cam72cam (dismissed):

internal/backend/remote-state/http/client.go
```diff
@@ -42,14 +42,13 @@ type httpClient struct {
 }
 
 func (c *httpClient) httpRequest(method string, url *url.URL, data *[]byte, what string) (*http.Response, error) {
-	// If we have data we need a reader
-	var reader io.Reader = nil
+	var body interface{}
 	if data != nil {
-		reader = bytes.NewReader(*data)
+		body = *data
```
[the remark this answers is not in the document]
ash2k: The code works differently in case of no body vs empty body. They are not the same thing at the moment. On the other hand, it always sets `Content-Type` to `application/json` and empty body is not a valid JSON, so this is unlikely a breaking change. Made the change.

internal/backend/remote-state/http/client.go
```diff
@@ -59,12 +58,12 @@ func (c *httpClient) httpRequest(method string, url *url.URL, data *[]byte, what
 	}
 
 	// Work with data/body
-	if data != nil {
+	if len(data) > 0 {
```
janosdebugs, line 61: I wonder if this is an equivalent change. `data != nil` still allows an empty slice to be passed in, while the new code doesn't. I wonder if anything relies on this behavior.

internal/backend/remote-state/http/client.go
```diff
@@ -59,12 +58,12 @@ func (c *httpClient) httpRequest(method string, url *url.URL, data *[]byte, what
...
 		req.Header.Set("Content-Type", "application/json")
-		req.ContentLength = int64(len(*data))
+		req.ContentLength = int64(len(data))
```
ollevche, line 63: We can skip it since `ContentLength` is also set in `NewRequest`.

internal/backend/remote-state/http/client.go
ash2k: Please see the discussion above. This is a private method and I've checked that nothing actually relies on that behavior.

cam72cam (approved):

PR #1458, opened 2024-04-03
Add unencrypted Method for migrations

### LINKED ISSUE #1365: [State Encryption] Syntax for migrating from unencrypted state/plan

cam72cam:
Resolves #1365 #1366 

## Target Release

1.7.0

### DIFF
--- a/internal/encryption/base.go
+++ b/internal/encryption/base.go
@@ -149,7 +149,7 @@ func (s *baseEncryption) decrypt(data []byte, validator func([]byte) error) ([]b
 				return data, nil
 			}
 		}
-		return nil, fmt.Errorf("decrypted payload provided without unencrypted method")
+		return nil, fmt.Errorf("encountered unencrypted payload without unencrypted method configured")
 	}
 
 	if es.Version != encryptionVersion {
--- a/internal/encryption/base.go
+++ b/internal/encryption/base.go
@@ -12,6 +12,7 @@ import (
 	"github.com/opentofu/opentofu/internal/encryption/config"
 	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
 	"github.com/opentofu/opentofu/internal/encryption/method"
+	"github.com/opentofu/opentofu/internal/encryption/method/unencrypted"
 
 	"github.com/hashicorp/hcl/v2"
 )
@@ -67,6 +68,7 @@ func newBaseEncryption(enc *encryption, target *config.TargetConfig, enforced bo
 	//
 	methods, diags := base.buildTargetMethods(base.encMeta)
 	base.encMethods = methods
+
 	return base, diags
 }
 
@@ -88,21 +90,12 @@ func IsEncryptionPayload(data []byte) (bool, error) {
 }
 
 func (s *baseEncryption) encrypt(data []byte, enhance func(basedata) interface{}) ([]byte, error) {
-	// No configuration provided, don't do anything
-	if s.target == nil {
-		return data, nil
-	}
+	encryptor := s.encMethods[0]
 
-	var encryptor method.Method = nil
-	if len(s.encMethods) != 0 {
-		// Use the pre-configured encryption method
-		encryptor = s.encMethods[0]
-	}
-
-	if encryptor == nil {
+	if unencrypted.Is(encryptor) {
 		// ensure that the method is defined when Enforced is true
 		if s.enforced {
-			return nil, fmt.Errorf("encryption of %q is enforced, and therefore requires a method to be provided", s.name)
+			return nil, fmt.Errorf("encryption of %q is enforced, and therefore requires a encryption method to be provided", s.name)
 		}
 		return data, nil
 	}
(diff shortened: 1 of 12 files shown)


internal/encryption/base.go
```diff
@@ -88,21 +90,12 @@ func IsEncryptionPayload(data []byte) (bool, error) {
...
-	if s.target == nil {
-		return data, nil
-	}
-
-	var encryptor method.Method = nil
-	if len(s.encMethods) != 0 {
-		// Use the pre-configured encryption method
-		encryptor = s.encMethods[0]
-	}
+	encryptor := s.encMethods[0]
```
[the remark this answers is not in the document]
cam72cam: Added a comment explaining why not

ollevche (dismissed):
Nice work, thank you!

The approach with unencrypted method is neat 🔥

internal/encryption/base.go
```diff
@@ -149,20 +140,16 @@ func (s *baseEncryption) decrypt(data []byte, validator func([]byte) error) ([]b
...
+		for _, method := range s.encMethods {
+			if unencrypted.Is(method) {
+				if s.enforced {
+					return nil, fmt.Errorf("unable to use unencrypted method when enforced = true")
+				}
 				return data, nil
 			}
 		}
-		return data, fmt.Errorf("decrypted payload provided without fallback specified")
+		return nil, fmt.Errorf("decrypted payload provided without unencrypted method")
```
cube2222, line 152: I think we should improve this error message. As a user I wouldn't understand it (I almost didn't understand it now).

How about: "encountered unencrypted payload without unencrypted method configured"

internal/encryption/method/unencrypted/method.go
```diff
@@ -0,0 +1,38 @@
...
+	return new(unenc), nil
+}
+
+type unenc struct{}
+
+func (a *unenc) Encrypt(data []byte) ([]byte, error) {
+	panic("Placeholder for type check!  Should never be called!")
+}
+func (a *unenc) Decrypt(data []byte) ([]byte, error) {
+	panic("Placeholder for type check!  Should never be called!")
```
cube2222, line 32: I'm missing something. Why are we using unencrypted.Is everywhere instead of just providing noop methods here that return what they receive without changes?

cube2222: Why does this need to be a special case other than some select validations.

cam72cam: To setup decryption methods, you need the embedded metadata.  With an unencrypted payload, you don't have any metadata and can't setup the methods.

It's sort of a chicken and egg problem that can't really be handled outside of base.go

cube2222: Ah, fair enough.

cube2222 (dismissed):

internal/encryption/base.go
```diff
@@ -88,21 +90,14 @@ func IsEncryptionPayload(data []byte) (bool, error) {
...
+	// buildTargetMethods above guarantees that there will be at least one encryption method.  They are not optional in the common target
+	// block, which is required to get to this code.
+	encryptor := s.encMethods[0]
 
-	if encryptor == nil {
+	if unencrypted.Is(encryptor) {
 		// ensure that the method is defined when Enforced is true
 		if s.enforced {
-			return nil, fmt.Errorf("encryption of %q is enforced, and therefore requires a method to be provided", s.name)
+			return nil, fmt.Errorf("encryption of %q is enforced, and therefore requires an encryption method to be provided", s.name)
```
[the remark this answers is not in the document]
cam72cam: Done!

Yantrio (approved):
It would be nice if we could also update the internal markdown documentation too.

Yantrio (approved):

PR #1479, opened 2024-04-10
fix: close manifest file

testwill:
close manifest file

Resolves #

## Target Release

1.7.0


ollevche (approved):
Good catch! Thank you for the contribution @testwill!

cam72cam (approved):

PR #1486, opened 2024-04-12
Rewrite outdated doc of function call with sensitive data

chenzejun:
During I working on #1485, I tried a lot function calls to reproduce the conservative behavior described in doc but failed, the example mayab outdated, and I tend to remove it.

```
locals {
  baz = {
    "a" = sensitive("cat")
    "b" = "dog"
  }
}
```
```
> local.baz
{
  "a" = (sensitive value)
  "b" = "dog"
}
> keys(local.baz)
[
  "a",
  "b",
]
>
```

related to #1485 

## Target Release

1.7.0


cam72cam (dismissed):

cam72cam:
Perhaps we should instead rewrite this block to specify that functions keep track of what values are sensitive from parameter to result.  Using keys() and values() as examples?

chenzejun:
> Perhaps we should instead rewrite this block to specify that functions keep track of what values are sensitive from parameter to result. Using keys() and values() as examples?

Sure, that's way better, PTAL again

cam72cam (approved):

Yantrio (approved):

ollevche (approved):

PR #1488, opened 2024-04-14
Allowed variable to pass inside `variables` block

### LINKED ISSUE #1213: Testing Feature - Allow to reference variables inside the `variables` block of a test file
### OpenTofu Version

```shell
1.6.0
```

### Use Cases

When defining default variables for the test runs, it would be useful to be able to use the actual variables for defining the defaults for the test runs

### Attempted Solutions

Currently, it's not possible to use `var`s in a top-level `variables` block in a test file

### Proposal

Support using `vars` in a top level `variables` block in a test file

main.tf:
```hcl
variable "var1" {
  default = true
}

variable "var2" {
}

resource "null_resource" "null" {

}

output "sss" {
  value = var.var2
}
```

main.tftest.hcl:
```hcl
variables {
  var2 = var.var1 ? "true" : "false"
}

run "first" {
  assert {
    condition = output.sss == false
    error_message = "Should work"
  }
}
```

### References

 - Related closed issue - https://github.com/opentofu/opentofu/issues/1091

siddharthasonker95:
Objective: Allow to reference variable inside the `variables` block of a test file.

Changes:
- Updated getEvalContextFromStates() to include config to create EvalContext for vars.
- Added relevant test cases.

Resolves #1213 

## Target Release

1.7.0


siddharthasonker95:
The fix in this PR has similar changes as mentioned in this [Added support to allow provider blocks in tests to access variables and outputs](https://github.com/opentofu/opentofu/pull/1347#top). I will update this [PR](https://github.com/opentofu/opentofu/pull/1347#top) once the above PR gets merged and include only the test case in this PR to resolve the issue.

internal/command/test.go
```diff
@@ -1065,8 +1066,18 @@ func getEvalContextFromStates(states map[string]*TestFileState) *hcl.EvalContext
 		}
 		runCtx[state.Run.Name] = cty.ObjectVal(outputs)
 	}
-	ctx := &hcl.EvalContext{Variables: map[string]cty.Value{"run": cty.ObjectVal(runCtx)}}
 
+	varCtx := make(map[string]cty.Value)
+	for _, variable := range config.Module.Variables {
+		varCtx[variable.Name] = variable.Default
+	}
```
RLRabinowitz, line 1073: This way we **only** use the default value of a variable. If a variable was explicitly set in the `tofu test` run, via a tfvars file of via a `TF_VAR_var1` environment variable, this value is completely ignored, and only the default value is used.

We should change this behaviour to allow for `TF_VAR_var1=false tofu test` to be passed down to the `variables` block in `var.` references

RLRabinowitz (changes requested):
Hey @siddharthasonker , sorry if the issue wasn't clear or explicit enough, but we shouldn't **only** support default values for variables

Please see the comment that I wrote in this review. We should allow for variable values to be provided in other manners, like tfvars file and `TF_VAR_` variables

internal/command/test.go
```diff
@@ -1038,21 +1041,16 @@ func buildInputVariablesForTest(run *moduletest.Run, file *moduletest.File, conf
...
-//
-// Parameters:
-//   - states: A map of TestFileState instances containing the state information.
-//
-// Returns:
-//   - *hcl.EvalContext: The constructed HCL evaluation context.
-func getEvalContextFromStates(states map[string]*TestFileState) *hcl.EvalContext {
+// getEvalContextFromStates constructs an hcl.EvalContext based on the provided map of
+// TestFileState instances and configuration. It extracts the relevant information from
+// the input parameters to create a context suitable for HCL evaluation.
```
siddharthasonker95, line 1051: I have refactored the code comment and removed parameters and return statements, as the variables themselves are self-explanatory.

internal/command/test.go
siddharthasonker95: I have revised the code and shared my approach [here.](https://github.com/opentofu/opentofu/pull/1488#issuecomment-2117165831)

internal/command/test.go
```diff
@@ -1065,9 +1066,31 @@ func getEvalContextFromStates(states map[string]*TestFileState) *hcl.EvalContext
...
+				val, valDiags := vv.ParseVariableValue(variable.ParsingMode)
+				if valDiags == nil {
+					varCtx[variable.Name] = val.Value
+					continue
+				}
+				diags = diags.Append(valDiags)
+			}
+		}
+		varCtx[variable.Name] = variable.Default
+	}
```
RLRabinowitz, line 1085: Could we use an abstraction like `prepareInputVariablesForAssertions` here? Looks like it does something very similar (populating variables and default values for test assertions)

It seems that you could extract a chunk of the `prepareInputVariablesForAssertions` code and use it here, instead of reimplementing it here

internal/command/test_test.go
```diff
@@ -1132,72 +1132,80 @@ Condition expression could not be evaluated at this time.
...
+			}
 
-OpenTofu has compared your real infrastructure against your configuration and
-found no differences, so no changes are needed.
+			command := &TestCommand{
+				Meta: meta,
+			}
 
-Success! 2 passed, 0 failed.
-`
```
RLRabinowitz, line 1191: Any reason why the current "expected" value of the `pass_with_local_variable` test has been reduced and does not contain the entire output, as it used to before this PR?

internal/command/testdata/test/pass_var_inside_variables/main.tftest.hcl
```diff
@@ -0,0 +1,10 @@
+variables {
+  var2 = var.var1 ? "true" : "false"
+}
```
RLRabinowitz, line 3: I'd suggest to add another test - We'd want one test checking an explicit value, and one test checking a default value

RLRabinowitz (commented):
Overall looks good 💪🏻 

Left a few comments, but after dealing with them we should be good to go

internal/command/testdata/test/pass_var_inside_variables/main.tftest.hcl
siddharthasonker95: sure, I have the added the test for default value in latest [commit](https://github.com/opentofu/opentofu/pull/1488/commits/0777ea9d965f9269039da0ad55654637576eb30d)

internal/command/test.go
siddharthasonker95: I have made the requested changes in latest [commit.](https://github.com/opentofu/opentofu/pull/1488/commits/0777ea9d965f9269039da0ad55654637576eb30d)

RLRabinowitz (approved):

internal/command/test.go
```diff
@@ -731,7 +734,10 @@ func (runner *TestFileRunner) plan(config *configs.Config, state *states.State,
 	references, referenceDiags := run.GetReferences()
 	diags = diags.Append(referenceDiags)
 
-	variables, variableDiags := buildInputVariablesForTest(run, file, config, runner.Suite.GlobalVariables, runner.States)
+	evalCtx, ctxDiags := getEvalContextForTest(runner.States, config, runner.Suite.GlobalVariables)
```
Evi1Pumpkin, line 737: Just making sure - I saw you added this call to the `runner.plan` and `runner.destory`, and I was wondering if we also need it in the `runner.apply`? 
Or maybe we don't need it because in the `runner.apply` stage we are relying on the plan that was created using the `runner.plan`, and the variables are already declared as part of the plan.

internal/command/test.go
```diff
@@ -1019,9 +1024,10 @@ func buildInputVariablesForTest(run *moduletest.Run, file *moduletest.File, conf
 		if file != nil {
 			if expr, exists := file.Config.Variables[name]; exists {
 				// If it's not set locally, it maybe set for the entire file.
-				variables[name] = unparsedVariableValueExpression{
+				variables[name] = testVariableValueExpression{
```
Evi1Pumpkin, line 1027: I didn't really understand this change, can you explain what happened before this change and what happens now?

Evi1Pumpkin (approved):
Splendid job! This is an awesome feature 💪

internal/command/test.go
siddharthasonker95: Yes, correct. In the case of `runner.apply`, we are using variables from the [plan](https://github.com/opentofu/opentofu/blob/7a713ccd833c10a315e2397da80923215da45332/internal/command/test.go#L504) that is part of [runner.plan](https://github.com/opentofu/opentofu/blob/7a713ccd833c10a315e2397da80923215da45332/internal/command/test.go#L723).

PR #1491, opened 2024-04-15
WIP: Allow configured providers to provide additional functions.

cam72cam:
Prior to this change a single unconfigured lazy provider instance was used per provider type to supply functions.  This used the functions provided by GetSchema.

With this change, provider function calls are detected and supplied via GraphNodeReferencer and are used in the ProviderTransformer to add dependencies between referencers and the providers that supply their functions.

With that information EvalContextBuiltin can now assume that all providers that require configuration have been configured by the time a particular scope is requested.  It uses either the configured instance or a unconfigured instance to supply the FunctionCall implementation.

At a high level, it allows providers to dynamically register functions based on their configurations.

main.lua
```lua
function main_a( input )
    local animal_sounds = {
        cat = 'meow',
        dog = 'woof',
        cow = 'moo'
    }
    return animal_sounds
end
```
main.tf
```hcl
terraform {
  required_providers {
    tester = {
      source  = "terraform.local/local/testfunctions"
      version = "0.0.1"
    }
  }
}

provider "tester" {
        lua = file("./main.lua")
}

output "test" {
        value = provider::tester::main_a(tomap({"foo": {"bar": 190}}))
}
```

Output:
```
Changes to Outputs:
  + test = {
      + cat = "meow"
      + cow = "moo"
      + dog = "woof"
    }
```

TODO: Testing, potential dynamic block integration

Resolves #TODO

## Target Release

1.7.0


internal/addrs/provider_function.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+)
+
+// ProviderFunction is the address of a provider defined function.
+type ProviderFunction struct {
+	referenceable
+	Name     string
+	Alias    string
```
Yantrio, line 11: tiny nitpick: can we call this ProviderAlias or something, so that nobody thinks it's referencing a function alias?

internal/lang/eval_test.go
```diff
@@ -879,21 +879,9 @@ func Test_enhanceFunctionDiags(t *testing.T) {
 		},
 		{
 			"Broken prefix",
-			"attr = magic::foo::bar::extra::missing_function(54)",
+			"attr = magic::foo::bar::extra::extra2::missing_function(54)",
 			"Invalid function format",
-			"Expected provider::<provider_name>::<function_name>, instead found \"magic::foo::bar::extra::missing_function\"",
-		},
-		{
```
Yantrio, line 886: how come we're deleting these tests?

Yantrio (dismissed):
Overall I like it. I'm tempted to say that we should have some sort of `ProviderFunctionFromString(input string) ProviderFunction` method that parses the addr, and then from there we can have a `IsCore()` method.

That could help the readability of the code.

LGTM so far, once the TODOs are done I'll be happy to review again. (Also we could do with some more tests)

internal/addrs/parse_ref.go
```diff
@@ -354,6 +353,32 @@ func parseRef(traversal hcl.Traversal) (*Reference, tfdiags.Diagnostics) {
...
+					Function: sp[2],
+				}
+			} else if len(sp) == 4 {
+				pf = ProviderFunction{
+					Name:     sp[1],
+					Alias:    sp[2],
+					Function: sp[3],
+				}
+			} else {
+				panic(root)
```
cam72cam, line 374: should be diags instead

internal/lang/references.go
```diff
@@ -83,5 +87,23 @@ func ReferencesInExpr(parseRef ParseRef, expr hcl.Expression) ([]*addrs.Referenc
...
+	}
 	return References(parseRef, traversals)
 }
+
+func filterProviderFunctions(funcs []hcl.Traversal) []hcl.Traversal {
+	pfuncs := make([]hcl.Traversal, 0, len(funcs))
+	for _, fn := range funcs {
+		if len(fn) > 0 {
+			if root, ok := fn[0].(hcl.TraverseRoot); ok {
+				if strings.HasPrefix(root.Name, "provider::") {
```
cube2222, line 102: Nit: we might want to pull out this check to a functon, we use it in many places.

internal/tofu/node_provider.go
```diff
@@ -41,7 +41,11 @@ func (n *NodeApplyableProvider) Execute(ctx EvalContext, op walkOperation) (diag
 	switch op {
 	case walkValidate:
 		log.Printf("[TRACE] NodeApplyableProvider: validating configuration for %s", n.Addr)
-		return diags.Append(n.ValidateProvider(ctx, provider))
+		diags = diags.Append(n.ValidateProvider(ctx, provider))
+		if diags.HasErrors() {
+			return diags
+		}
+		fallthrough
```
cube2222, line 48: How about, instead of fallthrough, moving walkValidate to the below case, and then have an `if` at the beginning. Fallthrough is easy to break in unexpected ways when moving code around.

internal/tofu/transform_provider.go
```diff
@@ -161,6 +161,52 @@ func (t *ProviderTransformer) Transform(g *Graph) error {
 			// Direct references need the provider configured as well as initialized
 			needConfigured[absPc.String()] = absPc
 		}
+		// Provider function references
```
cube2222, line 164: I think it would be good to expand this comment to explain that resources/datasources are referencers, and they might use functions in their attributes, and this will then make sure we have edges between the resource/datasource and the provider whose function they use.

cube2222:
WIP review: looking good!

internal/lang/eval_test.go
cam72cam: The functionality no longer exists within eval.go.  It will be hit much earlier on in context tests (which are currently commented out)

internal/tofu/context_plugins.go
```diff
@@ -22,50 +21,14 @@ import (
...
-		// paths which build a tofu.Context.
-		schema, err := cp.ProviderSchema(addr)
-		if err != nil {
-			return nil, err
-		}
-
-		funcs[addr] = providerFunctions(addr, schema.Functions, factory)
-	}
-	return funcs, nil
+	}, nil // TODO remove error from this function call!
```
cam72cam, line 31: This will be done in a followup PR to minimize churn.

internal/tofu/transform_provider.go
```diff
@@ -32,8 +32,10 @@ func transformProviders(concrete ConcreteProviderNodeFunc, config *configs.Confi
 		&ProviderTransformer{
 			Config: config,
 		},
+		// After schema transformer, we can add function references
+		//  &ProviderFunctionTransformer{Config: config},
 		// Remove unused providers and proxies
-		&PruneProviderTransformer{},
+		//  &PruneProviderTransformer{},
```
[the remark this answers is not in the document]
cam72cam: It's to show that logically these transformers must be run after, but are not part of the main group due to dependencies within the transformer chain.

internal/lang/eval.go
```diff
@@ -224,32 +219,23 @@ func (s *Scope) enhanceFunctionDiags(diags hcl.Diagnostics) hcl.Diagnostics {
...
+					// complete mismatch or invalid prefix
+					enhanced.Summary = "Invalid function format"
+					enhanced.Detail = err.Error()
+				}
 			} else {
-				// Func not in provider
-				enhanced.Summary = "Function not found in provider"
-				enhanced.Detail = fmt.Sprintf("Function %q was not registered by provider named %q of type %q", funcName, providerName, addr)
+				enhanced.Summary = "Unknown function namespace"
+				enhanced.Detail = fmt.Sprintf("Function %q does not exist within any known namespace", fn)
```
cube2222, line 236: Nit: Shouldn't this be more like "The namespace of function %q does not exist" or something like that?

Cause the current frame is "within any known namespace" even though I believe it **contains** the namespace, so it's not like you're iterating over namespaces to check if it exists anywhere.

internal/tofu/context_functions.go
```diff
@@ -3,55 +3,132 @@ package tofu
...
+	provider := ctx.Provider(absPc)
 
+	if provider == nil {
+		// Configured provider (NodeApplyableProvider) not required via transform_provider.go.  Instead we should use the unconfigured instance (NodeEvalableProvider) in the root.
+
+		// Make sure the alias is valid
+		validAlias := pf.Alias == ""
+		if !validAlias {
+			for _, alias := range pr.Aliases {
+				println(alias.Alias)
```
cube2222, line 47: leftover println

internal/tofu/context_functions.go
```diff
@@ -3,55 +3,132 @@ package tofu
...
+				if alias.Alias == pf.Alias {
+					validAlias = true
+					break
+				}
+			}
+			if !validAlias {
+				return nil, diags.Append(&hcl.Diagnostic{
+					Severity: hcl.DiagError,
+					Summary:  "Unknown function provider",
+					Detail:   fmt.Sprintf("Provider %q does not have alias %q", pf.Name, pf.Alias),
```
cube2222, line 57: Nit: No provider %q instance with alias %q

or alternatively

does not have **an instace with** alias

internal/tofu/context_functions.go
```diff
@@ -3,55 +3,132 @@ package tofu
...
+					Subject:  rng.ToHCL().Ptr(),
+				})
+			}
+		}
+
+		provider = ctx.Provider(addrs.AbsProviderConfig{Provider: pr.Type})
 		if provider == nil {
-			log.Printf("[TRACE] tofu.contextFunctions: Initializing function provider %q", addr)
-			provider, err = factory()
+			// This should not be possible
```
cube2222, line 65: Should we add "BUG: " to the error message in that case?

internal/tofu/context_functions.go
```diff
@@ -3,55 +3,132 @@ package tofu
...
+	if schema.Diagnostics.HasErrors() {
+		return nil, schema.Diagnostics
+	}
+	spec, ok := schema.Functions[pf.Function]
+	if !ok {
+		// During the validate operation, providers are not configured and therefore won't provide
+		// a comprehensive GetFunctions list
+		// Validate is built around unknown values already, we can stub in a placeholder
+		if op == walkValidate {
+			// Configured provider functions are not available during validate
```
cube2222, line 86: Do I understand correctly that this will actually only happen for **dynamically added** functions of configured providers, and will work fine for the ones that statically resolve during getschema?

internal/tofu/context_functions.go
```diff
@@ -3,55 +3,132 @@ package tofu
...
+			return nil, specs.Diagnostics
+		}
+
+		// If the function isn't in the custom GetFunctions list, it must be undefined
+		spec, ok = specs.Functions[pf.Function]
+		if !ok {
+			return nil, diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Function not found in provider",
+				Detail:   fmt.Sprintf("Function %q was not registered by provider%q", pf.Function, absPc.String()),
```
cube2222, line 116: Missing space between provider and %q?

internal/tofu/transform_provider.go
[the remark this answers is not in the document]
cube2222: I would suggest adding something like what you wrote now to the comment.

That is, "Please use these in the following way" etc.

internal/tofu/transform_provider.go
```diff
@@ -245,6 +247,127 @@ func (t *ProviderTransformer) Transform(g *Graph) error {
 	return diags.Err()
 }
 
+// ProviderFunctionTransformer is a GraphTransformer that maps resources to providers
+// within the graph. This will error if there are any resources that don't map
```
cube2222, line 251: I would add that it only does so for resources that use functions from the given provider.

cube2222 (dismissed):
Left a couple nit comments.

LGTM overall!

Yantrio (approved):
Overall it looks good to me. I have some concerns around the performance of the graph transformation stuff, but I would need to benchmark and run it to see how performant it is.

My comments are mostly small nitpicks, which I would prefer are addressed, but I'll leave that up to others to decide based on timings.

Nice work! 👍

go.sum
```diff
@@ -370,6 +370,8 @@ github.com/bmatcuk/doublestar/v4 v4.6.0/go.mod h1:xBQ8jztBU6kakFMg+8WGxn0c6z1fTS
 github.com/bradleyfalzon/ghinstallation/v2 v2.1.0 h1:5+NghM1Zred9Z078QEZtm28G/kfDfZN/92gkDlLwGVA=
 github.com/bradleyfalzon/ghinstallation/v2 v2.1.0/go.mod h1:Xg3xPRN5Mcq6GDqeUVhFbjEWMb4JHCyWEeeBGEYQoTU=
 github.com/bwesterb/go-ristretto v1.2.3/go.mod h1:fUIoIZaG73pV5biE2Blr2xEzDoMj7NFEuV9ekS419A0=
+github.com/cam72cam/hcl/v2 v2.0.0-20240416130056-03228b26f391 h1:X8ZCyI4kvTg3sm0colNG6x+rz4Ntb76ZoFwUyiuE3k0=
```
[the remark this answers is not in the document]
Yantrio: I will eat some humble pie here: I've now read the PR description where you discuss this

> As this may take some time for the HCL Team to review, I will likely move the forked code under the OpenTofu umbrella before this is merged.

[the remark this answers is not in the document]
cube2222: Yeah, let's please move it to the opentofu org.

internal/tofu/context_functions.go
cam72cam: Correct!

cube2222 (dismissed):
LGTM!

internal/tofu/transform_provider.go
```diff
@@ -255,24 +255,35 @@ type ProviderFunctionTransformer struct {
...
+	if t.Config == nil {
+		// This is probably a test case, inherited from ProviderTransformer
+		log.Printf("[WARN] Skipping provider function transformer due to missing config")
+		return nil
+	}
+
+	// Locate all providers in the graph
+	providers := providerVertexMap(g)
+
+	// LuT of (modulepath, providername) -> provider vertex
```
cube2222, line 269: What is LuT?

cube2222 (approved):

PR #1492, opened 2024-04-15
Support `for_each` syntax in `import` block

### LINKED ISSUE #1033: Add support for `for_each` in the `import` block

Evi1Pumpkin:
Resolves #1033

Plan for implementing `for_each` for the `import` block:
1. Prepare `ImportTarget` types and codebase for more dynamic addresses (addresses that are not fully known from the get-go) ✅
2. Allow for having a more dynamic address in the `to` field of an `import` block ✅
3. Add `for_each` field for `import` block, and adapt the code for resolving multiple import targets from the same import configuration   **<--- You are here 🚜**

## What this PR does
This PR is adding support for the `for_each` syntax in `import` blocks:
1. I added the `for_each` attribute to the `import` block (as optional).
2. Changed `ImportResolver` to expand imports if they include a for_each attribute and ensure import 'id' and 'to' have access to the 'for_each' key and value during their evaluation.
3. I used a similar technique to `dynamic` blocks when it comes to resolving the for_each `each.key` and `each.value`. We are adding the `RepetitionData` (a single combination of for_each key and value) to the evaluation scope, so they'll be known as part of the evaluation.
4. I added a new `Reference` of the import for_each to `NodeAbstractResource`, to make sure the evaluation will run in the correct dependency order.
5. To keep compatibility, I added support for the `tuple` type in for_each. It's only accessible while using `import` blocks with tor_each.

## Target Release

1.7.0


internal/tofu/context_plan2_test.go
```diff
@@ -4533,6 +4533,191 @@ import {
 	}
 }
 
+func TestContext2Plan_importForEach(t *testing.T) {
```
Evi1Pumpkin, line 4536: I didn't include "unhappy" tests like having a null key/value, sensitive key/value, or unknown key/value, as they fall under existing tests in either `eval_for_each_test.go` `TestEvaluateForEachExpression_errors()` and `context_plan2_test.go` `TestContext2Plan_importToInvalidDynamicAddress()`.
I tested them all manually with all possible for_each types (`map`, `object`, `set`, `tuple`), but didn't want to bloat our existing test cases with duplicated cases. Let me know if you still think we should add these tests for the for_each scenario.

Yantrio: I think we should add them, mainly because it would help us identify regressions in the future, especially if eval_for_each_test.go somehow gets refactored. I would rather have some duplication of tests if it gives us faith that we're testing properly.

internal/tofu/context_import.go
```diff
@@ -152,6 +192,13 @@ func (ri *ImportResolver) ResolveImport(importTarget *ImportTarget, ctx EvalCont
 		ID:     importId,
 	}
 
+	if keyData == EvalDataForNoInstanceKey {
+		log.Printf("[TRACE] importResolver: resolved an expanded import target %s", importAddress)
+	} else {
+		log.Printf("[TRACE] importResolver: resolved a singular import target %s", importAddress)
+
+	}
```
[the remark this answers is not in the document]
Evi1Pumpkin: Yes, this is a mistake on my part, you got it right 🙃 
Fixed it https://github.com/opentofu/opentofu/pull/1492/commits/1a192d39c608827f79acd7ccedce1e0a996ad389

internal/tofu/context_plan2_test.go
cube2222: Though I think this doesn't block the PR from the beta, and we can possibly add it in a follow-up PR (if it's a lot of work; if it's not let's just do it).

internal/tofu/eval_context.go
```diff
@@ -124,14 +124,20 @@ type EvalContext interface {
 	// evaluating. Set this to nil if the "self" object should not be available.
 	EvaluateExpr(expr hcl.Expression, wantType cty.Type, self addrs.Referenceable) (cty.Value, tfdiags.Diagnostics)
 
+	// EvaluateExprWithRepetitionData takes the given HCL expression and evaluates it
+	// to produce a value. it's different from EvaluateExpr as it takes into
+	// consideration any repetition key (a single combination of each.key and
+	// each.value of a for_each argument) that should be a part of the scope.
+	EvaluateExprWithRepetitionData(expr hcl.Expression, wantType cty.Type, keyData InstanceKeyEvalData) (cty.Value, tfdiags.Diagnostics)
```
cube2222, line 131: This is not a blocker for merging, but I wonder if what we're factoring out here (two lines of code) is worth further expanding this interface.

Maybe instead we could just have a helper function evaluateExprWithRepetitionData that takes the EvalContext and all needed params as arguments? Cause this is not exposing any new functionality, it's just a deterministic combination of other interface methods, that will be identical for each implementation of this interface.

So in other words, I'd suggest turning this into a helper function, outside of this interface.

cube2222 (approved):
Nice one!

There's one thing I'm not sure about code-architecture wise, but it's not blocking. We can discuss and clean up in a follow-up PR.

internal/tofu/eval_for_each.go
```diff
@@ -86,6 +88,17 @@ func evaluateForEachExpressionValue(expr hcl.Expression, ctx EvalContext, allowU
...
+	var isAllowedType bool
+	var allowedTypesMessage string
+	if allowTuple {
+		isAllowedType = ty.IsMapType() || ty.IsSetType() || ty.IsObjectType() || ty.IsTupleType()
+		allowedTypesMessage = "map, set of strings, or a tuple"
+	} else {
+		isAllowedType = ty.IsMapType() || ty.IsSetType() || ty.IsObjectType()
+		allowedTypesMessage = "map, or set of strings"
+
+	}
```
Yantrio, line 100:
```suggestion
		allowedTypesMessage = "map, or set of strings"
	}
```

Yantrio (changes requested):

internal/tofu/eval_for_each.go
```diff
@@ -40,7 +40,9 @@ func evaluateForEachExpression(expr hcl.Expression, ctx EvalContext) (forEach ma
 
 // evaluateForEachExpressionValue is like evaluateForEachExpression
 // except that it returns a cty.Value map or set which can be unknown.
-func evaluateForEachExpressionValue(expr hcl.Expression, ctx EvalContext, allowUnknown bool) (cty.Value, tfdiags.Diagnostics) {
+// The 'allowTuple' argument is used to support evaluating for_each from tuple
+// values, and is currently supported when using for_each in import blocks.
+func evaluateForEachExpressionValue(expr hcl.Expression, ctx EvalContext, allowUnknown bool, allowTuple bool) (cty.Value, tfdiags.Diagnostics) {
```
ollevche, line 45: minor suggestion: a lot of `evaluateForEachExpressionValue` calls look like this:

```
evaluateForEachExpressionValue(expr, ctx, true, true)
```

It's not obvious what literal boolean values stand for and it could be easier to read with some constants like:

```
evaluateForEachExpressionValue(expr, ctx, unknownsAllowed, tuplesAllowed)
```

ollevche (dismissed):
Looks good to me, great work!

I left a minor suggestion regarding readability, feel free to ignore it.

internal/tofu/eval_context.go
cam72cam: Agreed

internal/tofu/eval_for_each.go
cam72cam: I agree, though would rather see this change in a followup PR.

cam72cam:
LGTM, would be happy with it as is but would prefer to try to resolve some of the above conversations before merge.

internal/tofu/eval_context.go
cube2222: Similar to how Christian did it here, actually https://github.com/opentofu/opentofu/pull/1491/files#diff-4523c066aef327f818e5e08f0699f9a8d47eee358158a2d5eba8581cb04587ccR18

Yantrio (dismissed):
Approving as my only comment is a whitespace comment. :D

internal/tofu/context_plan2_test.go
Evi1Pumpkin: It shouldn't take long, but I'll push those tests as part of a different PR since we want this one merged by tomorrow.

siddharthasonker95 (approved):
LGTM, the comments are really helpful.

internal/tofu/eval_for_each.go
Evi1Pumpkin: Done https://github.com/opentofu/opentofu/pull/1492/commits/a3dbb14e8d96501e42e40a992590e905773cd935

internal/tofu/eval_context.go
cube2222: LGTM now!

Though to be clear, I'm fine with using this function also for module for_each if you want.

cube2222 (approved):
LGTM, feel free to merge.

internal/tofu/eval_context.go
Evi1Pumpkin: I agree with this suggestion.
I moved `EvaluateExprWithRepetitionData()` to a helper function inside `eval_import.go` - [link](https://github.com/opentofu/opentofu/pull/1492/commits/b2d04184ff9323967107f1b3d89af690c0880360).
Once I extract the `EvaluateImportAddress()` logic from the `eval_context.go` (in the next PR), it'll be even cleaner as all this import logic will be contained in `eval_import.go`.

> Though to be clear, I'm fine with using this function also for module for_each if you want (I see you reverted the change there completely).
Yeah, It's really minor, but it just feels clean that everything will be contained in the `eval_import.go` file.
And it only dries up two lines so it's not really important IMO.

internal/tofu/eval_for_each.go
Evi1Pumpkin: Totally agree. I'll make sure to add this change to the follow-up PR.

PR #1493, opened 2024-04-15
Review and order locked struct fields 

### LINKED ISSUE #1482: [Refactoring] Review and order locked struct fields
As discussed in https://github.com/opentofu/opentofu/pull/1439#discussion_r1555971840, we want to reorganize field list to follow the "mutex hat" rule? I.e. placing mutexes on top of the fields they protect.

This should be reviewed throughout the codebase.

1garo:
Resolves #1482

## What this PR does

Review and apply hat rule for mutexes on struct fields all over the codebase.

## Target Release

1.7.0

---------------------------


1garo:
There are some cases where the `mutex` is not very specific on each field it is protecting, then I fell that applying the hat rule could give maybe a false impression that the mutex is specific for that field and only.

The code below is a example of that, I thought on putting mutex on top of `Outputs` but felt wrong:
```go
// ModuleState is used to track all the state relevant to a single
// module. Previous to Terraform 0.3, all state belonged to the "root"
// module.
type ModuleState struct {
	// ...
	Outputs map[string]*OutputState `json:"outputs"`
	// ...
	Resources map[string]*ResourceState `json:"resources"`
	// ...
	Dependencies []string `json:"depends_on"`

	mu sync.Mutex
}
```

cam72cam (approved):

cam72cam (approved):

cam72cam:
Thanks @1garo !

Issue #1204, opened 2024-01-29
Testing Feature - Support overrides for data sources / resources / modules


RLRabinowitz:
### OpenTofu Version

```shell
1.6.0
```

### Use Cases

Mocking can be very helpful and powerful when testing. In addition to the request to [allow for mocking providers in tests](https://github.com/opentofu/opentofu/issues/1155), we should also be able to create more granular mocks, for specific resources / modules / data sources

### Attempted Solutions

No way currently to mock specific resources / modules / data sources in test runs

### Proposal

Create `override_module`, `override_resource` and `override_data` blocks, which could be defined globally in the test file / under the `mock_provider` block / under a `run` block, and could be used to completely override the module/resource/data source.

When using an `override_module`/`override_resource`/`override_data`, this module/resource/data wouldn't actually be created using the provider, it only immediately return with the `values`/`outputs` defined in the override block

```hcl
override_resource {
  target = aws_s3_bucket.my_bucket
  values = {
    body = "my_text"
  }
}

override_data {
  target = data.aws_s3_bucket.my_bucket
  values = {
    body = "my_text"
  }
}

override_module {
  target = module.my_module
  outputs = {
    output1 = "my_output
  }
}
```

### References

- Reference to docs of a similar ability in Terraform - https://developer.hashicorp.com/terraform/language/tests/mocking#overrides
- Related to

Issue #1155, opened 2024-01-19
Support for mocking providers in opentofu test


tv-sister:
### OpenTofu Version

```shell
OpenTofu v1.6.0
```

### Use Cases

Test OpenTofu code offline.

### Attempted Solutions

Using fake credentials in your provider in the testfile does not always work, the newly implemented _mock_ provider by Terraform has a broader usecase, as it allows faking data sources etc.

https://developer.hashicorp.com/terraform/language/tests/mocking

For example, using fake credentials and trying to retrieve an AMI with

```
data "aws_ami" "ubuntu" {
  most_recent = true

  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-*"]
  }

  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }

  owners = ["099720109477"] # Canonical
}
```

Returns the following error

```
reading EC2 AMIs: AuthFailure: AWS was not able to validate the provided access credentials
```

### Proposal

Add support for mocking providers, resources, and data sources.

cube2222:
Thanks for submitting the issue @tv-sister!

We've agreed that this makes sense for compatibility purposes, and the core team will pick it up with the goal being a release of this alongside 1.7.0.

RLRabinowitz:
Also - keep this possible crash in mind - https://github.com/hashicorp/terraform/issues/34601

PR #1499, opened 2024-04-16
add override implementation for testing framework

ollevche:
Resolves https://github.com/opentofu/opentofu/issues/1204

This PR adds initial implementation for `override_resource`, `override_data` and `override_module` blocks for testing framework.

## Implementation details

Overriding data and managed resources is implemented by avoiding calls to the underlying provider methods: `ReadResource`, `PlanResourceChange`, `ApplyResourceChange`, and `ReadDataSource`. It's implemented via a wrapper of the Provider interface, which is injected in a constructor (node's `getProvider` function). The wrapper has additional knowledge of when / why it's called so it mocks needed resources / data in a format of one-time use. It will be extended later to cover `mock_provider` and `mock_resource` / `mock_data` implementation.

Overriding modules is implemented by not adding a module configuration to the DAG and making the output nodes use static values from `override_module` block instead of evaluating the original expression.

## Reasoning on testing

The implementation requires a lot of small changes across the codebase and the only large chunk of code is actual mock generation. I created an e2e test to ensure the override feature works well in general. It could be extended later to test similar behavior of `mock_providers` and `mock_resource`. Unit tests cover the mock generation mechanism to ensure attributes and blocks are generated correctly for different edge cases. Also, newly created addr parsing functions are also covered with unit tests and can be reused later on.

## TransformForTest note

I think the `TransformForTest` function is a bit hard to extend so we would need to consider minor refactoring in the case of a major extension in the future.

## Target Release

1.8.0


internal/addrs/module.go
```diff
@@ -177,6 +177,25 @@ func (m Module) configRemovableSigil() {
...
 }
 
+// ParseModule parses a module address from the given traversal,
+// which has to contain only the module address with no resource/data/variable/etc.
+// This function only supports module addresses without instance keys (as the
+// returned Module struct doesn't support instance keys) and will return an
+// error if it encounters one.
+func ParseModule(traversal hcl.Traversal) (Module, tfdiags.Diagnostics) {
+	mod, remain, diags := parseModulePrefix(traversal)
+	if len(remain) != 0 {
```
cam72cam, line 187: Perhaps also check that `!diags.HasErrors()` ?

When diags.HasErrors(), any results from a function should be considered undefined.

cam72cam: I see you did that for `ParseConfigResource` below

ollevche: Yep, missed this one, thank you! I was still getting used to diags instead of errors by the time I was writing this line :)

Fixed in ff1754d396ecae51c1ac70046647dc7b880e2e13

internal/configs/test_file.go
```diff
@@ -537,6 +654,181 @@ func decodeTestRunOptionsBlock(block *hcl.Block) (*TestRunOptions, hcl.Diagnosti
 	return &opts, diags
 }
 
+func decodeOverrideResourceBlock(block *hcl.Block, mode addrs.ResourceMode) (*OverrideResource, hcl.Diagnostics) {
```
cam72cam, line 657: It looks like there's some duplication between the next few functions that could be refactored out?

ollevche: Yep, there's an almost identical function that could be moved out. Thanks!

[8b7f372](https://github.com/opentofu/opentofu/pull/1499/commits/8b7f372415f9a4dccbfad433212f9a6a5fbaa1bf)

internal/configs/config.go
```diff
@@ -955,8 +985,193 @@ func (c *Config) TransformForTest(run *TestRun, file *TestFile) (func(), hcl.Dia
...
+		resources = append(resources, r)
+	}
+
+	return resources, diags
+}
+
+func mergeOverridenModules(runModules, fileModules []*OverrideModule) ([]*OverrideModule, hcl.Diagnostics) {
+	// runAddrs is a unique set of module addresses in run block.
+	// It's already validated for duplicates previously.
+	runAddrs := make(map[string]struct{})
```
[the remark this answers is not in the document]
ollevche: It stands for module addresses in run block. It could be `runModuleAddrs` or `addrsInRun` if just `runAddrs` is not clear

internal/configs/config.go
```diff
@@ -955,8 +985,193 @@ func (c *Config) TransformForTest(run *TestRun, file *TestFile) (func(), hcl.Dia
...
+				output.OverrideValue = nil
+			}
+		}
+	}, diags
+}
+
+func mergeOverridenResources(runResources, fileResources []*OverrideResource) ([]*OverrideResource, hcl.Diagnostics) {
+	// runAddrs is a unique set of resource addresses in run block.
+	// It's already validated for duplicates previously.
+	runAddrs := make(map[string]struct{})
```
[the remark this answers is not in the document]
ollevche: The same as in `mergeOverridenModules`: It stands for resource addresses in run block. It could be `runResourceAddrs` or `addrsInRun` if just `runAddrs` is not clear

internal/configs/resource.go
```diff
@@ -49,6 +50,14 @@ type Resource struct {
 	// If this is nil, then this resource is essentially public.
 	Container Container
 
+	// IsOverridden indicates if the resource is being overriden. It's used in
+	// testing framework to not call the underlying provider.
+	IsOverridden bool
```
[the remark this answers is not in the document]
ollevche: Yep, it could be done this way. We will need to check for a nil map value (not overridden), a zero-length map (overridden with no values), and a filled map (overridden with values). I just find it harder to understand that way from the reader's perspective.

Also, I think it's easy to misuse nil and zero-length maps this way so an additional flag comes in handy.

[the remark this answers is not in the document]
siddharthasonker95: that makes sense... just wanted to know the idea behind this.
Great work Thanks!

internal/configs/config.go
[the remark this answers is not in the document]
ollevche: 7e162882a7c5616aa71cd116cc67197bd5d997d7

internal/configs/config.go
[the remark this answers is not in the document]
ollevche: 7e162882a7c5616aa71cd116cc67197bd5d997d7

ollevche:
I purposely left some failing lint checks to discuss if we need to tweak its settings. These functions looks fine to me so I suggest to increase the limit of complexity check. @opentofu/core-engineers what do you think?

internal/configs/test_file.go
```diff
@@ -537,6 +662,154 @@ func decodeTestRunOptionsBlock(block *hcl.Block) (*TestRunOptions, hcl.Diagnosti
...
+	if attr, exists := content.Attributes["outputs"]; exists {
+		outputs, moreDiags := parseObjectAttrWithNoVariables(attr)
+		mod.Outputs, diags = outputs, append(diags, moreDiags...)
+	}
+
+	return mod, diags
+}
+
+func parseObjectAttrWithNoVariables(attr *hcl.Attribute) (map[string]cty.Value, hcl.Diagnostics) {
+	if vars := attr.Expr.Variables(); len(vars) != 0 {
```
cam72cam, line 738: Is this explicitly needed? The Expr.Value(nil) below should produce a similar diagnostic?

cam72cam:
Current Thoughts:
1) Is there address parsing code that could be re-used from import/moved/removed blocks?
2) How does this handle for_each/count interactions?  Can we get a test for this?
3) All of the new addr functions should have comprehensive unit tests.
4) The mockValueComposer needs to be moved to it's own file, possibly in the hcl2shim package where there's already similar functionality?

The module node mocking seems reasonable, but the resource node mocking feels tacked on due to all of the special casing.  I know we talked about this in the past and I think we should think about potential alternative solutions here.  We might be able to inject a custom provider that only knows how to handle that specific resource instance?  This could also be useful when we get to provider mocking.

Overall I think that this work makes sense and is really good so far!

internal/configs/test_file.go
ollevche: Agree, removed this call in c398e309df9d282bf765ba5578fef6573de032c0

ollevche:
@cam72cam thank you for your thoughts and suggestions!

1. I can't reuse public functions directly, since they are specific to the concrete block (i.e. `ParseMoveEndpoint`, `ParseRemoveEndpoint`) and also I can't reuse `ParseAbsResourceInstance` (which is used in import) because it allows instance addresses. So I added public counterparts for existing private functions (`addrs.ParseConfigResource`, `addrs.ParseModule`) to parse resource and module addresses and return existing public structs concerning their internal logic (like disallow instance address). I think this way it could be better reused in the future. WDYT?

2. We mock each existing instance of resource / data / module. If there's also values / outputs field present to use in mocking - we will use it for each instance. I'll extend the e2e test to include such a scenario. Agree.

3. In the case, that we are moving with the current implementation, I'll add unit tests for these functions. Agree.

4. Yep, the mocking code needs to be moved somewhere else. Agree.

5. Yep, we discussed the option of custom provider implementation earlier. I will reconsider this approach once more!

cam72cam (approved):

cam72cam (dismissed):

cam72cam (dismissed):

internal/configs/hcl2shim/mock_value_composer.go
```diff
@@ -0,0 +1,353 @@
...
+			))
+		}
+	}
+
+	mockAttrs := make(map[string]cty.Value)
+
+	impliedTypes := schema.ImpliedType().AttributeTypes()
+
+	for k, attr := range schema.Attributes {
+		// If the value present in configuration - just use it.
```
Evi1Pumpkin, line 94: I want to validate I understand this scenario correctly -
From what I understand if I have the resource:
```
resource "something" "hi" {
    id = "2342"
}
```
If I have an "override_resource" block for this resource, I won't be able to override the `id` attribute?

internal/configs/hcl2shim/mock_value_composer.go
```diff
@@ -0,0 +1,353 @@
...
+			mockAttrs[k] = cv
+			addPotentialDefaultsWarning(k, "The field is ignored since overriding configuration values not allowed.")
+			continue
+		}
+
+		// Non-computed attributes can't be generated
+		// so we set them from configuration only.
+		if !attr.Computed {
+			mockAttrs[k] = cty.NullVal(attr.Type)
+			addPotentialDefaultsWarning(k, "The field is ignored since overriding non-computed fields not allowed.")
```
Evi1Pumpkin, line 105:
```suggestion
			addPotentialDefaultsWarning(k, "The field is ignored since overriding non-computed fields is not allowed.")
```

internal/configs/hcl2shim/mock_value_composer.go
```diff
@@ -0,0 +1,353 @@
...
+
+	mockAttrs := make(map[string]cty.Value)
+
+	impliedTypes := schema.ImpliedType().AttributeTypes()
+
+	for k, attr := range schema.Attributes {
+		// If the value present in configuration - just use it.
+		if cv, ok := configMap[k]; ok && !cv.IsNull() {
+			mockAttrs[k] = cv
+			addPotentialDefaultsWarning(k, "The field is ignored since overriding configuration values not allowed.")
```
Evi1Pumpkin, line 97:
```suggestion
			addPotentialDefaultsWarning(k, "The field is ignored since overriding configuration values is not allowed.")
```

internal/configs/hcl2shim/mock_value_composer.go
```diff
@@ -0,0 +1,353 @@
...
+			}
+		}
+
+		// If there's no value in defaults, we generate our own.
+		v, ok := mvc.getMockValueByType(impliedTypes[k])
+		if !ok {
+			diags = diags.Append(tfdiags.WholeContainingBody(
+				tfdiags.Error,
+				"Failed to generate mock value",
+				fmt.Sprintf("Mock value cannot be generated for dynamic type. Please, specify `%v` field explicitly in configuration.", k),
```
Evi1Pumpkin, line 133:
```suggestion
				fmt.Sprintf("Mock value cannot be generated for dynamic type. Please specify the `%v` field explicitly in the configuration.", k),
```

internal/configs/test_file.go
```diff
@@ -537,6 +662,143 @@ func decodeTestRunOptionsBlock(block *hcl.Block) (*TestRunOptions, hcl.Diagnosti
...
+		return traversal, &configRes, diags
+	}
+
+	res := &OverrideResource{
+		Mode: mode,
+	}
+
+	content, diags := block.Body.Content(overrideResourceBlockSchema)
+
+	if attr, exists := content.Attributes["target"]; exists {
```
Evi1Pumpkin, line 688: Isn't the `target` attribute in the block required? Feels like we should throw an error id it's missing/

internal/configs/test_file.go
```diff
@@ -537,6 +662,143 @@ func decodeTestRunOptionsBlock(block *hcl.Block) (*TestRunOptions, hcl.Diagnosti
...
+		}
+
+		return traversal, target, diags
+	}
+
+	mod := &OverrideModule{}
+
+	content, diags := block.Body.Content(overrideModuleBlockSchema)
+
+	if attr, exists := content.Attributes["target"]; exists {
```
Evi1Pumpkin, line 723: Same as before - Isn't the `target` attribute in the block required? Feels like we should throw an error if it's missing.

ollevche: Same - Yes, it is required and we throw an error if it's missing. The check is done before this function is called.

It is marked as required in the schema: https://github.com/opentofu/opentofu/blob/override-for-testing/internal/configs/test_file.go#L893

internal/configs/test_file.go
Evi1Pumpkin: Awesome, missed that :)

internal/tofu/provider_for_test_framework.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (p providerForTest) PlanResourceChange(r providers.PlanResourceChangeRequest) providers.PlanResourceChangeResponse {
+	if r.Config.IsNull() {
+		return providers.PlanResourceChangeResponse{
+			PlannedState: r.ProposedNewState, // null
+		}
+	}
+
+	if p.plannedChange != nil {
```
Evi1Pumpkin, line 52: I don't really understand this part of the `planned change` and how it affects the overridden resource. Can you explain a little?

Evi1Pumpkin (approved):
Splendid job! This is a really useful feature 🥇

internal/tofu/provider_for_test_framework.go
ollevche: Sure!

In general, `providerForTest` uses configuration values (from `values` / `outputs` field in `override_` block) and automatically generates some missing computed values. When doing so, it uses some rules on how the values should be generated and the value is random for strings.

After this, tofu checks if the provider generated the correct response. It's the point when the logic breaks without this `if`. We need to use the same generated value for multiple sequential provider calls (i.e. plan is called twice).

Currently, `providerForTest` is meant to be used once inside the `NodeAbstractResourceInstance` so we just pass the previously planned value to the `providerForTest`.

It will be changed to the proper cache when extending the `providerForTest` to support `mock_provider` feature. This one: https://github.com/opentofu/opentofu/issues/1155

I left a comment explaining it a bit more here: https://github.com/opentofu/opentofu/blob/override-for-testing/internal/tofu/provider_for_test_framework.go#L21

Evi1Pumpkin: Thanks for the explanation, this is much clearer! 🙏

PR #1503, opened 2024-04-18
Issue #1033: Docs for for_each on import

### LINKED ISSUE #1033: Add support for `for_each` in the `import` block

ghost:
Resolves #1033

## Target Release

1.7.0

### DIFF
--- a/website/docs/language/import/index.mdx
+++ b/website/docs/language/import/index.mdx
@@ -164,6 +164,6 @@ output "id" {
 
 :::note
 
-[Generaging configuration](/docs/language/import/generating-configuration/) is currently not possible when using `for_each` on `import blocks.
+[Generating configuration](/docs/language/import/generating-configuration/) is currently not possible when using `for_each` on `import blocks.
 
 :::


Yantrio (dismissed):
I'm happy with this, I would recommend we wait for an approval from @Evi1Pumpkin too though.

Do we also have to mention tuples support?

website/docs/language/import/index.mdx
```diff
@@ -131,3 +132,32 @@ import {
   id = "i-abcd1234"
 }
 ```
+
+### Importing multiple resources
+
+You can import multiple resources with one import block by using a `for_each` expression. This expression accepts a [set or a map](/docs/language/expressions/types/) and provides the `each.key` and `each.value` variables to access the individual elements.
```
Evi1Pumpkin, line 138: We also support `tuple` in the case of for_each. Using a tuple can be helpful when importing to a more complex address like `module.mod[each.value.moduleKey].test_object.a[each.value.resourceKey]`.

website/docs/language/import/index.mdx
```diff
@@ -131,3 +132,32 @@ import {
   id = "i-abcd1234"
 }
 ```
+
+### Importing multiple resources
```
Evi1Pumpkin, line 136: Maybe we should mention that you cannot use the `Generate configuration` capability when we are importing resource using `for_each`

website/docs/language/import/index.mdx
```diff
@@ -131,3 +132,32 @@ import {
...
+resource "random_id" "test_id" {
+  byte_length = 8
+  count = 2
+}
+
+import {
+  to = random_id.test_id[tonumber(each.key)]
+  id = each.value
+  for_each = {
+    for idx, item in var.server_ids: idx => item
```
Evi1Pumpkin, line 156: This is an interesting example (I didn't even know it works), but shouldn't we use a more common way of calling for a for_each in our docs? Something more similar to [this](https://opentofu.org/docs/language/meta-arguments/for_each/#using-sets) for example.
I don't want people to be confused.

website/docs/language/import/index.mdx
Yantrio: We should open an issue for that too. It would be a shame to not support that feature.

website/docs/language/import/index.mdx
Evi1Pumpkin: Makes sense.
I used the `random_integer` resource while developing the resource. But I don't think it can be a very good example:
```
locals {
    numbers = {
      "first" = "0,1,0"
      "second" = "0,1,0"
    }
}

resource "random_integer" "a" {
  for_each = local.numbers
  min =	1
  max = 50000
}
```
Overall, I don't have a strong opinion about it, so I'm leaving that decision to you.

website/docs/language/import/index.mdx
```diff
@@ -131,3 +132,38 @@ import {
...
+}
+
+output "id" {
+  value = random_id.test_id.*.b64_url
+}
+```
+
+:::note
+
+[Generaging configuration](/docs/language/import/generating-configuration/) is currently not possible when using `for_each` on `import blocks.
```
ollevche, line 167:
```suggestion
[Generating configuration](/docs/language/import/generating-configuration/) is currently not possible when using `for_each` on `import blocks.
```

ollevche (approved):

PR #1505, opened 2024-04-18
Introduce locking on the hook mocks to avoid data race failures

### LINKED ISSUE #1504: Data race errors thrown from mock erroneously
A recent addition some locking was missing around import hooks.

https://github.com/opentofu/opentofu/actions/runs/8733823228/job/23963252522?pr=1503#step:6:153

We should fix this by introducing locking.

Yantrio:
Resolves #1504

## Target Release
1.7.0


Evi1Pumpkin (approved):
Thanks for fixing this!

PR #1520, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better

Resolves[https://github.com/opentofu/opentofu/issues/1521](url)


ChengenH:
> use errors.New to replace fmt.Errorf with no parameters will much better

PR #1524, opened 2024-04-21
Refactor: Use uuid.NewRandom() for LockInfo ID generation

### LINKED ISSUE #1523: Refactor LockInfo ID generation
Current LockInfo ID generation in `internal/states/statemgr/locker.go` uses a UUID string created in a non-cryptographically secure way.

By refactoring to use the more performant and cryptographically safe `NewRandom()` from Google's uuid module the same behavior can be achieved.

mikelolasagasti:
This commit updates the LockInfo ID generation in the statemgr package to use `uuid.NewRandom()` from the `github.com/google/uuid` package instead of the previous approach using `uuid.FormatUUID(buf)` from `github.com/hashicorp/go-uuid`. The new method generates UUIDs in a more performant manner, even though cryptographic security isn't a requirement for this specific use case. This change improves efficiency without sacrificing reliability and ensures consistency with other UUID generation practices in the codebase.

Resolves #1523

## Target Release

1.7.0

### DIFF
--- a/internal/states/statemgr/locker.go
+++ b/internal/states/statemgr/locker.go
@@ -18,7 +18,7 @@ import (
 	"text/template"
 	"time"
 
-	uuid "github.com/hashicorp/go-uuid"
+	"github.com/google/uuid"
 	"github.com/opentofu/opentofu/version"
 )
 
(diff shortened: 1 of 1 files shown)


skyzyx (commented):
Is there a good way to add a benchmark for the `NewLockInfo()` function? So that we can test the previous implementation against the current one and more adequately validate the performance improvement?

Evi1Pumpkin:
Hey @mikelolasagasti!
Thank you for your contribution. But for next time, please take a look at our [contribution guide](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md#contributing-to-opentofu) to understand the process of opening a Pull Request in the OpenTofu project.
Unfortunately, we decided not to accept this contribution because of the following reason:
1. This is, at best, a very minor improvement.
2. Currently, we don't have an easy way to test the performance improvement.
And still, we appreciate your time and effort in submitting this Pull Request!

PR #1552, opened 2024-04-25
Migration guides

ghost:
This PR adds detailed migration guides as per the [TSC decision 2024-02-17](https://github.com/opentofu/opentofu/blob/main/TSC_SUMMARY.md#2024-02-17)

### DIFF
--- a/website/docs/intro/migration/terraform-1.5-or-lower.mdx
+++ b/website/docs/intro/migration/terraform-1.5-or-lower.mdx
@@ -15,8 +15,8 @@ switching from Terraform to OpenTofu.
 
 ## Step 0: Prepare a disaster recovery plan
 
-Although OpenTofu 1.6 is very similar to Terraform 1.5.7, it is a non-trivial change affecting your infrastructure. Make
-sure you have an up to date and *tested* disaster recovery plan.
+Although OpenTofu 1.6.2 is very similar to Terraform 1.5.7, make sure you have an up to date and *tested* disaster
+recovery plan.
 
 ## Step 1: Upgrading Terraform
 
@@ -41,7 +41,7 @@ Terraform has compared your real infrastructure against your
 configuration and found no differences, so no changes are needed.
 ```
 
-## Step 3: Install OpenTofu
+## Step 3: Install OpenTofu 1.6.2
 
 As a first step, please [follow the installation instructions for the OpenTofu CLI tool](install/index.mdx). Make sure
 to install OpenTofu 1.6.2. Please test if you can successfully execute the `tofu` command and that it returns the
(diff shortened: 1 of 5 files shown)


website/docs/intro/migration/terraform-1.5-or-lower.mdx
```diff
@@ -1,119 +1,131 @@
...
+$ tofu plan
+
+...
+
+No changes. Your infrastructure matches the configuration.
+
+OpenTofu has compared your real infrastructure against your
+configuration and found no differences, so no changes are needed.
+```
+
```
cube2222, line 94: I would suggest also doing an empty tofu apply after the tofu plan, as that rewrites the statefile.

It's not strictly necessary, but safer.

cube2222: Really, I would replace step 7 with that - empty apply.

website/docs/intro/migration/terraform-1.8.mdx
```diff
@@ -0,0 +1,152 @@
...
+The OpenTofu `removed` block works differently from the Terraform variant. Please [review the documentation](../../language/resources/syntax.mdx#removing-resources) and make the following changes:
+
+1. Remove the `lifecycle` block. If you used the `lifecycle` &rarr; `destroy = true` setting, remove the entire `removed` block. Verify that the code still works as intended after the migration.
+
+### Testing changes
+
+If you are using the `terraform test` feature, you will need to perform the following changes:
+
+1. If you are using `mock_provider`, restructure your tests to work without this feature (see [OpenTofu issue #1155](https://github.com/opentofu/opentofu/issues/1155))
+2. If you are using `override_resource`, `override_data`, or `override_module`, restructure your tests to work without these features.
```
cube2222, line 87: Should link to https://github.com/opentofu/opentofu/issues/1204?

website/docs/intro/migration/terraform-1.6.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+When migrating from Terraform 1.6.x, please migrate to OpenTofu 1.6.2 first, then
+[upgrade your OpenTofu installation](../upgrading.mdx) to the latest version.
+
+OpenTofu 1.6.2 is largely compatible with Terraform 1.6.x with a few minor changes needed. This migration guide will
+take you through the process of switching from Terraform to OpenTofu.
+
+## Step 0: Prepare a disaster recovery plan
+
+Although OpenTofu 1.6.2 is very similar to Terraform 1.6.6, it is a non-trivial change affecting your infrastructure.
```
cube2222, line 18: I'm not sure about the "non-trivial" bit, especially since all over the place we're calling it a easy drop-in replacement.

WDYT about just 
> Although OpenTofu 1.6.2 is very similar to Terraform 1.6.6, make sure you have an up to date and *tested* disaster recovery plan.

But not a strong opinion, fine with keeping it this way too.

cube2222 (dismissed):

website/docs/intro/migration/terraform-1.5-or-lower.mdx
```diff
@@ -1,119 +1,131 @@
...
+...
+
+No changes. Your infrastructure matches the configuration.
+
+Terraform has compared your real infrastructure against your
+configuration and found no differences, so no changes are needed.
+```
+
+## Step 3: Install OpenTofu
```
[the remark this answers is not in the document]
Yantrio: (same comment applies to other files too), I think its best if our ToC has a rough overview that people can skim too.

Yantrio (approved):
LGTM Overall, just tiny nitpicks from me :D Nice work!

Yantrio (approved):

PR #1554, opened 2024-04-25
Add `-show-sensitive` flag to tofu plan, apply, state-show and output commands

### LINKED ISSUE #839: Add -show-sensitive flag to enable viewing sensitive data in output
### OpenTofu Version

```shell
Not applicable for feature request.
```

### Use Cases

There are multiple cases where TF knows a sensitive value but refuses to display it, such as:

- viewing TF created secret values
- approving sensitive diffs

While the default behaviour should be to hide secret values, users should be free to decide when to reveal these and should be given the option to do so.

### Attempted Solutions

There are a few workarounds to getting secrets out ot TF, but all of these are painful, especially seeing sensitive diffs for a plan:

1. Run `tofu state pull > state.json` then look through the state file for the secret value
2. Use `output` with `nonsensitive()` to print secret value
3. Save the plan into a file then write a script to make sense of it 

### Proposal

The following commands should accept a `-show-sensitive` flag, and reveal sensitive values when this flag is present:

1. tofu plan -show-sensitive
2. tofu apply -show-sensitive
3. tofu state show kubernetes_secret.foo -show-sensitive
4. tofu output -show-sensitive

### References

_No response_

siddharthasonker95:
Objective: Add support for a `-show-sensitive` flag, and reveal sensitive values when this flag is present:

`tofu plan -show-sensitive`
`tofu apply -show-sensitive`
`tofu state show kubernetes_secret.foo -show-sensitive`
`tofu output -show-sensitive`
`tofu show -show-sensitive`

Changes:
* Added `-show-sensitive` argument for plan, apply, state show, show and output commands.
* Added relevant test cases for `-show-sensitive` arg.

In this Pull Request, to display sensitive values I’m adding ShowSensitive attribute in Renderer.

Resolves #839 

## Target Release

1.9.0

### DIFF
--- a/internal/command/apply_test.go
+++ b/internal/command/apply_test.go
@@ -2190,23 +2190,8 @@ func TestApply_showSensitiveArg(t *testing.T) {
 		t.Fatalf("bad status code: \n%s", output.Stderr())
 	}
 
-	expected := `Changes to Outputs:
-  [32m+[0m[0m notsensitive = "Hello world"
-  [32m+[0m[0m sensitive    = "Hello world"
-
-You can apply this plan to save these new output values to the OpenTofu
-state, without changing any real infrastructure.
-[0m[1m[32m
-Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
-[0m[0m[1m[32m
-Outputs:
-
-[0mnotsensitive = "Hello world"
-sensitive = "Hello world"`
-
-	actual := strings.TrimSpace(output.Stdout())
-	if diff := cmp.Diff(actual, expected); len(diff) > 0 {
-		t.Fatalf("got incorrect output\n %v", diff)
+	if got, want := output.Stdout(), "sensitive    = \"Hello world\""; !strings.Contains(got, want) {
+		t.Fatalf("got incorrect output, want %q, got:\n%s", want, got)
 	}
 }
 
@@ -2231,23 +2216,8 @@ func TestApply_withoutShowSensitiveArg(t *testing.T) {
 		t.Fatalf("bad status code: \n%s", output.Stderr())
 	}
 
-	expected := `Changes to Outputs:
-  [32m+[0m[0m notsensitive = "Hello world"
-  [32m+[0m[0m sensitive    = (sensitive value)
-
-You can apply this plan to save these new output values to the OpenTofu
-state, without changing any real infrastructure.
-[0m[1m[32m
-Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
-[0m[0m[1m[32m
-Outputs:
-
-[0mnotsensitive = "Hello world"
-sensitive = <sensitive>`
-
-	actual := strings.TrimSpace(output.Stdout())
-	if diff := cmp.Diff(actual, expected); len(diff) > 0 {
-		t.Fatalf("got incorrect output\n %v", diff)
+	if got, want := output.Stdout(), "sensitive    = (sensitive value)"; !strings.Contains(got, want) {
+		t.Fatalf("got incorrect output, want %q, got:\n%s", want, got)
 	}
 }
 
--- a/internal/command/apply_test.go
+++ b/internal/command/apply_test.go
@@ -2172,7 +2172,7 @@ func TestApply_showSensitiveArg(t *testing.T) {
 	testCopyDir(t, testFixturePath("apply-sensitive-output"), td)
 	defer testChdir(t, td)()
 
-	p := applyFixtureProvider()
+	p := testProvider()
 	view, done := testView(t)
 	c := &ApplyCommand{
 		Meta: Meta{
@@ -2180,44 +2180,27 @@ func TestApply_showSensitiveArg(t *testing.T) {
 			View:             view,
 		},
 	}
+
+	statePath := testTempFile(t)
 
 	args := []string{
+		"-state", statePath,
+		"-auto-approve",
 		"-show-sensitive",
 	}
+
 	code := c.Run(args)
 	output := done(t)
 	if code != 0 {
-		t.Fatalf("bad status code: \n%s", output.Stderr())
+		t.Fatalf("bad: \n%s", output.Stderr())
 	}
 
-	if got, want := output.Stdout(), "sensitive    = \"Hello world\""; !strings.Contains(got, want) {
-		t.Fatalf("got incorrect output, want %q, got:\n%s", want, got)
+	stdout := output.Stdout()
+	if !strings.Contains(stdout, "notsensitive = \"Hello world\"") {
+		t.Fatalf("bad: output should contain 'notsensitive' output\n%s", stdout)
 	}
-}
-
-func TestApply_withoutShowSensitiveArg(t *testing.T) {
-	td := t.TempDir()
-	testCopyDir(t, testFixturePath("apply-sensitive-output"), td)
-	defer testChdir(t, td)()
-
-	p := applyFixtureProvider()
-	view, done := testView(t)
-	c := &ApplyCommand{
-		Meta: Meta{
-			testingOverrides: metaOverridesForProvider(p),
-			View:             view,
-		},
-	}
-
-	args := []string{}
-	code := c.Run(args)
-	output := done(t)
-	if code != 0 {
-		t.Fatalf("bad status code: \n%s", output.Stderr())
-	}
-
-	if got, want := output.Stdout(), "sensitive    = (sensitive value)"; !strings.Contains(got, want) {
-		t.Fatalf("got incorrect output, want %q, got:\n%s", want, got)
+	if !strings.Contains(stdout, "sensitive    = \"Hello world\"") {
+		t.Fatalf("bad: output should contain 'sensitive' output\n%s", stdout)
 	}
 }
 
--- a/internal/command/apply_test.go
+++ b/internal/command/apply_test.go
@@ -2167,6 +2167,90 @@ func TestApply_warnings(t *testing.T) {
 	})
 }
 
+func TestApply_showSensitiveArg(t *testing.T) {
+	td := t.TempDir()
+	testCopyDir(t, testFixturePath("apply-sensitive-output"), td)
+	defer testChdir(t, td)()
+
+	p := applyFixtureProvider()
+	view, done := testView(t)
+	c := &ApplyCommand{
+		Meta: Meta{
+			testingOverrides: metaOverridesForProvider(p),
+			View:             view,
+		},
+	}
+
+	args := []string{
+		"-show-sensitive",
+	}
+	code := c.Run(args)
+	output := done(t)
+	if code != 0 {
+		t.Fatalf("bad status code: \n%s", output.Stderr())
+	}
+
+	expected := `Changes to Outputs:
+  [32m+[0m[0m notsensitive = "Hello world"
+  [32m+[0m[0m sensitive    = "Hello world"
+
+You can apply this plan to save these new output values to the OpenTofu
+state, without changing any real infrastructure.
+[0m[1m[32m
+Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
+[0m[0m[1m[32m
+Outputs:
+
+[0mnotsensitive = "Hello world"
+sensitive = "Hello world"`
+
+	actual := strings.TrimSpace(output.Stdout())
+	if diff := cmp.Diff(actual, expected); len(diff) > 0 {
+		t.Fatalf("got incorrect output\n %v", diff)
+	}
+}
+
+func TestApply_withoutShowSensitiveArg(t *testing.T) {
+	td := t.TempDir()
+	testCopyDir(t, testFixturePath("apply-sensitive-output"), td)
+	defer testChdir(t, td)()
+
+	p := applyFixtureProvider()
+	view, done := testView(t)
+	c := &ApplyCommand{
+		Meta: Meta{
+			testingOverrides: metaOverridesForProvider(p),
+			View:             view,
+		},
+	}
+
+	args := []string{}
+	code := c.Run(args)
+	output := done(t)
+	if code != 0 {
+		t.Fatalf("bad status code: \n%s", output.Stderr())
+	}
+
+	expected := `Changes to Outputs:
+  [32m+[0m[0m notsensitive = "Hello world"
+  [32m+[0m[0m sensitive    = (sensitive value)
+
+You can apply this plan to save these new output values to the OpenTofu
+state, without changing any real infrastructure.
+[0m[1m[32m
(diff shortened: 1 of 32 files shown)


internal/command/jsonstate/state.go
```diff
@@ -495,6 +495,13 @@ func marshalResources(resources map[string]*states.Resource, module addrs.Module
 	return ret, nil
 }
 
+// SetSensitive marks the SensitiveValues in Resource as nil to display the sensitive values.
+// This function is only used when the -show-sensitive flag is passed as an argument in
+// tofu state show command.
+func (r *Resource) SetSensitive() {
```
Evi1Pumpkin, line 501: The name is a little bit confusing, maybe we should call it `SetNonsensitive()` instead?

internal/command/views/operation.go
```diff
@@ -95,6 +95,12 @@ func (v *OperationHuman) EmergencyDumpState(stateFile *statefile.File, enc encry
 }
 
 func (v *OperationHuman) Plan(plan *plans.Plan, schemas *tofu.Schemas) {
+	// If the -show-sensitive argument is provided in the tofu plan/apply command,
+	// then mark the sensitive variable in plan.Changes.Outputs as false.
+	if v.view.showSensitive && plan.Changes != nil {
+		resetSensitiveVariables(plan.Changes.Outputs)
```
Evi1Pumpkin, line 101: I read the issue, and I'm unsure if we should only show sensitive outputs when it comes to `plan` / `apply` or the whole plan.
From the [issue](https://github.com/opentofu/opentofu/issues/839):
> There are multiple cases where TF knows a sensitive value but refuses to display it, such as:
1. viewing TF created secret values
2. approving sensitive diffs


From reading this, I understand that all sensitive changes should be seen, not only outputs, unless I misunderstood.

internal/command/jsonstate/state.go
siddharthasonker95: agreed, how about `SetSensitiveAsNil()`?

Evi1Pumpkin: `SetSensitiveAsNil()` can work, but from reading it, the first thing that pops into my head is that we took a specific sensitive value and replaced it with nil value instead.
Maybe `MakeAllValuesNonsensitive`?

siddharthasonker95: sounds good, thanks

siddharthasonker95: I have removed this method as I think it's no longer needed and we can reset the sensitive value in [state_show.go](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/state_show.go#L176) itself.

RLRabinowitz (commented):
Skimming through the use-cases, it seems like we'd also want a similar flag for `tofu show` (and not just `tofu state show`).

This did not appear in the original issue, but is probably warranted if we are introducing the flag for `tofu state show`

siddharthasonker95:
> Skimming through the use-cases, it seems like we'd also want a similar flag for `tofu show` (and not just `tofu state show`).
> 
> This did not appear in the original issue, but is probably warranted if we are introducing the flag for `tofu state show`

Sure, I will add support for `tofu show`. Do we need this flag for `tofu refresh` as well?

internal/command/state_show.go
```diff
@@ -33,13 +33,17 @@ func (c *StateShowCommand) Run(args []string) int {
...
+
 	if err := cmdFlags.Parse(args); err != nil {
 		c.Streams.Eprintf("Error parsing command-line flags: %s\n", err.Error())
 		return 1
 	}
 	args = cmdFlags.Args()
-	if len(args) != 1 {
-		c.Streams.Eprint("Exactly one argument expected.\n")
+	if len(args) == 0 {
+		c.Streams.Eprint("Atleast one argument expected.\n")
```
RLRabinowitz, line 46: I think that this change is wrong. The arguments have nothing to do with the flags. The one expected argument is the address of a resource, nothing to do with the new flag

internal/command/output_test.go
```diff
@@ -325,3 +325,71 @@ func TestOutput_stateDefault(t *testing.T) {
...
+		t.Fatalf("bad: \n%s", output.Stderr())
+	}
+
+	actual := strings.TrimSpace(output.Stdout())
+	if actual != "foo = <sensitive>" {
+		t.Fatalf("bad: %#v", actual)
+	}
+}
+
+func testStateForOutput() *states.State {
```
RLRabinowitz, line 386: Without context, this function is unclear. One would probably assume they could use this as the state for all tests in this suite

The name of it should be more precise, probably something like `stateWithSensitiveValue`

internal/command/state_show_test.go
```diff
@@ -270,6 +271,119 @@ func TestStateShow_configured_provider(t *testing.T) {
...
+    foo = "value"
+    id  = "bar"
+}`
+	actual := strings.TrimSpace(output.Stdout())
+	if diff := cmp.Diff(actual, expected); len(diff) > 0 {
+		t.Fatalf("got incorrect output\n %v", diff)
+	}
+}
+
+func testStateForStateShow() *states.State {
```
RLRabinowitz, line 349: Same as my previous comment, name is unclear in this context. It should be clear in context `stateWithSensitiveValue`

internal/command/state_show_test.go
```diff
@@ -270,6 +271,119 @@ func TestStateShow_configured_provider(t *testing.T) {
...
+				Provider: addrs.NewDefaultProvider("test"),
+				Module:   addrs.RootModule,
+			},
+		)
+	})
+
+	return state
+}
+
+func testProviderForStateShow() *providers.GetProviderSchemaResponse {
```
RLRabinowitz, line 371: Same as my previous comment, name is unclear in this context. It should be clear in context `providerWithSensitiveValue`

internal/command/views/operation.go
```diff
@@ -284,6 +290,19 @@ func (v *OperationJSON) Diagnostics(diags tfdiags.Diagnostics) {
...
+		output.BeforeSensitive = nil
+		output.AfterSensitive = nil
+		outputs[i] = output
+	}
+
+	for i := range resourceChanges {
+		resourceChanges[i].Change.BeforeSensitive = nil
+		resourceChanges[i].Change.AfterSensitive = nil
+	}
+}
```
RLRabinowitz, line 304: I'm not entirely sure the approach of altering the "Sensitive" values of outputs and resources in the input of the renderer is the best approach here. It might have some unwanted side-effects.

An example would be the following one. Assume you already have the following configuration:
```hcl
variable "var1" {
}

resource "null_resource" "a" {
  triggers = {
    a = var.var1
  }
}
```
And this configuration is applied with the value `"X"` for `var1`, and is in state. Later on, you change the configuration, and only change the variable to be sensitive:
```hcl
variable "var1" {
  sensitive = true
}

resource "null_resource" "a" {
  triggers = {
    a = var.var1
  }
}
```

Running a plan and providing the same value for `var1`, would result in an attempted change of the resource, including a warning that a field will now become sensitive:
```
  ~ update in-place

  # null_resource.a will be updated in-place
  ~ resource "null_resource" "a" {
        id       = "3504371271674334511"
      ~ triggers = {
          # Warning: this attribute value will be marked as sensitive and will not
          # display in UI output after applying this change. The value is unchanged.
          ~ "a" = (sensitive value)
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

However, if you run `tofu plan --show-sensitive`, the plan does not mention that the field will become sensitive, and becomes unclear:
```
  ~ update in-place

  # null_resource.a will be updated in-place
  ~ resource "null_resource" "a" {
        id       = "3504371271674334511"
        # (1 unchanged attribute hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

I'm not entirely sure that this is the only effect of overriding the sensitive values that are the input of the renderer. We should probably discuss what the expected behaviour here should be

RLRabinowitz: Well `RenderHuman` only affects the rendered output though. Currently, it can only tell the difference between a sensitive value and a non-sensitive value, and show a diff accordingly.

An arguably better approach would be to let it receive `showSensitive` as input, and then decide how to render the plan accordingly

Evi1Pumpkin: That's a good point @RLRabinowitz. It can be very confusing for the user to see an unexplainable change. Plus, afterward, when he runs commands without the `--show-sensitive` flag, he'll be surprised to see that the attribute is suddenly sensitive.
So IIMO this is unacceptable behaviour, and we should try and fix it.

Maybe instead of changing the `BeforeSensitive` and `AfterSensitive`, we can trickle down the `showSensitive` value as an input to the [sensitiveRenderer](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/computed/renderers/sensitive.go#L25) (it's responsible for showing the `this attribute value will no longer be marked as sensitive` [warning](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/computed/renderers/sensitive.go#L46)) and add a proper warning the fits our situation.
I don't think it should be hard to do that as the [sensitiveRenderer.RenderHuman()](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/computed/renderers/sensitive.go#L32) already accepts [RenderHumanOpts](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/computed/diff.go#L75) struct. From what I see we'll want to trickle down the `showSensitive` value similarly to how `Colorize` is passed (from the `jsonformat.Renderer` [here](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/views/operation.go#L105) to eventually the `RenderHumanOpts` [here](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/computed/diff.go#L76)).

siddharthasonker95: I chose to implement resetSensitiveVariables in [/view/operation.go](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/views/operation.go#L293). My approach was to keep the showSensitive flag, which is used for view or display, separate from RenderHuman. I wanted this flag to be part of the view layer exclusively and to handle sensitive values within the view layer only.

Thank you for @RLRabinowitz pointing out that edge case. Resetting the values of `BeforeSensitive` and `AfterSensitive` also impacted [CheckForSensitive()](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/structured/sensitive.go#L50), where the [action](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/structured/sensitive.go#L87) was not updated due to the empty values of BeforeSensitive and AfterSensitive.

siddharthasonker95: Thanks, @Evi1Pumpkin , for your suggestion. Including the showSensitive flag in [jsonformat.Renderer](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/views/operation.go#L104) and using it in [RenderHumanOpts](https://github.com/opentofu/opentofu/blob/94cd210f21d448024bd6d63183390a1e683f7548/internal/command/jsonformat/computed/diff.go#L75) is a better approach. I tried this locally, and it worked as expected. I will update the PR accordingly once I've finalised the changes.

internal/command/apply_test.go
```diff
@@ -2167,6 +2167,60 @@ func TestApply_warnings(t *testing.T) {
...
+	if code != 0 {
+		t.Fatalf("bad status code: \n%s", output.Stderr())
+	}
+
+	if got, want := output.Stdout(), "sensitive    = \"Hello world\""; !strings.Contains(got, want) {
+		t.Fatalf("got incorrect output, want %q, got:\n%s", want, got)
+	}
+}
+
+func TestApply_withoutShowSensitiveArg(t *testing.T) {
```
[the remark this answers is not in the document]
siddharthasonker95: The `TestApply_showSensitiveArg` test includes the show-sensitive flag, which displays the sensitive values as expected. On the other hand, the `TestApply_withoutShowSensitiveArg` test does not include the `show-sensitive` flag, so sensitive values are not displayed.

Both tests are run on the same test file, with and without the `show-sensitive` flag. I have separated them into distinct test functions for better clarity.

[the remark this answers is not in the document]
RLRabinowitz: I was referring to the existing  (not added in this PR) `TestApply_sensitiveOutput`. As far as I can tell, it's pretty much the same as  `TestApply_withoutShowSensitiveArg`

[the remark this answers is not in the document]
siddharthasonker95: got it. fixed it in latest [commit](https://github.com/opentofu/opentofu/pull/1554/commits/47a9c0f953094fb864d133ae7b8b8027dd51ced5)

internal/command/output_test.go
siddharthasonker95: makes sense. I have renamed the function to `stateWithSensitiveValueForOutput` to ensure uniqueness across test files in the same package. We could name it as `TestOutput_stateWithSensitiveValue`, but it might be misleading since this is just a helper function not a test case.

RLRabinowitz (dismissed):

cam72cam (dismissed):

ansromanov:
> > Skimming through the use-cases, it seems like we'd also want a similar flag for `tofu show` (and not just `tofu state show`).
> > This did not appear in the original issue, but is probably warranted if we are introducing the flag for `tofu state show`
> 
> Sure, I will add support for `tofu show`. Do we need this flag for `tofu refresh` as well?

The issue is that `tofu show -json` already shows sensitive values by default, and there is no way to change this behavior at the moment:

```
❯ tofu version                      
OpenTofu v1.7.2
on darwin_arm64

❯ tofu show | grep -i Outputs -A5     
Outputs:

registration_token = (sensitive value)

❯ tg show -json | jq .values.outputs
{

  "registration_token": {
    "sensitive": true,
    "value": "XXXXXXXXXXX",
    "type": "string"
  }
}
```

I propose to make `-show-sensitive` a boolean flag to have the possibility to override such behavior.
If it is out of scope for this PR, I could create a separate issue for this.

cam72cam:
Answered in https://github.com/opentofu/opentofu/issues/839#issuecomment-2167544113

ollevche (dismissed):

ollevche (approved):

Evi1Pumpkin (approved):
LGTM!

PR #1558, opened 2024-04-26
Improve performance of `AttachResourceConfigTransformer` on big graphs

### LINKED ISSUE #1559: Performance degradation of `AttachResourceConfigTransformer` on big graphs
### OpenTofu Version

```shell
1.7.0
```

### Use Cases

We need to evaluate big graphs, and this part of the code has N^2 complexity, leading to significant performance degradation during plan/apply

### Attempted Solutions

Will be fixed by PR

### Proposal

_No response_

### References

_No response_

alexott:
The current implementation of `AttachResourceConfigTransformer` has performance of `O(N^2)` to the number of vertices in the graph, and this leads to significant performance degradation on huge graphs.

The given implementation decreases the complexity to `O(N)` by performing direct lookup of a resource in the `ManagedResources` or `DataResources` depending on the object type.

Resolves #1559

## Target Release

1.8.0

### DIFF
--- a/internal/tofu/transform_attach_config_resource.go
+++ b/internal/tofu/transform_attach_config_resource.go
@@ -56,17 +56,19 @@ func (t *AttachResourceConfigTransformer) Transform(g *Graph) error {
 			m = config.Module.ManagedResources
 		} else if addr.Resource.Mode == addrs.DataResourceMode {
 			m = config.Module.DataResources
+		} else {
+			panic("unknown resource mode: " + addr.Resource.Mode.String())
 		}
-		if m != nil {
-			coord := addr.Resource.String()
-			if r, ok := m[coord]; ok && r.Addr() == addr.Resource {
-				log.Printf("[TRACE] AttachResourceConfigTransformer: attaching to %q (%T) config from %#v", dag.VertexName(v), v, r.DeclRange)
-				arn.AttachResourceConfig(r)
-				if gnapmc, ok := v.(GraphNodeAttachProviderMetaConfigs); ok {
-					log.Printf("[TRACE] AttachResourceConfigTransformer: attaching provider meta configs to %s", dag.VertexName(v))
-					if config.Module != nil && config.Module.ProviderMetas != nil {
-						gnapmc.AttachProviderMetaConfigs(config.Module.ProviderMetas)
-					}
+		coord := addr.Resource.String()
+		if r, ok := m[coord]; ok && r.Addr() == addr.Resource {
+			log.Printf("[TRACE] AttachResourceConfigTransformer: attaching to %q (%T) config from %#v", dag.VertexName(v), v, r.DeclRange)
+			arn.AttachResourceConfig(r)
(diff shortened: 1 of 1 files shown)


internal/tofu/transform_attach_config_resource.go
```diff
@@ -50,63 +51,23 @@ func (t *AttachResourceConfigTransformer) Transform(g *Graph) error {
...
-					log.Printf("[TRACE] AttachResourceConfigTransformer: no provider metas defined for %s", dag.VertexName(v))
-					continue
+		if m != nil {
+			coord := addr.Resource.String()
+			if r, ok := m[coord]; ok && r.Addr() == addr.Resource {
+				log.Printf("[TRACE] AttachResourceConfigTransformer: attaching to %q (%T) config from %#v", dag.VertexName(v), v, r.DeclRange)
+				arn.AttachResourceConfig(r)
+				if gnapmc, ok := v.(GraphNodeAttachProviderMetaConfigs); ok {
+					log.Printf("[TRACE] AttachResourceConfigTransformer: attaching provider meta configs to %s", dag.VertexName(v))
+					if config.Module != nil && config.Module.ProviderMetas != nil {
```
cam72cam, line 67: I don't believe `config.Module != nil` can ever be false at this point due to the above access.  It looks like the code this was based on had some extraneous checks.

alexott: I agree - I tried to keep the original logic as close as possible...

cam72cam: This is also missing the "[TRACE] AttachResourceConfigTransformer: no provider metas defined for %s" error message when ProviderMetas is nil.

internal/tofu/transform_attach_config_resource.go
```diff
@@ -50,63 +51,23 @@ func (t *AttachResourceConfigTransformer) Transform(g *Graph) error {
...
-					continue
-				}
-				if config.Module == nil {
-					log.Printf("[TRACE] AttachResourceConfigTransformer: no module in config for %s", dag.VertexName(v))
-					continue
-				}
-				if config.Module.ProviderMetas == nil {
-					log.Printf("[TRACE] AttachResourceConfigTransformer: no provider metas defined for %s", dag.VertexName(v))
-					continue
+		if m != nil {
```
cam72cam, line 60: Instead of checking if m != nil, you could add an else to the mode check with a panic if the mode is not set to the two valid values.  It's slightly more defensive in case that enum is ever extended.

alexott: Done

alexott:
@cam72cam I've addressed the comments. Please let me know if something else is required...

cam72cam (approved):

cam72cam:
@alexott It appears that your latest commit is not signed, could you please address that?

alexott:
Sorry, forgot about it - I need to configure it on the repository level

cube2222 (approved):
LGTM, thanks for working on this!

PR #1591, opened 2024-05-01
Make persist interval for remote state backend configurable

### LINKED ISSUE #1568: Make persistence interval configurable for the remote state backend
### OpenTofu Version

```shell
1.7.0
```

### Use Cases

When state size increases, the persistence of the state becomes a bottleneck as it happens synchronously. Currently, the persist interval is [hardcoded in local apply backend](https://github.com/opentofu/opentofu/blob/main/internal/backend/local/backend_apply.go#L87) to 20 seconds, and as state size increases, it becomes a significant amount of time spent on the persisting. 

For remote apply backend it's possible to disable intermediate persistence completely, but it's not possible to do for the local apply backend.

### Attempted Solutions

I have code changes that allow to specify a new persist interval via an environment variable.

### Proposal

Make the persist interval configurable (or allow you to disable it completely) instead of using a hardcoded value. Open question: How to specify it: Is it enough to provide support via an environment variable, or will we need to add support for the command-line option?

Also, maybe it makes sense to make actual cloud write async? 

### References

_No response_

alexott:
By default, the remote state backend persists state to the remote storage every 20 seconds—this value is hardcoded. But when the state is big (> 100Mb), this operation takes a long time and consumes a significant portion of these 20 seconds. This PR allows to make the persist interval configurable via the `TF_STATE_PERSIST_INTERVAL` environment variable, which allows to specify how often persistence should happen.

Please guide what part of documentation needs to be updated to document it...

P.S. I thought about adding the ability to completely disable this persist interval like it's done when the remote execution backend is used - but this could be a topic for discussion.

Resolves #1568

## Target Release

1.8.0

### DIFF
--- a/internal/backend/local/backend_apply.go
+++ b/internal/backend/local/backend_apply.go
@@ -31,7 +31,7 @@ var testHookStopPlanApply func()
 
 var (
 	defaultPersistInterval                 = 20
-	persistIntervalEnvironmentVariableName = "TF_BACKEND_PERSIST_INTERVAL_SECONDS"
+	persistIntervalEnvironmentVariableName = "TF_STATE_PERSIST_INTERVAL"
 )
 
 func getEnvAsInt(envName string, defaultValue int) int {
--- a/internal/backend/local/backend_apply.go
+++ b/internal/backend/local/backend_apply.go
@@ -10,6 +10,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/opentofu/opentofu/internal/addrs"
@@ -26,6 +28,22 @@ import (
 
 // test hook called between plan+apply during opApply
 var testHookStopPlanApply func()
+
+var (
+	defaultPersistInterval                 = 20
+	persistIntervalEnvironmentVariableName = "TF_BACKEND_PERSIST_INTERVAL_SECONDS"
+)
+
+func getEnvAsInt(envName string, defaultValue int) int {
+	if val, exists := os.LookupEnv(envName); exists {
+		parsedVal, err := strconv.Atoi(val)
+		if err == nil {
+			return parsedVal
+		}
+		panic(fmt.Sprintf("Can't parse value '%s' of environment variable '%s'", val, envName))
+	}
+	return defaultValue
+}
 
 func (b *Local) opApply(
 	stopCtx context.Context,
@@ -84,7 +102,11 @@ func (b *Local) opApply(
 	// stateHook uses schemas for when it periodically persists state to the
 	// persistent storage backend.
 	stateHook.Schemas = schemas
-	stateHook.PersistInterval = 20 * time.Second // arbitrary interval that's hopefully a sweet spot
+	persistInterval := getEnvAsInt(persistIntervalEnvironmentVariableName, defaultPersistInterval)
+	if persistInterval < 0 {
+		panic(fmt.Sprintf("Can't use negative value for %s: %d", persistIntervalEnvironmentVariableName, persistInterval))
+	}
+	stateHook.PersistInterval = time.Duration(persistInterval) * time.Second // arbitrary interval that's hopefully a sweet spot
 
 	var plan *plans.Plan
 	// If we weren't given a plan, then we refresh/plan
--- a/internal/backend/local/backend_apply.go
+++ b/internal/backend/local/backend_apply.go
@@ -103,8 +103,9 @@ func (b *Local) opApply(
 	// persistent storage backend.
 	stateHook.Schemas = schemas
 	persistInterval := getEnvAsInt(persistIntervalEnvironmentVariableName, defaultPersistInterval)
-	if persistInterval < 0 {
-		panic(fmt.Sprintf("Can't use negative value for %s: %d", persistIntervalEnvironmentVariableName, persistInterval))
+	if persistInterval < defaultPersistInterval {
+		panic(fmt.Sprintf("Can't use value lower than %d for env variable %s, got %d",
+			defaultPersistInterval, persistIntervalEnvironmentVariableName, persistInterval))
 	}
 	stateHook.PersistInterval = time.Duration(persistInterval) * time.Second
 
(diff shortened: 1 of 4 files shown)


alexott:
@cam72cam who can review this PR? thank you

cam72cam:
Hey @alexott! I was waiting for the synthetic test case you had mentioned so we could validate these changes are having the impact you are aiming for.

alexott:
@cam72cam I put code and instructions here: https://gist.github.com/alexott/138aa4de1d4be7f905b37170d20b6494

cam72cam:
@alexott is there any functional difference between that and using count?
```hcl
variable "n" {
        type = number
}

resource "random_integer" "rint" {
  count = var.n
  min = 1
  max = 50000
}
resource "random_bytes" "rbytes" {
  count = var.n
  length = 64
}
resource "random_id" "rid" {
  count = var.n
  byte_length = 20
}
resource "random_password" "rpass" {
  count = var.n
  length           = 20
}
resource "random_string" "rstr" {
  count = var.n
  length           = 20
  special          = true
}
resource "random_uuid" "ruuid" {
  count = var.n
}
```

cam72cam:
I think I've come up with a better synthetic use case:
```bash
echo "Generating large file..."
dd if=/dev/urandom of=./rdata.raw count=1k
echo "Converting large file to b64..."
cat rdata.raw | base64 -w 0 > rdata.b64
```

```hcl
resource "tfcoremock_simple_resource" "copier" {
        count = 500
        string = file("./rdata.b64")
}
```

That creates a state file that's over 300MB and spends half of it's time writing to terraform.tfstate (confirmed by setting backend to "inmem").  I'll give that a try with a S3 backend in a bit and see how that works.

What do you think of that @alexott?

alexott:
Really, to make it reproducible we need to have a lot of resources with a big state - creating a big file helps to create big initial state, but then to test periodic persistence we need to have many resources

cam72cam:
I've documented my confirmation of the significant performance increase here: https://github.com/opentofu/opentofu/issues/1660#issuecomment-2120687744

cam72cam:
I'm going to close this as there is a much larger discussion that needs to happen in the linked issue.  Thanks for starting the discussion and your patience @alexott!

cam72cam:
@alexott It appears I posted my comment and closed the wrong PR, my sincere apologies!

cam72cam (dismissed):

cam72cam:
I've raised an issue with the team about the env var global linting.  You did things the "idiomatic" way, which was the right call.

internal/backend/local/backend_apply.go
```diff
@@ -84,7 +102,11 @@ func (b *Local) opApply(
 	// stateHook uses schemas for when it periodically persists state to the
 	// persistent storage backend.
 	stateHook.Schemas = schemas
-	stateHook.PersistInterval = 20 * time.Second // arbitrary interval that's hopefully a sweet spot
+	persistInterval := getEnvAsInt(persistIntervalEnvironmentVariableName, defaultPersistInterval)
+	if persistInterval < 0 {
+		panic(fmt.Sprintf("Can't use negative value for %s: %d", persistIntervalEnvironmentVariableName, persistInterval))
+	}
+	stateHook.PersistInterval = time.Duration(persistInterval) * time.Second // arbitrary interval that's hopefully a sweet spot
```
cam72cam, line 109: Can we move the ` // arbitrary interval that's hopefully a sweet spot` comment above the defaultPersistInterval var?

alexott: Done

cam72cam (approved):

internal/backend/local/backend_apply.go
cam72cam: Thanks!

Yantrio (dismissed):
I'm happy to approve this. It will need a changelog entry and documentation improving but we can do that in a different PR?

WDYT @janosdebugs @cam72cam

alexott:
I can add both docs & changelog - just maybe need some guidance

cam72cam:
We will probably want a entry in https://opentofu.org/docs/cli/config/environment-variables/ describing the environment variable and some additional information in https://opentofu.org/docs/cli/commands/apply/ linking to that environment variable.

alexott:
Thank you! I'll add corresponding docs

alexott:
Documentation is added

cam72cam:
I hate to ask at this juncture, but does this make more sense as a CLI flag or a option in the backend configuration?

Thinking through the three different permutations of this:

## Environment Variable
* Can be set globally in a CI system easily
* May only make sense for some projects within a organization
* Is not easily discoverable (people tend not to read the docs first)
* Would be easy for a user to forget

## CLI Flag
* Trickier to set globally in a CI system, but not too hard
* May only make sense for some projects within a organization
* Is reasonably easy to discover via --help
* Would be easy for a user to forget

## Backend option
* Could be added as a global backend option
* A bit trickier to code, uses PartialContent concepts
* Added at the project level, not the CI or org level
* Is not easily discoverable (people tend not to read the docs first)
* Once it's added it does not require additional work from users

alexott:
For this specific use case we thought that environment variables could be least invasive...

cam72cam:
I'll bring this up at triage tomorrow morning as we have not accepted this issue yet and will discuss in depth with the core team.

cam72cam:
I should have double checked the issue before performing the review I did at this level.

cam72cam (dismissed):

Yantrio (dismissed):

website/docs/cli/commands/apply.mdx
```diff
@@ -93,6 +93,10 @@ For configurations using
 `tofu apply` also accepts the legacy options
 [`-state`, `-state-out`, and `-backup`](../../language/settings/backends/local.mdx#command-line-arguments).
 
+### Environment variables
+
+You can further customize behavior of `apply` command by using [environment variables](../config/environment-variables.mdx).  For example, the [TF_STATE_PERSIST_INTERVAL](../config/environment-variables.mdx##tf_state_persist_interval) environment variable allows to specify the interval between state persistence.
```
siddharthasonker95, line 98:
```suggestion
You can further customize behavior of `apply` command by using [environment variables](../config/environment-variables.mdx).  For example, the [TF_STATE_PERSIST_INTERVAL](../config/environment-variables.mdx#tf_state_persist_interval) environment variable allows to specify the interval between state persistence.
```

siddharthasonker95 (approved):

cam72cam (approved):

PR #1595, opened 2024-05-02
wrap JSON strings as jsonencode calls on generate-config-out

### LINKED ISSUE #1371: When generating configuration, wrap JSON data with `jsonencode`
### OpenTofu Version

```shell
1.6.2
```

### Use Cases

Currently, generated code that contains stringified JSON is generated as a stringified JSON string. For example, a generated `aws-iam-group-policy` would have a `policy` that's `"{\"Version\":\"2012-10-17\",\"Statement\";[{...}]}"`.

It would look nicer if it was instead:
```
  policy = jsonencode({
    Version = "2012-10-17"
    Statement = [{
      ...
    }]
  })
```

Just for a better aesthetic look for generated configuration

### Attempted Solutions

.

### Proposal

See "Use Cases"

### References

Related issue in the Terraform repository - https://github.com/hashicorp/terraform/issues/34444

ollevche:
Resolves #1371

This PR makes `tofu plan` with `generate-config-out` flag check for JSON strings and, if possible, replace them with `jsonencode` function call.

## Target Release

1.7.2

### DIFF
--- a/internal/genconfig/generate_config_test.go
+++ b/internal/genconfig/generate_config_test.go
@@ -425,6 +425,53 @@ resource "tfcoremock_simple_resource" "empty" {
   single = null
 }`,
 		},
+		"jsonencode_wrapping": {
+			schema: &configschema.Block{
+				Attributes: map[string]*configschema.Attribute{
+					"juststr": {
+						Type:     cty.String,
+						Optional: true,
+					},
+					"jsonobj": {
+						Type:     cty.String,
+						Optional: true,
+					},
+					"jsonarr": {
+						Type:     cty.String,
+						Optional: true,
+					},
+				},
+			},
+			addr: addrs.AbsResourceInstance{
+				Module: nil,
+				Resource: addrs.ResourceInstance{
+					Resource: addrs.Resource{
+						Mode: addrs.ManagedResourceMode,
+						Type: "tfcoremock_simple_resource",
+						Name: "example",
+					},
+					Key: nil,
+				},
+			},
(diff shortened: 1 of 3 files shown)


internal/genconfig/generate_config_test.go
```diff
@@ -423,6 +423,53 @@ resource "tfcoremock_simple_resource" "empty" {
...
+					},
+					"jsonobj": {
+						Type:     cty.String,
+						Optional: true,
+					},
+					"jsonarr": {
+						Type:     cty.String,
+						Optional: true,
+					},
+				},
```
cam72cam, line 443: Can we add a test case with a sensitive object, and one with a sensitive object field?

Evi1Pumpkin (approved):
Looking great!

internal/genconfig/generate_config_test.go
cam72cam: :+1: on the test case.  My only remaining question is if the generated config should include sensitive values?  What was the behavior before?

ollevche: Generated config doesn't include sensitive values, it just sets null with a comment: https://github.com/opentofu/opentofu/blob/main/internal/genconfig/generate_config.go#L161

Regarding the test case - I was adding a couple of additional tests when I discovered a panic (in existing code). It is caused by non-sensitive objects with an inner field value marked as sensitive. 

I'm investigating if that's a real case and how to fix it if so.

cam72cam (approved):

PR #1598, opened 2024-05-03
Don't do a DeepCopy in the remote state backend to avoid double copying

### LINKED ISSUE #1580: Excessive DeepCopy usage in case of using remote state backend
### OpenTofu Version

```shell
main branch
```

...

### Debug Output

...

### Expected Behavior

We should avoid doing unnecessary `DeepCopy` on the state

### Actual Behavior

When persisting the state, we're calling `DeepCopy` on state in the [internal/tofu/update_state_hook.go](https://github.com/opentofu/opentofu/blob/main/internal/tofu/update_state_hook.go#L16) and then in the [remote state backend](https://github.com/opentofu/opentofu/blob/main/internal/states/remote/state.go#L107), and with big states this causes significant slowness.

### Steps to Reproduce

Have a terraform code with a lot of resources (10ks), run apply with remote state backend configured.If you have Terraform code with a lot of resources (10ks), run apply with the 

### Additional Context

_No response_

### References

_No response_

alexott:
The `updateStateHook` performs `DeepCopy` to create a `state` object that is then pushed to the `StateMgr.WriteState`.  If we use remote state backend, then it's doing `DeepCopy` again leading to increased pressure on the garbage collection in case of big states (couple of hundreds megabytes).

Resolves #1580

## Target Release

1.8.0

### DIFF
--- a/internal/states/remote/state.go
+++ b/internal/states/remote/state.go
@@ -101,10 +101,8 @@ func (s *State) WriteState(state *states.State) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	// We create a deep copy of the state here, because the caller also has
-	// a reference to the given object and can potentially go on to mutate
-	// it after we return, but we want the snapshot at this point in time.
-	s.state = state.DeepCopy()
+	// We don't create a DeepCopy because it's already created in the `update_state_hook`
+	s.state = state
 
 	return nil
 }


cam72cam:
My main concern with this PR is mirrored in the commit message in 4d53eaa6df2f2f597e18c352ae0ef8baa3e4ed32.  At least at the time, the caller was untrusted and assumptions around the deep copy are likely baked deep into the codebase.

Honestly, this won't be easy to review and verify due to the lack of testing of edge cases in a lot of the codebase.

A different approach might be to introduce the idea of "immutable state".  The first caller would make a copy of the state, mark it as immutable and then pass it down the stack.  The receivers would then be able to see if the state is immutable or not and act accordingly.  You would also have an option to make a "MutableCopy" that would return a mutable state.

```go
func (s *State) Immutable() ImmutableState {
   return ImmutableState{state: s.DeepCopy()}
}

type ImmutableState struct {
  state *State
}

func (s *ImmutableState) Mutable() State {
   return s.state.DeepCopy()
}
```
This approach could be introduced in a case-by-case basis and would clearly show what caller/callee assumptions are being made.  Long term both should probably be an interface, but that's a massive refactoring effort that's not worth it at the moment.


I'm not saying your deductions are wrong, rather that it's a complex area and making assumptions explicit is the only path forward that I see where we don't accidentally introduce bugs later down the line.

alexott:
@cam72cam I completely agree with your analysis - we didn't analyze all possible paths, and there could be potential regressions. I also though about potential implementation of state that will do DeepCopy on demand...

Also, another potential place to remove DeepCopy is update state hook: https://github.com/alexott/opentofu/pull/1/files#diff-33cdcc966b79a205d486fd80536296251e6be1276eddf32e8a0ed9ce33ce92f7L15 - it looks like all callees of it are making DeepCopy anyway.

cam72cam:
Let's move back to the issue and discuss there as we need to have a deeper discussion about methods before a PR can be worked in.

cam72cam:
I'm going to close this as there is a much larger discussion that needs to happen in the linked issue. Thanks for starting the discussion and your patience @alexott!

PR #1611, opened 2024-05-06
Removing noisy and always failing jobs

ghost:
This PR removes the almost always failing action that verifies if an issue is linked. It also removes the reminder to the PR assignee, since we don't follow that process anymore.

## Target Release

1.8, 1.7, 1.6


.github/workflows/pr-opened.yml
```diff
@@ -1,22 +0,0 @@
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # v6.4.1
-        with:
-          script: |
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: "Reminder for the PR assignee: If this is a user-visible change, please update the changelog as part of the PR."
```
Evi1Pumpkin, line 21: Are you sure we want to remove that? That's a good reminder to update the changelog. I personally forgot a few times, but this message was helpful.

ollevche:
Why do we want to remove the linked issues check? Shouldn't we link PRs in 99% of cases?

I also agree with @Evi1Pumpkin about the CHANGELOG check. However, we could add an "if" to not send a reminder if the CHANGELOG file is already edited.

Evi1Pumpkin (approved):

ollevche (approved):

Issue #1528, opened 2024-04-21
Issues with loopable import blocks and CLI on windows platform


ShadowScorpion:
### Did you test the provider-defined functions?

Tested, worked

### Additional feedback / details on the provider-defined functions

Tested with multiple corefunc options. Works well.

### Did you test the `for_each` on the import block?

Tested, did not work / had problems

### Additional feedback / details on the import block

Tested on multiple variants, but facing problems, each time when trying to import.

Platform: 
```
./tofu -v
OpenTofu v1.7.0-beta1
on linux_amd64
+ provider registry.opentofu.org/hashicorp/aws v5.46.0
```

Steps to reproduce:
- Create on AWS the SSM parameter resources (working code):

```
variable "ssm_parameters_names_list" {
  type    = list(string)
  default = ["/import/test1", "/import/test2"]
}

variable "ssm_parameters_values_list" {
  type    = list(string)
  default = ["test1text", "test2text"]
}

variable "ssm_parameters_ids" {
  type    = list(string)
  default = ["arn:aws:ssm:eu-west-1:123456789012:parameter/import/test1", "arn:aws:ssm:eu-west-1:123456789012:parameter/import/test2"]
}

resource "aws_ssm_parameter" "test" {
  count = length(var.ssm_parameters_names_list)

  name  = var.ssm_parameters_names_list[count.index]
  type  = "String"
  value = var.ssm_parameters_values_list[count.index]
}

output "ssm_ids" {
  value = aws_ssm_parameter.test[*].id
}
```

```
`tofu init ; tofu apply
...
Apply complete! Resources: 2 added, 0 changed, 0 destroyed.

Outputs:

ssm_ids = [
  "/import/test1",
  "/import/test2",
]
```

- Remove the state file and any cache
```
rm -f ./terraform.tfstate ; rm -rf ./.*`

- Add import block:
`import {
  for_each = { for idx, name in var.ssm_parameters_names_list : idx => name }
  to = aws_ssm_parameter.test[each.key]
  id = each.value
}
```

And after using this block I'm getting the problem:

> │ Error: Configuration for import target does not exist
> │
> │ The configuration for the given import aws_ssm_parameter.test["0"] does not exist. All target instances must have an associated configuration to be imported.
> ╵
> ╷
> │ Error: Configuration for import target does not exist
> │
> │ The configuration for the given import aws_ssm_parameter.test["1"] does not exist. All target instances must have an associated configuration to be imported.

[...]

PR #1645, opened 2024-05-12
Add more tests for import for_each

Evi1Pumpkin:
1. Added more tests for the unhappy flow of import for_each
2. Extracted booleans to consts when calling the function `evaluateForEachExpressionValue()`

Resolves #

## Target Release

1.8.0

### DIFF
--- a/internal/tofu/context_plan2_test.go
+++ b/internal/tofu/context_plan2_test.go
@@ -4533,6 +4533,171 @@ import {
 	}
 }
 
+func TestContext2Plan_importToInvalidDynamicAddress(t *testing.T) {
+	type TestConfiguration struct {
+		Description         string
+		expectedError       string
+		inlineConfiguration map[string]string
+	}
+	configurations := []TestConfiguration{
+		{
+			Description:   "To address index value is null",
+			expectedError: "Import block 'to' address contains an invalid key: Import block contained a resource address using an index which is null. Please ensure the expression for the index is not null",
+			inlineConfiguration: map[string]string{
+				"main.tf": `
+variable "index" {
+  default = null
+}
+
+resource "test_object" "a" {
+  count = 1
+  test_string = "foo"
+}
+
+import {
+  to   = test_object.a[var.index]
+  id   = "123"
+}
+`,
+			},
+		},
+		{
+			Description:   "To address index is not a number or a string",
+			expectedError: "Import block 'to' address contains an invalid key: Import block contained a resource address using an index which is not valid for a resource instance (not a string or a number). Please ensure the expression for the index is correct, and returns either a string or a number",
+			inlineConfiguration: map[string]string{
+				"main.tf": `
+locals {
+  index = toset(["foo"])
+}
+
+resource "test_object" "a" {
+  for_each = toset(["foo"])
+  test_string = "foo"
+}
+
+import {
+  to   = test_object.a[local.index]
+  id   = "123"
+}
+`,
+			},
+		},
+		{
+			Description:   "To address index value is sensitive",
+			expectedError: "Import block 'to' address contains an invalid key: Import block contained a resource address using an index which is sensitive. Please ensure indexes used in the resource address of an import target are not sensitive",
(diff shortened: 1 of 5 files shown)


internal/tofu/context_plan2_test.go
```diff
@@ -4533,6 +4533,171 @@ import {
 	}
 }
 
+func TestContext2Plan_importToInvalidDynamicAddress(t *testing.T) {
```
Evi1Pumpkin, line 4536: I didn't change any of these tests, just reordered and moved it higher in the file

internal/tofu/context_plan2_test.go
```diff
@@ -4718,92 +4883,230 @@ import {
 	}
 }
 
-func TestContext2Plan_importToInvalidDynamicAddress(t *testing.T) {
+func TestContext2Plan_importWithInvalidForEach(t *testing.T) {
```
Evi1Pumpkin, line 4886: These are the new test cases

internal/tofu/context_plan2_test.go
```diff
@@ -4718,92 +4883,230 @@ import {
...
+  for_each = local.map
+  to = test_object.a[each.key]
+  id = each.value
+}
+`,
+			},
+		},
+		{
+			Description:   "for_each expression is null",
+			expectedError: "Invalid for_each argument: The given \"for_each\" argument value is unsuitable: the given \"for_each\" argument value is null. A map, set of strings, or a tuple is allowed.",
```
ollevche, line 4945: Just a minor readability comment: we could use backticks for strings that contain double quotes to not escape them.

Like this: \`Invalid for_each argument: The given "for_each" argument value is unsuitable: the given "for_each" argument value is null. A map, set of strings, or a tuple is allowed.\`

ollevche (approved):
Looks good, thank you!

Do we also have happy path test cases for both `count` and `for_each` in `import` block?

cam72cam (dismissed):

internal/tofu/context_plan2_test.go
Evi1Pumpkin: Fixed https://github.com/opentofu/opentofu/pull/1645/commits/53a62c802bb61cb68963a2727a2fee590fc37cf3

Evi1Pumpkin:
@ollevche Yes, we do. We have happy flow tests for for_each [here](https://github.com/opentofu/opentofu/blob/dec88808d4fc742c6f48ab4ff647f6c3e6e7e251/internal/tofu/context_plan2_test.go#L4701) :)
Regarding `count`, currently the import block does not support the `count` syntax.

ollevche:
@Evi1Pumpkin are you sure `import` block doesn't support `count`? We had [this minor issue](https://github.com/opentofu/opentofu/issues/1528) reported and I tested both `count` and `for_each` in `import`. It worked just fine!

Evi1Pumpkin:
> @Evi1Pumpkin are you sure import block doesn't support count? We had https://github.com/opentofu/opentofu/issues/1528 reported and I tested both count and for_each in import. It worked just fine!

Now I'm intrigued! 🙃 
The `import` [block schema](https://github.com/opentofu/opentofu/blob/53a62c802bb61cb68963a2727a2fee590fc37cf3/internal/configs/import.go#L105) doesn't support the count attribute at all, and instead supports the `for_each` attribute.
And when I test tofu with the following code:
```
import {
  count = 1 
  to   = random_integer.a[count.index]
  id = 2342
}
```

I get the following error:
```
│ Error: Unsupported argument
│
│   on main.tf line 244, in import:
│  244:   count = 1
│
│ An argument named "count" is not expected here.
```

Maybe you tested a scenario where the `count` attribute is on the resource, and not on the `import` block?

ollevche:
@Evi1Pumpkin yep, my bad! You are completely right!

cam72cam:
:+1: once the test fixes are addressed and the linter is passing

cam72cam (approved):

ollevche (approved):

PR #1674, opened 2024-05-24
chore: use errors.New to replace fmt.Errorf with no parameters will much better.

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better.

### DIFF
--- a/cmd/tofu/main.go
+++ b/cmd/tofu/main.go
@@ -8,6 +8,7 @@ package main
 import (
 	"context"
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


ChengenH:
> Hi @ChengenH this seems to be a duplicate pull request of #1520. Could you please explain why using `errors.New` instead of `fmt.Errorf` will benefit OpenTofu?

People tend to think that fmt.Errorf() need to format your error message prints something .
I think errors.Newf() is more clear.

PR #1675, opened 2024-05-24
Adding AI-disabling settings to the project

ghost:
In line with the recent updates to the contributing guidelines, this change adds settings files that disable prominent AI coding assistants when working in OpenTofu. The reason for this change is that coding assistants, such as GitHub Copilot, have likely been trained on the Terraform codebase and have a high likelihood of emitting training data due to the very specific nature of HCL and CTY.

## Target Release

1.8, 1.7, 1.6

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "github.copilot.enable": {
+    "*": false
+  }
+}


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "github.copilot.enable": {
+    "*": true,
```
cube2222, line 3: Doesn't this enable it for everything other than plaintext and markdown?

cube2222 (dismissed):

cube2222 (approved):

PR #1676, opened 2024-05-24
Fixes #1594: Incorrect https_proxy behavior in the S3 backend

### LINKED ISSUE #1594: s3 state backend no longer honours `https_proxy` environment variable
### OpenTofu Version

```shell
OpenTofu v1.6.2
on linux_amd64
```

```hcl
terraform {
  backend "s3" {
    # A configuration using Amazon S3
  }
}
```

### Debug Output

Not needed

### Expected Behavior

When operating in an environment where access to the internet is via a proxy server, and with an `https_proxy` environment variable defined, OpenTofu should use the configured proxy server, **as HashiCorp Terraform v1.5.7 correctly does**.

### Actual Behavior

OpenTofu disregards the environment variable. It tries to reach S3 directly, and eventually times out and fails.

### Steps to Reproduce

Do any command which tries to talk to the state backend (e.g. `tofu init`, `tofu plan`) in an environment with an `s3` backend configured, a `https_proxy` environment variable set, and no access to the S3 service other than through the proxy.

### Additional Context

The environment variables `http_proxy`, `HTTP_PROXY`, etc. are not well standardized. The blog post https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/ provides a good survey of the state of the world.

One of the conclusions there is that **the lowercased variables are preferred by many systems**.

OpenTofu is currently breaking from the prevailing direction by honouring the uppercased variable names exclusively, and this **differs from HashiCorp Terraform** behaviour.

For consistency with other widely deployed software, including HashiCorp Terraform, I believe OpenTofu should honour the lowercased versions of the proxy environment variables.

### References

Looking through Git history, it appears likely this was introduced by #770.

ghost:
Resolves #1594

## Target Release

1.8.0

## Progress

- [x] S3 test backend for integration tests
- [x] DynamoDB test backend for integration tests
- [X] STS test backend for integration tests
- [X] KMS test backend for integration tests
- [x] HTTP/HTTPS proxy backend for integration tests
- [X] Test case proving the inadequate proxy usage.
- [x] Bugfix
- [x] Linting and cleanup

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -752,9 +752,7 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 		StsEndpoint:             customEndpoints["sts"].String(obj),
 		StsRegion:               stringAttr(obj, "sts_region"),
 		Token:                   stringAttr(obj, "token"),
-		HTTPProxy:               aws.String(stringAttrDefaultEnvVar(obj, "http_proxy", "http_proxy", "HTTP_PROXY")),
-		HTTPSProxy:              aws.String(stringAttrDefaultEnvVar(obj, "https_proxy", "https_proxy", "HTTPS_PROXY")),
-		NoProxy:                 stringAttrDefaultEnvVar(obj, "no_proxy", "no_proxy", "NO_PROXY"),
+		HTTPProxyMode:           awsbase.HTTPProxyModeSeparate,
 		Insecure:                boolAttr(obj, "insecure"),
 		UseDualStackEndpoint:    boolAttr(obj, "use_dualstack_endpoint"),
 		UseFIPSEndpoint:         boolAttr(obj, "use_fips_endpoint"),
@@ -780,6 +778,16 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 
 	if val, ok := stringAttrOk(obj, "shared_credentials_file"); ok {
 		cfg.SharedCredentialsFiles = []string{val}
+	}
+
+	if val, ok := stringAttrOk(obj, "http_proxy"); ok {
+		cfg.HTTPProxy = &val
+	}
+	if val, ok := stringAttrOk(obj, "https_proxy"); ok {
+		cfg.HTTPSProxy = &val
+	}
+	if val, ok := stringAttrOk(obj, "no_proxy"); ok {
+		cfg.NoProxy = val
 	}
 
 	if value := obj.GetAttr("assume_role"); !value.IsNull() {
--- a/internal/backend/remote-state/s3/backend_proxy_test.go
+++ b/internal/backend/remote-state/s3/backend_proxy_test.go
@@ -20,6 +20,7 @@ func TestS3ProxyBehavior(t *testing.T) {
 	aws := testutils.AWS(t)
 
 	t.Run("direct-connect", func(t *testing.T) {
+		// This is a baseline test that confirms the connection works against the AWS testutils without a proxy.
 		testutils.SetupTestLogger(t)
 		config := map[string]any{
 			"access_key":       aws.AccessKey(),
@@ -40,6 +41,8 @@ func TestS3ProxyBehavior(t *testing.T) {
 		runBackendTests(t, config)
 	})
 	t.Run("proxy", func(t *testing.T) {
+		// This test case tests with a proxy.
+		testutils.SetupTestLogger(t)
 		var proxyTestCases = map[string]struct {
 			modifyConfig func(t *testing.T, config map[string]any, proxy testutils.HTTPProxyService)
 		}{
@@ -49,6 +52,8 @@ func TestS3ProxyBehavior(t *testing.T) {
 			"direct-config": {
 				modifyConfig: func(_ *testing.T, config map[string]any, proxy testutils.HTTPProxyService) {
 					config["http_proxy"] = proxy.HTTPProxy().String()
+					// Note: this is using the HTTP proxy intentionally since the "https_proxy" refers to the
+					// environment variables!
 					config["https_proxy"] = proxy.HTTPProxy().String()
 				},
 			},
--- a/internal/testutils/ca.go
+++ b/internal/testutils/ca.go
@@ -21,7 +21,7 @@ import (
 	"time"
 )
 
-const caKeySize = 1024
+const caKeySize = 2048
 const expirationYears = 10
 
 // CA creates an x509 CA certificate that can produce certificates for testing purposes.
--- a/internal/testutils/aws.go
+++ b/internal/testutils/aws.go
@@ -99,7 +99,6 @@ func newAWSTestService(t *testing.T, services []awsServiceFixture) AWSTestServic
 		HostAccessPorts: nil,
 		Image:           "localstack/localstack",
 		Env: map[string]string{
-			"USE_SSL":         "1",
 			"LOCALSTACK_HOST": fmt.Sprintf("localhost:%d", localStackPort),
 			"SERVICES":        strings.Join(ids, ","),
 			// Eager loading is on because we need to provision test fixtures anyway.


cam72cam:
Note: We will want to make sure any changes are mirrored in the aws encryption key_provider.

internal/testutils/ca.go
```diff
@@ -0,0 +1,147 @@
...
+
+	cert := &x509.Certificate{
+		SerialNumber: c.serial,
+		Subject:      config.Subject,
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(0, 0, 1),
+		SubjectKeyId: []byte{1},
+		ExtKeyUsage:  config.ExtKeyUsage,
+		KeyUsage:     x509.KeyUsageDigitalSignature,
+		DNSNames:     config.Hosts,
```
cipherboy, line 113: I haven't looked to see if this passes CI yet, but this looks much better!

Yantrio (dismissed):

Evi1Pumpkin:
Hey @janosdebugs, it'll be easier to review if you provide some more details about the problem and how you decided to solve it. Plus, some more details about the testing approach you took.

internal/backend/remote-state/s3/backend.go
```diff
@@ -750,9 +756,9 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 		// Note: we don't need to read env variables explicitly because they are read implicitly by aws-sdk-base-go:
```
Evi1Pumpkin, line 756: Should we remove or alter this comment? Because we are reading the env variables explicitly

ollevche (dismissed):
Thank you, this is awesome!

.github/workflows/checks.yml
```diff
@@ -83,8 +83,7 @@ jobs:
 
       - name: "Unit tests"
         run: |
-          go test ./...
-
+          go test -timeout 20m ./...
```
cam72cam, line 86: This more than doubles our "Quick Check" time:

https://github.com/opentofu/opentofu/actions/runs/9506261377
vs
https://github.com/opentofu/opentofu/actions/runs/9504270127

It adds 1 minute of overhead to Unit Tests linux_amd64, but adds 10!! minutes to linux_arm.  If we adopt this, we can not run these checks on emulated architectures.

cam72cam (changes requested):

go.mod
```diff
@@ -17,17 +17,22 @@ require (
 	github.com/apparentlymart/go-userdirs v0.0.0-20200915174352-b0c018a67c13
 	github.com/apparentlymart/go-versions v1.0.1
 	github.com/armon/circbuf v0.0.0-20190214190532-5111143e8da2
+	github.com/aws/aws-sdk-go v1.44.122
```
cam72cam, line 20: I would rather not pull in aws-sdk-go v1.x.x here.  We should not include this functionality in the testing package as we don't use it.

internal/backend/remote-state/s3/backend.go
```diff
@@ -34,7 +34,13 @@ import (
 	"github.com/zclconf/go-cty/cty/gocty"
 )
 
+// New creates an S3 backend.
 func New(enc encryption.StateEncryption) backend.Backend {
+	return NewTyped(enc)
+}
+
+// NewTyped returns a typed Backend pointer for the S3 backend.
+func NewTyped(enc encryption.StateEncryption) *Backend {
```
cam72cam, line 43: This is only used in a single test.  Why not just replace it with a cast in that test case?

internal/backend/remote-state/s3/backend_proxy_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"testing"
+
+	"github.com/opentofu/opentofu/internal/backend"
+	"github.com/opentofu/opentofu/internal/encryption"
+	"github.com/opentofu/opentofu/internal/states/remote"
+	"github.com/opentofu/opentofu/internal/testutils"
+)
+
+func TestS3ProxyBehavior(t *testing.T) {
+	testutils.SetupTestLogger(t)
```
cam72cam, line 18: Why is this not done from inside testutils.AWS(t)?

internal/backend/testing.go
```diff
@@ -24,7 +24,14 @@ import (
 
 // TestBackendConfig validates and configures the backend with the
 // given configuration.
+// Deprecated: Use TestTypedBackendConfig instead.
 func TestBackendConfig(t *testing.T, b Backend, c hcl.Body) Backend {
```
Yantrio, line 28: What's the point of making this stuff generic? it feels like it's not utilized too much. I think we could just keep the concrete typing and chalk it up to YAGNI

internal/backend/remote-state/s3/backend.go
```diff
@@ -746,16 +752,12 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
...
-		NoProxy:              stringAttrDefaultEnvVar(obj, "no_proxy", "NO_PROXY"),
-		Insecure:             boolAttr(obj, "insecure"),
-		UseDualStackEndpoint: boolAttr(obj, "use_dualstack_endpoint"),
-		UseFIPSEndpoint:      boolAttr(obj, "use_fips_endpoint"),
+		HTTPProxy:               aws.String(stringAttrDefaultEnvVar(obj, "http_proxy", "http_proxy", "HTTP_PROXY")),
+		HTTPSProxy:              aws.String(stringAttrDefaultEnvVar(obj, "https_proxy", "https_proxy", "HTTPS_PROXY")),
+		NoProxy:                 stringAttrDefaultEnvVar(obj, "no_proxy", "no_proxy", "NO_PROXY"),
+		Insecure:                boolAttr(obj, "insecure"),
+		UseDualStackEndpoint:    boolAttr(obj, "use_dualstack_endpoint"),
+		UseFIPSEndpoint:         boolAttr(obj, "use_fips_endpoint"),
```
maxb, line 760: Hi - I'm the person who originally reported this issue in #1594.

I disagree that this is the correct fix.

Ultimately, the bug was caused by re-implementing logic that a lower library could already perform, and re-implementing it incorrectly.

This changes doubles down on the reimplementation... and I can see diverges from it in at least one way, still.

Even if this code was changed to fully match the behaviour of the underlying library, it would still be needless duplication, and a hypothetical future bug if the underlying library was ever updated.

Please see my comment from 16th May on the original issue: https://github.com/opentofu/opentofu/issues/1594#issuecomment-2116200078 in which I suggested how this might be fixed instead.

Thanks.

maxb: I can test this tomorrow.

internal/backend/remote-state/s3/backend.go
```diff
@@ -746,16 +752,10 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
...
-		// Note: we don't need to read env variables explicitly because they are read implicitly by aws-sdk-base-go:
-		// see: https://github.com/hashicorp/aws-sdk-go-base/blob/v2.0.0-beta.41/internal/config/config.go#L133
-		// which relies on: https://cs.opensource.google/go/x/net/+/refs/tags/v0.18.0:http/httpproxy/proxy.go;l=89-96
-		HTTPProxy:            aws.String(stringAttrDefaultEnvVar(obj, "http_proxy", "HTTP_PROXY")),
-		HTTPSProxy:           aws.String(stringAttrDefaultEnvVar(obj, "https_proxy", "HTTPS_PROXY")),
-		NoProxy:              stringAttrDefaultEnvVar(obj, "no_proxy", "NO_PROXY"),
-		Insecure:             boolAttr(obj, "insecure"),
-		UseDualStackEndpoint: boolAttr(obj, "use_dualstack_endpoint"),
-		UseFIPSEndpoint:      boolAttr(obj, "use_fips_endpoint"),
+		HTTPProxyMode:           awsbase.HTTPProxyModeSeparate,
```
ronnyorot, line 755: Why did we add this configuration? What is the implication?

Evi1Pumpkin, line 755: Why did we add this configuration? What is the implication?

internal/backend/remote-state/s3/backend.go
Evi1Pumpkin: If we remove the explicit setting of these attributes, then we should not delete the comment in 750-752, and maybe even clarify it a little more and mention which environment variables we are implicitly getting. We can move this comment to the added logic in 783-792, it might make more sense there.

maxb: `tofu` from this branch successfully completes init/plan in my environment where release versions fail.

PR #1695, opened 2024-05-31
fix tofu installation tests

ollevche:
This PR fixes tests for installation scripts.

1. `test-install-instructions.sh` reports success status for test group even if there are fails. Syntax error in bash script.

2. Tests randomly fail due to non-OK exit code from `rm`, when there is no `install-opentofu.sh`. Due to the parent folder being shared and concurrent execution, it's possible for other tests to remove this script. Fixed by adding `-f` to `rm` in examples. 

PS I didn't check the brew test due to mismatched supported processors. Running docker with ubuntu image on Apple M2 chip (ARM) produces the following error: `Homebrew on Linux is not supported on ARM processors`.

## Target Release

1.8.0

## Checklist

- [X] I have locally started the website as [described here](https://github.com/opentofu/opentofu/blob/main/website/README.md) and checked my changes.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -258,13 +258,13 @@ jobs:
           path: ".licensei.cache"
           key: licensei-cache-${{ hashFiles('go.sum') }}
  
- installation-instructions:
-   name: "Test Installation Instructions"
-   runs-on: ubuntu-latest
+  installation-instructions:
+    name: "Test Installation Instructions"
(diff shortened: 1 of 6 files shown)


ollevche:
BTW should we also try to enable an automated check back? It was removed in https://github.com/opentofu/opentofu/pull/1090. This PR potentially resolves the underlying issue (it was not the initial goal)

RLRabinowitz (dismissed):

PR #1711, opened 2024-06-07
add early validation for enforced methods

### LINKED ISSUE #1710: tofu writes empty state if encryption configuration is invalid
### OpenTofu Version

```shell
OpenTofu v1.7.0-dev
on darwin_arm64
```

V1:
```hcl
terraform {
  encryption {
      passphrase = "imnotgonnachangeyou"
    }

    method "aes_gcm" "new_method" {
      keys = key_provider.pbkdf2.mykey
    }

    state {
      method = method.aes_gcm.new_method
    }
  }
}

resource "null_resource" "example-1" {
}
```

V2:
```hcl
terraform {
  encryption {
      passphrase = "imnotgonnachangeyou"
    }

    method "aes_gcm" "new_method" {
      keys = key_provider.pbkdf2.mykey
    }

    method "unencrypted" "migrate" {}

    state {
      enforced = true

      method = method.unencrypted.migrate

      fallback {
        method = method.aes_gcm.new_method
      }
    }
  }
}

resource "null_resource" "example-1" {
}
```

### Debug Output

Screenshot 2024-06-07 at 17 23 31

### Expected Behavior

tofu produces an error due to an invalid encryption configuration, however, the state file remains untouched.

### Actual Behavior

tofu produces an error due to an invalid encryption configuration, and writes empty `terraform.tfstate` and `errored.tfstate`

### Steps to Reproduce

1. tofu init with V1 configuration
2. tofu apply with V1 configuration
3. tofu apply with V2 configuration

### Additional Context

_No response_

### References

_No response_

ollevche:
Resolves #1710

## Target Release

1.8.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/encryption/targets.go
+++ b/internal/encryption/targets.go
@@ -13,6 +13,7 @@ import (
 	"github.com/opentofu/opentofu/internal/encryption/config"
 	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
 	"github.com/opentofu/opentofu/internal/encryption/method"
+	"github.com/opentofu/opentofu/internal/encryption/method/unencrypted"
 	"github.com/opentofu/opentofu/internal/encryption/registry"
 	"github.com/zclconf/go-cty/cty"
 )
@@ -58,7 +59,21 @@ func (base *baseEncryption) buildTargetMethods(meta map[keyprovider.Addr][]byte)
 	}
 
 	methods, targetDiags := builder.build(base.target, base.name)
-	return methods, append(diags, targetDiags...)
+	diags = append(diags, targetDiags...)
+
+	if base.enforced {
+		for _, m := range methods {
+			if unencrypted.Is(m) {
+				return nil, append(diags, &hcl.Diagnostic{
+					Severity: hcl.DiagError,
+					Summary:  "Unencrypted method is forbidden",
+					Detail:   "Unable to use `unencrypted` method since the `enforced` flag is used.",
+				})
+			}
+		}
+	}
+
+	return methods, diags
 }
 
 // build sets up a single target for encryption. It returns the primary and fallback methods for the target, as well


cam72cam:
Looks good!  Would like to see some tests to confirm (I should have added them last time I touched this code).

We could also probably remove the check inside baseEncryption.encrypt which checks for the unencrypted method as the check should have already happened.  Alternatively we could leave that check just to be safe and add a comment that it should have theoretically already been hit

internal/encryption/targets.go
```diff
@@ -58,7 +59,21 @@ func (base *baseEncryption) buildTargetMethods(meta map[keyprovider.Addr][]byte)
 	}
 
 	methods, targetDiags := builder.build(base.target, base.name)
-	return methods, append(diags, targetDiags...)
+	diags = append(diags, targetDiags...)
+
+	if base.enforced {
+		for _, m := range methods {
```
cam72cam, line 65: Hmm, this is a little tricky.  When encrypting, enforced just applies to the first method.  When decrypting it applies to all methods.

ollevche: I agree tofu should check the first one only on encryption. However, I am missing the point with decryption.

Currently, when decrypting it only applies to _a chosen method_ so it's okay to have an `unencrypted` method in the configuration if the state is not plaintext.

But why do we restrict reading plaintext state in the first place? Consistency?

I think I also miss the point of the `enforced` flag. After the migration, users should just remove the `unencrypted` method everywhere and it's a better option than checking `enforced` flag.

Could you please explain a bit? I think we also need to extend the doc with some guidance on the flag.

ollevche: @cam72cam ^

ollevche: In other words:

If a user wants to block the `unencrypted` method from being used on both encryption and decryption, they use `enforced` flag (current implementation, my understanding).

Why not remove the method from the configuration then?

cam72cam: > But why do we restrict reading plaintext state in the first place? Consistency?

Security.  If someone can bypass state encryption entirely by replacing the contents with an unencrypted payload, they can inject providers and all sorts of other nasty stuff.

ollevche: @cam72cam yep, that matches my understanding as well. But why would you keep an `unencrypted` method with an `enforced` flag after the migration? It makes a lot more sense to just remove the method completely.

ollevche: So there is a good use-case for this feature when users put their configuration into the env variable. The configuration is being merged and having an enforced flag in the file makes sure the configuration from env is read and valid.

Here is more info: https://github.com/opentofu/opentofu/blob/main/docs/state_encryption.md#environment-configuration

I believe we want to have no `unencrypted` methods in the configuration at all in this scenario so it's okay to just check the whole method list (with no ties to the operation). WDYT?

cam72cam: I believe that is valid.

cam72cam (approved):

cam72cam (approved):

internal/encryption/method/aesgcm/aesgcm.go
```diff
@@ -123,3 +123,8 @@ func (a aesgcm) getGCM(key []byte) (cipher.AEAD, error) {
 	}
 	return gcm, nil
 }
+
+func Is(m method.Method) bool {
```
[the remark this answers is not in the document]
ollevche: It's used in newly introduced test: https://github.com/opentofu/opentofu/pull/1711/files#diff-89663d138a7de3a7669ab0efd13ac851af80d5f7a6a4179b6e8e250682f5ee1bR40

Currently, it's hard to tell what algorithm is implemented under `method.Method` since there is no trace back to a descriptor. I would expect having `ID()` as a part of `method.Method` interface, but it may break the segregation logic.

`Is` doesn't need to be a part of the descriptor, but I can move it to the `_test.go` so it's only available for testing purposes. I can also just check via reflection, but that's more fragile.

PR #1716, opened 2024-06-10
RFC: Tofu Version Compatibility

### LINKED ISSUE #1708: Needs OpenTofu image version compatible with terraform version 1.7.x
### OpenTofu Version

```shell
1.7.1
```

### Use Cases

I am trying to update GCS bucket module version from [GCP Cloud foundation fabric](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/v31.0.0/modules/gcs) to use soft_delete_retention parameter which requires terraform [version](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/v31.0.0/modules/gcs/versions.tf) required_version = ">= 1.7.4"  and our terraform version is required_version = ">= 1.1"  so there is no issue I can see but the issue is OpenTofu version compatible with terraform. We are using OpenTofu image version 1.7.1

but I have found this information from OpenTofu website: "Right now, OpenTofu is a drop-in replacement for Terraform, as it's compatible with Terraform versions 1.5.x and most of 1.6.x"

The pipeline error is below

> Error: Unsupported OpenTofu Core version
> │ 
> │   on .terraform/modules/bucket_customer_pdf/modules/gcs/versions.tf line 16, in terraform:
> │   16:   required_version = ">= 1.7.4"
> │ 
> │ Module module.bucket_customer_pdf (from
> │ git::https://github.com/GoogleCloudPlatform/cloud-foundation-fabric.git//modules/gcs?ref=v31.0.0)
> │ does not support OpenTofu version 1.7.1. To proceed, either choose another
> │ supported OpenTofu version or update this version constraint. Version
> │ constraints are normally set for good reason, so updating the constraint
> │ may lead to other errors or unexpected behavior.

### Attempted Solutions

I can see there is no way to work with the latest GCS bucket parameter from GCP Cloud foundation fabric module. One thing is that I need to use GCS resources directly from Terraform rather than using that module. This might work but I have tried that yet.

### Proposal

So we need the OpenTofu compatible version with Terrafrom 1.7.x

### References

1. https://opentofu.org/
2. https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/v31.0.0/modules/gcs#retention-policy-soft-delete-policy-and-logging
3. https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/v31.0.0/modules/gcs/versions.tf

cam72cam:
In this RFC, we investigate current problems around software versioning and propose some potential solutions.

This is contingent on the [.tofu extension RFC](https://github.com/opentofu/opentofu/pull/1699) [merged]

Resolves #1708

## Target Release

1.10.0

## Checklist

### DIFF
--- a/rfc/20240610-Tofu-Version-Compatibility.md
+++ b/rfc/20240610-Tofu-Version-Compatibility.md
@@ -41,7 +41,7 @@ From the perspective of a module, root or child, there are two different ways in
 1) Disabled for all child modules of the current module.  This represents someone who has lots of child modules that have not yet added "Tofu Versions" and they are comfortable with the risk of bypassing the "Terraform Version" in all of them (and their children).
 2) Disabled for a specific child module of the current module.  This represents someone who is pulling a small number of child modules that have not yet added "Tofu Versions" and they would like to limit the blast radius of ignoring the "Terraform Version" check in this limited area of the current module.
 
-For for the first scenario, a flag called "bypass_terraform_version_requirement" is added to the `product{}` block.  If set to true, all child modules of the current module will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
+For the first scenario, a flag called "bypass_terraform_version_requirement" is added to the `product{}` block.  If set to true, all child modules of the current module will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
 
 For the second scenario, a flag called "bypass_terraform_version_requirement" is added to the specific `module{}` block.  If set to true, that specific module and all of it's children will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
 
--- a/rfc/20240610-Tofu-Version-Compatibility.md
+++ b/rfc/20240610-Tofu-Version-Compatibility.md
@@ -16,7 +16,7 @@ At the time of writing this RFC, OpenTofu 1.7.x has very similar features to Ter
 
 ## Proposed Solution
 
-As we introduce support for `.tofu` files, we open up the ability to add new language features without worrying as much about compatibility in shared modules. If we can differentiate between "OpenTofu Version" and "Terraform Version", the above scenarios become much easier.
+As we [introduce support](20240529-OpenTofu-Specific-Code-Override.md) for `.tofu` files, we open up the ability to add new language features without worrying as much about compatibility in shared modules. If we can differentiate between "OpenTofu Version" and "Terraform Version", the above scenarios become much easier.
 
 As the [provider compatibility check is deprecated](https://github.com/hashicorp/terraform-plugin-framework/blob/8a09cef0bb892e2d90c89cad551ce12a959f1a2d/provider/configure.go#L28-L32) and the state's "terraform_version" field is not currently used in OpenTofu, we will not discuss them in this RFC.
 
@@ -85,7 +85,7 @@ Either path is easy to both implement and thoroughly test.
 
 The `configs.Module.CheckCoreVersionRequirements()` function is where we will need to emit the warnings and corresponding debug log entries.
 
-As discussed in the `.tofu` RFC, we will want to quickly work on support for this featu
+As discussed in the `.tofu` RFC, we will want to quickly work on support for this feature.
 
 ### Open Questions
 
--- a/rfc/20240610-Tofu-Version-Compatibility.md
+++ b/rfc/20240610-Tofu-Version-Compatibility.md
@@ -1,6 +1,6 @@
 # Tofu Version Compatibility
 
-Issue: https://github.com/opentofu/opentofu/issues/1708 <!-- Ideally, this issue will have the "needs-rfc" label added by the Core Team during triage -->
+Issue: https://github.com/opentofu/opentofu/issues/1708
 
 As OpenTofu and Terraform diverge, the concept of "software version" for compatibility becomes murky.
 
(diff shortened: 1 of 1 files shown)


rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,109 @@
...
+What happens if only the `version.tf` file exists in this module with no tofu overwrite? This is a common case for modules who have not yet considered explicit OpenTofu migration.
+
+As a user, I would expect that the "Terraform Required Version" would have an equivalent "OpenTofu Required Version". This is not a 1-1 mapping, but is a reasonable stop-gap solution.
+
+As this is a not a 1-1 version mapping and a "best effort" solution, I would expect to see a warning:
+
+```shell
+$ tofu init
+...
+Warning: Using v1.7.x in 'terraform -> required_version' as equivalent to current tofu version 1.6.5!
```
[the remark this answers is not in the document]
cam72cam: It's tricky though, we will need to communicate why the "required_version" constraint was not met in some scenarios.

We could split it into two messages, a warning for specifying `terraform -> required_version` and one that's more detailed when the constraint is not met.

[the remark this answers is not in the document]
ollevche: I think we need to reconsider the version mapping approach. It will become obsolete as soon as Terraform releases an incompatible feature.

After that, OpenTofu will still depend on module authors to update their configuration with alternative `.tofu` implementations.

Also, required version could be incompatible, but an actual configuration is fine to run with OpenTofu and Terraform (if it doesn't use incompatible features).

ImIOImI:
Although I don't think this is critical now because very few popular module authors require very recent versions of Terraform, as the two projects start to diverge it will become increasingly more important. Its better to have a plan in place and be able to revise it while its less important than implement a new system when it is critical.

ghost (changes requested):

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+* State compatibility ("terraform_version" field)
+
+At the time of writing this RFC, OpenTofu 1.7.x has very similar features to Terraform 1.8.x (though not identical). As the "software version" requirements are updated in the previous examples, the exact requirements become murky. A module may require features from "Terraform 1.7.4", but would be perfectly happy with "OpenTofu 1.6.5" (as an example).
+
+> [!NOTE]
+> As we are unable to inspect the Hashicorp Terraform Repository in depth, all version compatibility must be based on Release Notes / Changelog Entries as well as user reports.
+
+## Proposed Solution
+
+As we introduce support for `.tofu` files, we open up the ability to add new language features without worrying as much about compatibility in shared modules. If we can differentiate between "OpenTofu Version" and "Terraform Version", the above scenarios become much easier.
```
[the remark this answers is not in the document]
cam72cam: I will once I can rebase this PR once the .tofu RFC is merged.

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+If we run into providers expecting a particular terraform version (even though this is deprecated), we may want to use the "Terraform Version Equivalent" in that API call.
+
+## Potential Alternatives
+
+Don't implement this and assume that all module and project authors will adopt the .tofu extension and keep the respective required_versions up to date.
+
+Implement an option to disable required_version checking.
+
+Try to guess the equivalent OpenTofu version for a specified Terraform version.
+
```
ImIOImI, line 106: Building a GitHub action was proposed as a possible alternative action as well. I particularly like this no matter what solution we land on because I'm assuming I can use it to guard against unexpected incompatibilities in modules I include in my code as well when I update versions.

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+If we run into providers expecting a particular terraform version (even though this is deprecated), we may want to use the "Terraform Version Equivalent" in that API call.
+
+## Potential Alternatives
+
+Don't implement this and assume that all module and project authors will adopt the .tofu extension and keep the respective required_versions up to date.
+
+Implement an option to disable required_version checking.
+
+Try to guess the equivalent OpenTofu version for a specified Terraform version.
```
skyzyx, line 105: IMO, we should favor being explicit over being implicit.

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+Either path is easy to both implement and thoroughly test.
+
+The `configs.Module.CheckCoreVersionRequirements()` function is where we will need to emit the warnings and corresponding debug log entries.
+
+### Open Questions
+
+Should there be an option (config, CLI, or ENV) to:
+* Disable the version checking altogether?
+* Don't check terraform required_versions
```
ollevche, line 91: This option would just ignore the warning, right?

rfc/20240610-Tofu-Version-Compatibility.md
ollevche: We also discussed some sort of compatibility table between versions and modules based on real user feedback. I know it's tricky, but still, it has some practical benefits. I would add those to the open questions block

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+* Treat both Tofu and Terraform version as identical instead of using our equivalent "terraform version" guess?
+* Do we show the guessed terraform version in `tofu -version`, maybe with a `-verbose` flag?
```
siddharthasonker95, line 93: This can be a good add on, as displaying the guessed Terraform version provides users with immediate feedback and ensures they are using the correct version for their configurations.

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+* Feature compatibility with config files in the root project and modules
+* Provider compatibility checks (deprecated but still used)
+* State compatibility ("terraform_version" field)
+
+At the time of writing this RFC, OpenTofu 1.7.x has very similar features to Terraform 1.8.x (though not identical). As the "software version" requirements are updated in the previous examples, the exact requirements become murky. A module may require features from "Terraform 1.7.4", but would be perfectly happy with "OpenTofu 1.6.5" (as an example).
+
+> [!NOTE]
+> As we are unable to inspect the Hashicorp Terraform Repository in depth, all version compatibility must be based on Release Notes / Changelog Entries as well as user reports.
+
+## Proposed Solution
```
RLRabinowitz, line 17: Maybe I missed where it is said in this RFC, but I couldn't find any place mentioning that the approach is to ignore the version constraint altogether or not. It's also a bit unclear whether this is the approach in this RFC, or if it is to maintain a compatibility table with Terraform versions

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+```hcl
+# version.tofu
+tofu {
+    required_version = "1.6.5"
+}
+```
+
+When loaded by:
+* Terraform, the `version.tofu` file would be ignored and the "Terraform Required Version" would be 1.7.4.
+* OpenTofu, the `version.tofu` file would be loaded instead of `version.tf` and the "OpenTofu Required Version" would be 1.6.5.
```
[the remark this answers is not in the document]
cam72cam: Is there any reason that we would not want to produce the warning below when someone uses `terraform -> required_version` even when `tofu -> required_version` has been specified?

rfc/20240610-Tofu-Version-Compatibility.md
cam72cam: Correct

rfc/20240610-Tofu-Version-Compatibility.md
cam72cam: @janosdebugs above reminded us that the TSC does not want a version compatibility table.   Perhaps we could link to the migration docs instead?

rfc/20240610-Tofu-Version-Compatibility.md
cam72cam: Hopefully clarified in b3b094db80

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+Either path is easy to both implement and thoroughly test.
+
+The `configs.Module.CheckCoreVersionRequirements()` function is where we will need to emit the warnings and corresponding debug log entries.
+
+### Open Questions
+
+Should there be an option (config, CLI, or ENV) to:
+* Disable the version checking altogether?
+* Don't check terraform required_versions
+* Treat both Tofu and Terraform version as identical instead of using our equivalent "terraform version" guess?
```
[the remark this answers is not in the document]
cam72cam: I struck out this option and added some reasoning in b3b094db80

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,106 @@
...
+
+We have two very similar paths here:
+* Split CoreVersionConstraints into two fields: `TofuVersionConstraints` and `TerraformVersionConstraints`
+* Add a flag into each VersionConstraint that understands if it is Tofu or Terraform specific.
+
+Either path is easy to both implement and thoroughly test.
+
+The `configs.Module.CheckCoreVersionRequirements()` function is where we will need to emit the warnings and corresponding debug log entries.
+
+### Open Questions
```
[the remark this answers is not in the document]
cam72cam: Good catch! b3b094db80

rfc/20240610-Tofu-Version-Compatibility.md
cam72cam: struck out in b3b094db80 with explanation

rfc/20240610-Tofu-Version-Compatibility.md
cam72cam: Noted in b3b094db80

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,109 @@
...
+
+We have two very similar paths here:
+* Split CoreVersionConstraints into two fields: `TofuVersionConstraints` and `TerraformVersionConstraints`
+* Add a flag into each VersionConstraint that understands if it is Tofu or Terraform specific.
+
+Either path is easy to both implement and thoroughly test.
+
+The `configs.Module.CheckCoreVersionRequirements()` function is where we will need to emit the warnings and corresponding debug log entries.
+
+As discussed in the `.tofu` RFC, we will want to quickly work on support for this featu
```
ImIOImI, line 88: Spelling... featu => feature

rfc/20240610-Tofu-Version-Compatibility.md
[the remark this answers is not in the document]
cam72cam: Rewritten

apparentlymart:
I found this older proposal just due to the rebase today, so I'm sorry that I'm coming in late and so might  be covering ground already covered, but I have a "Potential Alternative" that seems worth discussing in the relevant section even if we don't decide to do it:

The following is starting from the idea that for any version number v1.6 or greater there is (more-or-less) no well-defined relationship between Terraform version numbers and OpenTofu version numbers.

Given that, I'd consider trying for a rule that achieves some behavior like the following, for situations where `required_version` is only declared in a `.tf` file and there is no overriding `.tofu` file:

- If the `required_version` only uses version bounds in 1.5.x or earlier, treat it as an OpenTofu version constraint.
- If the `required_version` mentions any version bound 1.6.x or later, assume it's for Terraform and have OpenTofu entirely ignore it.

The above rules are naive and simplistic, and so would probably require some further nuance in practice.

Some common situations that I think _would_ be handled by these rules are:

- Exact version: `1.5.0` (though of course this would always fail in OpenTofu, since there _is_ no OpenTofu 1.5)
- Lower-bound only, acceptable: `>= 1.5.0` (if we assume that OpenTofu remains backward-compatible with Terraform 1.5 and earlier -- which is our goal -- then it's fine for OpenTofu to accept this version constraint as matching all available OpenTofu releases)
- Lower-bound only, ignored: `>= 1.6.0` (assume that this is talking about _Terraform_ 1.6.0, because it's in a `.tf` file, and so this module has effectively no constraint at all on which OpenTofu versions are compatible)
- Pessimistic: `~> 1.5.0` (again, this would always fail in OpenTofu because there are no OpenTofu versions that match this constraint)

The trickier cases are constraints like `>= 1.5.0, < 1.7.0`. What is the author intending to communicate with a version constraint like that?

---

Overall I feel like the risk of ignoring a version constraint we're unsure about might be a little overstated in this proposal. Version constraints like this are -- in public shared modules, at least -- most commonly used to give better feedback when a module is using a feature that would cause some sort of syntax or semantic error in older versions, since "you are using the wrong version" is a clearer signal than a random syntax error.

It is true that sometimes people use `terraform_version` in _root modules_ with an exact constraint or pessimistic constraint to prevent mistakes where operators might use the wrong version. In that situation there _is_ some risk that someone might accidentally use OpenTofu instead of Terraform or vice-versa, but there's only so much we can do about that as long as Terraform itself won't reject a module that's written for OpenTofu.

Given that, I'd personally hope for a solution more like my alternative above where we just make a best effort to interpret what's communicated in a module but "fail open" if a module doesn't seem to have been written with OpenTofu in mind, rather than introducing obscure language features that allow one module to change how another module's version constraints are interpreted.

All of this is opinions weakly held, though. I don't have any major objections to the proposal, but if we've considered a more "best effort" solution like I described above and have rejected it then I'd like to see some words about _why_ we rejected it in the "Potential Alternatives" section. It might also help to be more explicit about what specific risks we are imagining when asserting that this is "potentially dangerous option", since not all dangers are equal in weight.

cam72cam:
@apparentlymart you are hitting this RFC at just the right time, no need to apologize.  I ended up rewriting the RFC this morning based on the original conversation and lessons learned since.

I didn't explicitly say this in the proposal, but my potentially naive approach would be to only flag "Terraform Versions" if the required_version check failed.  I think that in practice, it would behave similarly to what you have described above.

I see it more as a flag that says "I know that there's a `Terraform Version` that's invalid within my project / module and I'm choosing to tell opentofu that it's not".  It would also not impact the root module's direct version constraint check (I think).

This kind of gets to my thoughts toward the end on of this should just be enabled by default.  As there is no well defined relationship between terraform and opentofu versions, the "Terraform Version" in the "OpenTofu" project doesn't really have any meaning.

My hesitation in just enabling this unconditionally is that some people may? be relying on some perceived feature compatibility in past terraform and opentofu versions.

ghost (changes requested):
I believe this RFC may possibly be using an oversized hammer to hit a tiny nail. The GCP issue is the only one where I personally heard of an incompatible version being declared and that has now been resolved thanks to the `.tofu` file support.

This RFC in its current form also creates a dangerous precedent to override how a module works from the outside (aspect-oriented programming style), which we have consistently refused and cautioned against in the past. This RFC would present a reason to reopen the AOP issues.

Before we commit to this course of action we should do some research to figure out how many modules declare a 1.6+ Terraform version at all. The data should be easy to obtain, the repositories are cloned as part of the search UI and we could index the required version into the dataset. I'm happy to go and do this research before adding more complexity to OpenTofu.

Personally, I would rather encourage module authors to declare a separate `.tofu` version instead of this RFC.

cam72cam:
@abstractionfactory I've got a local checkout I can dig though and post a report here.

cam72cam:
Here are the stats:
> Total: 33119, preRelease: 15309, preFork: 30493, postFork: 31182, unknown: 6
> preReleaseOnly: 1330, preForkOnly: 593, postForkOnly: 1290

Here is a list of all of the modules that have upgraded to post fork requirements:
https://gist.github.com/cam72cam/c9eac230887c53e69ed13ed2f6e2082a

cam72cam:
With that data in mind, it's clear that modules are being updated to newer terraform/tofu versions and we should have something in place to handle this situation.

cam72cam:
I don't follow @abstractionfactory?  Why would we exclude version requirements that can still have critical bugfixes?

cam72cam:
Part of that depends on our release cycle.  910 out of the 1290 post-fork-only versions require 1.9 and would *not* function prior to last week and would have been out of luck (other than a fork) for about 6 months.

Even still, the "Terraform Version" identified there is still not explicitly representing a "Tofu Version" and bug-for-bug compatibility is not certain.  As time goes on it will mean less and less.

I understand your hesitance about "modifying" modules from the outside, and I have expressed similar concerns in the past.  The question in my mind is how we treat an existing feature that is known to be misleading and potentially dangerous.  That's the root of the open question IMO and even if we don't merge this RFC (in any form) I'd like to have a clear understanding (documented) idea of what "Terraform Version" means in the context of OpenTofu

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,232 @@
...
+
+
+From the perspective of a module, root or child, there are two different ways in which "Terraform Version" should be ignored:
+1) Disabled for all child modules of the current module.  This represents someone who has lots of child modules that have not yet added "Tofu Versions" and they are comfortable with the risk of bypassing the "Terraform Version" in all of them (and their children).
+2) Disabled for a specific child module of the current module.  This represents someone who is pulling a small number of child modules that have not yet added "Tofu Versions" and they would like to limit the blast radius of ignoring the "Terraform Version" check in this limited area of the current module.
+
+For for the first scenario, a flag called "bypass_terraform_version_requirement" is added to the `product{}` block.  If set to true, all child modules of the current module will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
```
ollevche, line 44:
```suggestion
For the first scenario, a flag called "bypass_terraform_version_requirement" is added to the `product{}` block.  If set to true, all child modules of the current module will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
```

ollevche:
I would like to summarize the ideas I just read and propose a combined version of what's been discussed already.

Given that we don't have a mapping of tf <> tofu versions higher than the forked one, I agree that we should somehow distinguish between what variant the original author meant exactly (tofu or terraform). This is currently possible via `.tofu` files extension, however, it requires module author to alter the configuration.

I also agree that we shouldn't allow altering module behavior outside of the module itself (i.e. via `bypass_terraform_version_requirement`). I support the idea that after-fork version constraints are irrelevant for OpenTofu if there is no clear sign that author meant tofu variant exactly. That's being said, it seems reasonable to me to start deprecating `terraform` block and introduce a `tofu` one.

The first step could be introducing a CLI flag to ignore the `terraform` block version constraint (in the root and all the child modules) and rely on the one specified in `tofu` block. The check could still provide a warning if there is only `terraform` block version constraint (basically, opt-in version of what Martin described).

I think this is not as risky as just starting to ignore the checks, because user needs to opt-in. Also, it doesn't require module authors to change their configuration immediately.

Once we see wide adoption of `tofu` block, we could completely deprecate `terraform` one to minimize the confusion.

WDYT?

cam72cam:
I think that makes a lot of sense @ollevche and matches my understanding of the issue.  The question would then become what would the opt-in look like?  Off the top of my head we have a flag in the `tofu{}` block, a .tofurc flag, an environment variable, and a cli flag.

I'm kind of partial to the `tofu{}` block solution if we were to go this route.  It would clearly opt a project into the more visible deprecation (warnings for now) treatment of the `terraform {}` block

ollevche:
I think specifying `tofu` block should be treated as opt-in (on per module basis), however it is not going to resolve the issue with the child modules with only `terraform` block. For that case, it shouldn't be in the configuration IMO. CLI flag, env var or .tofurc flag are nice options.

ImIOImI:
Personally speaking, I understand the utility of using a `versions.tofu` file to override version requirements... but I can say from personal experience that overrides in configs very often confuse people. I've laid out some rather simple override patterns to see them become completely confused and ignored by colleagues multiple times.

If we can have all the version requirements in a `.tf` file living side by side with the rest of the tofu code, I think that would be best from a code readability standpoint. IMO, I think that would be a much more explicit signal about what's actually happening.

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,232 @@
...
+
+
+From the perspective of a module, root or child, there are two different ways in which "Terraform Version" should be ignored:
+1) Disabled for all child modules of the current module.  This represents someone who has lots of child modules that have not yet added "Tofu Versions" and they are comfortable with the risk of bypassing the "Terraform Version" in all of them (and their children).
+2) Disabled for a specific child module of the current module.  This represents someone who is pulling a small number of child modules that have not yet added "Tofu Versions" and they would like to limit the blast radius of ignoring the "Terraform Version" check in this limited area of the current module.
+
+For the first scenario, a flag called "bypass_terraform_version_requirement" is added to the `product{}` block.  If set to true, all child modules of the current module will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
```
ollevche, line 44:
```suggestion
For the first scenario, a flag called "bypass_terraform_version_requirement" is added to the `project{}` block.  If set to true, all child modules of the current module will only error on "Tofu Versions" and instead produce warnings for "Terraform Versions".
```

rfc/20240610-Tofu-Version-Compatibility.md
```diff
@@ -0,0 +1,232 @@
...
+
+
+Warning: Ignored terraform version requirement
+  on: something/i/dont/want/to/fork/versions.tf in module "childZ"
+  <suggestion to fix or fork that module>
+
+```
+
+
+#### Product scoped behavior:
```
Gogotchuri, line 115:
```suggestion
#### Module-scoped behavior:
```

Gogotchuri (approved):

cam72cam:
Closing this as subsumed by https://github.com/opentofu/opentofu/pull/3085

PR #1723, opened 2024-06-15
initwd Windows tests fixes

eduzgun:
## Target Release

1.8.0

## Checklist

### Go checklist

 

Related to #1201 

This PR helps resolves the initwd package tests failing on Windows. The errors included Windows filepath errors and a Tempdir Cleanup RemoveAll error. The solution for TempDir RemoveAll Cleanup error was to do `runtime.GC()` which runs a garbage collection and blocks other processes. This worked because it most likely released the open file handles that were not needed anymore. In the `WriteSnapshotToDir` function I also closed the file after creating it.

## Important
Another test was failing which was `TestModuleInstaller_symlink` and I was getting this error:
```
--- FAIL: TestModuleInstaller_symlink (0.01s)
    module_install_test.go:377: failed to copy fixture to temporary directory: symlink ..\child_a\ C:\Users\ED\AppData\Local\Temp\terraform-configload2464638177\modules\child_a: A required privilege is not held by the client.
```
What I had to do is enable developer mode on Windows (11), see https://github.com/Schniz/fnm/issues/338. If it's not possible to enable developer mode in the production environment then we could skip this test for Windows runtime only.
```
    t.Skip("Skipping symlinks test on Windows")
}
```

Tests in this package are passing on Windows and Linux machine
initwd pass

### DIFF
--- a/internal/initwd/from_module_test.go
+++ b/internal/initwd/from_module_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -281,6 +282,7 @@ func TestDirFromModule_rel_submodules(t *testing.T) {
 	}
 	t.Cleanup(func() {
 		os.Chdir(oldDir)
+		runtime.GC()
 	})
 
 	hooks := &testInstallHooks{}
--- a/internal/initwd/from_module_test.go
+++ b/internal/initwd/from_module_test.go
@@ -282,6 +282,8 @@ func TestDirFromModule_rel_submodules(t *testing.T) {
 	}
 	t.Cleanup(func() {
 		os.Chdir(oldDir)
+		// Trigger garbage collection to ensure that all open file handles are closed.
+		// This prevents "TempDir RemoveAll cleanup: The process cannot access the file because it is being used by another process" errors during TempDir cleanup.
 		runtime.GC()
 	})
 
--- a/internal/initwd/from_module_test.go
+++ b/internal/initwd/from_module_test.go
@@ -283,8 +283,11 @@ func TestDirFromModule_rel_submodules(t *testing.T) {
 	t.Cleanup(func() {
 		os.Chdir(oldDir)
 		// Trigger garbage collection to ensure that all open file handles are closed.
-		// This prevents "TempDir RemoveAll cleanup: The process cannot access the file because it is being used by another process" errors during TempDir cleanup.
-		runtime.GC()
+		// This prevents "TempDir RemoveAll cleanup: The process cannot access the file
+		//because it is being used by another process" errors during TempDir cleanup on Windows.
+		if runtime.GOOS == "windows" {
+			runtime.GC()
+		}
 	})
 
 	hooks := &testInstallHooks{}
--- a/internal/initwd/from_module_test.go
+++ b/internal/initwd/from_module_test.go
@@ -283,8 +283,7 @@ func TestDirFromModule_rel_submodules(t *testing.T) {
 	t.Cleanup(func() {
 		os.Chdir(oldDir)
 		// Trigger garbage collection to ensure that all open file handles are closed.
-		// This prevents "TempDir RemoveAll cleanup: The process cannot access the file
-		//because it is being used by another process" errors during TempDir cleanup on Windows.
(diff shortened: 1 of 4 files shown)


internal/initwd/from_module_test.go
```diff
@@ -281,6 +282,7 @@ func TestDirFromModule_rel_submodules(t *testing.T) {
 	}
 	t.Cleanup(func() {
 		os.Chdir(oldDir)
+		runtime.GC()
```
cam72cam, line 285: Can you explain / add a comment why we need manual GC here?

internal/initwd/module_install_test.go
```diff
@@ -739,10 +740,15 @@ func TestModuleInstaller_fromTests(t *testing.T) {
 	_, diags := inst.InstallModules(context.Background(), ".", "tests", false, false, hooks)
 	assertNoDiagnostics(t, diags)
 
+	// Use backslashes (\) for Windows paths in ModuleAddr
+	expectedModuleAddr := "test.tests.main.setup"
+	if runtime.GOOS == "windows" {
+		expectedModuleAddr = "test.tests\\main.setup"
```
cam72cam, line 746: This is very strange, I wonder how `.` is being replaced with `\` in the module install hook

internal/initwd/from_module_test.go
eduzgun: Sure, there are many TempDir RemoveAll cleanup errors that are coming up when running tests on Windows. This is because a file is being written to or created without properly being closed. As a result, TempDir cannot delete the temp files on Windows because they are blocked by another process. `t.cleanup()` is run when the test completes so it doesn't not affect the logic inside the test. By manually calling `runtime.GC()`, it triggers finalisers to run. If a file has no more references and a finaliser is set to close the file then `runtime.GC()` will cause this file handle, that is no longer being used, to close. Then the temporary file in Windows can be deleted.

internal/initwd/module_install_test.go
eduzgun: I agree, I got stuck there looking in the code. Myabe it's the way the testdata`testdata/local-module-from-test` is laid out?

internal/initwd/from_module_test.go
cam72cam: Thanks!

internal/initwd/module_install_test.go
cam72cam: Honestly, this seems like it might actually be a bug.  I'm digging into it a bit further.

cam72cam: I'm looking at ModuleInstaller.installLocalModule and moduleInstallWalker, which leads me to configs.ModuleRequest.Path (joined with '.').

internal/configs/config_build.go:buildTestModules is probably where the req.Path is set and includes some logic around splitting on "/".  I'm guessing that's where the bug is.

eduzgun: Yes you were correct. I've added a commit to this PR which uses filepath instead of path to solve that issue. Did a little test to show that `path.base()` always assumes the file system is UNIX:
```
var a_path_linux string = "bananas/main.tftest.hcl"
var a_path_windows string = "bananas\\main.tftest.hcl"

fmt.Println("The linux:", path.Base(a_path_linux))
fmt.Println("The windows: ", path.Base(a_path_windows))

fmt.Println("The linux filepath base:", filepath.Base(a_path_linux))
fmt.Println("The windows filepath base: ", filepath.Base(a_path_windows))

Output:
The linux: main.tftest.hcl
The windows:  bananas\main.tftest.hcl
The linux filepath base: main.tftest.hcl
The windows filepath base:  main.tftest.hcl
```

eduzgun:
Also, is it a better idea to do 
```
t.Cleanup(func() {
        runtime.GC()
    }
})
```
so we only do garbage collection on Windows? I'm assuming these tests were already passing on Linux. If so, we don't have to touch manual garbage collection for Linux.

eduzgun:
Okay, I've updated it now. Thanks for helping

cam72cam:
@eduzgun it looks like your latest commit is missing the DCO signoff

cam72cam (approved):

cam72cam:
@eduzgun looks like one minor linting issue and then we are good to merge :+1:

PR #1726, opened 2024-06-17
Enhancement: flag to treat warnings as errors

### LINKED ISSUE #757: Add a flag to treat warnings as errors

bunniseng:
Added a pedantic flag which enables warnings to be treated as errors.

Changes
- xx

Resolves #757

## Target Release

1.8.0


bunniseng:
This is now ready for review.

Screenshot shows running a command which produces a warning and an exit code of 0
view-new-normal

Screenshot demonstrates running a command with the pedantic global option toggled which produces an error and an exit code of 1.
view-new-pedantic-mode

I have covered legacy UI and new view components and json output.

bunniseng:
This is now ready for review.

cmd/tofu/commands.go
```diff
@@ -122,233 +142,307 @@ func initCommands(
...
+					Meta: meta,
+				},
 			}, nil
 		},
 
 		"console": func() (cli.Command, error) {
-			return &command.ConsoleCommand{
-				Meta: meta,
+			return &pedanticWrapper{
+				PedanticCommand: &command.ConsoleCommand{
```
cam72cam, line 154: What are your thoughts on something like:
 https://gist.github.com/cam72cam/11b16e6ba97f714e5f42b02664e23171 to reduce the copy/paste changes to commands.go?

bunniseng: cool, this is a really nice generic way to wrap the commands instead of having an inline per-command wrapping, I'll update the code with this approach.

bunniseng: This has been updated.

cam72cam:
@bunniseng now that we have sorted out the expectations of this PR in the linked issue, could you update it for the recent changes to main?  I'm hoping to get this reviewed this week / early next.  Thank you for your patience!

cmd/tofu/main.go
```diff
@@ -521,3 +487,65 @@ func mkConfigDir(configDir string) error {
...
+		// Capture -v and --version as the version option
+		if optionName == "-v" || optionName == "--version" {
+			optionName = optionVersion
+		}
+
+		// We have a few scenarios to cater for to maintain backwards compatibility for options and args processing:
+		// 1. Capture options listed before the command which are not part of the global options as args
+		// 2. Capture the command and options listed after the command as args
+		// 3. Capture the version option as a global option regardless of where it was specified
+		if _, ok := globalOptions[optionName]; !ok && !commandFound || commandFound && optionName != optionVersion {
```
cam72cam, line 532: Can we rename `ok` to `isGlobalOption`?

cmd/tofu/main_test.go
```diff
@@ -70,7 +70,7 @@ func TestMain_cliArgsFromEnv(t *testing.T) {
 			"cli string has blank values",
 			[]string{testCommandName, "bar", "", "baz"},
 			"-foo bar",
-			[]string{"-foo", "bar", "bar", "", "baz"},
```
cam72cam, line 73: Why was this test changed?  It appears that this was intentional previously?

cam72cam:
@bunniseng I've been thinking about this approach.  I think it is close, but there are some scenarios which we need to handle better.

Scenario:
```
tofu -pedantic apply -auto-approve
<plan output>
<plan warning>
(should exit 1 here but does not!)
<apply output>
exit 1 due to previous warnings.
```

I think there are two solutions to this problem:

## Add helper to view for enhanced diags.HasErrors()

Before:
```go
if diags.HasErrors() {
    view.ShowDiagnostics(diags)
    return 1
}
```

After:
```go
if view.HasErrors(diags) {
    view.ShowDiagnostics(diags)
    return 1
}
```

## Add wrapper around view.ShowDiagnostics() that handles error codes

Before:
```go
if diags.HasErrors() {
    view.ShowDiagnostics(diags)
    return 1
}
```

After:
```go
// HandleDiags(diags, error code if needed)
if code := view.HandleDiagnostics(diags, 1); code != 0 {
    return code
}
```



Both of these approaches pull in some similar concepts from your original PR.  The idea here is that we only want to do this warning -> error check in the command package.  In the other internal packages it could change the logic in extreme ways.  In the command package it's much easier to understand / validate the repercussions.


Additionally, both of these approaches remove the need for PedanticRunner/PedanticCommand and the deeply nested coupling that results from it.

I'm going to ping the @opentofu/core-engineers to also weigh in on this.

Thank you again for your incredible patience with us as we work through this issue with you!

cam72cam:
@bunniseng sorry for the delay, I've spoken with the core engineers about this and they would be happy with either approach I proposed.  I think after those changes, we should be close to merging this in finally!  Thank you again for your patience!

cam72cam:
@bunniseng looks like good progress, let me know when you would like me to re-review!

cmd/tofu/main_test.go
cam72cam: I think it is valid, as you can run:

`tofu -foo bar bar '' baz`

internal/backend/remote/backend.go
```diff
@@ -759,6 +759,10 @@ func (b *Remote) fetchWorkspace(ctx context.Context, organization string, name s
 
 // Operation implements backend.Enhanced.
 func (b *Remote) Operation(ctx context.Context, op *backend.Operation) (*backend.RunningOperation, error) {
+	if op.View == nil {
+		panic("Operation called with nil View")
```
cam72cam, line 763: Why is this panic here?

internal/cloud/backend.go
```diff
@@ -778,6 +778,10 @@ func (b *Cloud) StateMgr(name string) (statemgr.Full, error) {
 
 // Operation implements backend.Enhanced.
 func (b *Cloud) Operation(ctx context.Context, op *backend.Operation) (*backend.RunningOperation, error) {
+	if op.View == nil {
+		panic("Operation called with nil View")
```
cam72cam, line 782: We should not have panics strewn throughout the code.

internal/command/views/view.go
```diff
@@ -136,6 +151,23 @@ func (v *View) Diagnostics(diags tfdiags.Diagnostics) {
 	}
 }
 
+// HasErrors accepts a set of Diagnostics and interrogates both the View and Diagnostics returning whether an error
+// has occurred or not.
+func (v *View) HasErrors(diags tfdiags.Diagnostics) bool {
+	if v.PedanticWarningFlagged {
+		// If a pedantic warning has been flagged then we report the view is in error state.
+		// PedanticWarningFlagged can be set from legacy UI / diagnostics rendering and new view diagnostics rendering.
```
cam72cam, line 159: I don't think we need to track PedanticWarningFlagged at all given the latest iteration of this PR?

I think we can remove it and simplify quite a lot of code?  Curious on your thoughts @bunniseng.

cam72cam: Digging a bit further into this, it appears you are trying to handle:
```
internal/backend/backend.go:            o.View.Diagnostics(diags)
internal/backend/local/backend.go:                      view.Diagnostics(diags)
internal/backend/local/backend_plan.go:                 op.View.Diagnostics(diags)
internal/backend/local/backend_refresh.go:                      op.View.Diagnostics(diags)
internal/backend/local/backend_apply.go:                        op.View.Diagnostics(diags)
internal/backend/local/backend_apply.go:                                op.View.Diagnostics(diags)
internal/backend/local/backend_apply.go:        op.View.Diagnostics(diags)
```

Is that correct?

ollevche: After reviewing this PR, I see only two places where legacy `Warn` function call is actually should be checked for pedantic (feel free to point me if I miss something):

* [warning in cloud backend](https://github.com/bunniseng/opentofu/blob/feature/757_warnings_to_errors_new/internal/cloud/backend_run_warning.go#L40)
* [warning in init command](https://github.com/bunniseng/opentofu/blob/feature/757_warnings_to_errors_new/internal/command/init.go#L880)

Can we handle these cases separately somehow (put messages as diags?) and then remove `PedanticWarningFlagged`? What do you all think?

ollevche (commented):
Hey @bunniseng, I appreciate your patience on this PR, it is a lot of work! Thank you 🙏 

Overall it looks pretty good, however, I agree we should consider simplifying `PedanticWarningFlagged` with `pedanticUI` and all the relevant checks. I put a comment under the respective thread with more thoughts.

Also, I think we should consider adding an e2e test for existing behaviour and the new one (with pedantic).

internal/command/views/view.go
ollevche: Just a small note on the [Warn call in cloud command](https://github.com/bunniseng/opentofu/blob/feature/757_warnings_to_errors_new/internal/command/cloud.go#L95). We have `return 1` after that call, so exit code is always non-zero in this case.

ollevche: I am in favor of removing it and refactoring those few Warn calls. I might be missing some of the command package quirks so let's see what @cam72cam thinks about the whole thread.

cam72cam:
@bunniseng this week got away from me.  I will try to take a review pass tonight, but it might slip to Monday.

cmd/tofu/main.go
```diff
@@ -287,15 +292,22 @@ func realMain() int {
...
+
+	// Attach the help option to the command args to activate help if it has been toggled
+	if _, isGlobalOptionSet := opts[optionHelp]; isGlobalOptionSet {
+		args = append(args, optionHelp)
+	}
+
+	// Attach the pedantic option to the command args to activate pedantic mode if it has been toggled
+	if _, isGlobalOptionSet := opts[optionPedantic]; isGlobalOptionSet {
+		args = append(args, optionPedantic)
 	}
```
apparentlymart, line 311: I think _appending_ here is a little risky because IIRC the option parsing done by many commands (using [the Go stdlib `flag` package](https://pkg.go.dev/flag)) expects to find all of the option arguments at the start of the argument slice, before any positional arguments.

That is why `mergeEnvArgs` is going to so much trouble to find a suitable place to insert the extra environment-variable-provided options, rather than just appending them all at the end:

https://github.com/opentofu/opentofu/blob/2424a632bef4933c0b7bf910a40fb59b12753e3b/cmd/tofu/main.go#L376-L418

Perhaps we could adapt that existing code to also support _this_ case of adding an extra option into the argument slice. Although the existing precedent for handling stuff like this is to pass another argument to `initCommands` above, and have that function directly set a field in the `command.Meta` object so that the `command` package code doesn't need to re-parse all this stuff:

https://github.com/opentofu/opentofu/blob/2424a632bef4933c0b7bf910a40fb59b12753e3b/cmd/tofu/main.go#L252-L256

(Notice that the `originalWd` argument is part of the behavior of the `-chdir` global option, which is most direct existing precedent for handling those directly rather than indirectly via re-parsing of the command line.)

Adapting the function that deals with the environment variables is probably the least disruptive change to what this PR is already doing, so I'm not opposed to that if it seems like that would get this done without having to significantly refactor the existing work. The main thing, I think, is to ensure that the additional options get added immediately after the subcommand, like `{"plan", "-pedantic", "whatever"}` instead of `{"plan", "whatever", "-pedantic"}`.

apparentlymart (commented):
@cam72cam asked me to take a closer look at the `package main` changes in particular, so my comment below was the result of that. I've not looked closely at the rest of this since others are already reviewing it.

Aside from this one quirk I think the changes here make sense and it's nice to finally generalize the way that global options are handled, vs. them each being a special case as before.

apparentlymart:
I've added the "blocked" label to this just to represent that (as discussed above) we're intending to write an RFC for this behavior before continuing with it. Hopefully this label will make the need for that RFC work more obvious when we [filter issues/PRs by the "blocked" label](https://github.com/opentofu/opentofu/labels/blocked).

Sorry again for the delay! We do still have this on the list for the core team to revisit but unfortunately we've not got to it yet. :confounded:

PR #1728, opened 2024-06-17
allow static evaluations in encryption configuration

### LINKED ISSUE #1538: Cannot use variable as input for passphrase
### Did you test the provider-defined functions?

Did not test

### Additional feedback / details on the provider-defined functions

_No response_

### Did you test the `for_each` on the import block?

Did not test

### Additional feedback / details on the import block

_No response_

### Did you test the state encryption?

Tested, did not work / had problems

### Additional feedback / details on the state encryption

Using this in a terraform file:

```
variable "STATE_ENCRYPTION_KEY" {
  type = string
}

terraform {
  encryption {
    key_provider "pbkdf2" "my_passphrase" {
      passphrase = var.STATE_ENCRYPTION_KEY
    }
    method "aes_gcm" "my_method" {
      keys = key_provider.pbkdf2.my_passphrase
    }
    method "unencrypted" "migration" {
    }
    state {
      method = method.aes_gcm.my_method
      #enforced = true
      fallback {
        method = method.unencrypted.migration
      }
    }
  }
}
```

yields

```
│ Error: Invalid key_provider reference
│
│   on main.tf line 21, in terraform:
│   21:       passphrase = var.STATE_ENCRYPTION_KEY
│
│ Expected reference in form key_provider.type.name
```

if i hardcode the passphrase it works.

### Did you test the removed block?

Did not test

### Additional feedback / details on the removed block

_No response_

### Did you test the testing feature changes?

Did not test

### Additional feedback / details on the testing feature changes

_No response_

### Did you test the built-in functions?

Did not test

### Additional feedback / details on the built-in functions

_No response_

### Did you test the CLI changes?

Did not test

### Additional feedback / details on the CLI changes

_No response_

### Would you like to add anything else?

_No response_

ollevche:
Resolves #1538.

This PR is waiting on https://github.com/opentofu/opentofu/pull/1718 to be merged.

## Target Release

1.8.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/module.go
+++ b/internal/configs/module.go
@@ -64,6 +64,9 @@ type Module struct {
 	// IsOverridden indicates if the module is being overridden. It's used in
 	// testing framework to not call the underlying module.
 	IsOverridden bool
+
+	// TODO: is it a good idea to put StaticEvaluator inside a Module?
+	StaticEvaluator *StaticEvaluator
 }
 
 // File describes the contents of a single configuration file.
@@ -186,20 +189,20 @@ func NewModule(primaryFiles, overrideFiles []*File, call StaticModuleCall, sourc
 	}
 
 	// Static evaluation to build a StaticContext now that module has all relevant Locals / Variables
-	eval := NewStaticEvaluator(mod, call)
+	mod.StaticEvaluator = NewStaticEvaluator(mod, call)
 
 	// If we have a backend, it may have fields that require locals/vars
 	if mod.Backend != nil {
 		// We don't know the backend type / loader at this point so we save the context for later use
-		mod.Backend.Eval = eval
+		mod.Backend.Eval = mod.StaticEvaluator
 	}
 	if mod.CloudConfig != nil {
-		mod.CloudConfig.eval = eval
+		mod.CloudConfig.eval = mod.StaticEvaluator
 	}
 
 	// Process all module calls now that we have the static context
 	for _, mc := range mod.ModuleCalls {
-		mDiags := mc.decodeStaticFields(eval)
+		mDiags := mc.decodeStaticFields(mod.StaticEvaluator)
 		diags = append(diags, mDiags...)
 	}
 
--- a/internal/configs/module.go
+++ b/internal/configs/module.go
@@ -65,7 +65,7 @@ type Module struct {
 	// testing framework to not call the underlying module.
 	IsOverridden bool
 
-	// TODO: is it a good idea to put StaticEvaluator inside a Module?
+	// StaticEvaluator is used to evaluate static expressions in the scope of the Module.
 	StaticEvaluator *StaticEvaluator
 }
 
--- a/internal/configs/static_evaluator.go
+++ b/internal/configs/static_evaluator.go
@@ -11,7 +11,6 @@ import (
 	"github.com/hashicorp/hcl/v2/hcldec"
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/lang"
-	"github.com/opentofu/opentofu/internal/tfdiags"
 	"github.com/zclconf/go-cty/cty"
 )
 
@@ -119,6 +118,7 @@ func (s StaticEvaluator) DecodeBlock(body hcl.Body, spec hcldec.Spec, ident Stat
 	return val, diags
 }
 
-func (s StaticEvaluator) EvalContext(ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
-	return s.scope(ident).EvalContext(refs)
+func (s StaticEvaluator) EvalContext(ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, hcl.Diagnostics) {
+	evalCtx, diags := s.scope(ident).EvalContext(refs)
+	return evalCtx, diags.ToHCL()
 }
--- a/internal/configs/static_evaluator.go
+++ b/internal/configs/static_evaluator.go
@@ -119,6 +119,10 @@ func (s StaticEvaluator) DecodeBlock(body hcl.Body, spec hcldec.Spec, ident Stat
 }
 
 func (s StaticEvaluator) EvalContext(ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, hcl.Diagnostics) {
-	evalCtx, diags := s.scope(ident).EvalContext(refs)
+	return s.EvalContextWithParent(nil, ident, refs)
+}
+
+func (s StaticEvaluator) EvalContextWithParent(parent *hcl.EvalContext, ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, hcl.Diagnostics) {
+	evalCtx, diags := s.scope(ident).EvalContextWithParent(parent, refs)
 	return evalCtx, diags.ToHCL()
 }
--- a/internal/configs/static_evaluator.go
+++ b/internal/configs/static_evaluator.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/hcl/v2/hcldec"
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/lang"
+	"github.com/opentofu/opentofu/internal/tfdiags"
 	"github.com/zclconf/go-cty/cty"
 )
 
@@ -116,4 +117,8 @@ func (s StaticEvaluator) DecodeBlock(body hcl.Body, spec hcldec.Spec, ident Stat
 	val, valDiags := hcldec.Decode(body, spec, ctx)
 	diags = append(diags, valDiags...)
 	return val, diags
+}
+
+func (s StaticEvaluator) EvalContext(ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
+	return s.scope(ident).EvalContext(refs)
 }
--- a/internal/encryption/keyprovider.go
+++ b/internal/encryption/keyprovider.go
@@ -10,7 +10,10 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/configs"
 	"github.com/opentofu/opentofu/internal/encryption/config"
+	"github.com/opentofu/opentofu/internal/lang"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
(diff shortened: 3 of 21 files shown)


internal/configs/module.go
```diff
@@ -64,6 +64,9 @@ type Module struct {
 	// IsOverridden indicates if the module is being overridden. It's used in
 	// testing framework to not call the underlying module.
 	IsOverridden bool
+
+	// TODO: is it a good idea to put StaticEvaluator inside a Module?
+	StaticEvaluator *StaticEvaluator
```
cam72cam, line 69: This is what I was leaning towards, though it might not be the most elegant solution.

The only other solution I can come up with is to construct a new static evaluator from the root config's module in meta_encryption.go

internal/encryption/keyprovider.go
```diff
@@ -88,76 +91,57 @@ func (e *targetBuilder) setupKeyProvider(cfg config.KeyProviderConfig, stack []c
...
-
-		for _, kpc := range e.cfg.KeyProviderConfigs {
-			// Find the key provider in the config
-			if kpc.Type == depType && kpc.Name == depName {
-				depDiags := e.setupKeyProvider(kpc, stack)
-				diags = append(diags, depDiags...)
-				break
-			}
-		}
+	refs, refDiags := lang.References(addrs.ParseRef, deps)
```
cam72cam, line 118: This removes the unused functionality that a key_provider can depend on another key provider?

internal/encryption/keyprovider.go
```diff
@@ -88,76 +91,57 @@ func (e *targetBuilder) setupKeyProvider(cfg config.KeyProviderConfig, stack []c
...
+	diags = append(diags, refDiags.ToHCL()...)
+	if diags.HasErrors() {
+		return diags
 	}
+
+	// TODO: Is StaticIdentifier correctly composed?
```
cam72cam, line 124: Yes, it could be `Name: "encryption.keyprovider.%s"` for more clarity?

internal/encryption/keyprovider.go
ollevche: No, I just removed the check that validates all the deps are refs to key providers.

If we want to remove the code relevant to provider chaining, the refactoring should be done in a separate PR.

ollevche: Oh, I missed the line where we call setupKeyProvider. I'll get it back slightly changing the validation logic

internal/configs/module.go
ollevche: I think `Module` is a better fit for `StaticEvaluator` since these values are more coherent.

internal/encryption/keyprovider.go
ollevche: I fixed this accidental removal in 2766e27

internal/encryption/keyprovider.go
ollevche: Thanks! Changed in https://github.com/opentofu/opentofu/pull/1728/commits/2766e2791383b3836f70c88d8f2da6393cd2dbc6

internal/encryption/keyprovider.go
cam72cam: I think you need to filter out deps that are not key_providers here.  lang.References will probably break on those traversals.

ollevche: Added in [0a61405](https://github.com/opentofu/opentofu/pull/1728/commits/0a61405e6b96fbe5db7c9e717678ae96625b9f99)

website/docs/language/state/examples/encryption/fallback_from_unencrypted.tf
```diff
@@ -1,12 +1,17 @@
...
 terraform {
   encryption {
     ## Step 1: Add the unencrypted method:
     method "unencrypted" "migrate" {}
 
     ## Step 2: Add the desired key provider:
     key_provider "pbkdf2" "mykey" {
-      # Change this to be at least 16 characters long:
-      passphrase = "changeme!"
+      ## You can also use variables here!
```
cam72cam, line 13: Maybe we could change these examples to use variables instead of locals?

internal/configs/static_evaluator.go
```diff
@@ -116,3 +117,7 @@ func (s StaticEvaluator) DecodeBlock(body hcl.Body, spec hcldec.Spec, ident Stat
 	diags = append(diags, valDiags...)
 	return val, diags
 }
+
+func (s StaticEvaluator) EvalContext(ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
```
cam72cam, line 121: Most functions on this struct return hcl.Diagnostics.  Perhaps we want to to the conversion here?

internal/encryption/keyprovider.go
```diff
@@ -130,23 +133,18 @@ func (e *targetBuilder) setupKeyProvider(cfg config.KeyProviderConfig, stack []c
...
-		for _, kpc := range e.cfg.KeyProviderConfigs {
-			// Find the key provider in the config
-			if kpc.Type == depType && kpc.Name == depName {
-				depDiags := e.setupKeyProvider(kpc, stack)
-				diags = append(diags, depDiags...)
-				break
-			}
+		kpc, ok := e.cfg.GetKeyProvider(depType, depName)
+		if !ok {
+			nonKeyProviderDeps = append(nonKeyProviderDeps, dep)
```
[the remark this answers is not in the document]
ollevche: Nice catch! Before that change, such refs were just ignored. I think it should be an error if no key providers found

internal/configs/static_evaluator.go
ollevche: Fixed in [0abfc55](https://github.com/opentofu/opentofu/pull/1728/commits/0abfc55900be650c85710f65b0c67f614976345d)

internal/encryption/keyprovider.go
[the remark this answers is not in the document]
ollevche: Add an error in [a6b8937](https://github.com/opentofu/opentofu/pull/1728/commits/a6b89378d2a7586ba4f59c84e215b7e24aa1c8fe)

website/docs/language/state/examples/encryption/fallback_from_unencrypted.tf
ollevche: Yep, changed back to var in [c821140](https://github.com/opentofu/opentofu/pull/1728/commits/c8211407d15e5356f837ef04376679389b56e3f0)

internal/encryption/keyprovider.go
```diff
@@ -130,23 +133,22 @@ func (e *targetBuilder) setupKeyProvider(cfg config.KeyProviderConfig, stack []c
...
+
+		kpc, ok := e.cfg.GetKeyProvider(depType, depName)
+		if !ok {
 			diags = append(diags, &hcl.Diagnostic{
 				Severity: hcl.DiagError,
-				Summary:  "Invalid key_provider reference",
-				Detail:   "Expected reference in form key_provider.type.name",
-				Subject:  dep.SourceRange().Ptr(),
+				Summary:  "Undefined Key Provider",
+				Detail:   fmt.Sprintf("Key provider %s.%s not found in the encryption configuration. Please, ensure you are referencing previously configured key provider.", depType, depName),
```
cam72cam, line 145:
```suggestion
Detail:   fmt.Sprintf("Key provider %s.%s is missing from the encryption configuration.", depType, depName),
Subject:  dep.SourceRange().Ptr(),
```

internal/encryption/targets_test.go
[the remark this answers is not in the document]
ollevche: Thanks, added in [a4f25c2](https://github.com/opentofu/opentofu/pull/1728/commits/a4f25c288cf40005fb55b8f6be479d540716acc2)

internal/encryption/keyprovider.go
```diff
@@ -156,8 +158,30 @@ func (e *targetBuilder) setupKeyProvider(cfg config.KeyProviderConfig, stack []c
...
+	diags = append(diags, evalDiags...)
+	if diags.HasErrors() {
+		return diags
+	}
+
+	// Injecting configured encryption values
+	// into the evalCtx to be usable.
+	evalCtx = evalCtx.NewChild()
+	evalCtx.Functions = e.ctx.Functions
+	evalCtx.Variables = e.ctx.Variables
```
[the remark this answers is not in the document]
ollevche: Yep, but `EvalContext` produced by `StaticEvaluator` is dependent on references specific to each key provider. So we can't create one beforehand and then populate it.

However, I was initially thinking about making an encryption `EvalContext` to be a parent of `StaticEvaluator`'s `EvalContext`. It now seems easier and more convenient to me so I added `EvalContextWithParent` function ([7a06a23](https://github.com/opentofu/opentofu/pull/1728/commits/7a06a23077aeb0ee876a4443f564ee08e4306130)) with relevant tests ([9f24e5a](https://github.com/opentofu/opentofu/pull/1728/commits/9f24e5a1773dacbb527a82b833d0227bed8bafa9)).

I may be missing something, so I'm happy to learn about alternatives.

cam72cam (dismissed):

internal/lang/eval.go
```diff
@@ -289,21 +289,32 @@ func (s *Scope) EvalReference(ref *addrs.Reference, wantType cty.Type) (cty.Valu
 // this type offers, but this is here for less common situations where the
 // caller will handle the evaluation calls itself.
 func (s *Scope) EvalContext(refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
-	return s.evalContext(refs, s.SelfAddr)
+	return s.evalContext(nil, refs, s.SelfAddr)
 }
 
-func (s *Scope) evalContext(refs []*addrs.Reference, selfAddr addrs.Referenceable) (*hcl.EvalContext, tfdiags.Diagnostics) {
+// EvalContextWithParent is exactly the same as EvalContext except the resulting hcl.EvalContext
+// will be derived from the given parental hcl.EvalContext.
```
RLRabinowitz, line 296: nit: I wonder if we should have a few more words here explaining what this actually means and does in effect.
Even `EvalContext`'s comments on `parent` and `NewChild` don't really explain what it means to use them

internal/configs/static_evaluator.go
```diff
@@ -116,3 +116,12 @@ func (s StaticEvaluator) DecodeBlock(body hcl.Body, spec hcldec.Spec, ident Stat
...
 }
+
+func (s StaticEvaluator) EvalContext(ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, hcl.Diagnostics) {
+	return s.EvalContextWithParent(nil, ident, refs)
+}
+
+func (s StaticEvaluator) EvalContextWithParent(parent *hcl.EvalContext, ident StaticIdentifier, refs []*addrs.Reference) (*hcl.EvalContext, hcl.Diagnostics) {
+	evalCtx, diags := s.scope(ident).EvalContextWithParent(parent, refs)
+	return evalCtx, diags.ToHCL()
+}
```
RLRabinowitz, line 127: Just a random thought, could probably just ignore and resolve.

I am now contemplating what's the best approach for `StaticEvaluator`, should we keep creating convenience functions like `DecodeExpression` and `DecodeBlock` or not?

Here, we could technically use a convenience method for `DecodeBody`, pretty similarly to the other methods here. The two exceptions being that 1. We do not want to use all references from the body. 2. We want to have the original eval context as parent

I guess we can just go ahead and see how `StaticEvaluator` starts shaping itself

internal/lang/eval_test.go
```diff
@@ -421,6 +421,47 @@ func TestScopeEvalContext(t *testing.T) {
 	}
 }
 
+// TestScopeEvalContextWithParent tests if the resulting EvalCtx has correct parent.
```
RLRabinowitz, line 424: nit: Technically the tests here probably should check that the EvalContext still contains the necessary vars and functions

RLRabinowitz (approved):

internal/configs/static_evaluator.go
ollevche: I don't have a reasoned answer here. I think the best we can do is to see where the `StaticEvaluator` goes and refactor it at some point.

internal/lang/eval.go
ollevche: I extended the comment to describe a little more and also named a few examples: [c3e8fda](https://github.com/opentofu/opentofu/pull/1728/commits/c3e8fda7caf88a09f6e832ab705740c8da33faf6).

internal/lang/eval_test.go
ollevche: I agree, I extended the test with value-based checks for Variables and Functions: [fa3d716](https://github.com/opentofu/opentofu/pull/1728/commits/fa3d716b993c49a1f814dd35d2dfce6a80741b51)

cam72cam (approved):

PR #1809, opened 2024-07-11
Improve ModuleExpansionTransformer performance

jonjohnsonjr:
The existing implementation took a very long time given a large number of Vertices or Modules in a graph.

This constructs a tree that can quickly return the set of vertices in a given module rather than iterating over every Vertex for every Module.

We have a fairly unusual graph with an obscene number of modules and vertices, so this change makes a fairly substantial impact on our performance.

Before: https://share.firefox.dev/3Y0qu9h
After: https://share.firefox.dev/4bHYBGd

Time spent in ModuleExpansionTransformer drops from ~27s to ~1s.

I haven't added any new tests because (in theory) this should not affect any functionality, just performance, but let me know if I should add some.

This `pathTree` implementation is not exposed to anything because I didn't want to change any interfaces, but it might make sense for [`tofu.Graph`](https://github.com/opentofu/opentofu/blob/f668c48ffd57d1f5f5dcd7543f179b373643081a/internal/tofu/graph.go#L23) to implement something like this? I would personally wait until we find other instances where we want vertex lookup by module before changing any APIs, but I'd defer to maintainers.

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/transform_module_expansion.go
+++ b/internal/tofu/transform_module_expansion.go
@@ -33,11 +33,21 @@ type ModuleExpansionTransformer struct {
 
 func (t *ModuleExpansionTransformer) Transform(g *Graph) error {
 	t.closers = make(map[string]*nodeCloseModule)
+
+	// Construct a tree for fast lookups of Vertices based on their ModulePath.
+	tree := &pathTree{
+		children: make(map[string]*pathTree),
+	}
+
+	for _, v := range g.Vertices() {
+		tree.addVertex(v)
+	}
+
 	// The root module is always a singleton and so does not need expansion
 	// processing, but any descendent modules do. We'll process them
 	// recursively using t.transform.
 	for _, cfg := range t.Config.Children {
-		err := t.transform(g, cfg, nil)
+		err := t.transform(g, cfg, tree, nil)
 		if err != nil {
 			return err
 		}
@@ -75,9 +85,19 @@ func (t *ModuleExpansionTransformer) Transform(g *Graph) error {
 	// Modules implicitly depend on their child modules, so connect closers to
 	// other which contain their path.
 	for _, c := range t.closers {
-		for _, d := range t.closers {
-			if len(d.Addr) > len(c.Addr) && c.Addr.Equal(d.Addr[:len(c.Addr)]) {
-				g.Connect(dag.BasicEdge(c, d))
+		// For a closer c with address ["module.foo", "module.bar", "module.baz"],
+		// we'll look up all potential parent modules:
+		//
+		// - t.closers["module.foo"]
+		// - t.closers["module.foo.module.bar"]
+		//
+		// And connect the parent module to c.
+		//
+		// We skip i=0 because c.Addr[0:0] == [], and the root module should not exist in t.closers.
+		for i := 1; i < len(c.Addr); i++ {
+			parentAddr := c.Addr[0:i].String()
+			if parent, ok := t.closers[parentAddr]; ok {
+				g.Connect(dag.BasicEdge(parent, c))
 			}
 		}
 	}
@@ -85,7 +105,7 @@ func (t *ModuleExpansionTransformer) Transform(g *Graph) error {
 	return nil
 }
 
-func (t *ModuleExpansionTransformer) transform(g *Graph, c *configs.Config, parentNode dag.Vertex) error {
+func (t *ModuleExpansionTransformer) transform(g *Graph, c *configs.Config, tree *pathTree, parentNode dag.Vertex) error {
 	_, call := c.Path.Call()
 	modCall := c.Parent.Module.ModuleCalls[call.Name]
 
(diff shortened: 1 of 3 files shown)


cam72cam:
Hi @jonjohnsonjr, thanks for the PR!  There is a lot of low hanging fruit ripe to be optimized in the graph transformation.

> graph with an obscene number of modules and vertices

Are you able to build a test scenario with a provider like `tfcoremock`?  As we review more of these much needed optimization PRs, we are starting to build up a suite of performance tests to simulate some of the larger and more complex configurations in use.

jonjohnsonjr:
> Are you able to build a test scenario with a provider like tfcoremock? As we review more of these much needed optimization PRs, we are starting to build up a suite of performance tests to simulate some of the larger and more complex configurations in use.

I'll try to put something together, it might take me a bit.

jonjohnsonjr:
Please forgive my ignorance here, I couldn't figure out a way to easily generate addressable resources without resorting to bash, I assume there's a better way to do this, but:

My `main.tf`:
```
module "generated-1" {
  source = "./generated"
}

module "generated-2" {
  source = "./generated"
}

module "generated-3" {
  source = "./generated"
}

module "generated-4" {
  source = "./generated"
}

module "generated-5" {
  source = "./generated"
}

module "generated-6" {
  source = "./generated"
}

module "generated-7" {
  source = "./generated"
}

module "generated-8" {
  source = "./generated"
}

module "generated-9" {
  source = "./generated"
}

module "generated-10" {
  source = "./generated"
}

```

In `submodule/main.tf`:
```
resource "tfcoremock_simple_resource" "iter" {
  count  = 1
  string = count.index
}
```

I then generated 1000 modules in `./generated/main.tf` that reference `./submodule`:

```bash
mkdir generated
for i in $(seq 1 1000); do echo module \"module-$i\" \{ source = \"../submodule\" \}; done > ./generated/main.tf
```

Before:
```
tofu plan  91.57s user 10.24s system 268% cpu 37.926 total
```

After:
```
tofu plan  57.09s user 9.55s system 352% cpu 18.882 total
```

cam72cam:
Hey, sorry for the late reply!  I'll dig into this tomorrow!  Test scenario looks good!

jonjohnsonjr:
Opened https://github.com/opentofu/opentofu-performance-testing/pull/1

When validating this I also noticed there was another n^2 loop for connecting the closers just below, so I've fixed that as well.

internal/command/terraform.tfstate
```diff
@@ -0,0 +1 @@
+{"version":4,"terraform_version":"1.9.0","serial":0,"lineage":"fake-for-testing","outputs":{"foo":{"value":"bar","type":"string","sensitive":true}},"resources":[],"check_results":null}
```
cam72cam, line 1: Looks like this was accidentally committed.

cam72cam:
Looks like some crashes in the tests

internal/tofu/transform_module_expansion.go
```diff
@@ -75,17 +85,26 @@ func (t *ModuleExpansionTransformer) Transform(g *Graph) error {
...
-			if len(d.Addr) > len(c.Addr) && c.Addr.Equal(d.Addr[:len(c.Addr)]) {
-				g.Connect(dag.BasicEdge(c, d))
+		// For a closer c with address ["module.foo", "module.bar", "module.baz"], we'll look up:
+		//
+		// t.closers["module.foo"]
+		// t.closers["module.foo.module.bar"]
+		//
+		// And connect the parent module to c.
+		//
+		// We skip i=0 because c.Addr[0:0].String() == "".
```
cam72cam, line 95: Perhaps we could clarify this note to mention that the root module ("") will not exist in t.closers and can be skipped?

internal/tofu/transform_module_expansion.go
```diff
@@ -75,17 +85,26 @@ func (t *ModuleExpansionTransformer) Transform(g *Graph) error {
 	// Modules implicitly depend on their child modules, so connect closers to
 	// other which contain their path.
 	for _, c := range t.closers {
-		for _, d := range t.closers {
-			if len(d.Addr) > len(c.Addr) && c.Addr.Equal(d.Addr[:len(c.Addr)]) {
-				g.Connect(dag.BasicEdge(c, d))
+		// For a closer c with address ["module.foo", "module.bar", "module.baz"], we'll look up:
```
cam72cam, line 88: we'll look up all potential parent modules.

internal/command/terraform.tfstate
jonjohnsonjr: Oh how embarrassing, thanks.

internal/tofu/transform_module_expansion.go
jonjohnsonjr: Done.

cam72cam:
Tests look good, I ran them against main for the same effect.

One minor change needed is to add the standard license headers to your test file.

Yantrio:
I'm happy to approve this as long as the license headers are updated 👍

jonjohnsonjr:
> One minor change needed is to add the standard license headers to your test file.

Whoops, copied them over from the test file I cribbed. Thanks!

cam72cam (approved):

jonjohnsonjr:
Friendly ping @Yantrio

cam72cam:
Thanks @jonjohnsonjr !

PR #1810, opened 2024-07-11
Only call StringWithNodeTypes for TRACE logs

jonjohnsonjr:
As it turns out StringWithNodeTypes is a fairly expensive call that happens quite a bit. Unless TF_LOG=TRACE, the result of that call is immediately discarded. This adds a check to avoid computing this at all if the current log level is not TRACE.

For a particularly large graph that I only interact with because someone is paying me to, this saves me about 7s on every invocation of `tofu`.

## Target Release

1.8.0

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ NEW FEATURES:
 
 ENHANCEMENTS:
 * Added `-show-sensitive` flag to tofu plan, apply, state-show and output commands to display sensitive data in output. ([#1554](https://github.com/opentofu/opentofu/pull/1554))
+* Improved performance for large graphs when debug logs are not enabled. ([#1810](https://github.com/opentofu/opentofu/pull/1810))
 
 BUG FIXES:
 * Ensure that using a sensitive path for templatefile that it doesn't panic([#1801](https://github.com/opentofu/opentofu/issues/1801))
--- a/internal/tofu/graph_builder.go
+++ b/internal/tofu/graph_builder.go
@@ -43,11 +43,14 @@ func (b *BasicGraphBuilder) Build(path addrs.ModuleInstance) (*Graph, tfdiags.Di
 		log.Printf("[TRACE] Executing graph transform %T", step)
 
 		err := step.Transform(g)
-		if thisStepStr := g.StringWithNodeTypes(); thisStepStr != lastStepStr {
-			log.Printf("[TRACE] Completed graph transform %T with new graph:\n%s  ------", step, logging.Indent(thisStepStr))
-			lastStepStr = thisStepStr
-		} else {
-			log.Printf("[TRACE] Completed graph transform %T (no changes)", step)
+
+		if logging.IsDebugOrHigher() {
+			if thisStepStr := g.StringWithNodeTypes(); thisStepStr != lastStepStr {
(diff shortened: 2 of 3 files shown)


cam72cam:
Hi @jonjohnsonjr, this looks like a reasonable deduction and proposed fix.  If we find that opting out of log message computation is a common pattern, we will eventually want to move it into a internal library.  For now however, this should be fine.

Is this reproduced with the same test scenario as #1809?

jonjohnsonjr:
> If we find that opting out of log message computation is a common pattern, we will eventually want to move it into a internal library.

Totally reasonable, happy to help with that if I find more instances of this.

> Is this reproduced with the same test scenario as https://github.com/opentofu/opentofu/pull/1809?

Yep! This is from calling `terraform plan` on a very large graph, so I expect it to be less pronounced for most users, but ~10% of my overall execution time was spent on `StringWithNodeTypes`.

Here's a link that filters the flamegraph down to just that function call: https://share.firefox.dev/3zEyOBu

(I'd show an "after", but it just doesn't have `StringWithNodeTypes` anywhere so it's not super interesting.)

jonjohnsonjr:
Looked into this some more and I think it might be worth switching entirely to `log/slog` because the current leveled logging implementation is really unfortunate for perf.

cam72cam:
We discussed this a bit in slack, slog does not have the TRACE level that we need.  If we refactor this, we would probably want to add functions to the logging package as a wrapper for whatever the eventual solution is.

Given that this is the primary hot path and is just two function calls, I think it makes sense to keep this PR simple for now.

I've put together a simple performance scenario which goes from 51s to 36s:
```hcl
cat main.tf 
resource "tfcoremock_simple_resource" "iter" {
        count = 1000
        string = "${count.index}"
}

resource "tfcoremock_simple_resource" "iter2" {
        count = 1000
        string = "${tfcoremock_simple_resource.iter[count.index].string}"
}

```

cam72cam:
I've added this scenario to our performance test repo: https://github.com/opentofu/opentofu-performance-testing/tree/master/large-graph-with-simple-dependencies

@jonjohnsonjr Could you add an entry in the CHANGELOG.md?  You will need to update this branch as we have edited the changelog for 1.9.0

jonjohnsonjr:
Thank you! Added a line to CHANGELOG.md, happy to modify exact wording to be more/less specific if that's too vague.

jonjohnsonjr:
```
--- FAIL: TestInit_cancelModules (0.01s)
    init_test.go:1846: wrong error message
        should contain: Module installation was canceled by an interrupt signal
        got:
        
        Error: Error accessing remote module registry
        
          on main.tf line 1:
           1: module "foo" {
        
        Failed to retrieve available versions for module "foo" (main.tf:1) from
        registry.does.not.exist: failed to request discovery document: Get
        "https://registry.does.not.exist/.well-known/terraform.json": dial tcp:
        lookup registry.does.not.exist: no such host.
```

This seems unrelated to my PR, maybe a flake?

cam72cam:
> This seems unrelated to my PR, maybe a flake?

Yeah, we have a few flaky tests that need to be fixed.  It passed on all other arch's so I'm not concerned.

cam72cam (approved):

cam72cam:
Thanks for your patience @jonjohnsonjr !

PR #1811, opened 2024-07-11
Avoid copies on calls to parser.Sources()

jonjohnsonjr:
This is passing around a `map[string][]byte`, which is fine, but the underlying [implementation](https://github.com/hashicorp/hcl/blob/360ae579460fab69e9939599af85c8f255a59007/hclparse/parser.go#L117-L128) makes a copy every time it's called, and it doesn't appropriately pre-allocate the size of the map to make that more efficient. This generates a lot of garbage, especially given that nothing modifies what is returned by Sources().

Instead, we can call the `Files()` method, which returns a `map[string]*hcl.File`, which still gives us access to the underlying `[]byte` via the `Bytes` field, but doesn't require making a copy of the map (or modifying upstream to fix this).

For my project, this drops the time spent in [`loadModule()`](https://github.com/opentofu/opentofu/blob/f668c48ffd57d1f5f5dcd7543f179b373643081a/internal/configs/config_build.go#L161-L206) from 2.2s to 1.4s.

## Target Release

1.8.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/format/diagnostic.go
+++ b/internal/command/format/diagnostic.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/hashicorp/hcl/v2"
 	viewsjson "github.com/opentofu/opentofu/internal/command/views/json"
 	"github.com/opentofu/opentofu/internal/tfdiags"
 
@@ -31,7 +32,7 @@ var disabledColorize = &colorstring.Colorize{
 // at all. Although the long-form text parts of the message are wrapped,
 // not all aspects of the message are guaranteed to fit within the specified
 // terminal width.
-func Diagnostic(diag tfdiags.Diagnostic, sources map[string][]byte, color *colorstring.Colorize, width int) string {
+func Diagnostic(diag tfdiags.Diagnostic, sources map[string]*hcl.File, color *colorstring.Colorize, width int) string {
 	return DiagnosticFromJSON(viewsjson.NewDiagnostic(diag, sources), color, width)
 }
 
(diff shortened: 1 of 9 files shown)


cam72cam:
Hi @jonjohnsonjr, how large of a configuration are we talking?  500, 5k, 50k files?

jonjohnsonjr:
Converting to draft because I missed broken tests, me culpa.

> Hi @jonjohnsonjr, how large of a configuration are we talking? 500, 5k, 50k files?

In total "only" 6161 files, but the `Sources()` method is called 5372 times, and the number of invocations grows ~linearly with the number of files (I'm less confident in this but it seems approximately correct in my case). Plotting `len(p.Sources()` looks like:

image

(X axis is [0, 5372] corresponding to Nth invocation)

Prior to this change, `Sources()` runtime increased with the number of files due to the copy, so the overall runtime is ~O(n^2) with respect to the number of source files.

After this change, `Sources()` is constant (effectively zero, since it's just an accessor), so it drops quite a bit.

jonjohnsonjr:
>   internal/command/views/json/diagnostic.go:136: Function 'NewDiagnostic' has too many statements (77 > 50) (funlen)

I am happy to refactor `NewDiagnostic` a bit to get under this lint threshold and/or `nolint` it (given that it was already over the threshold prior to my change). Also happy to send that refactoring as a separate PR, let me know what y'all would prefer.

cam72cam:
At this point, I'd recommend `nolint` and would definitely welcome a followup PR (though not required).

jonjohnsonjr:
> At this point, I'd recommend nolint and would definitely welcome a followup PR (though not required).

Done, will follow-up after.

jonjohnsonjr:
Opened a draft PR here: https://github.com/opentofu/opentofu/pull/1818

cam72cam:
So looking into this a bit further, it seems that it is O(D * F), where D is the number of diagnostics and F is the number of files.  Your change makes it closer to O(D).

Here's my somewhat extreme test scenario: https://github.com/opentofu/opentofu-performance-testing/pull/2

cam72cam (approved):

Yantrio (approved):

PR #1875, opened 2024-07-30
Adds multi line support to console

Andrew-Hayes:
Resolves # 

## Target Release

1.9.0

## Checklist

### Go checklist

 


internal/command/console.go
```diff
@@ -179,26 +179,38 @@ func (c *ConsoleCommand) Run(args []string) int {
...
-			return 0
+		line := strings.TrimSpace(scanner.Text())
+
+		// we update the state with the new line, so if we have open
+		// brackets we know not to execute the command just yet
+		consoleState.UpdateState(line)
+
+		// we check if there is no escaped new line at the end, or any open brackets
+		// if we have neither, then we can execute
+		if !strings.HasSuffix(line, "\\") && consoleState.BracketsOpen() <= 0 && len(line) != 0 {
```
cam72cam, line 195: I need to download and play with it, but I'm curious about the len(line) != 0:
```
> 5 + 4 \\

```

Wouldn't that get stuck?

cam72cam: > 4 + 5
9
> 4 + 5 \\
 
╷
│ Error: Invalid character
│ 
│   on <console-input> line 1:
│   (source code not available)
│ 
│ This character is not used within the language.
╵

╷
│ Error: Extra characters after expression
│ 
│   on <console-input> line 1:
│   (source code not available)
│ 
│ An expression was successfully parsed, but extra characters were found after it.
╵

cam72cam:
Looks like some odd mixed EOL handling:
```
> {
. x = 4
. }
{
  "x" = 4
}
> {
. x = 4 \\
. }
╷
│ Error: Invalid character
│ 
│   on <console-input> line 2:
│   (source code not available)
│ 
│ This character is not used within the language.
╵

╷
│ Error: Missing attribute separator
│ 
│   on <console-input> line 2:
│   (source code not available)
│ 
│ Expected a newline or comma to mark the beginning of the next attribute.
╵


>  
```

internal/command/console_state.go
```diff
@@ -0,0 +1,93 @@
...
+
+// UpdateState updates the state of the console with the latest line data
+func (c *consoleBracketState) UpdateState(line string) {
+	if len(line) == 0 {
+		// we can skip empty lines
+		return
+	}
+	c.buffer = append(c.buffer, strings.TrimSuffix(line, "\\"))
+
+	tokens, _ := hclsyntax.LexConfig([]byte(line), "<console-input>", hcl.Pos{Line: 1, Column: 1})
```
cam72cam, line 60: Here you are not using the trimmed line, causing the error I posted oh the main PR.  the TrimSuffix should be used to update line, not just in the buffer append.

internal/command/console_state.go
```diff
@@ -0,0 +1,93 @@
...
+	// we calculate open brackets, braces and parentheses by the diff between each count
+	var total int
+	total += c.oBrace - c.cBrace
+	total += c.oBracket - c.cBracket
+	total += c.oParentheses - c.cParentheses
+	return total
+}
+
+// UpdateState updates the state of the console with the latest line data
+func (c *consoleBracketState) UpdateState(line string) {
```
cam72cam, line 53: What do you think about encapsulating a lot of the bracket and `\\` functionality in here.  The consumers of this only care about: Is there a command I need to execute, What's my indentation level for the prompt?

```
// Returns command if there is one to execute and the level of indentation needed for the prompt
func (c *consoleBracketState) UpdateState(line string) (string, int) {
```

We could then make BracketsOpen and ClearState internal to the package and not used by consumers.  I'd ideally like to encapsulate as much of this as possible.

internal/command/console.go
Andrew-Hayes: wait... isn't `\\` just escaping the slash, not the new line. So the command we send to execute is `4 + 5` with a random slash after it.. which is invalid..

internal/command/console_state.go
Andrew-Hayes: yeah, i can move bracketsOpen in. However, do we not need ClearState external so the consumer makes the decision when to execute and subsequently when to reset?

Andrew-Hayes: i went with the approach: return the current full command and and open brackets each time we call update. then within update we clear the buffer if open brackets is <= 0

cam72cam (approved):

internal/command/console_state.go
```diff
@@ -0,0 +1,101 @@
...
+// or if any escaped new lines
+// in the console and we should hold off on processing the commands
+// it returns 3 states:
+// -1 is returned the is an incorrect amount of brackets.
+// for example "())" has too many close brackets
+// 0 is returned if the brackets are closed.
+// for examples "()" or "" would be in a close bracket state
+// >=1 is returned for the amount of open brackets.
+// for example "({" would return 2. "({}" would return 1
+func (c *consoleBracketState) commandInOpenState() int {
```
ollevche, line 33: nit: I would prefer a new int type like `commandStatus` with a few constants to represent potential statuses like
* `commandStatusMissingBrackets`
* `commandStatusClosedBrackets`
* `commandStatusPendingBrackets`

It would be easier to read instead of int comparison.

ollevche (approved):
Looks good, thank you! Left a nit comment, that can be ignored :)

internal/command/console_state.go
Andrew-Hayes: I did have this originally, but found it was nicer UX to match the prompt to the number of open brackets. So two open brackets appears as:

```
> {
. (
.. [
...  
```

so started using ints

PR #1879, opened 2024-08-01
Allow usage of static variables in lifecycles

### LINKED ISSUE #304: Open existing Terraform Issue [27360]: [Add support for lifecycle meta-argument in modules ]

[[LINK_TO_ISSUE](https://github.com/hashicorp/terraform/issues/27360)]

**Additional context**

> Terraform currently only allows the lifecycle meta-argument to be used within the declaration of a resource. It would be really useful if users were able to specify lifecycle blocks in modules that can then be applicable to some/all of the resources within that module.
> The main use-case I have is being able to use the [ignore_changes](https://www.terraform.io/docs/configuration/meta-arguments/lifecycle.html#ignore_changes) to instruct terraform to ignore changes to resources or particular attributes of resources.

cam72cam:
We may want to extend this to support statefull contexts someday, but for now this is a reasonable start.

Resolves #304, https://github.com/opentofu/opentofu/issues/1329

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/module_merge.go
+++ b/internal/configs/module_merge.go
@@ -235,8 +235,8 @@ func (r *Resource) merge(or *Resource, rps map[string]*RequiredProvider) hcl.Dia
 	if r.Mode == addrs.ManagedResourceMode {
 		// or.Managed is always non-nil for managed resource mode
 
-		if or.Managed.lcContent != nil {
-			r.Managed.lcContent = or.Managed.lcContent
+		if or.Managed.lifecycleBody != nil {
+			r.Managed.lifecycleBody = or.Managed.lifecycleBody
 		}
 
 		if or.Managed.Connection != nil {
--- a/internal/configs/module_merge.go
+++ b/internal/configs/module_merge.go
@@ -235,22 +235,18 @@ func (r *Resource) merge(or *Resource, rps map[string]*RequiredProvider) hcl.Dia
 	if r.Mode == addrs.ManagedResourceMode {
 		// or.Managed is always non-nil for managed resource mode
 
+		if or.Managed.lcContent != nil {
+			r.Managed.lcContent = or.Managed.lcContent
+		}
+
 		if or.Managed.Connection != nil {
 			r.Managed.Connection = or.Managed.Connection
-		}
-		if or.Managed.CreateBeforeDestroySet {
-			r.Managed.CreateBeforeDestroy = or.Managed.CreateBeforeDestroy
-			r.Managed.CreateBeforeDestroySet = or.Managed.CreateBeforeDestroySet
 		}
 		if len(or.Managed.IgnoreChanges) != 0 {
 			r.Managed.IgnoreChanges = or.Managed.IgnoreChanges
 		}
 		if or.Managed.IgnoreAllChanges {
 			r.Managed.IgnoreAllChanges = true
-		}
-		if or.Managed.PreventDestroySet {
-			r.Managed.PreventDestroy = or.Managed.PreventDestroy
-			r.Managed.PreventDestroySet = or.Managed.PreventDestroySet
 		}
 		if len(or.Managed.Provisioners) != 0 {
 			r.Managed.Provisioners = or.Managed.Provisioners
--- a/internal/configs/resource.go
+++ b/internal/configs/resource.go
@@ -645,60 +645,95 @@ func (r *Resource) decodeLifecycleIgnoreChanges(eval *StaticEvaluator, attr *hcl
 	//   ignore_changes = ["*"]
 
 	// Potential inputs:
-	//
+	// * -> ERROR
+	// "*" -> ERROR
+	// all -> IgnoreAllChanges = true
+	// "all" -> IgnoreAllChanges = true
+	// [all] -> [all]
+	// ["all"] -> [all]
+	// [] -> []
+	// [foo] -> [foo]
+	// ["foo"] -> [foo]
+	// [var.bar] -> EITHER value of var.bar if it exists or var.bar traversal. This scenario is ambiguous
+	// [foo, "bar"] -> [foo, bar]
+	// [foo, var.bar] -> [foo, var.bar]
+	// ["foo", var.bar] -> [foo, ${var.bar}]
+	// concat(["foo"], ["bar"]) -> [foo, bar]
+	// concat([foo], [bar]) -> ERROR!
+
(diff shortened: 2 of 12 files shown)


internal/configs/resource.go
```diff
@@ -68,6 +68,7 @@ type ManagedResource struct {
 	Connection   *Connection
 	Provisioners []*Provisioner
 
+	lcContent           *hcl.BodyContent
```
ollevche, line 71: I would prefer something more explicit like `lifecycleBody`.

`Content` is more abstract IMO and `lc` doesn't add much context outside of this issue. I would need to check usages of this variable to ensure I understand correctly.

internal/configs/resource.go
```diff
@@ -647,6 +577,163 @@ func decodeReplaceTriggeredBy(expr hcl.Expression) ([]hcl.Expression, hcl.Diagno
 	return exprs, diags
 }
 
+// nolint: funlen, gocognit, cyclop, nolintlint // Legacy code moved from decode
```
ollevche, line 580: I understand it is not a brand-new function. However, would you mind refactoring it a bit to pass linting?

internal/configs/resource.go
```diff
@@ -647,6 +576,210 @@ func decodeReplaceTriggeredBy(expr hcl.Expression) ([]hcl.Expression, hcl.Diagno
...
+	// ["all"] -> [all]
+	// [] -> []
+	// [foo] -> [foo]
+	// ["foo"] -> [foo]
+	// [var.bar] -> EITHER value of var.bar if it exists or var.bar traversal. This scenario is ambiguous
+	// [foo, "bar"] -> [foo, bar]
+	// [foo, var.bar] -> [foo, var.bar]
+	// ["foo", var.bar] -> [foo, ${var.bar}]
+	// concat(["foo"], ["bar"]) -> [foo, bar]
+	// concat([foo], [bar]) -> ERROR!
```
Yantrio, line 662: I think we should add unit tests that cover each of these to define the behavior more strongly and also to avoid regressing in the future.

internal/configs/resource.go
```diff
@@ -647,6 +576,210 @@ func decodeReplaceTriggeredBy(expr hcl.Expression) ([]hcl.Expression, hcl.Diagno
...
+
+	if hcl.ExprAsKeyword(attr.Expr) == "all" {
+		r.Managed.IgnoreAllChanges = true
+		return diags
+	}
+
+	// Try evaluating as static expression
+	ident := StaticIdentifier{Module: eval.call.addr, Subject: fmt.Sprintf("%s.%s.lifecycle.%s", r.Type, r.Name, attr.Name), DeclRange: attr.Range}
+	val, valDiags := eval.Evaluate(attr.Expr, ident)
+	if valDiags.HasErrors() {
```
Yantrio, line 672: I agree with the linter here, let's try and invert this if so we're not nesting so much

internal/configs/resource.go
```diff
@@ -647,6 +576,210 @@ func decodeReplaceTriggeredBy(expr hcl.Expression) ([]hcl.Expression, hcl.Diagno
...
+			}
+		} else {
+			// Invalid Expression
+			return diags.Extend(valDiags)
+		}
+	} else {
+		if val.Type().IsListType() || val.Type().IsSetType() || val.Type().IsTupleType() {
+			for _, item := range val.AsValueSlice() {
+				if item.Type() != cty.String {
+					diags = diags.Append(&hcl.Diagnostic{
```
Yantrio, line 707: can we attach a subject here or a srcrange somehow?

internal/configs/resource.go
```diff
@@ -647,6 +576,210 @@ func decodeReplaceTriggeredBy(expr hcl.Expression) ([]hcl.Expression, hcl.Diagno
...
+	}
+
+	var ignoreAllRange hcl.Range
+	for _, trav := range r.Managed.IgnoreChanges {
+		if root, ok := trav[0].(hcl.TraverseAttr); ok && root.Name == "*" {
+			ignoreAllRange = trav.SourceRange()
+			diags = append(diags, &hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid ignore_changes wildcard",
+				Detail:   "The [\"*\"] form of ignore_changes wildcard is was deprecated and is now invalid. Use \"ignore_changes = all\" to ignore changes to all attributes.",
```
Yantrio, line 764:
```suggestion
				Detail:   "The [\"*\"] form of ignore_changes wildcard is deprecated and now invalid. Please use \"ignore_changes = all\" to ignore changes to all attributes.",
```

Yantrio (dismissed):
LGTM, I just request that we add some unit tests around the resourec.go logic that we've moved around and added to, to ensure we're covering it all nicely.

internal/configs/resource.go
cam72cam: I'd like to keep this message consistent as it was simply moved from the existing code.

internal/configs/resource.go
cam72cam: That's not easy at all, IMO this is the cleanest way to implement this.

internal/configs/resource.go
cam72cam: Added, as well as a few additional cases for 100% coverage of this complex functionality.

Yantrio (approved):
Approving assuming you'll add docs before merging

ghost (approved):

internal/configs/module_merge.go
```diff
@@ -235,23 +235,19 @@ func (r *Resource) merge(or *Resource, rps map[string]*RequiredProvider) hcl.Dia
...
 		if len(or.Managed.IgnoreChanges) != 0 {
 			r.Managed.IgnoreChanges = or.Managed.IgnoreChanges
 		}
 		if or.Managed.IgnoreAllChanges {
 			r.Managed.IgnoreAllChanges = true
 		}
-		if or.Managed.PreventDestroySet {
-			r.Managed.PreventDestroy = or.Managed.PreventDestroy
-			r.Managed.PreventDestroySet = or.Managed.PreventDestroySet
-		}
```
ollevche, line 254: could you please explain why those are removed?

ollevche (approved):

virtualroot (approved):
Approving PR because I love this feature 🖤

cam72cam:
Closing as we are trying to use dynamic evaluation for this feature and this branch is wildly out of date.

PR #1890, opened 2024-08-05
Fix repetitive diagnosis output in init response

### LINKED ISSUE #1812: tofu init response repetitive diags when configuration has extra label
### OpenTofu Version

```shell
Any/Latest
```

```hcl
terraform {
  extra_label{
    required_providers {
      azurerm = {
        source  = "hashicorp/azurerm"
        version = "~> 3.0.2"
      }
    }
  }
}
```

### Debug Output

2024-07-12T10:59:02.775+0800 [INFO]  OpenTofu version: 1.8.0 dev
2024-07-12T10:59:02.784+0800 [DEBUG] using github.com/hashicorp/go-tfe v1.36.0
2024-07-12T10:59:02.784+0800 [DEBUG] using github.com/opentofu/hcl/v2 v2.0.0-20240416130056-03228b26f391
2024-07-12T10:59:02.784+0800 [DEBUG] using github.com/hashicorp/terraform-svchost v0.1.1
2024-07-12T10:59:02.784+0800 [DEBUG] using github.com/zclconf/go-cty v1.14.4
2024-07-12T10:59:02.784+0800 [INFO]  Go runtime version: go1.21.4
2024-07-12T10:59:02.784+0800 [INFO]  CLI args: []string{"E:\\Test_workspace\\test-template\\test-it\\tofu.exe", "init"}
2024-07-12T10:59:02.787+0800 [DEBUG] Attempting to open CLI config file: C:\Users\AppData\Roaming\terraform.rc
2024-07-12T10:59:02.787+0800 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2024-07-12T10:59:02.787+0800 [DEBUG] checking for credentials in "C:\\Users\\AppData\\Roaming\\terraform.d\\plugins"
2024-07-12T10:59:02.787+0800 [DEBUG] ignoring non-existing provider search directory terraform.d/plugins
2024-07-12T10:59:02.787+0800 [DEBUG] will search for provider plugins in C:\Users\AppData\Roaming\terraform.d\plugins
2024-07-12T10:59:02.790+0800 [DEBUG] ignoring non-existing provider search directory C:\Users\AppData\Roaming\HashiCorp\Terraform\plugins
2024-07-12T10:59:02.791+0800 [DEBUG] Found the config directory: C:\Users\AppData\Roaming\terraform.d
2024-07-12T10:59:02.791+0800 [INFO]  CLI command args: []string{"init"}

### Expected Behavior

The configuration diags should only occur once

### Actual Behavior

$ ./tofu.exe init

Initializing the backend...
OpenTofu encountered problems during initialization, including problems
with the configuration, described below.

The OpenTofu configuration must be valid before initialization so that
OpenTofu can determine which modules and providers need to be installed.
╷
│ Error: Unsupported block type
│
│   on main.tf line 2, in terraform:
│    2:   extra_label{
│
│ Blocks of type "extra_label" are not expected here.
╵

╷
│ Error: Unsupported block type
│
│   on main.tf line 2, in terraform:
│    2:   extra_label{
│
│ Blocks of type "extra_label" are not expected here.

### Steps to Reproduce

tofu init

### Additional Context

_No response_

### References

_No response_

buraksenn:
In the issue #1812 it was explained and demonstrated clearly that init response was outputting duplicate diagnosis error. I have added `Merge` method to diagnostics to not add exact equal diagnostics to prevent duplicates.

Resolves #1812 

## Target Release

1.9.0

## Checklist

### Go checklist

### DIFF
--- a/internal/tfdiags/diagnostics.go
+++ b/internal/tfdiags/diagnostics.go
@@ -9,7 +9,6 @@ import (
 	"bytes"
 	"fmt"
 	"path/filepath"
-	"reflect"
 	"sort"
 	"strings"
 
@@ -102,17 +101,17 @@ func (diags Diagnostics) Append(new ...interface{}) Diagnostics {
 	return diags
 }
 
-func (diags Diagnostics) Merge(other Diagnostics) Diagnostics {
+func (diags Diagnostics) StrictDeduplicateMerge(other Diagnostics) Diagnostics {
 	if len(diags) == 0 {
 		return other
 	}
 
 	for _, d := range diags {
 		for _, o := range other {
-			if reflect.DeepEqual(d, o) {
-				continue
+			isEqual := d.Description().Equal(o.Description()) && d.Severity() == o.Severity() && d.Source().Equal(o.Source())
+			if DoNotConsolidateDiagnostic(d) || DoNotConsolidateDiagnostic(o) || !isEqual {
+				diags = append(diags, o)
 			}
-			diags = append(diags, o)
 		}
 	}
 
--- a/internal/tfdiags/diagnostics.go
+++ b/internal/tfdiags/diagnostics.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"fmt"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strings"
 
@@ -96,6 +97,23 @@ func (diags Diagnostics) Append(new ...interface{}) Diagnostics {
 	// here, but we'll make sure of that so callers can rely on empty == nil
 	if len(diags) == 0 {
 		return nil
+	}
+
+	return diags
+}
+
+func (diags Diagnostics) Merge(other Diagnostics) Diagnostics {
+	if len(diags) == 0 {
+		return other
+	}
+
+	for _, d := range diags {
+		for _, o := range other {
+			if reflect.DeepEqual(d, o) {
+				continue
+			}
+			diags = append(diags, o)
+		}
 	}
 
 	return diags
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -292,7 +292,7 @@ func (c *InitCommand) Run(args []string) int {
 
 	// Now, we can check the diagnostics from the early configuration and the
 	// backend.
-	diags = diags.Append(earlyConfDiags.Merge(backDiags))
+	diags = diags.Append(earlyConfDiags.StrictDeduplicateMerge(backDiags))
 	if earlyConfDiags.HasErrors() {
 		c.Ui.Error(strings.TrimSpace(errInitConfigError))
 		c.showDiagnostics(diags)
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -293,7 +293,6 @@ func (c *InitCommand) Run(args []string) int {
 	// Now, we can check the diagnostics from the early configuration and the
 	// backend.
 	diags = diags.Append(earlyConfDiags)
-	diags = diags.Append(backDiags)
 	if earlyConfDiags.HasErrors() {
 		c.Ui.Error(strings.TrimSpace(errInitConfigError))
 		c.showDiagnostics(diags)
@@ -303,6 +302,7 @@ func (c *InitCommand) Run(args []string) int {
 	// Now, we can show any errors from initializing the backend, but we won't
 	// show the errInitConfigError preamble as we didn't detect problems with
 	// the early configuration.
+	diags = diags.Append(backDiags)
 	if backDiags.HasErrors() {
 		c.showDiagnostics(diags)
 		return 1
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -292,7 +292,7 @@ func (c *InitCommand) Run(args []string) int {
 
 	// Now, we can check the diagnostics from the early configuration and the
 	// backend.
-	diags = diags.Append(earlyConfDiags)
+	diags = diags.Append(earlyConfDiags.Merge(backDiags))
 	if earlyConfDiags.HasErrors() {
 		c.Ui.Error(strings.TrimSpace(errInitConfigError))
 		c.showDiagnostics(diags)
@@ -302,7 +302,6 @@ func (c *InitCommand) Run(args []string) int {
 	// Now, we can show any errors from initializing the backend, but we won't
 	// show the errInitConfigError preamble as we didn't detect problems with
 	// the early configuration.
-	diags = diags.Append(backDiags)
 	if backDiags.HasErrors() {
 		c.showDiagnostics(diags)
 		return 1
--- a/internal/command/init_test.go
+++ b/internal/command/init_test.go
@@ -2995,6 +2995,33 @@ func TestInit_moduleVersion(t *testing.T) {
 	})
 }
 
+func TestInit_invalidExtraLabel(t *testing.T) {
+	td := t.TempDir()
+	testCopyDir(t, testFixturePath("init-syntax-invalid-extra-label"), td)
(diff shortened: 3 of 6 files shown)


ollevche (changes requested):
Hey @buraksenn, thank you for working on this issue!

Ignoring `backDiags` when we print `earlyConfDiags` may require users to run init multiple times just to find out there are more errors that need fixing.

I would suggest trying to consolidate those diags even if they are errors. It needs to be done delicately so we don't silence any useful messages, that are not duplicated.

That will also help with failed tests, since we don't change the existing behaviour. 

Happy to answer any questions and discuss potential concerns!

buraksenn:
> Hey @buraksenn, thank you for working on this issue!
> 
> Ignoring `backDiags` when we print `earlyConfDiags` may require users to run init multiple times just to find out there are more errors that need fixing.
> 
> I would suggest trying to consolidate those diags even if they are errors. It needs to be done delicately so we don't silence any useful messages, that are not duplicated.
> 
> That will also help with failed tests, since we don't change the existing behaviour.
> 
> Happy to answer any questions and discuss potential concerns!

That made sense to me as well but I saw that we are not consolidation error level if source is nil so thought that this would be the fix. Then the resolution would be to expand consolidateWarnings or add another method such as consolidateErrors to prevent duplicates right @ollevche ?

cam72cam:
Hmm, I've done some related work on one of my branches: https://github.com/opentofu/opentofu/commit/032c9ff91d0f2c1de4e0b8b4a54e17629a97cbd8#diff-cb9acfeab1fa4de42a8453d53172343075c9db7dfbd41d59569050760f455132

I can take some time today to get that PR ready for review.  It could make sense to rebase this work on top of my changes once they are reviewed and merged.

Ericwww:
I would also prefer the opinion that consolidate duplicated error. For some details, we could have a discussion on how to judge diagnostics is equal (or duplicated) 😄

cam72cam:
Long term, we should revise how and where we are loading configuration (only once).  That depends on a *lot* of refactoring in the command package that we have not had time to start or review.

The question is do we want to do a "generic" merge like in https://github.com/opentofu/opentofu/pull/1894 or do we want to have a more explicit "merge between these two sets" logic, which might make more sense here.

cam72cam:
> I agree it makes more sense to have some explicit merge logic, that will result in a unique diag list from the previous two (backDiags and earlyConfDiags).

In that case @ollevche, it probably wouldn't make sense for this PR to be based on mine, though it could be used as a reference.

buraksenn:
So I'm not sure which approach I should take right now. I would be glad to do the work related to consolidating etc but it would be great to clarify the approach. @cam72cam should I wait for your PR to be merged to continue?

cam72cam:
@buraksenn You should be able to continue on this branch with the approach of merging two different lists of diagnostics, removing any exact duplicates.  This would probably look something like:
`diags = diags.Append(earlyConfDiags.Merge(backDiags))`

buraksenn:
> @buraksenn You should be able to continue on this branch with the approach of merging two different lists of diagnostics, removing any exact duplicates. This would probably look something like: `diags = diags.Append(earlyConfDiags.Merge(backDiags))`

Thanks @cam72cam , I will continue with this. Appreciate it

internal/tfdiags/diagnostics.go
```diff
@@ -101,6 +102,23 @@ func (diags Diagnostics) Append(new ...interface{}) Diagnostics {
...
 }
 
+func (diags Diagnostics) Merge(other Diagnostics) Diagnostics {
+	if len(diags) == 0 {
+		return other
+	}
+
+	for _, d := range diags {
+		for _, o := range other {
+			if reflect.DeepEqual(d, o) {
```
buraksenn, line 112: I was thinking about adding Equal method but `Diagnostic` interface has many structs/interfaces that also has other struct/interface dependencies. For the comparison check it would become a overkill for adding that much `Equal()` functions to many structs. That's why I have used `reflect.DeepEqual` here

buraksenn:
@ollevche can you review again if possible?

ollevche:
@buraksenn I'll take a look at the PR today. Thank you for working on this one!

internal/tfdiags/diagnostics.go
ollevche: `reflect.DeepEqual` feels a bit of an overkill here. Also, it could fail with wrapped diags.

Would you mind adapting the logic in [ConsolidateWarnings](https://github.com/opentofu/opentofu/blob/main/internal/tfdiags/consolidate_warnings.go#L25) here?

I think we can treat two diags as duplicates if:
* their `Severity`s are equal
* their `Source.Subject`s are present and the whole `Source`s are equal
* their `Description`s are equal

I would also check if `DoNotConsolidateDiagnostic` function allows merging the diag.

ollevche (commented):
Thank you for the contribution!

It looks pretty good, special kudos for adding a test. However, I think we need to change the comparison logic to something similar to `ConsolidateWarnings`' implementation.

Left a comment with more details.

internal/tfdiags/diagnostics.go
cam72cam: I've gone back and forth on this.  We know we are using overlapping code paths to produce duplicate diagnostics.  I'm ok with this being a bit stricter than ConsolidateWarnings.  I think the DeepEqual is probably a bit overkill, but @ollevche's list makes sense, perhaps with the Ranges added as well.

cam72cam: I also think that DoNotConsolidateDiagnostic can be ignored for this particular function, as it is a special purpose function only for this scenario.

internal/tfdiags/diagnostics.go
```diff
@@ -101,6 +102,23 @@ func (diags Diagnostics) Append(new ...interface{}) Diagnostics {
 	return diags
 }
 
+func (diags Diagnostics) Merge(other Diagnostics) Diagnostics {
```
cam72cam, line 105: Perhaps name this StrictDedeuplicate or similar to signal that it is made for a very specific purpose?

internal/tfdiags/diagnostics.go
ollevche: Yep, I agree `DoNotConsolidateDiagnostic` check is optional here.

> perhaps with the Ranges added as well.

`Source`s should be equal and `Source.Subject`s (ranges) should be non-zero.

buraksenn: Added Equal() methods to used structs and refactored the function

internal/tfdiags/diagnostics.go
buraksenn: Changed it to StrictDeduplicateMerge

cam72cam (approved):

internal/tfdiags/diagnostics.go
```diff
@@ -101,6 +101,23 @@ func (diags Diagnostics) Append(new ...interface{}) Diagnostics {
 	return diags
 }
 
+func (diags Diagnostics) StrictDeduplicateMerge(other Diagnostics) Diagnostics {
```
ollevche, line 104: Can we say that diags are duplicates if there is no source range set in them?

Those diags (same description and severity) could reference different parts of configuration and should not be treated as duplicates.

cam72cam: If everything else is identical, and there is no source range, I don't see how hiding one of two duplicate errors would cause any difficulty.

ollevche (approved):

cam72cam:
Thanks @buraksenn !

PR #1892, opened 2024-08-07
#300: allow square brackets syntax and pass aliases map to graph

ollevche:
Part of #300 

This PR is not meant to be merged to main directly. There is also lots of TODOs left.

Main purpose of the PR is to allow using square brackets with `each.value` / `each.key` when referencing providers:

```hcl
provider random {
  alias = "a"
}

provider random {
  alias = "b"
}

provider random {
  alias = "c"
}

module "mod" {
  for_each = toset(["a", "b", "c"])

  source = "./mod"
  providers = {
    random.moda = random[each.value]
    random.modb = random[each.value]
  }
}

resource "random_id" "all" {
  byte_length = 10

  for_each = toset(["a", "b", "c"])

  provider = random[each.value]
}
```

With such configuration, proper list of `PassedProviderConfig` with `ProviderConfigRefMapping` will be passed to graph transformers. Also, `ProviderConfigRefMapping` will be passed for configuration of `data` and `resource` blocks.

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/resource.go
+++ b/internal/configs/resource.go
@@ -696,26 +696,6 @@ func NewProviderConfigRefMapping(p *Provider) *ProviderConfigRefMapping {
 	return m
 }
 
-// TODO/Oleksandr: review the calls and provide a proper mapping
-func NewProviderConfigMappingFromRef(ref *ProviderConfigRef) *ProviderConfigRefMapping {
-	m := &ProviderConfigRefMapping{
-		Name:         ref.Name,
-		NameRange:    ref.NameRange,
-		providerType: ref.providerType,
-	}
-
-	if ref.Alias == "" {
-		return m
-	}
-
-	m.Aliases = map[addrs.InstanceKey]string{
-		addrs.NoKey: ref.Alias,
-	}
-	m.AliasRange = ref.AliasRange
-
-	return m
-}
-
 // TODO/Oleksandr: review calls and provide a proper alias
 func (m *ProviderConfigRefMapping) GetNoKeyAlias() string {
 	return m.Aliases[addrs.NoKey]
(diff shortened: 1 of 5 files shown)


internal/configs/resource.go
```diff
@@ -696,39 +705,171 @@ func NewProviderConfigRefMapping(p *Provider) *ProviderConfigRefMapping {
...
+				Detail:   "",
+				Subject:  m.Alias.Range().Ptr(),
+			})
+		}
+
+		m.Aliases = map[addrs.InstanceKey]string{
+			addrs.NoKey: aliasVal.AsString(),
+		}
+
+		return nil
```
cam72cam, line 743: return diags, don't want to drop warnings

internal/configs/resource.go
```diff
@@ -696,39 +705,171 @@ func NewProviderConfigRefMapping(p *Provider) *ProviderConfigRefMapping {
...
 	}
 
-	if ref.Alias == "" {
-		return m
+	var diags hcl.Diagnostics
+
+	// There is no for each so we just put alias under NoKey.
+	if forEach == nil {
+		// TODO/Oleksandr: use static eval for alias expr?
```
cam72cam, line 722: We did outline this as a possibility in the RFC, though it is a nice to have.

internal/configs/resource.go
```diff
@@ -696,39 +705,171 @@ func NewProviderConfigRefMapping(p *Provider) *ProviderConfigRefMapping {
...
-	m.Aliases = map[addrs.InstanceKey]string{
-		addrs.NoKey: ref.Alias,
+	// TODO/Oleksandr: Is static identifier valid if we evaluate for_each for alias?
+	instances, instancesDiags := eval.Evaluate(forEach, StaticIdentifier{
+		Module:    eval.call.addr,
+		Subject:   "alias",
+		DeclRange: m.Alias.Range(),
+	})
+	diags = diags.Extend(instancesDiags)
+	if instancesDiags.HasErrors() {
```
cam72cam, line 753: This piece is a bit interesting, if we can't evaluate for_each here, we are not allowed to use each.value in the providers expression.  However, that does not necessarily mean there's a config issue.  I glossed over this scenario in my prototype.

Ex:
```hcl
module "foo" {
  for_each = data.some.stuff # Static evaluation not allowed
  providers {
    random = random["foo"]
    other = random[local.other] # Similar question to L722 above
  }
}
```

An alternative approach would to use expr dependencies to see if for_each is needed to evaluate.

Pseudo:
```
refs = lang.ReferencesInExpr(aliasExpr)
if for_each in refs {
  try eval for_each
  if fail eval, return diags
  else, alias mapping logic
  return mapping
}
return simple alias
```

cam72cam: Though that does depend on the question of if we want to allow raw strings or locals in the expression

ollevche: That's an excellent point, thank you!

We discussed it with @Evi1Pumpkin and decided it makes sense to have {NoKey: "alias"} mapping for cases when there is no instance references in the alias expression.

I also added `HasAlias` and `HasInstanceRefsInAlias` shortcuts to easily check if we need to iterate over all the instance keys on the graph part.

Updates are in [c661554](https://github.com/opentofu/opentofu/pull/1892/commits/c661554a4171c400383e43537e0982d810304ef1)

cam72cam (approved):

PR #399, opened 2023-09-14
[Issue 364] Stop Consolidation of Warnings

Hefeweizen:
Change:
add a flag to show all instances of similar warnings

Rationale:
Multiple instances of similar warnings are consolidated into one entry, with notation that there may be more.  This obscures the location of those additional warnings.  If one desires to change those warnings, one must iteratively fix them, rerunning the plan after each change, to get the tooling to show you these changes.

This new flag stops the consolidation.

Issue: https://github.com/opentffoundation/opentf/issues/364
Resolves #364

## Target Release

1.6.0

## Draft CHANGELOG entry

Add `-verbose-warnings` flag, optionally stopping the consolidation of similar warnings, instead listing the full details for every warning individually.

### NEW FEATURES | UPGRADE NOTES | ENHANCEMENTS | BUG FIXES | EXPERIMENTS

-  `opentf plan -verbose-warnings`: show details on all warnings.

### Example Output
```
$ opentf plan 
...
╷
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket.b1,
│   on main.tf line 14, in resource "aws_s3_bucket" "b1":
│   14: resource "aws_s3_bucket" "b1" {
│ 
│ Use the aws_s3_bucket_versioning resource instead
│ 
│ (and 3 more similar warnings elsewhere)
╵
$ opentf plan -verbose-warnings
...
╷
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket.b1,
│   on main.tf line 14, in resource "aws_s3_bucket" "b1":
│   14: resource "aws_s3_bucket" "b1" {
│ 
│ Use the aws_s3_bucket_versioning resource instead
╵
╷
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket.b1,
│   on main.tf line 16, in resource "aws_s3_bucket" "b1":
│   16:   acl    = "private"
│ 
│ Use the aws_s3_bucket_acl resource instead
╵
╷
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket.b2,
│   on main.tf line 23, in resource "aws_s3_bucket" "b2":
│   23: resource "aws_s3_bucket" "b2" {
│ 
│ Use the aws_s3_bucket_versioning resource instead
╵
╷
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket.b2,
│   on main.tf line 25, in resource "aws_s3_bucket" "b2":
│   25:   acl    = "private"
│ 
[...]

PR #1894, opened 2024-08-07
Warning and Error consolidation CLI options

### LINKED ISSUE #364: Show all warnings
### OpenTF Version

```shell
1.0
```

### Use Cases

When upgrading a provider eg the aws provider, many warnings can occur, but only the first of them is shown. Eg 
```
Warning: Argument is deprecated
  on .terraform/modules/spa.s3_logs/main.tf line 1, in resource "aws_s3_bucket" "this":
   1: resource "aws_s3_bucket" "this" {
Use the aws_s3_bucket_acl resource instead

(and 21 more similar warnings elsewhere)
```
There is currently no option to show those similar warnings. All these warnings could be in different files, so we are forced to first address the only one shown, then re-run the terraform plan. Sometimes this requires many cycles. TEDIOUS!

### Attempted Solutions

Run `terraform validate`. 

However this does not seem to catch these warnings, possibly because the warnings are not about invalid code, just deprecation warnings. 

### Proposal

Several possibilities: 
- Add a switch on the plan and apply commands, eg `--show-all-warnings`. This switch should work even if the output is json. 
- Add a new command, `show-warnings`: if the users runs an apply or a plan and sees deprecation warnings, then they could run this new command to get the details. This makes sense if there are steps of the plan / apply can be skipped when the user just wants these (eg one would expect that it would be sufficient for terraform to parse the code, and ask the provider for any warnings about it, without actually having to check the cloud resources status, which could save a ton of time)

### References

- The issue is mentioned in a few places but the biggest discussion I found is https://discuss.hashicorp.com/t/show-all-warnings/5606/18. This demonstrates that HC is not interested in implementing this much-needed and almost trivial capability.

### LINKED ISSUE #1480: Merge duplicate diagnostics to avoid diagnostics flood
### OpenTofu Version

```shell
any
```

### Use Cases

When I use the template to run a plan 
```
output "out" {
  value = [for i in range(50): length([for i in range(50):bad()])]
}
```
I get 2500 same diagnostics :
image

### Attempted Solutions

merge them
or wait upstream(hcl) fix it

### Proposal

_No response_

### References

_No response_

cam72cam:
Originally started by @Hefeweizen in #399, rebased and continued by @cam72cam.  Additional tests added by @Ericwww.

Resolves #1480
Resolves #364 

## Target Release

1.9.0

## Checklist

### Go checklist

 


internal/tfdiags/consolidate_warnings.go
```diff
@@ -7,112 +7,123 @@ package tfdiags
...
+		panic(fmt.Sprintf("Invalid diagnostic severity: %#v", wg.Severity()))
+	}
+
 	switch extraCount {
 	case 1:
-		msg = "(and one more similar warning elsewhere)"
+		msg = fmt.Sprintf("(and one more similar %s elsewhere)", diagType)
 	default:
-		msg = fmt.Sprintf("(and %d more similar warnings elsewhere)", extraCount)
+		msg = fmt.Sprintf("(and %d more similar %ss elsewhere)", extraCount, diagType)
```
Ericwww, line 126: some typo here? 

`%ss` => `%s`

cam72cam: I don't think so:

error -> errors
warning -> warnings

internal/tfdiags/consolidate_warnings.go
Ericwww, line 1: Maybe the filename should change also.

The function in this file is not only consolidating warnings now  :)

internal/tfdiags/consolidate_warnings.go
Ericwww: Ah, I got you.

internal/tfdiags/consolidate_warnings.go
cam72cam: Git does not pick up on the rename properly when I do so.  I can do that in a follow-up PR to separate the changes.

internal/tfdiags/consolidate_warnings_test.go
```diff
@@ -96,7 +96,7 @@ func TestConsolidateWarnings(t *testing.T) {
 
 	// We're using ForRPC here to force the diagnostics to be of a consistent
 	// type that we can easily assert against below.
-	got := diags.ConsolidateWarnings(2).ForRPC()
+	got := diags.Consolidate(2, Warning).ForRPC()
```
ollevche, line 99: nit: I would prefer constants for `1` and `2` when calling `Consolidate`. It isn't clear to me why exactly this number should be passed.

ollevche (approved):
LGTM! Left a minor nit comment (feel free to ignore)

internal/tfdiags/consolidate_warnings_test.go
cam72cam: Other than this test, it's always 1.  The function was originally built to support consolidation at different count thresholds.  I see this as a legacy option that I didn't want to touch in this PR, even though the arg should probably be removed.

PR #1900, opened 2024-08-08
Add exclude flag support

RLRabinowitz:
Adds support for exclude flag.

Works very similarly to the `-target` flag, but does the opposite.
- We basically find all excluded targets, and all targets that depend on them, and then pick all other targetable nodes in the graph. Then, act the same way as with targets - Target all picked nodes and their descendants, and target any output that all of its dependencies are targeted
- There's validation that `-target` and `-exclude` flags cannot be used together. This possibly could be changed and adapted in the future to work well together
- We error if an `-exclude` flag was provided for a remote run, so as not to confuse users, thinking that a run would exclude specific resources when actually it doesn't

Note: Most of the code in this PR is inspired by the current implementation of the target flag. In most cases, you'd see the target flag's relevant implementation right before the newly added exclude flag implementation.

Resolves https://github.com/opentofu/opentofu/issues/426

## Target Release

1.9.0

## Checklist

### Go checklist

 


internal/backend/remote/backend_context.go
```diff
@@ -27,8 +27,9 @@ func (b *Remote) LocalRun(op *backend.Operation) (*backend.LocalRun, statemgr.Fu
 	var diags tfdiags.Diagnostics
 	ret := &backend.LocalRun{
 		PlanOpts: &tofu.PlanOpts{
-			Mode:    op.PlanMode,
-			Targets: op.Targets,
+			Mode:     op.PlanMode,
+			Targets:  op.Targets,
+			Excludes: op.Excludes,
```
RLRabinowitz, line 32: By browsing the code, looks like this isn't really needed, as in any call of the `LocalRun` method there is never `Targets` or `Excludes` on the operation.

internal/cloud/backend_context.go
```diff
@@ -27,8 +27,9 @@ func (b *Cloud) LocalRun(op *backend.Operation) (*backend.LocalRun, statemgr.Ful
 	var diags tfdiags.Diagnostics
 	ret := &backend.LocalRun{
 		PlanOpts: &tofu.PlanOpts{
-			Mode:    op.PlanMode,
-			Targets: op.Targets,
+			Mode:     op.PlanMode,
+			Targets:  op.Targets,
+			Excludes: op.Excludes,
```
RLRabinowitz, line 32: By browsing the code, looks like this isn't really needed, as in any call of the `LocalRun` method there is never `Targets` or `Excludes` on the operation.

internal/tofu/context_apply_test.go
```diff
@@ -7225,8 +7224,8 @@ func TestContext2Apply_targetedDestroy(t *testing.T) {
 
 	// the module instance should remain
 	mod = state.Module(addrs.RootModuleInstance.Child("child", addrs.NoKey))
-	if len(mod.Resources) != 1 {
-		t.Fatalf("expected 1 resources, got: %#v", mod.Resources)
+	if mod != nil {
```
RLRabinowitz, line 7227: This test was wrong

internal/command/arguments/extended.go
```diff
@@ -119,8 +120,44 @@ func (o *Operation) Parse() tfdiags.Diagnostics {
...
+
+func parseRawTargetsAndExcludes(targets []string, excludes []string) ([]addrs.Targetable, []addrs.Targetable, tfdiags.Diagnostics) {
+	var parsedTargets []addrs.Targetable
+	var parsedExcludes []addrs.Targetable
+	var diags tfdiags.Diagnostics
+
+	if len(targets) > 0 && len(excludes) > 0 {
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"-target and -exclude flags used together",
```
cam72cam, line 136: Suggestion: `"Invalid combination of arguments"`

internal/tofu/context_apply_test.go
cam72cam: Could you explain how this test was wrong?

RLRabinowitz: Basically, the state was built in such a way were dependencies were wrong. The actual configuration file marks the module as dependent on the standalone resource. However, this was not reflected in the state file.

By actually simulating an apply instead of building the state file (which some tests do), I was able to get the actual expected behaviour here

Changed a comment here:
82b9d01d245dae95a021f87cd2041cb843f326ef

internal/command/arguments/extended.go
RLRabinowitz: Done in 90554e23b7f1e4f193e86a6a454f895925587aca

internal/tofu/transform_targets.go
```diff
@@ -13,34 +13,42 @@ import (
...
 }
 
 // TargetsTransformer is a GraphTransformer that, when the user specifies a
-// list of resources to target, limits the graph to only those resources and
-// their dependencies.
+// list of resources to target, or a list of resources to exclude, limits the
+// graph to only those resources and their dependencies (or in the case of
+// excludes - limits the graph to all resources that are not excluded or not
+// dependent on excluded resources).
 type TargetsTransformer struct {
```
Yantrio, line 31: Can I ask why you decided to add it to this existing transformer instead of keeping it separate? For navigation of the code I would expect to find a `ExcludesTransformer` somewhere and this may be confusing

internal/tofu/transform_targets.go
```diff
@@ -73,10 +81,115 @@ func (t *TargetsTransformer) selectTargetedNodes(g *Graph, addrs []addrs.Targeta
...
+		// Only resource and resource instance nodes can be targeted.
+		return nil
+	}
+}
+
+// Returns a set of targeted nodes, after excluding resources. An excluded resource
+// is either addressed directly, address indirectly via its container, or it's
+// dependent on an excluded node. The rest are targeted nodes.
+func (t *TargetsTransformer) removeExcludedNodes(g *Graph, excludes []addrs.Targetable) dag.Set {
+	targetedNodes := make(dag.Set)
```
Yantrio, line 108: im not a huge fan of the naming here. I don't think the explicit opposite of exclude is `targetted` but instead maybe `includedNodes` ? 

I got confused reading this thinking that we're mixing up target and exclude logic here.

internal/tofu/transform_targets.go
```diff
@@ -73,10 +81,115 @@ func (t *TargetsTransformer) selectTargetedNodes(g *Graph, addrs []addrs.Targeta
...
 	// side effects from the targeted nodes, these are added because outputs
 	// cannot be targeted on their own.
+	//
+	// Note: This behaviour has some quirks, as there are specific cases where
+	// you would think an output should not be updated, but it is
+	// For example, when there's a module call with an input that is dependent
+	// on a root resource, and only the root resource is targeted, any output
+	// that depends on a module output might be updated, if said module output
+	// does not depend on any resource of the module itself.
+	// Right now, we will not change this behaviour
```
Yantrio, line 188: can you explain in a comment here about why we're not doing this, and when we could do this please?

internal/tofu/transform_targets.go
```diff
@@ -115,14 +228,70 @@ func (t *TargetsTransformer) selectTargetedNodes(g *Graph, addrs []addrs.Targeta
...
 		}
 	}
 
-	return targetedNodes, nil
+	return targetedOutputNodes
+}
+
+func (t *TargetsTransformer) nodeIsExcluded(vertexAddr addrs.Targetable, excludes []addrs.Targetable) bool {
+	for _, excludeAddr := range excludes {
+		// The behaviour here is a bit different from targets.
```
Yantrio, line 243: This comment assumes that the reader has knowledge of what TargetContains() does. Maybe it's best to move this comment into that method?

internal/tofu/transform_targets.go
```diff
@@ -115,14 +228,70 @@ func (t *TargetsTransformer) selectTargetedNodes(g *Graph, addrs []addrs.Targeta
...
+
+func (t *TargetsTransformer) nodeDescendantsExcluded(vertexAddr addrs.Targetable, excludes []addrs.Targetable) bool {
+	for _, excludeAddr := range excludes {
+		// The behaviour here is a bit different from targets.
+		// Before expansion - We'd like to only exclude resources that were excluded by module or resource.
+		//   If the excluded target is an AbsResourceInstance, then we'd want to skip exclude until we expand the resource
+		// After expansion - We'd like to exclude any vertex that contains the exclude address
+		//   Since before expansion the vertexAddr is without an index, then if the excludeAddr is an instance, it will
+		//   only contain vertexAddr if its key is NoKey
+		// So - a simple TargetContains here should be enough, both before and after expansion
```
Yantrio, line 267:
```suggestion
```

Yantrio (changes requested):
Looks good overall, I left some comments about the approach and testing. I peronsally think that we should be handling it separate from the `target` logic.

 Ill try and find some time later today to pull the branch and test it locally.

internal/plans/internal/planproto/planfile.proto
```diff
@@ -66,6 +66,11 @@ message Plan {
     // configuration.
     repeated string target_addrs = 5;
 
+    // An unordered set of exclude addresses to exclude when applying. If no
+    // target addresses are present, the plan applies to the whole
+    // configuration.
+    repeated string exclude_addrs = 6;
```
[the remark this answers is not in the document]
RLRabinowitz: At least going through the history of the file, before it was moved back in 2021 to this file from a different file, 6 was never used.

However, that shouldn't really matter. I saw throughout the git history changes toe the values here, and that makes sense, as IIRC plan files can only be ingested by an OpenTofu binary of the same version as the one that has created it

internal/tofu/transform_targets.go
RLRabinowitz: 09ae600ac387435db24b5d23c5ad66be430d89b8

internal/tofu/transform_targets.go
RLRabinowitz: 09ae600ac387435db24b5d23c5ad66be430d89b8

Just removed the comment, as it makes more sense in the function under it.
That function also follows up with an explanation on the addresses used, so it should clarify that

internal/tofu/transform_targets.go
RLRabinowitz: 0e8d58db30ff3ad339b1d8dd409847dd9eff6c9d

internal/tofu/transform_targets.go
RLRabinowitz: Is that helpful?

f11054456890284c2edd29bc81f85c39f785981f

internal/tofu/transform_targets.go
RLRabinowitz: We'd probably prefer this being a single transformer, as right now "targeting" is a concept that both `-target` and `-exclude` flag contribute to. And, if in the future we'd want to support both `-target` and -`exclude` flag together, it would make more sense for the transformer to be singular.

I've renamed it to TargetingTransformer - 07779eb70d1c7e048f35c3eb8e1bf5c88f785df4

internal/command/arguments/plan_test.go
```diff
@@ -160,6 +162,79 @@ func TestParsePlan_targets(t *testing.T) {
...
+		"two excludes": {
+			args: []string{"-exclude=foo_bar.baz", "-exclude", "module.boop"},
+			want: []addrs.Targetable{foobarbaz.Subject, boop.Subject},
+		},
+		"invalid traversal": {
+			args:    []string{"-exclude=foo."},
+			want:    nil,
+			wantErr: "Dot must be followed by attribute name",
+		},
+		"invalid target": {
```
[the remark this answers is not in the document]
RLRabinowitz: 234833a617f32b063780389eaf166449d861ee83

internal/command/arguments/extended.go
```diff
@@ -85,20 +89,17 @@ type Operation struct {
...
 	forceReplaceRaw []string
 	destroyRaw      bool
 	refreshOnlyRaw  bool
 }
 
-// Parse must be called on Operation after initial flag parse. This processes
-// the raw target flags into addrs.Targetable values, returning diagnostics if
-// invalid.
-func (o *Operation) Parse() tfdiags.Diagnostics {
+func parseRawTargets(targets []string) ([]addrs.Targetable, tfdiags.Diagnostics) {
```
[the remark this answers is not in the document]
RLRabinowitz: c1a461628681a7c01675c11af8e0de85955623ed

internal/command/apply_destroy_test.go
```diff
@@ -386,290 +386,222 @@ func TestApply_destroyPath(t *testing.T) {
...
-
-	backupActualStr := strings.TrimSpace(backupStateFile.State.String())
-	backupExpectedStr := strings.TrimSpace(originalState.String())
-	if backupActualStr != backupExpectedStr {
-		t.Fatalf("bad:\n\nactual:\n%s\n\nexpected:\nb%s", backupActualStr, backupExpectedStr)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Create a temporary working directory that is empty
+			td := t.TempDir()
+			testCopyDir(t, testFixturePath("apply-destroy-targeted"), td)
```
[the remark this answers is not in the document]
RLRabinowitz: 2afb6c5ef82a6473007c114a6e043210d8ade67f

internal/command/plan_test.go
```diff
@@ -1366,6 +1366,92 @@ func TestPlan_targetFlagsDiags(t *testing.T) {
 	}
 }
 
+// Config with multiple resources, targeted plan with exclude
+func TestPlan_excluded(t *testing.T) {
+	td := t.TempDir()
+	testCopyDir(t, testFixturePath("apply-targeted"), td)
```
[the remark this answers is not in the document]
RLRabinowitz: cbfd781ad1ec456c87e338c3cde4c221db8eb2a7

internal/command/arguments/plan_test.go
```diff
@@ -160,6 +162,79 @@ func TestParsePlan_targets(t *testing.T) {
...
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			got, diags := ParsePlan(tc.args)
+			if len(diags) > 0 {
+				if tc.wantErr == "" {
+					t.Fatalf("unexpected diags: %v", diags)
+				} else if got := diags.Err().Error(); !strings.Contains(got, tc.wantErr) {
+					t.Fatalf("wrong diags\n got: %s\nwant: %s", got, tc.wantErr)
+				}
```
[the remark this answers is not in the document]
RLRabinowitz: a2bbdf7f08eddcd42e88bc6a75a4e635597fe172

internal/command/arguments/extended.go
```diff
@@ -85,20 +89,20 @@ type Operation struct {
...
+// addrs.Targetable
+// This is used for parsing the input of -target and -exclude flags
+func parseTargetables(rawTargetables []string) ([]addrs.Targetable, tfdiags.Diagnostics) {
+	var targetables []addrs.Targetable
 	var diags tfdiags.Diagnostics
 
-	o.Targets = nil
-
-	for _, tr := range o.targetsRaw {
+	for _, tr := range rawTargetables {
```
Yantrio, line 105: the error message here still says "Invalid Target" but it could be exclude.

cam72cam (dismissed):

internal/command/arguments/extended.go
RLRabinowitz: 2076f53523e55944bcc4a18871125a5890f2c502

cam72cam (dismissed):

cam72cam (approved):

PR #1911, opened 2024-08-13
300/provider foreach

Andrew-Hayes:
Resolves # [300](https://github.com/opentofu/opentofu/issues/300)

This PR adds for_each support to providers. To achieve that we have made a few changes

- We evaluate a provider and if it has a for_each statement we create multiple instances of the provider based on the for_each iteration.
- Within the provider we now have a ProviderBlock and Provider. The ProviderBlock is populated from the provider block in the hcl and then we evaluate that. This is then translated to 1 or more "Provider" 's that contain the instance specific data. 
- We have moved the evaluation checks for the for_each and count of a resource into a "lang" sub package. So it can be re-used. We then make use of this for provider for_each evaluations. 
- Added unit tests for the new changes

## Target Release

1.9.0

## Checklist

### Go checklist

 


internal/configs/resource.go
```diff
@@ -668,6 +668,24 @@ func decodeProviderConfigRef(expr hcl.Expression, argName string) (*ProviderConf
 	expr, shimDiags = shimTraversalInString(expr, false)
 	diags = append(diags, shimDiags...)
 
+	if idxexpr, ok := expr.(*hclsyntax.IndexExpr); ok {
```
cam72cam, line 671: I'm not sure why these changes are in this PR

internal/configs/provider.go
```diff
@@ -70,17 +77,25 @@ func decodeProviderBlock(block *hcl.Block) (*Provider, hcl.Diagnostics) {
...
-		}
+	if attr, exists := content.Attributes["for_each"]; exists {
+		provider.ForEach = attr.Expr
+	}
+
+	if attr, exists := content.Attributes["count"]; exists {
+		provider.Count = attr.Expr
+	}
+
+	if provider.Count != nil && provider.ForEach != nil && provider.AliasExpr != nil {
```
cam72cam, line 92: This expression is incorrect

internal/configs/provider.go
```diff
@@ -145,6 +160,109 @@ func decodeProviderBlock(block *hcl.Block) (*Provider, hcl.Diagnostics) {
...
+		diags = append(diags, evalDiags.ToHCL()...)
+		if evalDiags.HasErrors() {
+			return nil, diags
+		}
+
+		var out []*Provider
+		for k, v := range forVal {
+			v := v
+			iter := *p
+			iter.Alias = k
```
cam72cam, line 195: We probably want to check `hclsyntax.ValidIdentifier(alias)`, not sure how the numeric count will handle that name.0 for example.

internal/configs/provider.go
Andrew-Hayes: yeah, i thought that when writing it. now to remember the logic lecture from first year of uni.

cam72cam: I'd recommend something like:
```
if (alias != nil && count != nil) || (alias != nil && for_each != nil) || (count != nill && for_each != nil) {
  // common error message
}
```

internal/tofu/eval_iteration_test.go
cam72cam, line 1: I wonder if it makes sense to migrate some/all these tests to the eval_checks package?

internal/configs/provider_iteration_test.go
```diff
@@ -0,0 +1,82 @@
...
+	if name, exists := mod.ProviderLocalNames[p]; !exists {
+		t.Fatal("provider FQN hashicorp/local not found")
+	} else {
+		if name != "local" {
+			t.Fatalf("provider localname mismatch: got %s, want local", name)
+		}
+	}
+
+	if len(mod.ProviderConfigs) != 3 {
+		t.Fatal("incorrect numver of providers")
```
cam72cam, line 30: number

internal/tofu/eval_iteration_test.go
Andrew-Hayes: yeah it does. some are tightly coupled though, even more so than the main code. but is def worth the effort.

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
 	return diags
 }
+
+func (p *Provider) generateForEachProviders(eval *StaticEvaluator) ([]*Provider, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+	if eval == nil {
+		return nil, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Iteration not allowed in test files",
+			Detail:   "for_each was declared as an provider attribute in a test file",
```
ollevche, line 343:
```suggestion
			Detail:   "for_each was declared as a provider attribute in a test file",
```

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+			Subject:  p.AliasExpr.Range().Ptr(),
+		})
+	}
+
+	foeachRefsFunc := func(refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
```
ollevche, line 348: foeachRefsFunc -> forEachRefsFunc

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
 }
+
+func (p *Provider) generateForEachProviders(eval *StaticEvaluator) ([]*Provider, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+	if eval == nil {
+		return nil, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Iteration not allowed in test files",
+			Detail:   "for_each was declared as an provider attribute in a test file",
+			Subject:  p.AliasExpr.Range().Ptr(),
```
ollevche, line 344:
```suggestion
			Subject:  p.ForEach.Range().Ptr(),
```

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+		return nil, diags
+	}
+
+	var out []*Provider
+	for k, v := range forVal {
+		if !hclsyntax.ValidIdentifier(k) {
+			return nil, diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid Identifier",
+				Detail:   fmt.Sprintf("The provided identifier %s is invalid", k),
```
ollevche, line 370: nit: Invalid for_each key identifier?

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+	return out, diags
+}
+
+func (p *Provider) generateCountProviders(eval *StaticEvaluator) ([]*Provider, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+	if eval == nil {
+		return nil, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Iteration not allowed in test files",
+			Detail:   "for_each was declared as an provider attribute in a test file",
```
ollevche, line 391:
```suggestion
			Detail:   "count was declared as a provider attribute in a test file",
```

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+			Subject:  p.AliasExpr.Range().Ptr(),
```
ollevche, line 392:
```suggestion
			Subject:  p.Count.Range().Ptr(),
```

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+	}
+
+	foeachRefsFunc := func(refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
+		var diags tfdiags.Diagnostics
+		evalContext, evalDiags := eval.EvalContext(StaticIdentifier{
+			Module:    eval.call.addr,
+			Subject:   fmt.Sprintf("provider.%s.for_each", p.Name),
+			DeclRange: p.ForEach.Range(),
+		}, refs)
+		return evalContext, diags.Append(evalDiags)
```
ollevche, line 355: Seems like we will have duplicated diags if we append them here and in `EvaluateForEachExpression`. The same for count

internal/configs/provider.go
```diff
@@ -290,3 +333,89 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+	var out []*Provider
+	countVal, evalDiags := evalchecks.EvaluateCountExpression(p.Count, countEvalFunc)
+	diags = append(diags, evalDiags.ToHCL()...)
+	if evalDiags.HasErrors() {
+		return nil, diags
+	}
+
+	for i := 0; i < countVal; i++ {
+		iter := *p
+		iter.Alias = fmt.Sprintf("[%d]", i)
```
ollevche, line 414: Do we want count providers to be available only via square brackets? Should we let the caller decide?

internal/eval_checks/diagnostics.go
```diff
@@ -3,7 +3,7 @@
 // Copyright (c) 2023 HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
-package tofu
+package evalchecks
```
ollevche, line 6: can we make the package name the same as the folder name?

ollevche (commented):
Overall it looks pretty solid, left a couple of questions and suggestions. Thank you!

internal/configs/provider.go
```diff
@@ -23,7 +25,12 @@ type Provider struct {
 	Name       string
 	NameRange  hcl.Range
 	Alias      string
-	AliasRange *hcl.Range // nil if no alias set
+	AliasExpr  hcl.Expression // nil if no alias set
+	AliasRange *hcl.Range     // nil if no alias set
+	ForEach    hcl.Expression
+	EachValue  *cty.Value
+	Count      hcl.Expression
+	CountIndex *cty.Value
```
Evi1Pumpkin, line 33: Two things that quite bother me with this implementation:
1. This struct is no longer as described in the comment above: `Provider represents a "provider" block in a module or file.` It sometimes represents a block and sometimes a provider "instance".
2. Now, it holds two responsibilities and changes based on the application's lifecycle. If `tofu` is in the stage of reading the configuration, this struct represents the `provider` block. If `tofu` is building the module tree and populating the providers (after calling `decodeStaticFields()`), the struct now describes a specific "instance" of the provider.

This implementation is quite confusing and different from the practices in the tofu repository. Maybe we should consider adding a different struct that will include the `EachValue` and `CountIndex`?

Evi1Pumpkin (commented):
Overall looking excellent, left a few comments

internal/configs/provider.go
cam72cam: 1. It's more so a "provider block" vs "provider block with instance data"
2. This is true, the expansion here feels a bit weird IMO and I noticed that during the prototype phase.  This change is the "low churn" option, but adds cognitive load to understanding provider configs in the configs package. 

There are technically 3 levels of config parsing that are all mixed together:
1. hcl/json text -> configs.File (HCL loading)
2. configs.File -> configs.Module (merging files)
3. configs.Module -> configs.Module (static evaluation)

Even before static evaluation was added, there were all sorts of odd validation bugs all throughout the code when dealing with overrides.  Static evaluation shines more of a light on this.

In this specific scenario, we would need:
```go

// File parsing produces
type configs.ProviderBlock struct {
  // 1-1 mapping to HCL
}

// Module struct from Files (normal+override) produces
type configs.ProviderData struct {
  // Closer to existing struct
  ForEach hcl.Expression
  Count hcl.Expression
}
// Validation is performed when constructing ProviderData

// func (ProviderData) expand(StaticContext) []Provider

// Expanded into one or more
type configs.Provider struct {
  // Matches existing struct
  
  ForEachValue *cty.Value
  CountValue *cty.Value
  // OR
  InstanceData instances.RepetitionData
}
```


For the purposes of this PR, we could merge the concept of ProviderBlock and ProviderData to reduce the potential set of changes as it is a less important distinction IMO.

internal/eval_checks/diagnostics.go
Andrew-Hayes: oh yeah. i just lazily threw the package here until we figured out where it should go.. then never did..

internal/configs/provider.go
ronnyorot: I agree ^
Also, I prefer the usage of `InstanceData instances.RepetitionData` over `ForEachValue` and `CountValue`.

Evi1Pumpkin: I agree ^
Also, I prefer the usage of InstanceData instances.RepetitionData over ForEachValue and CountValue.

ollevche: We already track some of the modernization/refactoring issues and I would suggest we document that down to address later. Both @Evi1Pumpkin and @cam72cam arguments sound pretty reasonable.

Breaking down config package structures into different stages such as loading (raw) and evaluation (ready for graph processing) makes perfect sense to me.

internal/configs/provider.go
```diff
@@ -70,17 +77,25 @@ func decodeProviderBlock(block *hcl.Block) (*Provider, hcl.Diagnostics) {
...
+	if attr, exists := content.Attributes["count"]; exists {
+		provider.Count = attr.Expr
+	}
+
+	if (provider.AliasExpr != nil && provider.Count != nil) || (provider.AliasExpr != nil && provider.ForEach != nil) || (provider.Count != nil && provider.ForEach != nil) {
+		diags = append(diags, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  `Invalid combination of "alias", "count" and "for_each"`,
+			Detail:   `The "alias", "count" and "for_each" arguments are mutually-exclusive, only one may be used.`,
+			Subject:  provider.AliasExpr.Range().Ptr(),
```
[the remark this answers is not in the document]
Andrew-Hayes: funny i did have them split but Christian advocated for a combined error message

internal/configs/provider.go
Andrew-Hayes: interesting. so i had made changes to resource.go so i could reference a provider like so: 

```
provider = local.0
```

this made use of https://github.com/opentofu/opentofu/blob/main/internal/addrs/instance_key.go#L66-L68 

this meant that it had to be wrapped in square brackets to match. However thats all been remove as it steps on the toes of what you are working on. So not sure if i still need them 🤔

internal/configs/provider.go
```diff
@@ -154,13 +204,6 @@ func (p *Provider) Addr() addrs.LocalProviderConfig {
 	}
 }
 
-func (p *Provider) moduleUniqueKey() string {
```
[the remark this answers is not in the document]
Andrew-Hayes: i had initially copied it from the original branch without thinking about it. So it is perhaps not needed, but as Addr().StringCompact() already exists is it duplicate code to have both? 🤔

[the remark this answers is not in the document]
Evi1Pumpkin: I don't mind, we can leave it like this. I was trying to understand the reason behind that and if I misunderstood or missed something. Thanks for the clarification!

internal/configs/provider.go
Andrew-Hayes: We discussed this offline. The agreement was to go with @cam72cam 's suggestion and do 2 structs: one that captures the understanding of "pre-evaluation" and one that is "post evaluation". Then we will revisit in future with an RFC on larger refactoring work.

internal/configs/provider.go
Andrew-Hayes: this func is called in EvaluateForEachExpression. So EvaluateForEachExpression returns this diags as it's own.

internal/configs/provider.go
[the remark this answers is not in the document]
ollevche: It is nit, you can just ignore that part of the comment

internal/configs/provider.go
ollevche: My bad, I thought `diags` variable is from the outer scope. Thanks!

internal/configs/provider.go
```diff
@@ -290,3 +354,104 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
 	}
 	return diags
 }
+
+func (p *ProviderBlock) toProvider() Provider {
+	return Provider{
+		ProviderCommon: ProviderCommon{
```
cam72cam, line 360: return Provider{ProviderCommon: p.ProviderCommon}

internal/configs/provider.go
```diff
@@ -154,11 +219,12 @@ func (p *Provider) Addr() addrs.LocalProviderConfig {
 	}
 }
 
-func (p *Provider) moduleUniqueKey() string {
-	if p.Alias != "" {
-		return fmt.Sprintf("%s.%s", p.Name, p.Alias)
+// Addr returns the address of the receiving provider configuration, relative
+// to its containing module.
+func (p *ProviderBlock) Addr() addrs.LocalProviderConfig {
```
cam72cam, line 224: This function is no longer used

internal/configs/provider.go
```diff
@@ -13,17 +13,14 @@ import (
...
 
-// Provider represents a "provider" block in a module or file. A provider
-// block is a provider configuration, and there can be zero or more
-// configurations for each actual provider.
-type Provider struct {
-	Name       string
-	NameRange  hcl.Range
-	Alias      string
-	AliasRange *hcl.Range // nil if no alias set
+type ProviderCommon struct {
```
ollevche, line 21: I understand, we already discussed this previously and I do not have a strong opinion here.

But...

Do we need `ProviderCommon`? It is not useful by itself and doesn't say anything about what it really is.

We could move `ProviderCommon` fields to `ProviderBlock` and tweak a `Provider` struct to include a `ConfigBlock` field of type `ProviderBlock` (instead of embedding). This looks good to me since we have both clean separation and source configuration values.

internal/configs/provider.go
```diff
@@ -290,3 +321,76 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
...
+				Detail:   fmt.Sprintf("The provided identifier %s is invalid", k),
+				Subject:  p.ForEach.Range().Ptr(),
+			})
+		}
+
+		v := v
+		iter := Provider{ProviderCommon: p.ProviderCommon}
+		iter.Alias = k
+		iter.InstanceData.EachValue = v
+		out = append(out, &iter)
```
ollevche, line 367: nit:

```suggestion
		out = append(out, &Provider{
			ProviderCommon:		p.ProviderCommon,
			Alias:			k,
			InstanceData:		instances.RepetitionData{
				EachValue:	v,
			},
		})
```

internal/configs/test_file.go
```diff
@@ -396,10 +397,22 @@ func loadTestFile(body hcl.Body) (*TestFile, hcl.Diagnostics) {
...
+				diags = append(diags, decodeDiags...)
+				if decodeDiags.HasErrors() {
+					continue
+				}
+				providers, diagsStatic := providerBlock.decodeStaticFields(nil)
+				diags = append(diags, diagsStatic...)
+				if diagsStatic.HasErrors() {
+					continue
+				}
+				for _, provider := range providers {
```
ollevche, line 413: Creating a provider list out of a provider block with a nil static eval should always result in a single provider, right?

nit: should we just make it `providers[0]` for clarity?

internal/evalchecks/eval_for_each.go
```diff
@@ -2,33 +2,40 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright (c) 2023 HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
+package evalchecks
 
-package tofu
+// TODO this is only here until i figure out where it should be
```
ollevche, line 7: Is it still relevant?

ollevche (commented):
Looks great, thank you! Left a couple of questions and several nit suggestions. Feel free to ignore them if you think they are irrelevant.

internal/configs/provider.go
Andrew-Hayes: i would tend to agree, i'm not a fan of embedding. But i'm ok with what we have

internal/evalchecks/eval_for_each.go
Andrew-Hayes: kind of. I still feel like there is maybe somewhere better it should be? but no-one seems to mind it being here...

internal/configs/config.go
```diff
@@ -966,15 +965,16 @@ func (c *Config) transformProviderConfigsForTest(run *TestRun, file *TestFile) (
...
+				ProviderCommon: ProviderCommon{
+					Name:          ref.InChild.Name,
+					NameRange:     ref.InChild.NameRange,
+					Version:       testProvider.Version,
+					Config:        testProvider.Config,
+					DeclRange:     testProvider.DeclRange,
+					IsMocked:      testProvider.IsMocked,
+					MockResources: testProvider.MockResources,
+				},
+				Alias: ref.InChild.Alias,
```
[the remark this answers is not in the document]
Andrew-Hayes: The range is part of the block as it's linked to where the expression lives. It didn't look like it was used for anything once we have the instance data etc. but I could be wrong.. 👀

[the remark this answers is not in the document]
ollevche: Yep, seems like it is not used anywhere. We can add it back anytime if needed though

internal/configs/provider.go
```diff
@@ -290,3 +325,79 @@ func checkProviderNameNormalized(name string, declrange hcl.Range) hcl.Diagnosti
 	}
 	return diags
 }
+
+func (p *ProviderBlock) generateForEachProviders(eval *StaticEvaluator) ([]*Provider, hcl.Diagnostics) {
```
Evi1Pumpkin, line 329: Can we keep the methods of the `ProviderBlock` struct close to each other and to the struct itself?
Or else it's a bit confusing when there are similar structs in the same file with mixed functions between them (also relevant for `providerBlockSchema`, which is only relevant to the `ProviderBlock`)

internal/evalchecks/eval_for_each.go
cam72cam: I suggested maybe in the lang package.  If it does not fit there, I think leaving it as it's own package works well enough for now.

internal/configs/test_file.go
```diff
@@ -396,10 +397,22 @@ func loadTestFile(body hcl.Body) (*TestFile, hcl.Diagnostics) {
...
+				_, decodeDiags := providerBlock.decodeStaticFields(nil)
+				diags = append(diags, decodeDiags...)
+				if decodeDiags.HasErrors() {
+					continue
+				}
+				providers, diagsStatic := providerBlock.decodeStaticFields(nil)
+				diags = append(diags, diagsStatic...)
+				if diagsStatic.HasErrors() {
+					continue
+				}
```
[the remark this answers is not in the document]
Andrew-Hayes: no idea how I missed this.. haha

internal/evalchecks/eval_for_each.go
Andrew-Hayes: import cycled in the lang package itself. but I moved it into a sub package and it felt like it might be more at home there?

internal/configs/provider.go
Andrew-Hayes: ok, new layout has the following reasons: 

- Schema at the bottom due to convention. 
- ProviderBlock functions above that to keep them grouped
- Provider functions above that to keep struct methods together
- Private functions above that
- Public functions above that. because that felt right, but is probably some java convention that I remember from the dark days.

internal/evalchecks/eval_for_each.go
cam72cam: Sounds good to me!

ollevche (dismissed):

internal/configs/provider.go
```diff
@@ -95,17 +200,8 @@ func decodeProviderBlock(block *hcl.Block) (*Provider, hcl.Diagnostics) {
...
-		if attr, exists := content.Attributes[name]; exists {
-			diags = append(diags, &hcl.Diagnostic{
-				Severity: hcl.DiagError,
-				Summary:  "Reserved argument name in provider block",
-				Detail:   fmt.Sprintf("The provider argument name %q is reserved for use by OpenTofu in a future version.", name),
-				Subject:  &attr.NameRange,
-			})
-		}
-	}
+	reserveredDiags := checkReserverNames(content)
```
cam72cam, line 203: checkReservedNames

internal/configs/provider.go
```diff
@@ -154,104 +294,108 @@ func (p *Provider) Addr() addrs.LocalProviderConfig {
...
-	if len(traversal) > 2 {
-		diags = diags.Append(&hcl.Diagnostic{
-			Severity: hcl.DiagError,
-			Summary:  "Invalid provider configuration address",
-			Detail:   "Extraneous extra operators after provider configuration address.",
-			Subject:  traversal[2:].SourceRange().Ptr(),
-		})
+	forEachRefsFunc := func(refs []*addrs.Reference) (*hcl.EvalContext, tfdiags.Diagnostics) {
+		var diags tfdiags.Diagnostics
+		evalContext, evalDiags := eval.EvalContext(StaticIdentifier{
```
cam72cam, line 347: This can just be a return, don't need additional vars here

internal/configs/provider.go
```diff
@@ -154,104 +294,108 @@ func (p *Provider) Addr() addrs.LocalProviderConfig {
...
-// then the returned address is invalid.
-func ParseProviderConfigCompactStr(str string) (addrs.LocalProviderConfig, tfdiags.Diagnostics) {
-	var diags tfdiags.Diagnostics
+	var out []*Provider
+	for k, v := range forVal {
+		if !hclsyntax.ValidIdentifier(k) {
+			return nil, diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid for_each key Identifier",
+				Detail:   fmt.Sprintf("The provided identifier %s is invalid", k),
```
cam72cam, line 367: Let's make this a better error message, maybe include badIdentifierDetail and rename "identifier" to  alias

internal/configs/provider.go
Andrew-Hayes: it needs them because EvalContext returns `hcl.Diagnostics` but we need `tfdiags.Diagnostics`. So have to do that append magic.

internal/configs/test_file.go
```diff
@@ -396,10 +397,17 @@ func loadTestFile(body hcl.Body) (*TestFile, hcl.Diagnostics) {
...
 			diags = append(diags, providerDiags...)
-			if provider != nil {
-				tf.Providers[provider.moduleUniqueKey()] = provider
+			if providerBlock != nil {
+				providers, diagsStatic := providerBlock.decodeStaticFields(nil)
+				diags = append(diags, diagsStatic...)
+				if diagsStatic.HasErrors() {
+					continue
+				}
+				// providers only ever has 1 element for nul static fields.
```
cam72cam, line 408: typos

internal/configs/test_file.go
```diff
@@ -396,10 +397,17 @@ func loadTestFile(body hcl.Body) (*TestFile, hcl.Diagnostics) {
...
+				provider := providers[0]
```
cam72cam, line 409: Even though the comment is currently true, let's do a for-loop instead just to be safe against future enhancements

internal/configs/provider.go
cam72cam: Ah, my mistake!

internal/configs/test_file.go
cam72cam: I asked to keep it as a loop to future proof it.

cam72cam (approved):
Approving with the caveat that additional tests and documentation will be coming in the larger #300 feature branch.

PR #1939, opened 2024-08-23
Add RFC for global provider cache locking

cam72cam:
Proof of concept: https://github.com/opentofu/opentofu/pull/1878

RFC for #1483 

## Target Release

1.9.0

## Checklist


ollevche (approved):

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
+# Safe Global Provider Cache Access
+
+Issue: https://github.com/OpenTofu/opentofu/issues/1483
+
+Many CI/CD systems download all providers for every single tofu action run. The global provider cache allows downloaded providers to be saved in between runs and linked into the local provider cache (in .terraform/). Unfortunately, the global provider cache does not have any locking around it and can fail when used in multiple simultaneous runs.
+
+Another common use case is terragrunt. When running `tofu init` on multiple projects simultaneously via the tool, there is a high likely hood of a conflicting write to a provider's directory. This has caused some interesting tooling workarounds on their end, which are not ideal.
```
Yantrio, line 7: It might be worth explaining that this is mainly when using terragrunt's `run-all` or `graph` commands.

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+
+Additionally, as we build out true e2e tests for OpenTofu, safe access to the global provider cache can dramatically reduce run times.
```
Yantrio, line 9:
```suggestion
Additionally, as we build out true e2e tests for OpenTofu, safe access to the global provider cache can dramatically reduce run times by allowing us to run these concurrently.
```

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Proposed Solution
+
+A filesystem level lock should be added that is safe for cross-process and in some scenarios cross-machine access. It will be best-effort and should rely on standard locking practices (not home grown) and be transparent to the user.
```
Yantrio, line 13: I'm a bit worried that this can be mis-understood. A "filesystem level lock" could also just be a lockfile written to disk.

We should mention that we're looking at using the native syscalls to lock files at the filesystem level here. 

Especially when we talk about lockfiles below. I don't want people to mix up the 2

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+- Using hashicorp/local v2.5.1 from the shared cache directory
+```
+
+Any number of tofu instances should be able to be run in parallel and still have safe access to the global cache directory.
+
+#### Scenarios which currently cause failures
+
+Multiple init/plans/apply without a primed cache.
+
+One possibility is that each project downloads and overwrites the same provider files.  The global provider cache scan is not a fast process and should be run as little as possible, it is therefore only run at the start of `tofu init`. ProjectA and ProjectB's inits may be run at the same time.  When this occurs, they will both download the full set of providers that they need and clobber each others files.  This is not ideal, but is only a waste of time/resources.
```
Yantrio, line 39: If we're doing a filesystem level lock, and the files are clobbered. does that mean that we could end up with a clash of locking? I think we should write a bit more information about the impact of this apart from just time/resources and explore possibilities a bit more incase people have these kinds of questions in the future.

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+
+Another possibility is that one project may already be planning/applying and have a live provider executable running, effectively locking it on disk.  If ProjectA is still downloading providers during init and ProjectB has a much smaller list and is already in plan, ProjectA may attempt to overwrite the provider binary that ProjectB is running and maintaining an execution lock on.
```
Yantrio, line 41: This is a huge problem and I don't think its a rare edge-case either. we should attempt to address this somehow.

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+$ rm .terraform/ -r
+$ TF_PLUGIN_CACHE_DIR=~/.tofu.d/plugin-cache/ tofu init
+Initializing provider plugins...
+- Reusing previous version of hashicorp/local from the dependency lock file
+- Using hashicorp/local v2.5.1 from the shared cache directory
+```
+
+Any number of tofu instances should be able to be run in parallel and still have safe access to the global cache directory.
+
+#### Scenarios which currently cause failures
```
Yantrio, line 35: I suspect that people may want to run multiple `terraform` executions in a container and have a common cache volume between them. Can we document situations where containers may crash and end up with an indefinite lock on them?

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+
+Mixture of missing platform/corrupt lock files with valid lock files.
+
+The provider lock file may have been generated on a system with a different architecture or may have become corrupt due to a variety of reasons. When this happens, the global cache may not match the lock file and force a re-download of the provider. This will cause potentially unexpected downloads in the above scenarios.
+
+### Technical Approach
+
+This proposal hinges on a stable and consistent cross-platform lock. The codebase already contains this in the form of locking local state files.  This code is battle tested and overall quite simple to use.  With some light refactoring, this filesystem locking code can be moved into it's own internal package and used both to lock providers and to lock the local state file.
+
+The existing file locking should be safe on any local filesystems, but should be used with caution on shared volumes such as legacy NFS shares who do not provide strong locking consistency.
```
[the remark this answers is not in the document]
Yantrio: We should document that we don't recommend NFS then

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,68 @@
...
+
+Within the provider installation code, the whole section which inspects and links to the current providers available should be locked at the provider level.  The lock should be done at this granularity specifically for the complicating factors mentioned above.
+
+Additionally, we should make the package installer smarter and able to check the files in the cache against the downloaded version. This prevents a bad cache entry from overwriting a valid entry which may already been in use by another process.
+
+### Open Questions
+
+* Is fnctl flock safe enough? It is industry standard and battle tested.  Are there any additional caveats that should be mentioned in the docs?
```
Yantrio, line 60: I would love if we could discuss options above. I'm a bit concerned that we've discussed that we do want to lock, but not *how* we want to lock until you mention fnctl here. 

I think this ties into my comment about explaining 'file system level locking' above.

Yantrio (changes requested):

yhakbar:
Hello from the Terragrunt maintainers!

We have some feedback that we've consolidated here so that you don't have a bunch of comments from different developers:

> Terragrunt/Gruntwork _[sic]_ provides a http provider mirror that can be run locally and maintains a provider cache (archives) external to OpenTofu itself. While this does have some advantages, it still requires each provider to be extracted into the local provider cache folder, taking additional time and space.  It may be a safer alternative to using a networked filesystem between systems.

This isn't technically correct (for one, we're _Gruntwork_, not _Gruntworks_ XD). Terragrunt downloads providers into a shared cache folder. and creates symlinks to providers for each `.terraform` directory. It actually stores the provider plugins within the cache in an identical manner, so no additional time or space is consumed (aside from the size of symlinks and the time to create them).

The Terragrunt provider cache server also doesn't technically act as an HTTP mirror from the perspective of OpenTofu. It technically operates as a locally running remote registry.

We'd like to be able to recommend that Terragrunt users using OpenTofu as the underlying IaC engine use this capability directly instead of using the Terragrunt provider cache server (assuming they don't have requirements like storing provider plugins on an NFS mounted drive, etc). There are some problems with the proposal that would prevent that right now, however.

One problem that the proposed solution does not address the fact that no cached providers will be used if the `.terraform.lock.hcl` file is not available in the current working directory. As OpenTofu currently functions, running `init` with a `TF_PLUGIN_CACHE_DIR` will only use cached providers if the `.terraform.lock.hcl` file exists in the current directory.

In the implementation of the Terragrunt provider cache server, we've worked around this issue by doing something like the following before resolving an `init` a second time:

```bash
tofu providers lock -fs-mirror=/shared/cache/dir
```

The `tofu providers lock` command doesn't use the `TF_PLUGIN_CACHE_DIR` directory at all, so using the `fs-mirror` is the workaround we've leveraged to force OpenTofu to respect the providers that have already been downloaded.

This has it's own issues, however, as `tofu providers lock` with `-fs-mirror` will throw an error if any provider plugin is missing from the cache, with an error like the following:

```
│ Error: Could not retrieve providers for locking
│
│ OpenTofu failed to fetch the requested providers for darwin_arm64 in order to calculate their checksums: some providers could not be installed:
```

All that to say, the solution should ensure that `init` and `providers lock` will work with or without a `.terraform.lock.hcl` file present in the local directory, and that partial availability of cached providers in the `TF_PLUGIN_CACHE_DIR` and the `.terraform.lock.hcl` should result in downloading the appropriate providers, and re-using available providers. It makes total sense to avoid editing the `.terraform.lock.hcl` file unless a user uses `tofu init -upgrade`, but in that scenario, existing plugins shouldn't be re-downloaded unless necessary.

These are corner cases you might have already evaluated, but we wanted to make sure they were addressed by this proposal to make it a viable alternative to the Terragrunt provider cache server.

rfc/20240824-provider-cache-locking.md
yhakbar: Not entirely accurate. 

Unless users are using [terragrunt-fetch-dependency-output-from-state](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-fetch-dependency-output-from-state), users can still end up writing providers to disk for multiple Units (the directories where `terragrunt.hcl` files live) concurrently for dependencies.

rfc/20240824-provider-cache-locking.md
cam72cam: This is under the section of current failure scenarios, not what we are proposing happen under the new setup

rfc/20240824-provider-cache-locking.md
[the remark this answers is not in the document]
cam72cam: ebf7dae4c4

rfc/20240824-provider-cache-locking.md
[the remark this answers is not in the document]
cam72cam: ebf7dae4c4

rfc/20240824-provider-cache-locking.md
cam72cam: ebf7dae4c4

rfc/20240824-provider-cache-locking.md
cam72cam: ebf7dae4c4

rfc/20240824-provider-cache-locking.md
mwos-sl: Also, this problem is not unique to `terragrunt run-all`. The same problem is in atlantis/jenkins with parallel stages, no matter if you use terragrunt or not.

mwos-sl:
I find this comment from the community member very valuable, so sharing it here to underline that simple solution could be probably also beneficial: https://github.com/hashicorp/terraform/pull/33479#issuecomment-1665138221

wosiu:
I find this comment from the community member very valuable, so sharing it here to underline that simple solution could be probably also beneficial: https://github.com/hashicorp/terraform/pull/33479#issuecomment-1665138221

rfc/20240824-provider-cache-locking.md
```diff
@@ -0,0 +1,69 @@
...
+* Is a filesystem lock syscall safe enough? It is industry standard and battle tested.  Are there any additional caveats that should be mentioned in the docs?
+  - Yes. As mentioned above, we should document that networked filesystems are not recommended for this use case.
+
+### Future Considerations
+
+A considerable amount of time is spent scanning the provider cache.  Instead, it should probably be refactored to only read the provider metadata necessary at any given time.  This will offer some significant performance bonuses on large configurations.
+
+## Potential Alternatives
+
+Terragrunt/Gruntwork provides a http provider mirror that can be run locally and maintains a provider cache (archives) external to OpenTofu itself. While this does have some advantages, it still requires each provider to be extracted into the local provider cache folder, taking additional time and space.  It may be a safer alternative to using a networked filesystem between systems.
```
wosiu, line 69: FWIW - It is not that great. Terrwgraunt starts local https proxy server within single invocation of `terragrunt run` / `terragrunt run-all` and stops the server on the end of the command (there is a ticket to improve this: https://github.com/gruntwork-io/terragrunt/issues/3363).

So currently, when you run in parallel multiple `terragrunt run` there is still racing. That's the case for atlantis+terragrunt setup for example. So I would not consider terragrunt as a "potential alternative". We do need this fixed on opentofu level.

cam72cam: Being honest, I'm not sure I'll have time to dig back into this until the new year.  I'll try though.

Yantrio (approved):

apparentlymart (approved):

PR #1950, opened 2024-08-27
Fix setops not used in block set renderer

buraksenn:
While implementing https://github.com/opentofu/opentofu/pull/1948 I have realized that `setOpts` is defined but is not used. I think this was a mistake and opened a PR for this. 

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonformat/computed/renderers/block.go
+++ b/internal/command/jsonformat/computed/renderers/block.go
@@ -79,7 +79,6 @@ func (renderer blockRenderer) RenderHuman(diff computed.Diff, indent int, opts c
 	for _, key := range attributeKeys {
 		attribute := renderer.attributes[key]
 		if importantAttribute(key) {
-
 			// Always display the important attributes.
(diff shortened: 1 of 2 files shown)


buraksenn:
Can I get reviews on this and #1948  when you have time @cam72cam, @ollevche ?

ollevche (dismissed):
Thanks for the fix! Looks like a typo indeed.

Small note for future readers: from the user perspective tofu now will put correct "# forces replacement" suffixes for some of the set blocks.

BTW @buraksenn do you mind putting a test case for this scenario?

buraksenn:
> Thanks for the fix! Looks like a typo indeed.
> 
> Small note for future readers: from the user perspective tofu now will put correct "# forces replacement" suffixes for some of the set blocks.
> 
> BTW @buraksenn do you mind putting a test case for this scenario?

Thanks @ollevche for the review. Added a test case for this. I've also debugged test case if setOpts is not passed `#forces replacement` is not outputted so I think this was unintentional

cam72cam (approved):

PR #1959, opened 2024-09-02
#300 add support for count in provider refs

ollevche:
This PR makes it possible to use count.index expression when referencing a provider.

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/resource.go
+++ b/internal/configs/resource.go
@@ -7,6 +7,7 @@ package configs
 
 import (
 	"fmt"
+	"math/big"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
@@ -119,9 +120,20 @@ func (r *Resource) HasCustomConditions() bool {
 
 func (r *Resource) decodeStaticFields(eval *StaticEvaluator) hcl.Diagnostics {
 	if r.ProviderConfigRef != nil {
-		return r.ProviderConfigRef.decodeStaticAlias(eval, r.ForEach)
+		return r.ProviderConfigRef.decodeStaticAlias(eval, r.getInstanceExpression())
 	}
 	return nil
+}
+
+func (r *Resource) getInstanceExpression() hcl.Expression {
+	switch {
+	case r.ForEach != nil:
+		return r.ForEach
+	case r.Count != nil:
+		return r.Count
+	default:
+		return nil
+	}
 }
 
 func decodeResourceBlock(block *hcl.Block, override bool) (*Resource, hcl.Diagnostics) {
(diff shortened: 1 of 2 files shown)


internal/configs/resource.go
```diff
@@ -813,56 +929,24 @@ func (m *ProviderConfigRefMapping) decodeStaticAlias(eval *StaticEvaluator, forE
...
+	switch {
+	case hasEachRefInAlias:
+		aliases, aliasDiags = generateForEachAliases(instances, aliasEvalCtx, m.Alias)
+	case hasCountRefInAlias:
+		aliases, aliasDiags = generateCountAliases(instances, aliasEvalCtx, m.Alias)
 	}
 
-	if !diags.HasErrors() {
+	diags = diags.Extend(aliasDiags)
+	if !aliasDiags.HasErrors() {
```
Andrew-Hayes, line 949: this is more of a tidy up on the existing code. should we tweak this last bit to be more idiomatic? 

```
diags = diags.Extend(aliasDiags)
if aliasDiags.HasErrors() {
    return diags
}

m.Aliases = aliases
return diags
```

internal/configs/resource.go
```diff
@@ -813,56 +929,24 @@ func (m *ProviderConfigRefMapping) decodeStaticAlias(eval *StaticEvaluator, forE
...
-		instanceKey, err := addrs.ParseInstanceKey(k)
-		if err != nil {
-			panic("decodeStaticAlias failed to parse for each instance keys: wrong expression?")
-		}
-
-		aliases[instanceKey] = aliasVal.AsString()
+	switch {
+	case hasEachRefInAlias:
+		aliases, aliasDiags = generateForEachAliases(instances, aliasEvalCtx, m.Alias)
+	case hasCountRefInAlias:
```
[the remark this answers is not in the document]
ollevche: That is okay because it is handled earlier [at line 898](https://github.com/opentofu/opentofu/pull/1959/files/1d469a886ae1feabfdf9d156a9d9bff2a2bbf03a#diff-107610f23331475f8c654225fe47be1f0e706b7acbea73c9b39af9fb9e0ebe13R898). I left a comment to make it more obvious in [8075c3f](https://github.com/opentofu/opentofu/pull/1959/commits/8075c3f26301c29062c5e2568f6bba9a762b71c5).

internal/configs/resource.go
ollevche: Thanks! Made it more idiomatic in [8075c3f](https://github.com/opentofu/opentofu/pull/1959/commits/8075c3f26301c29062c5e2568f6bba9a762b71c5)

internal/configs/resource.go
```diff
@@ -740,7 +752,97 @@ func (m *ProviderConfigRefMapping) GetNoKeyAlias() string {
...
+
+	aliases := make(map[addrs.InstanceKey]string)
+
+	instanceIter := instances.ElementIterator()
+	for instanceIter.Next() {
+		k, v := instanceIter.Element()
+
+		instanceEvalCtx := evalCtx.NewChild()
+		instanceEvalCtx.Variables = map[string]cty.Value{
+			"each": cty.ObjectVal(map[string]cty.Value{
```
cam72cam, line 770: This does not make sense to me.  This only applies for for_each and not count?

internal/configs/resource.go
```diff
@@ -813,59 +929,29 @@ func (m *ProviderConfigRefMapping) decodeStaticAlias(eval *StaticEvaluator, forE
...
 	}
 
-	if !diags.HasErrors() {
-		m.Aliases = aliases
+	diags = diags.Extend(aliasDiags)
+	if aliasDiags.HasErrors() {
+		return diags
 	}
 
+	m.Aliases = aliases
```
cam72cam, line 954: We should at some point validate that these aliases are valid identifiers, either in this PR or a followup.

internal/configs/resource.go
```diff
@@ -813,59 +929,29 @@ func (m *ProviderConfigRefMapping) decodeStaticAlias(eval *StaticEvaluator, forE
 		})
 	}
 
-	// TODO/Oleksandr: Is static identifier valid if we evaluate for_each for alias?
-	instances, instancesDiags := eval.Evaluate(forEach, aliasStaticID)
+	instances, instancesDiags := eval.Evaluate(instanceExpr, aliasStaticID)
```
cam72cam, line 932: This should probably use evalchecks (or where ever it ends up landing) to properly handle decoding the each/count data.  I'd be happy with a TODO for later.

internal/configs/resource.go
ollevche: Previously, we didn't have such a validation. Also, I think we need just to check if provider exists - it will cover everything else. What do you think?

internal/configs/resource.go
ollevche: Thanks! Left a todo for later, when we update feature branch with [the Andrew's PR](https://github.com/opentofu/opentofu/pull/1911)

cam72cam (approved):

PR #1960, opened 2024-09-02
#300: fix some hacks from the first commits

ollevche:
This PR is based on https://github.com/opentofu/opentofu/pull/1959 and aimed at resolving hacks from the first commits (for example, InParentTODO calls).

Some of the fixes await the latest changes from graph execution work by @Evi1Pumpkin. So it will be done in a separate PR after the merge.

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonconfig/config.go
+++ b/internal/command/jsonconfig/config.go
@@ -279,6 +279,7 @@ func marshalProviderConfigs(
 		for _, ppc := range mc.Providers {
 			// These provider names include aliases, if set
 			moduleProviderName := ppc.InChild.String()
+			// TODO/Oleksandr: update this field to resolved provider after graph execution
 			parentProviderName := ppc.InParentTODO().String()
 
 			// Look up the provider FQN from the module context, using the non-aliased local name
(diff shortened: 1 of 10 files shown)


internal/command/jsonconfig/config.go
```diff
@@ -279,6 +279,7 @@ func marshalProviderConfigs(
 		for _, ppc := range mc.Providers {
 			// These provider names include aliases, if set
 			moduleProviderName := ppc.InChild.String()
+			// TODO/Oleksandr: update this field to resolved provider after graph execution
```
cam72cam, line 282: I think long term we need to change the jsonconfig format here.  The only major impact to this is the plan format , which is not supported between tofu versions anyways.

internal/configs/config.go
```diff
@@ -484,7 +484,14 @@ func (c *Config) addProviderRequirements(reqs getproviders.Requirements, recurse
 					})
 					continue
 				}
-				if i.ProviderConfigRef.Name != target.ProviderConfigRef.Name || i.ProviderConfigRef.Alias != target.ProviderConfigRef.GetNoKeyAlias() {
+
+				// We want to compare aliases only if there is no `each` or `count` inside a resource provider reference.
+				haveIncompatibleAliases := true
+				if !target.ProviderConfigRef.HasInstanceRefsInAlias() {
```
cam72cam, line 490: I wonder if a pattern like `alias, ok := target.ref.KnownAlias(); ok` would make more sense here?  I also stink at naming functions.

cam72cam: Given that import blocks are only allowed in the root module, we should know the exact provider alias and can treat any unknowns here as a bug?

internal/configs/resource.go
```diff
@@ -699,7 +715,7 @@ type ProviderConfigRefMapping struct {
 
 // HasAlias returns true if the provider is referenced by alias.
 func (m *ProviderConfigRefMapping) HasAlias() bool {
-	return len(m.Aliases) != 0
+	return m == nil || len(m.Aliases) != 0
```
cam72cam, line 718: Why would HasAlias return true if m == nil?

internal/configs/config.go
ollevche: We treat it as an error because the provider could be set in both resource and import blocks. If that is the case, providers (and aliases) should match. If the provider alias contains `each` / `count`, they will be different for sure so we error. Does it answer your question?

internal/configs/resource.go
ollevche: That's a typo, thanks! [302cdc0](https://github.com/opentofu/opentofu/pull/1960/commits/302cdc036887666de1113dace20e45d50f3dfc24)

cam72cam: In what scenario is m nil?

cam72cam (approved):

PR #1963, opened 2024-09-03
#300 - support in graph for each on providers in resources and modules

Evi1Pumpkin:
### Explanation about the code:
This code changes the graph part to support for_each on providers in resources / modules.

**How does it works:**

1. The [first changed transformer](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR955-R977) that runs is `ProviderConfigTransformer`. Now this transformer iterates over the `pair.InParentMapping.Aliases` (before it was a single alias) and adds the one / multiple targets based on the different aliases to the `graphNodeProxyProvider` ([here](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR977)). If we have multiple targets configured here, it means that in the configuration we have a `for_each` on the providers on the module specifically.
2. The second changed transformer is the `ProviderTransformer`. This transformer (also others) is running multiple times in different stages of the `tofu` command, and can run on abstract resources, expanded resource instances, and other types of resources. This transformer calls `ProvidedBy()` on each resource / resource instance in the configuration to get their provider configuration. Now the  `ProvidedBy()` can either return:
    1. `exactProvider` in the case where the provider was already resolved in the previous transformer iterations, or it is taken from the state. In this case we know that the provider is absolute, and we don’t need to recursively resolve it as part of this transformer. That is why we can treat it in a [specific logic](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR277-R292) that breaks early on, and sets the provider [directly on the resource](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR287), on the correct level (resource / instance). We’ll talk a little more about the provider level in regarding to the changes on `node_resource_abstract.go` and their importance.
    2. `map[addrs.InstanceKey]addrs.ProviderConfig` that was inferred from the config.
    It can represent a single or many providers that each can be a proxy provider (that will be [recursively resolved to an absolute provider as part of this transformer](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR297) or an absolute provider address. If we have multiple providers returned by `ProvidedBy()` in the map, it means that in the configuration we have a `for_each` on the providers on the resource specifically.
    Another crucial difference to understand is that the `map[addrs.InstanceKey]addrs.ProviderConfig` can be either resolved in this transformer to:
        1. A single provider that we’ll [set straight on the resource](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR322) as a `ResourceProvider` (this is how the code used to work until this feature).
        2. Multiple providers, all can be potential providers to expanded resource instances. But we need to get into the expansion of instances (while walking the graph) to actually match the correct provider to the specific instance. We are setting these potential providers on the resource [here](https://github.com/opentofu/opentofu/pull/1963/files#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4dR324).
3. `NodeAbstractResource` also changed:
    1. It can now either [contain](https://github.com/opentofu/opentofu/pull/1963/files#diff-4697def02e73712956845a2a67e54292c5a6b41e4b3f57ffe6e75a3da55326d4R76-R86):
        1. A `ResolvedResourceProvider`, if not null it means the providers is set on the whole resource like how it used to (no `for_each` on providers in the resource and/or modules containing this resource).
        2. `potentialProviders` that will later be resolved in the `resolveInstanceProvider()` function based on a specific `addrs.AbsResourceInstance` param ([here](https://github.com/opentofu/opentofu/pull/1963/files#diff-4697def02e73712956845a2a67e54292c5a6b41e4b3f57ffe6e75a3da55326d4R311)). The result of this function will eventually be part of the `NodeAbstractResourceInstance` as the `ResolvedInstanceProvider`. The  `resolveInstanceProvider()` function is used in the `nodeExpandPlannableResource` as part of the expansion of resources to resource instances ([example](https://github.com/opentofu/opentofu/pull/1963/files#diff-762df315496c62c8a4a5d661f20193aa14f055fb3b5af179b2c39f18cec246b4R131)).
    2. Until now, the provider was always set on the resource level. Now because it might exist on the instance level, functions that used to access the provider directly, need to receive the `resolvedProvider`  as a parameter. We are talking about the `ProvidedByImpl` ([here](https://github.com/opentofu/opentofu/pull/1963/files#diff-4697def02e73712956845a2a67e54292c5a6b41e4b3f57ffe6e75a3da55326d4R336)) and `readResourceInstanceState` ([here](https://github.com/opentofu/opentofu/pull/1963/files#diff-4697def02e73712956845a2a67e54292c5a6b41e4b3f57ffe6e75a3da55326d4R541)) and `readResourceInstanceStateDeposed` ([here](https://github.com/opentofu/opentofu/pull/1963/files#diff-4697def02e73712956845a2a67e54292c5a6b41e4b3f57ffe6e75a3da55326d4R582)).
4. `NodeAbstractResourceInstance`:
    1. [Contains](https://github.com/opentofu/opentofu/pull/1963/files#diff-d7e3ce5e9bea4873c53893bb9e469e7d3084f81f6cab75ba05698e3c3390c7e7R53) a `ResolvedInstanceProvider` , if not null it means the providers is set on the specific instance (in the configuration we have a `for_each` on providers in the resource and/or modules containing this resource).
    2. `ResolvedProvider()` [function](https://github.com/opentofu/opentofu/pull/1963/files#diff-d7e3ce5e9bea4873c53893bb9e469e7d3084f81f6cab75ba05698e3c3390c7e7R83) that returns the correct provider of the resource, whether it is a  `ResolvedResourceProvider` or a `ResolvedInstanceProvider` . This code also validates if a provider exist for the instance or if both the `ResolvedResourceProvider` and `ResolvedInstanceProvider` are somehow set (only one can be set).
    3. Many alterations to different methods to use the new `ResolvedProvider()` function to get the correct provider ( `ResolvedResourceProvider` / `ResolvedInstanceProvider`).
5. Changes to all the other files starting with `node_resource_*.go`:
    1. Many alterations to different methods to use the new `ResolvedProvider()` function to get the correct provider ( `ResolvedResourceProvider` / `ResolvedInstanceProvider` ).
6. Changes affecting reading / writing from the state file:
    1. The resource in the state can either contain:
        1. A `ProviderConfig`([link](https://github.com/opentofu/opentofu/pull/1963/files#diff-ec2a72741b3f1384c6cdafe17112d7e7ebe54aac03612b915376d00b305db979R32), the old provider configuration that is set on the whole resource level. If this prop doesn’t exist, it means that the providers are set in each instance object.
        2. An `InstanceProvider` ([link](https://github.com/opentofu/opentofu/pull/1963/files#diff-fd4be33b7dd624708618d5e9c021a44493369b44d0202904883d435e4089df55R71)), the new provider configuration that is set on the specific instance level. If this prop doesn’t exist, the provider is set on the resource level.
    2. The main consideration I had was to keep the state as compatible as possible to previous versions, before this feature. So if we don’t use a  a `for_each` on providers in the resource and/or modules containing this resource, then the state should be identical to how it was before of this feature.
    3. The approach I took is to always pass both `ResourceProvider` and `InstanceProvider`, during the writing / reading process, assuming only one of them will be set and the other will be empty. If one is empty, the correlating prop will be omitted from state, upon write. And be read as empty, upon read.
    I added validations to [read](https://github.com/opentofu/opentofu/pull/1963/files#diff-ec2a72741b3f1384c6cdafe17112d7e7ebe54aac03612b915376d00b305db979R62-R68) and [write](https://github.com/opentofu/opentofu/pull/1963/files#diff-7d1ad7530ce0b73d0e3cfd7b991689fc72d594d485c4c15d761734be82a3a3d8R156-R161) processes to ensure that we don’t have a resource without a provider, or to ensure we don’t have both `ResourceProvider` and `InstanceProvider` set for a resource.
    4. We have a little bug with this approach that if an existing resource in the state had a `ResourceProvider` and its configuration changed and now it uses `InstanceProvider` (or vice versa), we need to explicitly delete the old provider configuration. This is a known bug that I still need to handle and is quite easy to fix.

### Flows that were altered and need some love in testing:
1. deposed flow
2. import flow - especially the import in the `ProvidedBy()` [here](https://github.com/opentofu/opentofu/blob/ead243e2296fb0fa268659aa51ea6fb20d37ee78/internal/tofu/node_resource_abstract.go#L378-L394)
3. state show command (resource / resource index)
4. taint command
5. untaint command
6. provider functions

### Last Todos (fixes be included in future PRs, as it'll be easier to reach these code paths in tests):
1. https://github.com/opentofu/opentofu/pull/1963/files#r1768245536
2. https://github.com/opentofu/opentofu/pull/1963/files#r1773592070
3. The bug I mentioned in the last clause of my code description ⬆️  (6 iv)

## Target Release

1.9.0

## Checklist

### Go checklist

 


internal/command/state_show.go
```diff
@@ -162,16 +162,24 @@ func (c *StateShowCommand) Run(args []string) int {
...
 	// check if the resource has a configured provider, otherwise this will use the default provider
 	rs := state.Resource(addr.ContainingResource())
-	absPc := addrs.AbsProviderConfig{
+	absResourceProvider := addrs.AbsProviderConfig{
 		Provider: rs.ProviderConfig.Provider,
 		Alias:    rs.ProviderConfig.Alias,
 		Module:   addrs.RootModule,
 	}
+
+	absInstanceProvider := addrs.AbsProviderConfig{
```
Evi1Pumpkin, line 171: We need to test this change thoroughly. 
I didn't understand why the `absResourceProvider` has its Module set to `addrs.RootModule`. I took the same approach with `absInstanceProvider`, but we should validate this is correct.

internal/states/resource.go
```diff
@@ -29,7 +29,7 @@ type Resource struct {
 	// with a provider configuration when the resource configuration block is
 	// not available, such as if it has been removed from configuration
 	// altogether.
-	ProviderConfig addrs.AbsProviderConfig
+	ProviderConfig addrs.AbsProviderConfig // Ronny TODO: should we rename to ProviderResourceConfig? (this doesn't impact the state file)
```
Evi1Pumpkin, line 32: We can rename it to `ProviderResourceConfig` just to make sure we won't be confused with `ProviderInstanceConfig`

internal/states/statefile/version4.go
```diff
@@ -710,7 +740,7 @@ type resourceStateV4 struct {
 	Type           string                  `json:"type"`
 	Name           string                  `json:"name"`
 	EachMode       string                  `json:"each,omitempty"`
-	ProviderConfig string                  `json:"provider"`
+	ProviderConfig string                  `json:"provider,omitempty"` // Ronny TODO: Do we want to omit? Or save an empty string / nil?
```
Evi1Pumpkin, line 743: This will affect whether we see the `provider` attribute on the resource in the state file

internal/tofu/node_resource_validate.go
```diff
@@ -281,8 +281,13 @@ var connectionBlockSupersetSchema = &configschema.Block{
 
 func (n *NodeValidatableResource) validateResource(ctx EvalContext) tfdiags.Diagnostics {
 	var diags tfdiags.Diagnostics
+	// TODO Ronny fix
```
Evi1Pumpkin, line 284: I still didn't fix this area. I believe I'll have to loop over all the potential providers and ensure I validate each one separately.

Evi1Pumpkin: NVM it is fixed

internal/states/resource.go
```diff
@@ -38,6 +38,42 @@ func (rs *Resource) Instance(key addrs.InstanceKey) *ResourceInstance {
...
+
+	resourceProvider = rs.ProviderConfig
+
+	// If the provider is set on the instance level, we can't get it from the resource
+	instance := rs.Instances[key]
+
+	if instance.Current.InstanceProvider.Provider.Type != "" {
+		instanceProvider = instance.Current.InstanceProvider
+	} else {
+		// Take the provider from an arbitrary item from the deposed map TODO Ronny: validate with deposed scenario
```
Evi1Pumpkin, line 55: Just a reminder to make sure we test the deposed acenario

internal/states/state.go
```diff
@@ -330,7 +330,12 @@ func (s *State) ProviderAddrs() []addrs.AbsProviderConfig {
 	m := map[string]addrs.AbsProviderConfig{}
 	for _, ms := range s.Modules {
 		for _, rc := range ms.Resources {
-			m[rc.ProviderConfig.String()] = rc.ProviderConfig
+			//m[rc.ProviderConfig.String()] = rc.ProviderConfig // Ronny TODO: I assume we always have an instance inside the resource. Is it a valid assumption?
```
Evi1Pumpkin, line 333: For now, I've always seen we have the instances set in this part, but I'm keeping this comment here as a reminder to make sure this is correct in all scenarios (or if you guys can think of a scenario when this is not true).

internal/states/statefile/version4.go
```diff
@@ -389,12 +408,17 @@ func writeStateV4(file *File, w io.Writer, enc encryption.StateEncryption) tfdia
 				continue
 			}
 
+			providerConfigValue := ""
+			if rs.ProviderConfig.Provider.Type != "" {
```
cam72cam, line 412: Maybe add a helper function IsValid() or IsSet() to the provider addr struct?

internal/states/resource.go
```diff
@@ -38,6 +38,42 @@ func (rs *Resource) Instance(key addrs.InstanceKey) *ResourceInstance {
...
+
+	// If the provider is set on the instance level, we can't get it from the resource
+	instance := rs.Instances[key]
+
+	if instance.Current != nil && instance.Current.InstanceProvider.Provider.Type != "" {
+		instanceProvider = instance.Current.InstanceProvider
+	} else {
+		// Take the provider from an arbitrary item from the deposed map
+		for _, deposedInstance := range instance.Deposed {
+			instanceProvider = deposedInstance.InstanceProvider
```
cam72cam, line 57: What's going in here?

internal/tofu/transform_provider.go
```diff
@@ -387,7 +475,7 @@ func (t *CloseProviderTransformer) Transform(g *Graph) error {
 	for _, p := range pm {
 		key := p.ProviderAddr().String()
 
-		// get the close provider of this type if we already created it
+		// get the close provider of this type if we alread created it
```
cam72cam, line 478: typo

internal/tofu/transform_provider.go
```diff
@@ -429,7 +517,7 @@ func (t *CloseProviderTransformer) Transform(g *Graph) error {
 // This transformer may create extra nodes that are not needed in practice,
 // due to overriding provider configurations in child modules.
 // PruneProviderTransformer can then remove these once ProviderTransformer
-// has resolved all of the inheritance, etc.
+// has resolved all of the inheritence, etc.
```
cam72cam, line 520: This was not a typo

internal/states/resource.go
Evi1Pumpkin: I clarified this part. Let me know if [this](https://github.com/opentofu/opentofu/pull/1963/commits/3a615ba0091d655498db9ac0907a2cb4e9cb7feb) is better.
But overall, this is an assumption based on how the code behaved before this change. The provider was set on the resource level, so all the deposed objects had the same provider.

internal/states/statefile/version4.go
Evi1Pumpkin: I was too lazy to do it, so I think this is about time :) https://github.com/opentofu/opentofu/pull/1963/commits#:~:text=Use%20new%20IsSet%20function%20to%20check%20if%20a%20provider%20is%20not%20set

internal/tofu/node_resource_import.go
```diff
@@ -30,6 +31,13 @@ type graphNodeImportState struct {
 	states []providers.ImportedResource
 }
 
+func (n *graphNodeImportState) SetPotentialProviders(potentialProviders []distinguishableProvider) {
+	// Technically we shouldn't get here as we don't support import with for_each on providers in resources / modules
+	// So we should panic here (ensure we have a code that validates it somewhere else) / implement a proper user error.
+	//TODO Ronny - don't forget to handle this scenario ^
+	panic("implement me")
```
Evi1Pumpkin, line 38: We need to ensure the user is notified with a proper error that this code path is currently not possible. In stage one, we won't support using the `import` block to import into resources with a `for_each` on providers in the resource and/or modules containing this resource.

internal/states/resource.go
ollevche: If I understand correctly, instance pending for destruction could still reference different providers, isn't it?

The comment of the `Deposed` field:

> // Deposed, if len > 0, contains any remote objects that were previously
> 	// represented by the corresponding resource instance but have been
> 	// replaced and are pending destruction due to the create_before_destroy
> 	// lifecycle mode.

internal/states/statefile/version4.go
```diff
@@ -139,9 +145,22 @@ func prepareStateV4(sV4 *stateV4) (*File, tfdiags.Diagnostics) {
 
 			instAddr := rAddr.Instance(key)
 
+			// The provider can either be set on the resource level or on the instance level
+			var instanceProviderAddr addrs.AbsProviderConfig
+			if isV4.InstanceProvider != "" {
```
ollevche, line 150: Let's also handle somehow (panic?) the case when provider is not set anywhere. This should never happen, but still better to handle it as early as possible.

internal/states/statefile/version4.go
ollevche: IMO we have to ensure other tools **fail** correctly if this field is something they rely on. I think it should be omit completely.

internal/tofu/node_resource_abstract.go
```diff
@@ -73,12 +73,23 @@ type NodeAbstractResource struct {
...
-	// The address of the provider this resource will use
-	ResolvedProvider addrs.AbsProviderConfig
-	// storedProviderConfig is the provider address retrieved from the
-	// state. This is defined here for access within the ProvidedBy method, but
-	// will be set from the embedding instance type when the state is attached.
-	storedProviderConfig addrs.AbsProviderConfig
+	// The address of the provider this resource will use.
+	// Can be empty if the provider is not set on the resource level but in the resource instance instead. In this case,
+	// the potentialProviders property will be populated.
+	// It can happen when we have a for_each on the providers in the resource or containing module.
```
ollevche, line 79:
```suggestion
	// It can happen when we have a for_each or count on the providers in the resource or containing module.
```

internal/tofu/node_resource_abstract.go
```diff
@@ -73,12 +73,23 @@ type NodeAbstractResource struct {
...
+	ResolvedResourceProvider addrs.AbsProviderConfig
+
+	// potentialProviders is a list of provider and their identifiers, meant to be resolved per each instance to
+	// calculate its provider.
+	// If the potentialProviders is populated, it means the ResolvedResourceProvider is empty and the provider will be
+	//defined per resource instance and not on the whole resource.
```
ollevche, line 85:
```suggestion
	// defined per resource instance and not on the whole resource.
```

internal/tofu/node_resource_abstract.go
```diff
@@ -290,33 +301,88 @@ func (n *NodeAbstractResource) DependsOn() []*addrs.Reference {
 	return result
 }
 
-func (n *NodeAbstractResource) SetProvider(p addrs.AbsProviderConfig) {
-	n.ResolvedProvider = p
+func (n *NodeAbstractResource) SetPotentialProviders(potentialProviders []distinguishableProvider) {
+	n.potentialProviders = potentialProviders
```
ollevche, line 305: do we also want to check if ResolvedResourceProvider is nil? or maybe set it to nil?

internal/tofu/node_resource_abstract.go
```diff
@@ -290,33 +301,88 @@ func (n *NodeAbstractResource) DependsOn() []*addrs.Reference {
...
+}
+
+// resolveInstanceProvider gets the specific expanded instance of the resource, and tries to calculate the resolved
+// InstanceProvider if a few potential providers exist. A few potential providers might exist if the resource
+// configuration, or one of the containing module's configuration, contains a for_each on the providers.
```
ollevche, line 310:
```suggestion
// configuration, or one of the containing module's configuration, contains a for_each or count in the provider reference.
```

internal/tofu/node_resource_abstract.go
```diff
@@ -290,33 +301,88 @@ func (n *NodeAbstractResource) DependsOn() []*addrs.Reference {
...
+func (n *NodeAbstractResource) resolveInstanceProvider(instance addrs.AbsResourceInstance) addrs.AbsProviderConfig {
+	for _, potentialProvider := range n.potentialProviders {
+		if potentialProvider.IsResourceInstanceMatching(instance) {
+			return potentialProvider.concreteProvider.ProviderAddr()
+		}
+	}
+
+	// Can happen if ResolvedResourceProvider is already set for the whole resource, and we don't have a specific
+	// provider set for the instances
+	return addrs.AbsProviderConfig{}
```
ollevche, line 320: should we use a boolean as a second return arg here to better communicate "failure"?

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -46,6 +46,11 @@ type NodeAbstractResourceInstance struct {
 	// During import we may generate configuration for a resource, which needs
 	// to be stored in the final change.
 	generatedConfigHCL string
+
+	// In the case where we have for_each on the provider on the containing
```
ollevche, line 50:
```suggestion
	// In the case where we have for_each or count on the provider on the containing
```

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -63,6 +68,44 @@ func NewNodeAbstractResourceInstance(addr addrs.AbsResourceInstance) *NodeAbstra
...
+	}
+
+	if !result.IsSet() {
+		panic(fmt.Sprintf("ResolvedProvider for %s cannot get a provider", n.Addr))
+	}
+
+	// Throw an error if ResolvedResourceProvider and ResolvedInstanceProvider exists at the same time.
+	if n.ResolvedInstanceProvider.IsSet() && n.ResolvedResourceProvider.IsSet() {
+		panic(fmt.Sprintf("ResolvedProvider for %s has a provider set for the resource and the resource's instance", n.Addr))
+	}
```
ollevche, line 104: I am wondering if we can restructure code in such a way so those checks are compile-time guaranteed.

internal/tofu/node_resource_import.go
ollevche: Just a note: we have a validation for this case as a part of configuration parsing.

internal/tofu/transform_provider.go
```diff
@@ -108,61 +118,78 @@ func (t *ProviderTransformer) Transform(g *Graph) error {
...
 	type ProviderRequest struct {
-		Addr  addrs.AbsProviderConfig
-		Exact bool // If true, inheritance from parent modules is not attempted
+		Addr        addrs.AbsProviderConfig
+		Exact       ExactProvider // If populated, inheritance from parent modules is not attempted
+		instanceKey addrs.InstanceKey
 	}
 	requested := map[dag.Vertex]map[string]ProviderRequest{}
 	needConfigured := map[string]addrs.AbsProviderConfig{}
 	for _, v := range g.Vertices() {
```
ollevche, line 127: can we reduce nesting inside of this loop?

internal/tofu/transform_provider.go
```diff
@@ -108,61 +118,78 @@ func (t *ProviderTransformer) Transform(g *Graph) error {
...
+
+					case addrs.LocalProviderConfig:
+						// ProvidedBy() return a LocalProviderConfig when the resource
+						// contains a `provider` attribute
+						absPc.Provider = pv.Provider()
+						modPath := pv.ModulePath()
+						if t.Config == nil {
+							absPc.Module = modPath
+							absPc.Alias = p.Alias
+							break
```
ollevche, line 172: just to clarify - this break is not supposed to end the loop, right?

i suggest to put an else statement here to make it more obvious to a reader

internal/tofu/transform_provider.go
```diff
@@ -568,14 +656,53 @@ func (n *graphNodeCloseProvider) DotNode(name string, opts *dag.DotOpts) *dag.Do
 	}
 }
 
+// distinguishableProvider is a representation of a concrete provider and identifiers that match the provider to a
+// specific resource instance. Because we might have for_each on providers in the resource/module, we cannot calculate
```
ollevche, line 660:
```suggestion
// specific resource instance. Because we might have for_each or count on providers in the resource/module, we cannot calculate
```

internal/tofu/transform_provider.go
```diff
@@ -613,7 +771,7 @@ type ProviderConfigTransformer struct {
 	// each provider node is stored here so that the proxy nodes can look up
 	// their targets by name.
 	providers map[string]GraphNodeProvider
-	// record providers that can be overridden with a proxy
+	// record providers that can be overriden with a proxy
```
ollevche, line 774: I don't think it's a typo
```suggestion
	// record providers that can be overridden with a proxy
```

ollevche (commented):
Great work, thank you! Left several questions and nit comments. Happy to re-review later and discuss open questions.

internal/states/resource.go
Evi1Pumpkin: I liked your idea https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/addrs/provider_config.go
```diff
@@ -343,6 +343,11 @@ func (m ModuleInstance) ProviderConfigAliased(provider Provider, alias string) A
 	}
 }
 
+// IsSet returns true if the Provider's Type is empty.
```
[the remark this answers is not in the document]
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/states/instance_object.go
```diff
@@ -49,6 +49,11 @@ type ResourceInstanceObject struct {
 	// destroy operations, we need to record the status to ensure a resource
 	// removed from the config will still be destroyed in the same manner.
 	CreateBeforeDestroy bool
+
+	// InstanceProvider, if non-nil, this instance has a specific provider different
+	// from the whole resource's provider. Happens in the case we have a
+	// for_each on providers in the resource / module
```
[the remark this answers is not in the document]
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/states/module.go
```diff
@@ -87,9 +90,9 @@ func (ms *Module) RemoveResource(addr addrs.Resource) {
 // Set obj to nil to remove the primary generation object altogether. If there
 // are no deposed objects then the instance will be removed altogether.
 //
-// The provider address is a resource-wide setting and is updated for all other
-// instances of the same resource as a side-effect of this call.
-func (ms *Module) SetResourceInstanceCurrent(addr addrs.ResourceInstance, obj *ResourceInstanceObjectSrc, provider addrs.AbsProviderConfig) {
+// The provider address can be either a resource-wide settings or set for the
+// specific instance, and it depends on the given resourceProvider / instanceProvider
+func (ms *Module) SetResourceInstanceCurrent(addr addrs.ResourceInstance, obj *ResourceInstanceObjectSrc, resourceProvider addrs.AbsProviderConfig, instanceProvider addrs.AbsProviderConfig) {
```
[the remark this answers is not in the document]
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/tofu/node_resource_abstract.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/tofu/node_resource_abstract.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/tofu/node_resource_abstract_instance.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/tofu/transform_provider.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/tofu/transform_provider.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/states/instance_object_src.go
```diff
@@ -64,6 +64,11 @@ type ResourceInstanceObjectSrc struct {
 	Status              ObjectStatus
 	Dependencies        []addrs.ConfigResource
 	CreateBeforeDestroy bool
+
+	// InstanceProvider, if non-nil, this instance has a specific provider different
```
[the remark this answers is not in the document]
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/93aaf45228ccf0e9341cf03727b505249eeb2f1c

internal/states/module.go
```diff
@@ -158,11 +180,17 @@ func (ms *Module) SetResourceInstanceCurrent(addr addrs.ResourceInstance, obj *R
...
+func (ms *Module) SetResourceInstanceDeposed(addr addrs.ResourceInstance, key DeposedKey, obj *ResourceInstanceObjectSrc, resourceProvider addrs.AbsProviderConfig, instanceProvider addrs.AbsProviderConfig) {
+	ms.SetResourceProvider(addr.Resource, resourceProvider)
 
 	rs := ms.Resource(addr.Resource)
 	is := rs.EnsureInstance(addr.Key)
+
+	// If the instanceProvider is not empty, populate the obj with its value
+	if instanceProvider.IsSet() {
+		obj.InstanceProvider = instanceProvider
+	}
```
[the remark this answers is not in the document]
Evi1Pumpkin: Good catch! https://github.com/opentofu/opentofu/pull/1963/commits/95c541cc3de61e60cc88bc570bd114c09d146948

internal/states/statefile/version4.go
Evi1Pumpkin: https://github.com/opentofu/opentofu/pull/1963/commits/1182ee49506a9cd491ca9f276b2ca784293a19b9

internal/tofu/node_resource_abstract.go
Evi1Pumpkin: We won't use this boolean in the calling functions, so I don't think we need it

internal/tofu/node_resource_abstract_instance.go
Evi1Pumpkin: Hmm, I can't think of a way to do it now, but there are parts I'm convinced we can refactor and want to do later after we are in alpha (like the provider transformer code, for example).

internal/tofu/transform_provider.go
Evi1Pumpkin: Probably. I think we should refactor this code eventually. But because it is a little delicate and quite critical, I believe we should first release it to alpha and then refactor.

internal/tofu/transform_provider.go
Evi1Pumpkin: This is indeed confusing, but the break is on the execution of the specifc switch case

internal/tofu/transform_destroy_edge.go
```diff
@@ -105,13 +105,13 @@ func (t *DestroyEdgeTransformer) tryInterProviderDestroyEdge(g *Graph, from, to
 
 		// we don't care about `exact` here, since we're only looking for any
 		// clue that the providers may differ.
-		p, _ := pc.ProvidedBy()
+		/*  p, _ := pc.ProvidedBy() // TODO Ronny fix
```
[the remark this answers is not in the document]
ollevche: Fixed by @cam72cam in https://github.com/opentofu/opentofu/pull/2031/commits/f64a640d192a612d1641e8bc3b2096023c116f3b

PR #1977, opened 2024-09-10
bug: add error handling for missing state lock config in S3, HTTP and OSS backends

### LINKED ISSUE #1852: force-unlock unlocks state even with no locking configured (S3 backend)
### OpenTofu Version

```shell
OpenTofu v1.9.0-dev
on darwin_arm64
+ provider registry.opentofu.org/hashicorp/aws v5.59.0
```

```hcl
terraform {
  backend "s3" {
    region = "<REGION>"
    bucket = "<BUCKET_NAME>"
    key = "state"
  }
}

provider "aws" {
  region = "<REGION>"
}

resource "null" "a" {}
```

### Debug Output

Normal output:

```
Do you really want to force-unlock?
  OpenTofu will remove the lock on the remote state.
  This will allow local OpenTofu commands to modify this state, even though it
  may still be in use. Only 'yes' will be accepted to confirm.

  Enter a value: yes

OpenTofu state has been successfully unlocked!

The state has been unlocked, and OpenTofu commands should now be able to
obtain a new lock on the remote state.
```

### Expected Behavior

I expect `tofu force-unlock anyvalue` to fail due to 1) disabled locking and 2) dummy lockID. Local backend fails with "state is not locked" error in this scenario.

### Actual Behavior

`tofu force-unlock anyvalue` results in a "successfully unlocked" output, which is misleading.

### Steps to Reproduce

You would need an S3 bucket configured to be used as a backend, then:
1. `tofu init`
2. `tofu force-unlock anyvalue`

### Additional Context

_No response_

### References

_No response_

g0dfl3sh:
Resolves #1852

Resolves # 

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/http/client.go
+++ b/internal/backend/remote-state/http/client.go
@@ -131,7 +131,7 @@ func (c *httpClient) Lock(info *statemgr.LockInfo) (string, error) {
 
 func (c *httpClient) Unlock(id string) error {
 	if c.UnlockURL == nil {
-		return nil
+		return fmt.Errorf("Unlock URL not specified in the backend configuration. State is not locked.")
 	}
 
 	resp, err := c.httpRequest(c.UnlockMethod, c.UnlockURL, c.jsonLockInfo, "unlock")
--- a/internal/backend/remote-state/oss/client.go
+++ b/internal/backend/remote-state/oss/client.go
@@ -362,7 +362,7 @@ func (c *RemoteClient) getLockInfo() (*statemgr.LockInfo, error) {
 }
 func (c *RemoteClient) Unlock(id string) error {
 	if c.otsTable == "" {
-		return nil
+		return fmt.Errorf("OTS Tablestore not specified in the backend configuration. State is not locked.")
 	}
 
 	lockErr := &statemgr.LockError{}
--- a/internal/backend/remote-state/s3/client.go
+++ b/internal/backend/remote-state/s3/client.go
@@ -424,7 +424,7 @@ func (c *RemoteClient) getLockInfo(ctx context.Context) (*statemgr.LockInfo, err
 
 func (c *RemoteClient) Unlock(id string) error {
 	if c.ddbTable == "" {
-		return nil
+		return fmt.Errorf("DynamoDB table not specified in the backend configuration. State is not locked.")
 	}
 
 	lockErr := &statemgr.LockError{}


LucasBadico (commented):
Good man, great work! I won't approve because I don't have all the context in mind. 

Look like you nailed it!

g0dfl3sh:
Thanks for pointing it out @abstractionfactory! I had the wrong linter selected in my VSCode extension. Should be good to go now.

g0dfl3sh:
@abstractionfactory Not sure why the unit tests for darwin fail now, but it seems unrelated to my changes. Would you be able to re-trigger?

g0dfl3sh:
Thanks! It worked this time; probably just a flaky test.

ollevche (changes requested):
Hello and thank you for the contribution!

The fix will help when force-unlocking the state, but it will break the configuration for existing users on `tofu apply`, when there is no locking enabled.

HTTP, OSS and S3 remote state backends support optional locking (based on configuration). When the locking is disabled, `Lock` and `Unlock` methods are going to be called anyway so it should not error in such cases.

I suggest looking into the force-unlock command implementation to provide a more isolated fix to not break anything outside the scope of the issue. We can check if the locking is disabled before trying to force-unlock and then provide a meaningful error.

I am happy to help if you have any further questions!

g0dfl3sh:
Thanks for the hints @ollevche! Going to start looking into that.

g0dfl3sh:
Hey @ollevche! Just pushed a new commit attempting to do what you asked for the `S3` backend. Please let me know if i am going the right path (if so, I will push the changes for `OSS` and `HTTP` as well).
I feel a little uneasy about declaring all those (what should be in theory) no-ops, but that was the only way I could get rid of the type errors I was getting.

ollevche (commented):
Thank you for the update! You are certainly going in the right direction.

I would introduce a separate interface similar to the following:

```go
type OptionalLocker interface {
    Locker
    IsLockingEnabled() bool
}
```

Then we could extend only OSS, HTTP and S3 backends to implement `IsLockingEnabled` function.

In the force unlock command we can assert if the implementation conforms `OptionalLocker` interface and if so check the output of `IsLockingEnabled`.

This is not an ideal solution, but that way we follow general code style and extend only what is needed to extend.

What do you think?

g0dfl3sh:
Thank you for the suggestion! That makes a whole lot more sense and will probably be cleaner. Will test this tomorrow and let you know.

g0dfl3sh:
Sorry for the delay; just pushed my changes. 
I had to add the `if backendConfig.Type == "s3" || backendConfig.Type == "oss" || backendConfig.Type == "http"` statement because otherwise it would break state locking for other backends such as GCS.
Let me know what you think when you have time!

g0dfl3sh:
@ollevche The unit tests errored out with `null pointer deference`. I have added an additional check for `backendConfig` to not be nil. Should hopefully pass now.

internal/command/unlock.go
```diff
@@ -24,6 +24,11 @@ type UnlockCommand struct {
 	Meta
 }
 
+type OptionalLocker interface {
+	statemgr.Locker
+	IsLockingEnabled() bool
+}
+
```
ollevche, line 31: Can we use `OptionalLocker` declared in the `statemgr` package?

internal/command/unlock.go
```diff
@@ -95,6 +100,17 @@ func (c *UnlockCommand) Run(args []string) int {
 
 	_, isLocal := stateMgr.(*statemgr.Filesystem)
 
+	if backendConfig != nil && (backendConfig.Type == "s3" || backendConfig.Type == "oss" || backendConfig.Type == "http") {
```
ollevche, line 103: I think we don't need to check configuration values here. Let's rely on the interface assertion check. Interface assertion + `IsLockingEnabled` call will cover it.

internal/command/unlock.go
```diff
@@ -95,6 +100,17 @@ func (c *UnlockCommand) Run(args []string) int {
...
+		if optionalLocker, ok := stateMgr.(OptionalLocker); ok {
+			// Now we can safely call IsLockingEnabled() on optionalLocker
+			if !optionalLocker.IsLockingEnabled() {
+				c.Ui.Error("Locking is not configured for this backend")
```
ollevche, line 107: It could be configured, but still disabled, so I suggest slightly rewording the message:

```suggestion
				c.Ui.Error("Locking is disabled for this backend")
```

internal/states/remote/state.go
```diff
@@ -280,6 +280,18 @@ func (s *State) Unlock(id string) error {
...
+	if s.disableLocks {
+		return false
+	}
+
+	if c, ok := s.Client.(OptionalClientLocker); ok {
+		lockInfo := c.IsLockingEnabled()
+		return lockInfo
+	}
+	return false
+}
```
ollevche, line 293: If `Client` implements `ClientLocker`, locking should be enabled:

```suggestion
func (s *State) IsLockingEnabled() bool {
	if s.disableLocks {
		return false
	}

        switch c := s.Client.(type) {
        // Client supports optional locking.
        case OptionalClientLocker:
                return c.IsLockingEnabled()
        // Client supports locking by default.
        case ClientLocker:
                return true
        // Client doesn't support any locking.
        default:
                return false
        }
}
```

ollevche (commented):
Thank you! I left a few small comments, but overall the PR looks good 👍

g0dfl3sh:
Thanks for the help! Should be good to go now.

g0dfl3sh:
The Compare check failed, but given the errors it seems unrelated to changes from this PR. Should we try a re-run?

internal/states/statemgr/locker.go
```diff
@@ -66,6 +66,11 @@ type Locker interface {
 	Unlock(id string) error
 }
 
+type OptionalLocker interface {
```
ollevche, line 69:
```suggestion
// OptionalLocker extends Locker interface to allow callers
// to know whether or not locking is actually enabled.
// This is useful for some of the backends, which support
// optional locking based on the configuration (such as S3,
// OSS and HTTP backends).
type OptionalLocker interface {
```

internal/states/remote/remote.go
```diff
@@ -33,6 +33,11 @@ type ClientLocker interface {
 	statemgr.Locker
 }
 
+type OptionalClientLocker interface {
```
ollevche, line 36:
```suggestion
// OptionalClientLocker is an optional interface that allows callers to
// to determine whether or not locking is actually enabled.
// See OptionalLocker for more details.
type OptionalClientLocker interface {
```

ollevche (commented):
Looks good, left a couple of suggestions to describe public interfaces, other than that - we are good to go!

g0dfl3sh:
Thanks for adding those! I need to make a habit out of adding comments early on as I always end up forgetting.

internal/states/remote/state.go
```diff
@@ -280,6 +280,24 @@ func (s *State) Unlock(id string) error {
 	return nil
 }
 
+func (s *State) IsLockingEnabled() bool {
+	if s.disableLocks {
+		return false
+	}
+
+	switch c := s.Client.(type) {
```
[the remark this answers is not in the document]
ollevche: Yep, it works that way, here is an example: https://go.dev/play/p/k2uRxaouup_G

ollevche:
> 2. Downcasting is usually code smell, I wonder if it would be possible to achieve this without doing it? (Liskov Substitution Principle and all.) I'm not super worried if it goes into the codebase like this, but it definitely makes the code harder to read.

This is not ideal and we should consider a larger refactoring of state manager interfaces, because currently, it makes little sense. However, I understand historical reasons behind this approach.

IMO it should not be part of this PR / bug fix.

g0dfl3sh:
Thanks for the feedback folks; I've learned a lot! Going to start working on adding tests.

g0dfl3sh:
I managed to write tests for most of the the use cases of the added code, but I wasn't able to write any meaningful tests for the actual behavior of the `force-unlock` command. It seems like whatever I tried it would default to using the local backend and bypassing the whole logic, which is the current behavior of`TestUnlock`.
Maybe that could be better integrated as acceptance tests? I'm really not sure, any pointer would be much appreciated!

g0dfl3sh:
Thanks a lot for the review @abstractionfactory! And if a decision is made that we need e2e tests for this I can definitely try to look into that as well.

g0dfl3sh:
Thanks for running the checks @ollevche! Just pushed a fix for the linting errors.

ollevche (approved):
@g0dfl3sh thanks for the contribution!

I tested changes manually with an S3 backend and it worked just fine.

I agree we would want to put an e2e test here as well, but that will be hard to achieve due to backend dependencies.

@abstractionfactory your S3 mocking test tools will be super useful here once they are merged!

g0dfl3sh:
Thanks for all the support folks! Do I need to backport this to any specific branch? I was just reading the `Backporting` section in `CONTRIBUTING.md` and I'm not sure if it applies here.

PR #1993, opened 2024-09-17
Fixes #1931: Dockerfile update / base image deprecation

### LINKED ISSUE #1931: Dockerimage update base image
### OpenTofu Version

```shell
irrelevant
```

### Use Cases

We use the docker image in our CI/CD environment and need to install additional dependencies. The dependencies are available in later versions of alpine but not in 3.18, thus we need to add the repositories of a later version which could lead to problems.

### Attempted Solutions

Use repositories from later versions.

### Proposal

If there are no reasons to stick to a specific alpine version, it would be nice if either `latest` (points to the latest stable version, not to edge) or `3` would be used.

### References

_No response_

ghost:
This PR implements #1931 and deprecates using the `ghcr.io/opentofu/opentofu` image as a base image.

The reason behind this change, as discussed in #1931, is security. For a secure base image, we would need to re-release this image at least every week, applying package upgrades, which we currently don't have the capacity to do. We are providing instructions in the OpenTofu documentation how to build an OpenTofu base image.

> [!NOTE]
> The OpenTofu image is **safe to use** if you are using it as a CLI tool. We are releasing a new image whenever there is a security update to OpenTofu itself, but we cannot guarantee that all packages (not used by OpenTofu) are up to date in this image at all times.

Base image users will see the following output starting with the `1.9` and `1.9.0` tags:

image

Resolves #1931
OpenTofu 1.10 follow-up: #1994

## Target Release

1.9.0

## Checklist

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -159,6 +159,14 @@ docker_manifests:
       - ghcr.io/opentofu/opentofu:{{ .Version }}-386
     skip_push: auto # Skips the push on pre-release versions, like 1.6.1-alpha1. See https://goreleaser.com/customization/docker_manifest/#customization
 
+  - name_template: ghcr.io/opentofu/opentofu:{{ .Major }}
+    image_templates:
+      - ghcr.io/opentofu/opentofu:{{ .Version }}-amd64
+      - ghcr.io/opentofu/opentofu:{{ .Version }}-arm64
+      - ghcr.io/opentofu/opentofu:{{ .Version }}-arm
+      - ghcr.io/opentofu/opentofu:{{ .Version }}-386
+    skip_push: auto # Skips the push on pre-release versions, like 1.6.1-alpha1. See https://goreleaser.com/customization/docker_manifest/#customization
+
   - name_template: ghcr.io/opentofu/opentofu:latest
     image_templates:
       - ghcr.io/opentofu/opentofu:{{ .Version }}-amd64
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 UPGRADE NOTES:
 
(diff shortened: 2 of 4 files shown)


cam72cam (approved):

ollevche (approved):

ProbstDJakob:
@abstractionfactory I know it is a bit late, but in the [docker.mdx](https://github.com/opentofu/opentofu/pull/1993/files#diff-5efba064586df64e4c64b69c43dd8b79ba7077e32936d8fe157d64058d9ee058R91) you give an example for building an image. As of now I think this should work (I am not quite sure and can not test it right now), but in the future it may break. The problem is that you are using Alpine as the image in the first stage and Ubuntu in the second. Alpine uses musl libc and Ubuntu (as most other distros) uses GNU libc. If a future release of tofu dynamically links to libc, this will prevent the resulting image to execute tofu. Also it might be simpler to not use a multi-stage build. First of all it will not introduce the problem and second it is easier to understand for beginners (advanced users could switch to a multi-stage build if they think it is appropriate).

Furthermore since you are using `ADD` you could write something like `ADD https://get.opentofu.org/install-opentofu.sh /install-opentofu.sh`.

ProbstDJakob:
If you never are going to link the binary dynamically to libc this should not be a problem and thus can be left as is. Though it might still be simpler to understand if there is only one stage, but that depends on who you going to ask I guess.

I think most people won't bother looking into the script. So for those people `ADD https://get.opentofu.org/install-opentofu.sh /install-opentofu.sh` might be the way to go. And yes, downloading the script, looking into it, and committing it to the own repo is better regarding security, but it involves complying to the license (which is by the way not included within the script) which is not always possibly depending on the licenses used. As well as regularly checking if the script has changed in a breaking manner (other URL endpoints used) or updated invocations of used tools which on their hand had a breaking change in the CLI (e.g. `curl` dumps the option `--tlsv1.2`) - or wait until the script breaks.

In my opinion the security is nice to have but might just be a false friend since the binary could still be compromised (which would affect way more machines than just the ones using the script), but I would totally get it if you decide to stick to the current version.

You're welcome. I am happy to assist as long as I have time.

PR #2023, opened 2024-09-26
Fixes #2022: Running external commands as a key provider

### LINKED ISSUE #2022: Support external programs as key providers
### OpenTofu Version

```shell
OpenTofu v1.8.2
on linux_amd64
+ provider registry.opentofu.org/carlpett/sops v1.1.1
```

### Use Cases

It should be possible to load the statefile-password dynamically from a file encrypted with SOPS. This prevents the password from being stored in plain text.

### Attempted Solutions

I have already tried it with the provider carlpett/sops, but this gives the error message “Error: Dynamic value in static context” as shown in Issue #2021.

### Proposal

_No response_

### References

_No response_

abstractionfactory:
Resolves #2022

This PR introduces a new key provider that you can configure as follows:

```hcl
terraform {
  encryption {
    key_provider "external" "foo" {
      command = ["./some_program", "some_parameter"]
    }
  }
}
```

This external program can provide OpenTofu with key material. It has to perform 3 steps.

### Step 1: Writing a header

The external program must write the following header to the standard output in a single line, followed by a newline:

```json
{"magic":"OpenTofu-External-Key-Provider","version":1}
```

### Step 2: Reading the input

The external program must read the input data from stdin matching the following JSON:

```json
{
  "external_data": {
    "key1": "value1",
    "key2": "value2"
  }
}
```

### Step 3: Writing the output

The external program must write the key material to the standard output in the following format:

```json
{
  "key": {
    "encryption_key": "newly generated base64-encoded encryption key",
    "decryption_key": "base64-encoded decryption key, if input meta was present, omitted otherwise"
  },
  "meta": {
    "external_data": {
      "key1": "value1",
      "key2": "value2"
    }
  }
}
```

## Chaining with PBKDF2

You can also use this key provider in conjunction with the PBKDF2 provider. This will allow you to read a password from a password manager and supply it as a passphrase:

```hcl
terraform {
  encryption {
    key_provider "external" "foo" {
      command = ["./some_program", "some_parameter"]
    }
    key_provider "pbkdf2" "passphrase" {
      chain = key_provider.external.foo
    }
  }
}
```

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/encryption/keyprovider/externalcommand/provider.go
```diff
@@ -0,0 +1,74 @@
...
+			Message: fmt.Sprintf("the external command exited with an error (%v)\n\nStdout:\n-------\n%s\n-------\nStderr:\n-------\n%s", err, stdout, stderr),
+		}
+	}
+
+	var result *Output
+	decoder := json.NewDecoder(bytes.NewReader(stdout.Bytes()))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&result); err != nil {
+		return keyprovider.Output{}, nil, &keyprovider.ErrKeyProviderFailure{
+			Message: fmt.Sprintf("the external command returned an invalid JSON response (%v)\n\nStdout:\n-------\n%s\n-------\nStderr:\n-------\n%s", err, stdout, stderr),
```
cam72cam, line 69: Could this logging result in leaking secrets?

website/docs/language/state/encryption.mdx
```diff
@@ -225,8 +226,180 @@ The OpenBao key provider is compatible with the last MPL-licensed version of Has
...
+```
+
+#### Creating outputs
+
+The external program must create an output that matches the following format:
+
+```json
+{
+  "key": {
+    "encryption_key": "base64-encoded encryption key",
```
cam72cam, line 281: add "newly generated"

website/docs/language/state/encryption.mdx
```diff
@@ -225,8 +226,180 @@ The OpenBao key provider is compatible with the last MPL-licensed version of Has
...
+	if inMeta != nil {
+		// TODO produce decryption key
+	}
+
+	output := Output{
+		// TODO: produce output
+	}
+	outputData, err := json.Marshal(output)
+	if err != nil {
+		log.Fatalf("Failed to stringify output: %v", err)
```
cam72cam, line 396: stringify -> encode

cam72cam (dismissed):

internal/encryption/keyprovider/externalcommand/provider.go
```diff
@@ -0,0 +1,74 @@
...
+	"os/exec"
+
+	"github.com/opentofu/opentofu/internal/encryption/keyprovider"
+)
+
+type keyProvider struct {
+	command []string
+}
+
+func (k keyProvider) Provide(rawMeta keyprovider.KeyMeta) (keysOutput keyprovider.Output, encryptionMeta keyprovider.KeyMeta, err error) { //nolint:nonamedreturns //This is a stupid rule.
```
ollevche, line 23: nit: there is no naked returns so naming return values is not needed

internal/encryption/keyprovider/externalcommand/provider.go
```diff
@@ -0,0 +1,74 @@
...
+				}
+			}
+		}
+		return keyprovider.Output{}, nil, &keyprovider.ErrKeyProviderFailure{
+			Message: fmt.Sprintf("the external command exited with an error (%v)\n\nStderr:\n-------\n%s", err, stderr),
+		}
+	}
+
+	var result *Output
+	decoder := json.NewDecoder(bytes.NewReader(stdout.Bytes()))
```
ollevche, line 65: nit: can we simplify it a bit?
```suggestion
	decoder := json.NewDecoder(stdout)
```

ollevche:
> I'm still considering if the external program should output a handshake line indicating a protocol version in case later we want to change the way these programs work.

I think this is a good idea, the handshake could be done as an exchange of request-response JSON objects to handle potentially more than just a version. This way we will allow external commands to also validate some protocol metadata.

yan12125:
Thank you, this works like a charm. However, I'm not quite sure what's the purpose of `external_data`. I got `null` during either encryption or decryption.

yan12125:
Thank you! I guess I don't need such a feature in my simple scenario :)

ollevche (dismissed):

internal/encryption/keyprovider/externalcommand/descriptor.go
```diff
@@ -0,0 +1,35 @@
...
+	keyprovider.Descriptor
+
+	TypedConfig() *Config
+}
+
+type descriptor struct {
+}
+
+func (f descriptor) ID() keyprovider.ID {
+	return "externalcommand"
```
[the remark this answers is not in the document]
cam72cam:
```hcl
key_provider "external" {
    command = "..."
}
```

internal/encryption/keyprovider/external/README.md
```diff
@@ -0,0 +1,22 @@
+# External command key provider
```
ollevche, line 1: ?
```suggestion
# External key provider

```

website/docs/language/state/encryption.mdx
```diff
@@ -161,7 +169,8 @@ The PBKDF2 key provider allows you to use a long passphrase as to generate a key
 
 | Option                   | Description                                                                                                                                             | Min.      | Default                            |
 |--------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------|-----------|------------------------------------|
-| passphrase *(required)*  | Enter a long and complex passphrase.                                                                                                                    | 16 chars. | -                                  |
+| passphrase *(required)*  | Enter a long and complex passphrase. Required if `chain` is not specified.                                                                              | 16 chars. | -                                  |
+| chain *(required)*       | Receive the passphrase from another key provider. Required if `passphrase` is not specified.                                                            |           | -                                  |
```
ollevche, line 173: Just a note for other readers, who might not be too familiar with the encryption implementation.

I was a bit surprised by not finding any code which makes it possible to evaluate chains (this PR doesn't touch anything outside of specific key providers). Then I realized we already have it in place but with no actual user-facing features.

`targetBuilder.setupKeyProviders` is what makes it possible so if anyone else got confused, reading through that function makes it more clear.

Also, I am wondering, should we put more tests into that chaining feature? No one used it before and we may introduce something unexpected by enabling this. WDYT?

ollevche: Is it possible to use chaining with non-experimental key providers?

ollevche (approved):
Everything looks good, except I am a bit worried about introducing changing here. I would put more testing effort for this feature.

If that's only usable with experimental key provider - I think this fine, otherwise we need more tests.

ollevche:
Here is the output of `echo -n "null" | ./testprovider.sh` in the `internal/encryption/keyprovider/external/testprovider/data` folder:

Screenshot 2025-01-08 at 16 24 49

ollevche:
INPUT in that case is `-n null`:

Screenshot 2025-01-08 at 16 29 14

website/docs/language/state/encryption.mdx
ollevche: Indeed, it doesn't seem to be much of a risk. I am good, thanks!

ollevche (dismissed):

cam72cam (dismissed):

cam72cam (approved):

ollevche (approved):

PR #2035, opened 2024-09-30
test: use regex for expected value in `TestUiHookPreApply_periodicTimer`

wyardley:
Resolve flaky test by using regex for expected value of `TestUiHookPreApply_periodicTimer`. Match any single digit for the number of seconds vs. hard-coding. This seems less fragile to me.

I could look into separately asserting on the total elapsed time, and / or use a smaller range of numbers in the regex if there's concern about this starting to take unexpectedly more time, but I'm assuming that's not mostly what this test is trying to check.

Example failure:
https://github.com/opentofu/opentofu/actions/runs/11072397888/job/30766524726

Note: I do get one golangci-lint failure within that specific function, but as best I can tell, it's not related to my changes.

## Target Release

1.9.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/views/hook_ui_test.go
+++ b/internal/command/views/hook_ui_test.go
@@ -126,15 +126,15 @@ func TestUiHookPreApply_periodicTimer(t *testing.T) {
 	close(uiState.DoneCh)
 	<-uiState.done
 
-	expectedOutput := `test_instance.foo: Modifying... [id=test]
-test_instance.foo: Still modifying... [id=test, 1s elapsed]
-test_instance.foo: Still modifying... [id=test, 2s elapsed]
-test_instance.foo: Still modifying... [id=test, 3s elapsed]
+	expectedRegexp := `test_instance\.foo: Modifying... \[id=test\]
+test_instance\.foo: Still modifying... \[id=test, \ds elapsed\]
+test_instance\.foo: Still modifying... \[id=test, \ds elapsed\]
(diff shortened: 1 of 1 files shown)


internal/command/views/hook_ui_test.go
```diff
@@ -126,15 +126,15 @@ func TestUiHookPreApply_periodicTimer(t *testing.T) {
...
+	expectedRegexp := `test_instance\.foo: Modifying... \[id=test\]
+test_instance\.foo: Still modifying... \[id=test, \ds elapsed\]
+test_instance\.foo: Still modifying... \[id=test, \ds elapsed\]
+test_instance\.foo: Still modifying... \[id=test, \ds elapsed\]
 `
 	result := done(t)
 	output := result.Stdout()
-	if output != expectedOutput {
-		t.Fatalf("Output didn't match.\nExpected: %q\nGiven: %q", expectedOutput, output)
+	if matched, _ := regexp.MatchString(expectedRegexp, output); !matched {
```
wyardley, line 136: Note: I used another example within the same file as reference here, but assume that's not significant enough to require a comment here?

cam72cam: That's a small idiomatic change, no comment necessary.  Thanks for checking!

cam72cam (approved):

cam72cam:
This was a bit of an occasional pain, thanks for submitting a patch!

ollevche (approved):

PR #2086, opened 2024-10-19
Fix exit command in console interactive mode

boaz0:
When typing "exit" the code does run "break" but since "switch" is the innermost statement and not "for", the program is still running and not exiting with status code 0.

To fix it, "return 0" when "exit" is typed instead of "break".

Resolves # 

## Target Release

1.9.0

## Checklist

### Go checklist

 

- [xI have run golangci-lint on my change and receive no errors relevant to my code.


cam72cam (approved):

internal/command/console_interactive.go
```diff
@@ -68,7 +68,7 @@ func (c *ConsoleCommand) modeInteractive(session *repl.Session, ui cli.Ui) int {
 				c.showDiagnostics(diags)
 			}
 			if exit {
-				break
+				return 0
```
ollevche, line 71: nit: I would like to change other break statements to returns just in case someone puts something after the loop in the future and doesn't account for this exact return.

boaz0:
@ollevche - i hope i addressed all your concerns. I ran unit tests to make sure nothing is broken and did play around to make sure the behavior is not broken.

ollevche (changes requested):
I think we shouldn't make any additional changes besides replacing `switch-case` with `if-else`. Sorry for not making it clear beforehand!

boaz0:
no worries @ollevche 
I apologize for the misunderstanding on my part as well :smiley_cat:

internal/command/console_interactive.go
```diff
@@ -57,18 +57,17 @@ func (c *ConsoleCommand) modeInteractive(session *repl.Session, ui cli.Ui) int {
...
 			l.SetPrompt(fmt.Sprintf("%s ", strings.Repeat(".", openState)))
-		default:
+		} else {
 			out, exit, diags := session.Handle(fullCommand)
 			if diags.HasErrors() {
 				c.showDiagnostics(diags)
 			}
 			if exit {
-				break
+				return 0
```
ollevche, line 70: Let's leave a break statement in there for convenience, it is not a problem since we changed `switch-case` to `if-else`

ollevche (approved):
Thank you for the fix, @boaz0!

ollevche:
We forgot to update changelog file for this PR and some of the others. I opened a separate PR for this: https://github.com/opentofu/opentofu/pull/2116

PR #2090, opened 2024-10-22
"http" backend to return existing lock metadata on lock conflict, rather than new lock metadata

haiderbari:
Resolves # Bug 2004 https://github.com/opentofu/opentofu/issues/2004

## Target Release

1.9.0, if possible back port

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/http/client_test.go
+++ b/internal/backend/remote-state/http/client_test.go
@@ -336,8 +336,16 @@ func TestHttpClient_lock(t *testing.T) {
 			}
 
 			lockID, err := client.Lock(tt.lockInfo)
-			if (err != nil) != (tt.expectedErrorMsg != nil) || (err != nil && err.Error() != tt.expectedErrorMsg.Error()) {
-				// One of errors is nil and the other is not OR if both are not nil but their messages differ
+			if tt.expectedErrorMsg != nil && err == nil {
+				// no expected error
+				t.Errorf("Lock() no expected error = %v", tt.expectedErrorMsg)
+			}
+			if tt.expectedErrorMsg == nil && err != nil {
+				// unexpected error
+				t.Errorf("Lock() unexpected error = %v", err)
+			}
+			if tt.expectedErrorMsg != nil && err.Error() != tt.expectedErrorMsg.Error() {
+				// mismatched errors
 				t.Errorf("Lock() error = %v, want %v", err, tt.expectedErrorMsg)
 			}
 			if lockID != tt.expectedStateLockID {
--- a/internal/backend/remote-state/http/client_test.go
+++ b/internal/backend/remote-state/http/client_test.go
@@ -259,6 +259,19 @@ func TestHttpClient_IsLockingEnabled(t *testing.T) {
 // Tests the Lock method for the HTTP client.
 // Test to see correct lock info is returned
 func TestHttpClient_lock(t *testing.T) {
+	stateLockInfoA := statemgr.LockInfo{
+		ID:        "ada-lovelace-state-lock-id",
+		Who:       "AdaLovelace",
+		Operation: "TestTypePlan",
+	}
+
+	trimString := func(str string) string {
+		// Anonymous Helper function to remove new line, tab
+		// and space characters from a string
+		space := regexp.MustCompile(`\s+`)
+		return space.ReplaceAllString(str, " ")
+	}
+
 	testCases := []struct {
 		name           string
 		lockMethod     string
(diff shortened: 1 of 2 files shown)


internal/backend/remote-state/http/client_test.go
```diff
@@ -251,3 +255,136 @@ func TestHttpClient_IsLockingEnabled(t *testing.T) {
...
+		return space.ReplaceAllString(str, " ")
+	}
+
+	testCases := []struct {
+		name           string
+		lockMethod     string
+		lockInfo       *statemgr.LockInfo
+		handler        http.HandlerFunc
+		validateResult func(lockID string, errorMessage error)
+	}{
```
ollevche, line 281: I would prefer a more clear separation between test logic and test data (inputs and outputs). I think the following structure covers test needs in a slightly simpler way (you would also need to change test run logic inside a loop):

```suggestion
	inputLockInfo := statemgr.LockInfo{
		ID:        "ada-lovelace-state-lock-id",
		Who:       "AdaLovelace",
		Operation: "TestTypePlan",
	}

	testCases := []struct {
		name				string
		lockResponseStatus	int
		lockResponseBody		[]byte
		expectedLockInfo		*statemgr.LockInfo // or just an ID
		expectedErrorMsg		string
	}{
	...
```

internal/backend/remote-state/http/client_test.go
```diff
@@ -251,3 +255,136 @@ func TestHttpClient_IsLockingEnabled(t *testing.T) {
...
+					t.Fatalf("Lock, expected locked: ID= to be %q, got %q", expectedStateLockID, errorMessage.Error())
+				}
+
+				if !strings.Contains(errString, fmt.Sprintf("ID: %s", expectedStateLockID)) {
+					t.Fatalf("Lock, expected ID: to be %q, got %q", expectedStateLockID, errorMessage.Error())
+				}
+
+				if !strings.Contains(errString, "Who: LinusTorvalds") {
+					t.Fatalf("Lock, expected Who: to be LinusTorvalds, got %q", errorMessage.Error())
+				}
```
ollevche, line 335: Can we compare two strings (error messages) directly?

PS with my previous suggestion, this comparison can be moved to a shared test run logic (inside a loop).

ollevche (commented):
Thank you for the contribution! I left a few suggestions to simplify tests.

internal/backend/remote-state/http/client_test.go
haiderbari: I have updated the tests base your suggestion

internal/backend/remote-state/http/client_test.go
haiderbari: I have updated the comparison checks

internal/backend/remote-state/http/client_test.go
```diff
@@ -251,3 +253,93 @@ func TestHttpClient_IsLockingEnabled(t *testing.T) {
...
+			lockResponseStatus:  http.StatusLocked,
+			lockResponseBody:    []byte(`{"ID":"linus-torvalds-http-remote-state-lock-id", "Path": "", "Operation": "TestTypePlan", "Who": "LinusTorvalds", "Created": "2024-08-15T09:00:26.00Z" }`),
+			expectedStateLockID: "",
+			expectedErrorMsg: &statemgr.LockError{
+				Info: &statemgr.LockInfo{
+					ID:        "linus-torvalds-http-remote-state-lock-id",
+					Who:       "LinusTorvalds",
+					Operation: "TestTypePlan",
+					Created:   time.Date(2024, time.August, 15, 9, 0, 26, 0, time.UTC),
+				},
```
ollevche, line 297: nit (could be ignored): we can move this as a separate variable and reuse it for json generation and in the error message.

internal/backend/remote-state/http/client_test.go
```diff
@@ -251,3 +253,93 @@ func TestHttpClient_IsLockingEnabled(t *testing.T) {
...
+			client := &httpClient{
+				LockURL:    lockURL,
+				LockMethod: "LOCK",
+				Client:     retryablehttp.NewClient(),
+			}
+
+			lockID, err := client.Lock(tt.lockInfo)
+			if tt.expectedErrorMsg != nil && err.Error() != tt.expectedErrorMsg.Error() {
+				t.Errorf("Lock() error = %v, want %v", err, tt.expectedErrorMsg)
+			}
```
ollevche, line 339: We must check for (1) unexpected error, (2) absence of expected error and (3) mismatched errors. Currently, we only check for (3) and panic on (2).

ollevche (changes requested):

internal/backend/remote-state/http/client_test.go
haiderbari: thanks for the feedback, I have already implemented it, I haven't pushed that change yet.

internal/backend/remote-state/http/client_test.go
ollevche: I think we can do it without a separate variable:
```go
			lockID, err := client.Lock(tt.lockInfo)
			if tt.expectedErrorMsg != nil && err == nil {
				// no expected error
			}
			if tt.expectedErrorMsg == nil && err != nil {
				// unexpected error
			}
			if tt.expectedErrorMsg != nil && err != nil && err.Error() != tt.expectedErrorMsg.Error() {
				// mismatched errors
			}			
```

haiderbari: I think this capture all use cases 

`if (err != nil) != (tt.expectedErrorMsg != nil) || (err != nil && err.Error() != tt.expectedErrorMsg.Error()) {
	// One of errors is nil and the other is not OR if both are not nil but their messages differ
}
`

ollevche: I would prefer a simpler-to-read option with multiple ifs, it also allows us to create different error messages based on different cases.

haiderbari: Hi @ollevche I have updated the tests base on your suggestions, thanks

internal/backend/remote-state/http/client_test.go
haiderbari: Moved it to it's own variable.

ollevche:
Hey @haiderbari could you please also fix the linting issue?

haiderbari:
> Hey @haiderbari could you please also fix the linting issue?

Hi @ollevche does the version matter for golanglint ci. When I ran the command it didn't flag that line for me or are you not using the default linting rules.

Issue #1367, opened 2024-03-11
`tofu test` - Should use `.tfvars` files from the test folder


RLRabinowitz:
### OpenTofu Version

```shell
1.6.2
```

### Use Cases

Using `.tfvars` files

Currently when running `tofu test`s, it picks up variables of `terraform.tfvars` and `*.auto.tfvars` files from the root configuration folder.
Being able to also use `terraform.tfvars` and `*.auto.tfvars` from the test folder would make sense, to help set values specific for test suites

### Attempted Solutions

A workaround could be using `TF_VAR_` environment variables, which isn't so nice, or setting static `variables` blocks in the `testtf.hcl` file, but that would necessitate copying that `variables` block between test files

### Proposal

Alongside using `terraform.tfvars`, `terraform.tfvars.json`, `*.auto.tfvars`, `*.auto.tfvars.json` files from the root directory, also take variables from the same files in the test folder.

If a variable appears both in a `.tfvars` file in the root configuration and in the test folder, the one in the test folder should take precedence. If a variable appears only in one `.tfvars` file, be it in the root configuration or in the test folder, then this variable should be available as such in the test

janosdebugs:
Thank you for volunteering @sanskruti-shahu , I've assigned you. Please read [the "I've been assigned an issue, now what?" section of the contribution guide](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md#ive-been-assigned-an-issue-now-what) before you begin working.

jasonyoung-pearl:
This change seems to result in tfvars in the `tests/` directory _always_ being automatically used, even when simply running `tofu apply`, not just `tofu test`. This means if we create a `test.auto.tfvars` in our module's `tests/` directory, they are included whenever the module is used. Hard to see how this could be expected behavior, so assuming this should be raised as a bug?

cam72cam:
Actually, I'm going to re-open this as there are a few bugs in this functionality

AdosH1:
Hey @cam72cam, I'd like to take a look at this
You mentioned a few bugs in this functionality - would you mind elaborating the specific issues?

From what I understand the current implementation simply overwrites variables with test ones regardless of the command used eg. `tofu apply` vs `tofu test`
Are there other problems in the implementation that you saw?

AdosH1:
@cam72cam 
On looking deeper, it seems you may have solved this issue in #2039 
[...]

PR #2128, opened 2024-11-01
test: ensure test vars are not used in tf plan

AdosH1:
Added test to ensure variables in a tests folder are not applied upon a `tf plan` command.
Related to work discussed in [#2106](https://github.com/opentofu/opentofu/pull/2106)
Stemming from issue #1367 
## Target Release

N/A

## Checklist

### Go checklist

 


internal/command/plan_test.go
```diff
@@ -220,6 +220,44 @@ func TestPlan_noState(t *testing.T) {
...
+	}
+	code := c.Run(args)
+	output := done(t)
+	if code != 0 {
+		t.Fatalf("bad: %d\n\n%s", code, output.Stderr())
+	}
+
+	plan := testReadPlan(t, outPath)
+	ctyTestVar := plan.VariableValues["testVar"]
+	psuedo, err := ctyTestVar.Decode(cty.DynamicPseudoType)
```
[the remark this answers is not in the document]
ollevche: This is fine - variables are wrapped as `cty.DynamicPseudoType`, but after decoding it is okay to use it as a `cty.String` in this case.

ollevche, line 249: `psuedo` should be renamed like `decodedExpectedVar` to be more readable.

internal/command/plan_test.go
```diff
@@ -220,6 +220,44 @@ func TestPlan_noState(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unable to decode plan variables, error: %v", err)
+	}
+
+	result := psuedo.GoString()
```
ollevche, line 254: You can use `psuedo.AsString()` here because it is fine to assume the decoded var is of `cty.String` type in that particular case.

internal/command/plan_test.go
```diff
@@ -220,6 +220,44 @@ func TestPlan_noState(t *testing.T) {
...
+	if !strings.Contains(result, "ValueFROMmain/tfvars") {
```
ollevche, line 255: Let's move the expected value as a separate constant to reduce duplication.

internal/command/plan_test.go
```diff
@@ -220,6 +220,44 @@ func TestPlan_noState(t *testing.T) {
...
+			View:             view,
+		},
+	}
+
+	outPath := filepath.Join(td, "test.plan")
+	args := []string{
+		"-out", outPath,
+	}
+	code := c.Run(args)
+	output := done(t)
```
ollevche, line 242: As an alternative approach, you can check the output of the plan command instead of the plan file. You would want to check if `test_instance.testRes.ami` equals to expected value (`ValueFROMmain/tfvars`) in the output. This alternative approach could simplify the test a bit.

ollevche (commented):
Thanks for the contribution @AdosH1! The test is fine already and I left a few suggestions to improve.

ollevche:
Also, @AdosH1 you have to add a DCO sign to all of your commits: [details](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md#signing-off-your-commits).

internal/command/plan_test.go
AdosH1: thanks, updated

internal/command/plan_test.go
AdosH1: agreed, updated :pray:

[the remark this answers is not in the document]
AdosH1: Thanks for the explanation, its weird to me `cty`'s own encode/decode functions don't quite work together as well as I'd hope

internal/command/plan_test.go
AdosH1: I've replaced this with a simple `contains` in the output string, as I don't want to dictate any formatting requirements in the output
Given the test case, there won't be anything else that would assume the value `valueFROMmain/tfvars`so I think this is okay - let me know if you think we need to be more specific

AdosH1:
Thank you for your comments @ollevche, I appreciate your support!
I've updated the PR based on your comments and simplified the assertion, let me know if there's anything else you think might need to be improved :pray:

internal/command/plan_test.go
ollevche: I don't think it is required, but I would use a more specific line to search for. Also, I would double-check if the output contains `ValueFROMtests/tfvars` just to be sure it doesn't.

ollevche:
Hey @AdosH1 I left a minor improvement comment. Otherwise, the PR looks good!

internal/command/plan_test.go
AdosH1: Updated to assert for the line `ami = "ValueFROMmain/tfvars"`
Added a check to ensure `ValueFROMtests/tfvars` doesn't exist in output

internal/command/plan_test.go
```diff
@@ -221,6 +221,41 @@ func TestPlan_noState(t *testing.T) {
...
+	args := []string{}
+	code := c.Run(args)
+	output := done(t)
+	if code != 0 {
+		t.Fatalf("bad: %d\n\n%s", code, output.Stderr())
+	}
+
+	// Verify values defined in the 'tests' folder are not used during a plan action
+	expectedValue := "ValueFROMmain/tfvars"
+	expectedResult := fmt.Sprintf("ami = \"%s\"", expectedValue)
```
ollevche, line 247: nit:

```suggestion
	expectedResult := `ami = "ValueFROMmain/tfvars"`
```

internal/command/plan_test.go
```diff
@@ -221,6 +221,41 @@ func TestPlan_noState(t *testing.T) {
...
+	view, done := testView(t)
+	c := &PlanCommand{
+		Meta: Meta{
+			testingOverrides: metaOverridesForProvider(p),
+			View:             view,
+		},
+	}
+
+	args := []string{}
+	code := c.Run(args)
```
ollevche, line 239: nit:

```suggestion
	code := c.Run([]string{})
```

AdosH1:
@ollevche Thanks for the suggestions, committed & ensured tests are passing

ollevche (approved):

PR #2183, opened 2024-11-19
Bump golang version to 1.22.8 (address CVE-2024-34156) (#2050)

ghost:
This is a backport of #2050 to 1.7 to bump the Go version.

## Target Release

1.7.0

## Checklist

### DIFF
--- a/.github/workflows/compare-snapshots.yml
+++ b/.github/workflows/compare-snapshots.yml
@@ -16,12 +16,16 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
 
+      - name: Determine Go version
+        id: go
+        uses: ./.github/actions/go-version
 
-      - name: Install Go
-        uses: actions/setup-go@v2
-        with: { go-version: '1.21.3' }
+      - name: Set up Go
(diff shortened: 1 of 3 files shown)


apparentlymart:
There are some changes in Go 1.22 that might be troublesome to backport, but all seem minor and debatable how much they matter:

- The `encoding/json` package now encodes control codes 8 and 12 in strings as `\b` and `\f` instead of `\u0008` and `\u000c`. The upstream `cty` delegates to `encoding/json` for its JSON serialization of strings, so OpenTofu's `jsonencode` function would be affected by this.

    If someone were using `jsonencode` in a location that causes a provider to force replacing an object then that could be troublesome. However, those particular control codes are probably pretty rare to use in practice, so perhaps that's okay.
- The behavior of the pure Go DNS resolver on Windows has changed to include the `%SystemRoot%\System32\drivers\etc\hosts` file. This would affect OpenTofu only if we're building for Windows with CGo disabled; I'm not sure if we are. Even if it does affect OpenTofu, it's arguably a good effect.
- The standard library's HTTP client now rejects responses with the `Content-Length` header set to the empty string, whereas before I believe it ignored that. If anyone is using OpenTofu with a server that returns responses that are broken in this specific way then this upgrade might break their use of that server, but again that seems relatively unlikely.


All of these seem pretty unlikely to cause anyone a practical problem, but perhaps still worth mentioning them in "upgrade notes" in the changelog of each release we backport this to. :thinking:

cam72cam (approved):

ollevche (approved):

PR #2256, opened 2024-12-05
Fix packagecloud push skip for alpha/beta/rc

cam72cam:
Anything that's released under a "v" tag and is not a pre-release, we should push to packagecloud.  The v check is probably not needed, but I wanted to minimize the changes to this pipeline.

branch | tag | prerelease | packagecloud
--- | --- | --- | ---
main | v1.9.0-alpha1 | yes | no
main | v1.9.0-beta1 | yes | no
v1.9 | v1.9.0-rc1 | yes | no
v1.9 | v1.9.0 | no | yes
v1.8 | v1.8.8 | no | yes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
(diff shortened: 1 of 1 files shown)


cam72cam:
Was barking up the wrong tree

PR #2289, opened 2024-12-12
Backport[v1.7]: Bump golang.org/x/crypto from 0.21.0 to 0.31.0

ghost:
Backporting #2287 to v1.7

> [!IMPORTANT]
> This issue does NOT affect OpenTofu because it does not run TLS servers, but to avoid unnecessary nagging from compliance checking tools, we should merge and backport this anyway.

## Target Release

1.7.0


cam72cam (approved):

ollevche:
@abstractionfactory something weird happened with DCO check

apparentlymart:
Unfortunately the `golang.org/x/...` modules tend to bundle both security fixes and functionality changes together in minor releases, and so it's not always safe to adopt their security fixes in an OpenTofu patch release.

I know of at least one situation where the Go team responded to feedback about this by backporting a security fix to an earlier minor version and making a patch release -- i.e. in this case pubilshing a `v0.21.1` that includes only the security fix and nothing else -- and so we could potentially pursue that sort of feedback again here, but probably not worth the bother if this was just a false-positive due to a low-quality scanning tool that doesn't use sufficient granularity when detecting whether a particular codebase is affected by an upstream problem.

It would probably be more productive to encourage those affected by the problem to encourage their security tool vendors to produce more accurate results with fewer false-positives; the data to do so is readily available in [the Go vulnerability database](https://pkg.go.dev/vuln/), which tracks vulnerabilities at the granularity of individual symbols rather than entire modules and so tools using it (like [`govulncheck`](https://go.dev/blog/govulncheck)) can produce considerably less noisy results.

ollevche:
For the same reason, https://github.com/opentofu/opentofu/pull/2311 is not backported to `v1.7`. It requires `golang.org/x/crypto` upgrade as well.

To be clear: OpenTofu is NOT affected by the vulnerability resolved in https://github.com/opentofu/opentofu/pull/2311.

PR #2324, opened 2025-01-02
Fix quadratic marshalPlannedValues

jonjohnsonjr:
As mentioned in https://github.com/opentofu/opentofu/issues/2252, Changes.ResourceInstance() has linear complexity in the number of resources being changed. Ideally, we will eventually address the root problem here, but given that the Resources slice is exposed, it's hard to make that change without breaking all the callers.

In this change, I aim to fix only the "tofu show" path, which already is reaching into Changes.Resources to construct its own module mappings. I just take this one step further by adding an additional map to make lookups constant time. Since we're visiting every change in order to marshal them, we were calling ResourceInstance for every change, which then iterated over every change again, hence the quadratic performance.

This cuts my "tofu show" time in half. What remains is just loading the plan files, some thorny go-cty stuff I don't want to touch, and finally the actual JSON marshalling. I don't think we can do much better without a significant undertaking (due to the go-cty stuff) beyond parallelizing some of this, which seems doable but I'll save for a later date.

[Before](https://profiler.firefox.com/public/vf0adrt0wn0vj1mbjzmz04bbhsh4bfak47qj6pr/flame-graph/?globalTrackOrder=0w9&hiddenGlobalTracks=1w9&hiddenLocalTracksByPid=24597-07d~24614-0wa~24615-0wc~24616-0we~24617-0we~24618-0w7~24619-0w6~24620-0w6~24624-0w7~24633-0w7&symbolServer=http%3A%2F%2F127.0.0.1%3A3000%2F05kfyd3pvqinq21ka2bfykzn3hh544z2mrsjkxk&thread=02w79wdfwh&v=10)

image

[After](https://profiler.firefox.com/public/gs8rjc0yhyb2e7epqjxn68j22afz4rbh7p52ewg/flame-graph/?globalTrackOrder=0w9&hiddenGlobalTracks=1w9&hiddenLocalTracksByPid=41251-0cd~41265-0wb~41266-0wc~41267-0wf~41268-0wd~41279-0w7~41288-0w9~41292-0w8~41295-0w7~41297-0w7&symbolServer=http%3A%2F%2F127.0.0.1%3A3000%2F02r6cpz6p74rwc85z4skk0jpg1y0da47hv55jgc&thread=02wcfwh&v=10)

image

Resolves # 

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonplan/values.go
+++ b/internal/command/jsonplan/values.go
@@ -98,63 +98,76 @@ func marshalPlannedOutputs(changes *plans.Changes) (map[string]Output, error) {
 func marshalPlannedValues(changes *plans.Changes, schemas *tofu.Schemas) (Module, error) {
 	var ret Module
 
-	// build two maps:
+	// build three maps:
 	// 		module name -> [resource addresses]
 	// 		module -> [children modules]
+	// 		resource addr -> change src
 	moduleResourceMap := make(map[string][]addrs.AbsResourceInstance)
 	moduleMap := make(map[string][]addrs.ModuleInstance)
+	changeMap := make(map[string]*plans.ResourceInstanceChangeSrc)
 	seenModules := make(map[string]bool)
 
 	for _, resource := range changes.Resources {
-		// If the resource is being deleted, skip over it.
 		// Deposed instances are always conceptually a destroy, but if they
 		// were gone during refresh then the change becomes a noop.
-		if resource.Action != plans.Delete && resource.DeposedKey == states.NotDeposed {
-			containingModule := resource.Addr.Module.String()
-			moduleResourceMap[containingModule] = append(moduleResourceMap[containingModule], resource.Addr)
+		if resource.DeposedKey != states.NotDeposed {
+			continue
+		}
 
-			// the root module has no parents
-			if !resource.Addr.Module.IsRoot() {
-				parent := resource.Addr.Module.Parent().String()
-				// we expect to see multiple resources in one module, so we
-				// only need to report the "parent" module for each child module
-				// once.
-				if !seenModules[containingModule] {
-					moduleMap[parent] = append(moduleMap[parent], resource.Addr.Module)
-					seenModules[containingModule] = true
-				}
+		changeMap[resource.Addr.String()] = resource
 
-				// If any given parent module has no resources, it needs to be
-				// added to the moduleMap. This walks through the current
-				// resources' modules' ancestors, taking advantage of the fact
-				// that Ancestors() returns an ordered slice, and verifies that
-				// each one is in the map.
-				ancestors := resource.Addr.Module.Ancestors()
-				for i, ancestor := range ancestors[:len(ancestors)-1] {
-					aStr := ancestor.String()
+		// If the resource is being deleted, skip over it.
+		if resource.Action == plans.Delete {
+			continue
+		}
 
-					// childStr here is the immediate child of the current step
-					childStr := ancestors[i+1].String()
-					// we likely will see multiple resources in one module, so we
-					// only need to report the "parent" module for each child module
-					// once.
-					if !seenModules[childStr] {
-						moduleMap[aStr] = append(moduleMap[aStr], ancestors[i+1])
-						seenModules[childStr] = true
-					}
-				}
+		containingModule := resource.Addr.Module.String()
+		moduleResourceMap[containingModule] = append(moduleResourceMap[containingModule], resource.Addr)
+
+		// the root module has no parents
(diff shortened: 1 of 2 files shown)


cam72cam:
Hi @jonjohnsonjr, would you be open to adding a test case in https://github.com/opentofu/opentofu-performance-testing/ to show the quadratic behavior / fix?

jonjohnsonjr:
Added both this PR and https://github.com/opentofu/opentofu/pull/2323 (more noticeable for that test case, actually) to https://github.com/opentofu/opentofu-performance-testing/pull/3

internal/command/jsonplan/values.go
```diff
@@ -98,18 +98,21 @@ func marshalPlannedOutputs(changes *plans.Changes) (map[string]Output, error) {
...
 	moduleMap := make(map[string][]addrs.ModuleInstance)
+	changeMap := make(map[string]*plans.ResourceInstanceChangeSrc)
 	seenModules := make(map[string]bool)
 
 	for _, resource := range changes.Resources {
 		// If the resource is being deleted, skip over it.
 		// Deposed instances are always conceptually a destroy, but if they
 		// were gone during refresh then the change becomes a noop.
 		if resource.Action != plans.Delete && resource.DeposedKey == states.NotDeposed {
+			changeMap[resource.Addr.String()] = resource
```
ollevche, line 115: nit: I would prefer `changeMap` content to be equal to what `plans.Changes.ResourceInstance` returns, i.e. all `NotDeposed` changes. Then, it makes sense to rename `changeMap` to what reflects its content better, e.g. `notDeposedChangeMap`. This way we keep the original behaviour and make the retrieval more efficient. WDYT?

internal/command/jsonplan/values.go
```diff
@@ -168,12 +171,12 @@ func marshalPlannedValues(changes *plans.Changes, schemas *tofu.Schemas) (Module
...
 // marshalPlanResources
-func marshalPlanResources(changes *plans.Changes, ris []addrs.AbsResourceInstance, schemas *tofu.Schemas) ([]Resource, error) {
+func marshalPlanResources(changeMap map[string]*plans.ResourceInstanceChangeSrc, ris []addrs.AbsResourceInstance, schemas *tofu.Schemas) ([]Resource, error) {
 	var ret []Resource
 
 	for _, ri := range ris {
-		r := changes.ResourceInstance(ri)
-		if r.Action == plans.Delete {
+		r, ok := changeMap[ri.String()]
+		if !ok || r.Action == plans.Delete {
```
ollevche, line 179: nit: `changes.ResourceInstance` returns nil if there is nothing found. Then, it would panic on `r.Action == plans.Delete` check. In the current implementation, we would skip the resource if it is not found, assuming this is a `plans.Delete` action. However, it might silence some critical dev error (because panic is hidden). I would prefer to keep the original behavior here.

ollevche (commented):
Thanks for contribution @jonjohnsonjr! I left a few nit comments, which should not be a problem during the usual execution, however might hide some developer mistakes in future extensions.

internal/command/jsonplan/values.go
jonjohnsonjr: Makes sense, I agree. Flipped some conditionals to avoid excessive indentation as well, thanks.

internal/command/jsonplan/values.go
jonjohnsonjr: Good point! Done.

jonjohnsonjr:
Addressing https://github.com/opentofu/opentofu/pull/2324#discussion_r1901657741 made this diff messier, I suggest viewing with whitespace ignored: https://github.com/opentofu/opentofu/pull/2324/files?diff=split&w=1

internal/command/jsonplan/values_test.go
```diff
@@ -366,6 +367,16 @@ func testSchemas() *tofu.Schemas {
 	}
 }
 
+func testChanges(changes *plans.Changes) map[string]*plans.ResourceInstanceChangeSrc {
+	ret := make(map[string]*plans.ResourceInstanceChangeSrc)
+	for _, resource := range changes.Resources {
+		if resource.Action != plans.Delete && resource.DeposedKey == states.NotDeposed {
```
ollevche, line 373:
```suggestion
		if resource.DeposedKey == states.NotDeposed {
```

jonjohnsonjr: Thanks!

cam72cam (approved):

ollevche (approved):
Thank you @jonjohnsonjr!

PR #2370, opened 2025-01-14
Moved for different types

### LINKED ISSUE #1369: Add support to moves between two resources of different types
The terraform-provider-sdk has added `MoveResourceState`, which allows the provider to implement state moves between two resources of different types.

We should support providers implementing this API, and allow moves between two sources of different types

Gogotchuri:
This is an implementation for moved when resource types are different.

## Assumptions
We've made a few assumptions in this PR:
1) `ResourceInstanceObjectSrc` won't be decoded until we call `NodeAbstractResourceInstance.readResourceInstanceState`.
2) We must run the same logic as `upgradeResourceState` regarding schema versioning validations and conformance.
3) There will not be a need to call `upgradeResourceState` if we call `moveResourceState`. Meaning - `MoveResourceState` upgrades the state itself.

## Implementation
`upgradeResourceState` and `moveResourceState` are quite similar code paths and I wanted to use existing validation and preparation logic already presented in the `upgradeResourceState` function.
I've abstracted away provider call from `upgradeResourceState` as the `providerStateTransform` type function.
Using the strategy pattern, we can pass different state transformation logic into the existing `upgradeResourceState` function, now renamed to `transformResourceState`. That way we are reusing existing logic and only swapping provider state transformation calls - `UpgradeResourceState` and `MoveResourceState`.

Resolves #1369 

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/tofu/node_resource_abstract.go
```diff
@@ -509,14 +509,17 @@ func (n *NodeAbstractResource) writeResourceState(ctx EvalContext, addr addrs.Ab
 	return diags
 }
 
+func resourceTypesDiffer(a, b addrs.AbsResourceInstance) bool {
+	return a.Resource.Resource.Type != b.Resource.Resource.Type || a.Resource.Resource.ImpliedProvider() != b.Resource.Resource.ImpliedProvider()
```
cam72cam, line 513: I gave you some incorrect feedback on modifying this function.  Type has the provider as a prefix and the original comparison was valid.  The ImpliedProvider call is not needed.

cam72cam:
Here's my investigation into the callers of `ResourceInstanceObjectSrc.Decode`:

* `NodeResourceAbstractInstance.readResourceInstanceState` (and deposed sibling)
  - Expected location to perform this work (similar to upgrade)
  - Tests should ensure that deposed instances are treated correctly
* `evaluateStateData.GetResource` 
  - Explicit comment that state upgrade should have already happened by this point (same as move)
* `Context.driftedResources`
  - In my testing, we hit a warning as the schema may not match between the from and to
  - driftedResources is "best effort"
  - I don't know if this is an issue or not
* `jsonstate/marshalResources`
  - I'm pretty sure this is a non-issue as the migration should have happened by now

internal/tofu/upgrade_resource_state.go
```diff
@@ -18,14 +18,78 @@ import (
...
 // upgradeResourceState will, if necessary, run the provider-defined upgrade
 // logic against the given state object to make it compliant with the
-// current schema version. This is a no-op if the given state object is
-// already at the latest version.
+// current schema version.
 //
 // If any errors occur during upgrade, error diagnostics are returned. In that
 // case it is not safe to proceed with using the original state object.
-func upgradeResourceState(addr addrs.AbsResourceInstance, provider providers.Interface, src *states.ResourceInstanceObjectSrc, currentSchema *configschema.Block, currentVersion uint64) (*states.ResourceInstanceObjectSrc, tfdiags.Diagnostics) {
+func upgradeResourceState(addr, prevAddr addrs.AbsResourceInstance, provider providers.Interface, objectSrc *states.ResourceInstanceObjectSrc, currentSchema *configschema.Block, currentVersion uint64) (*states.ResourceInstanceObjectSrc, tfdiags.Diagnostics) {
```
ollevche, line 31: nit: I would prefer to introduce a separate struct here to replace these 6 func args

internal/tofu/node_resource_abstract.go
```diff
@@ -533,11 +536,16 @@ func (n *NodeAbstractResourceInstance) readResourceInstanceState(ctx EvalContext
 		// Shouldn't happen since we should've failed long ago if no schema is present
 		return nil, diags.Append(fmt.Errorf("no schema available for %s while reading state; this is a bug in OpenTofu and should be reported", addr))
 	}
-	src, upgradeDiags := upgradeResourceState(addr, provider, src, schema, currentVersion)
+
+	if prevAddr := n.prevRunAddr(ctx); resourceTypesDiffer(addr, prevAddr) {
```
ollevche, line 540: nit: I've got a bit confused here. It looks like we are trying to move resource state if the types are different automatically. However this is not the case since types could be different only if there is a `moved` block in the configuration. I would prefer a comment here with a short explanation.

ollevche (commented):
Thank you @Gogotchuri! The PR looks good, the idea to reuse most of the upgrade logic is absolutely valid and well implemented 👍  

Regarding the assumption of `ResourceInstanceObjectSrc.Decode` - I reviewed all the calls and I don't think there is a problem with that. I was uncertain about the `evaluationStateData.GetResource`, however the comment there seems valid.

I am not approving, just because this is a draft PR. Happy to re-review once it is ready!

internal/tofu/node_resource_abstract.go
Gogotchuri: You are correct. This automatically detects type differences and moves if the types don't match. We could cross-check with the move list, which should already be generated by this point. I'll try to do that; it's probably better to be explicit about it. Good catch!

internal/tofu/upgrade_resource_state.go
Gogotchuri: Yeah, I was thinking the same, 6 arguments are too much, and long, will do.

internal/tofu/node_resource_abstract.go
ollevche: I don't think we need to explicitly check that. `prevRunAddr` internally checks move table, which is based on the moved blocks:

https://github.com/opentofu/opentofu/blob/4936a893217a6ceeea4ba2c8837fbf0112038757/internal/tofu/node_resource_abstract_instance.go#L2749-L2752

It is just something unobvious from the first sight so the comment would be helpful.

internal/refactoring/move_validate.go
```diff
@@ -160,18 +160,6 @@ func ValidateMoves(stmts []MoveStatement, rootCfg *configs.Config, declaredInsts
 					StmtRange: stmt.DeclRange,
 				})
 			}
-
-			// Resource types must match.
-			if resourceTypesDiffer(absFrom, absTo) {
```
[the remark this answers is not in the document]
Gogotchuri: Done

internal/tofu/node_resource_abstract.go
Gogotchuri: Changed it, also added a condition to detect if the resource has been moved

internal/tofu/upgrade_resource_state.go
Gogotchuri: Refactored this part, and wrapped transform arguments in a separate function, so we now have 3 args instead of six.
I think it is better for readability this way.

internal/tofu/node_resource_abstract.go
Gogotchuri: Done

internal/tofu/node_resource_abstract.go
```diff
@@ -509,22 +509,26 @@ func (n *NodeAbstractResource) writeResourceState(ctx EvalContext, addr addrs.Ab
 	return diags
 }
 
-func resourceTypesDiffer(a, b addrs.AbsResourceInstance) bool {
-	return a.Resource.Resource.Type != b.Resource.Resource.Type || a.Resource.Resource.ImpliedProvider() != b.Resource.Resource.ImpliedProvider()
+func resourceTypeMoved(evalCtx EvalContext, newAddr, oldAddr addrs.AbsResourceInstance) bool {
```
ollevche: nit: I would prefer a bit longer, but more descriptive name: `isResourceMovedToDifferentType`

ollevche:
Agree with @cam72cam on the drifted resources - we need to ensure the messaging is understandable from user perspective.

Gogotchuri:
> Agree with @cam72cam on the drifted resources - we need to ensure the messaging is understandable from user perspective.

Could you provide more details? What messaging is not understandable currently? @ollevche

internal/provider-simple/provider.go
```diff
@@ -70,6 +70,16 @@ func (s simple) ValidateDataResourceConfig(req providers.ValidateDataResourceCon
 	return resp
 }
 
+func (s simple) MoveResourceState(req providers.MoveResourceStateRequest) (resp providers.MoveResourceStateResponse) {
```
ollevche, line 73: nit: this function seems to be unused

internal/tofu/node_resource_abstract.go
```diff
@@ -533,11 +536,31 @@ func (n *NodeAbstractResourceInstance) readResourceInstanceState(ctx EvalContext
...
+	} else {
+		src, diags = upgradeResourceState(stateTransformArgs{
+			currentAddr:          addr,
+			prevAddr:             prevAddr,
+			provider:             provider,
+			objectSrc:            src,
+			currentSchema:        schema,
+			currentSchemaVersion: currentVersion,
+		})
+	}
```
ollevche, line 559: nit: args could be a separate variable to reduce duplication

internal/tofu/node_resource_abstract_test.go
```diff
@@ -230,7 +234,8 @@ func TestNodeAbstractResource_ReadResourceInstanceState(t *testing.T) {
 
 			ctx.ProviderProvider = providers.Interface(mockProvider)
 
-			got, readDiags := test.Node.readResourceInstanceState(ctx, test.Node.NodeAbstractResource.Addr.Resource.Instance(addrs.NoKey).Absolute(addrs.RootModuleInstance))
+			//TODO same change as line 220, please double check
```
ollevche, line 237: Those addr changes look good to me: code under test should behave the same way, and manual addresses match the state.

ollevche (commented):
The PR looks good, however lacks some more tests specific to the feature itself. Left a few more nit suggestions.

Regarding the messaging for drift changes - from the `driftedResources` code and what @cam72cam wrote, I assume we generate warnings about changed schemas, however it looks expected to me thus being misleading for the end user. I tried to reproduce the warning, however gpg provider is not the best one to simulate drifting.

internal/provider-simple/provider.go
Gogotchuri: Implemented just in case. Since it was needed for interface compatibility and the implementation is so simple.

Gogotchuri:
@ollevche You are correct, we are lacking tests even for the pre-existing state upgrade part. I'm working on them.
@cam72cam, a little explanation of how to reproduce the unclear warning you are getting would help.

cam72cam:
I believe what I did to hit the warnings block is try to move a "tfcoremock_simple_resource" into a gpg_key_pair from "terraform-provider-gpg/gpg".  It also produced a clear error about a invalid move target.

internal/tofu/upgrade_resource_state.go
```diff
@@ -18,82 +18,142 @@ import (
...
-		// uint64 instead of int64, but unsigned integers aren't friendly
-		// to all protobuf target languages so in practice we use int64
-		// on the wire. In future we will change all of our internal
-		// representations to int64 too.
-		Version: int64(src.SchemaVersion),
-	}
-
-	if stateIsFlatmap {
-		req.RawStateFlatmap = src.AttrsFlat
+	if args.objectSrc.SchemaVersion != args.currentSchemaVersion {
```
cam72cam, line 150: I'm not sure this check makes sense for the moved path

Gogotchuri: Done, move into upgrateStateTranform

internal/tofu/upgrade_resource_state.go
```diff
@@ -102,29 +162,29 @@ func upgradeResourceState(addr addrs.AbsResourceInstance, provider providers.Int
...
 	// anyway for robustness, e.g. for in-process providers.
-	newValue := resp.UpgradedState
-	if errs := newValue.Type().TestConformance(currentSchema.ImpliedType()); len(errs) > 0 {
+	if errs := newState.Type().TestConformance(args.currentSchema.ImpliedType()); len(errs) > 0 {
 		for _, err := range errs {
 			diags = diags.Append(tfdiags.Sourceless(
 				tfdiags.Error,
 				"Invalid resource state upgrade",
-				fmt.Sprintf("The %s provider upgraded the state for %s from a previous version, but produced an invalid result: %s.", providerType, addr, tfdiags.FormatError(err)),
+				fmt.Sprintf("The %s provider upgraded the state for %s from a previous version, but produced an invalid result: %s.", providerType, args.currentAddr, tfdiags.FormatError(err)),
```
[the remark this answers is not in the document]
Gogotchuri: Changed some wording

internal/tofu/node_resource_abstract_test.go
```diff
@@ -178,7 +180,17 @@ func TestNodeAbstractResourceSetProvider(t *testing.T) {
 	}
 }
 
-func TestNodeAbstractResource_ReadResourceInstanceState(t *testing.T) {
+type readResourceInstanceStateTest struct {
```
cam72cam, line 183: We usually define these inline

internal/tofu/context_plan2_test.go
```diff
@@ -4289,7 +4780,7 @@ resource "test_object" "b" {
 				},
 			},
 			ResourceTypes: map[string]providers.Schema{
-				"test_object": providers.Schema{
```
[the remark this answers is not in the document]
Gogotchuri: It was a redundant type and the IDE auto-corrected it.
I should revert it since it is unrelated.

internal/tofu/context_plan2_test.go
```diff
@@ -1391,6 +1391,497 @@ func TestContext2Plan_movedResourceBasic(t *testing.T) {
 	})
 }
 
+func customTestSchema(attrs map[string]*configschema.Attribute) providers.Schema {
```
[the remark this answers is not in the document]
Gogotchuri: Renamed

internal/provider-simple-v6/provider.go
```diff
@@ -71,8 +71,12 @@ func (s simple) ValidateDataResourceConfig(req providers.ValidateDataResourceCon
 	return resp
 }
 
-func (p simple) UpgradeResourceState(req providers.UpgradeResourceStateRequest) (resp providers.UpgradeResourceStateResponse) {
-	ty := p.schema.ResourceTypes[req.TypeName].Block.ImpliedType()
+func (s simple) MoveResourceState(_ providers.MoveResourceStateRequest) providers.MoveResourceStateResponse {
+	panic("not implemented")
```
cam72cam, line 75: Why implement this in provider-simple, but not provider-simple-v6?

internal/refactoring/move_execute.go
```diff
@@ -159,7 +159,9 @@ func ApplyMoves(stmts []MoveStatement, state *states.State) MoveResults {
 							newInst := newAddr.Instance(key)
 							recordOldAddr(oldInst, newInst)
 						}
+
```
cam72cam, line 162: whitespace

internal/tofu/upgrade_resource_state.go
```diff
@@ -18,82 +18,143 @@ import (
...
+		// We only do state transformation for managed resources.
+		// This was a part of the normal workflow in older versions and
+		// returned early, so we are only going to log the error for now.
+		log.Printf("[ERROR] data resource %s should not require state transformation", args.currentAddr)
+		return args.objectSrc, nil
+	}
+
+	// TODO: This should eventually use a proper FQN.
+	providerType := args.currentAddr.Resource.Resource.ImpliedProvider()
+	if args.objectSrc.SchemaVersion > args.currentSchemaVersion {
```
cam72cam, line 144: I don't think this check makes sense for moved.  The schema version between resources/providers during a move has no bearing here.  This should be moved to the upgrade case only

internal/tofu/upgrade_resource_state.go
```diff
@@ -102,29 +163,29 @@ func upgradeResourceState(addr addrs.AbsResourceInstance, provider providers.Int
...
-	newValue := resp.UpgradedState
-	if errs := newValue.Type().TestConformance(currentSchema.ImpliedType()); len(errs) > 0 {
+	if errs := newState.Type().TestConformance(args.currentSchema.ImpliedType()); len(errs) > 0 {
 		for _, err := range errs {
 			diags = diags.Append(tfdiags.Sourceless(
 				tfdiags.Error,
-				"Invalid resource state upgrade",
-				fmt.Sprintf("The %s provider upgraded the state for %s from a previous version, but produced an invalid result: %s.", providerType, addr, tfdiags.FormatError(err)),
+				"Invalid resource state transformation",
+				fmt.Sprintf("The %s provider changed the state for %s from a previous version, but produced an invalid result: %s.", providerType, args.currentAddr, tfdiags.FormatError(err)),
```
cam72cam, line 171: This message is still a bit off.  It refers to "from a previous version", we could probably just drop that clause.

Gogotchuri: Yeah, that seems better

internal/tofu/upgrade_resource_state.go
```diff
@@ -18,82 +18,142 @@ import (
...
 
 		// TODO: The internal schema version representations are all using
 		// uint64 instead of int64, but unsigned integers aren't friendly
 		// to all protobuf target languages so in practice we use int64
 		// on the wire. In future we will change all of our internal
 		// representations to int64 too.
-		Version: int64(src.SchemaVersion),
+		//nolint:gosec // this will be refactored eventually
+		Version:         int64(args.objectSrc.SchemaVersion),
+		RawStateFlatmap: args.objectSrc.AttrsFlat,
```
cam72cam, line 102: What happened to checking stateIsFlatMap from the original function?

cam72cam (approved):

cam72cam:
Last thought is adding a comment to provider interface moved call.  We currently call it on a configured provider for convenience, but that is not a long term guarantee and may differ from other tooling.

internal/tofu/upgrade_resource_state.go
Gogotchuri: Removed due to redundancy of the check.
But for the sake of backward compatibility, let's keep it.

cam72cam (approved):

ollevche (dismissed):
Looks good, thanks!

I suggest to extend `moved` documentation page to include newly added use-case ([link](https://opentofu.org/docs/language/modules/develop/refactoring/#moved-block-syntax)).

ollevche (approved):
Thanks!

PR #2411, opened 2025-01-22
make pg backend acquire schema-based global locks

### LINKED ISSUE #2218: pg backend new state causes lock conflicts
### OpenTofu Version

```shell
OpenTofu v1.7.1
on linux_amd64

I am aware this is an old version, but the backend has not changed since as you will see with the provided lines later on.
```

```hcl
terraform {
  backend "pg" {}
  required_version = "~> 1.7"
}
```

Any module using the postgres backend will hit this problem

### Debug Output

N/A

### Expected Behavior

The pg backend should not hold a global lock when trying to create a new workspace. One unrelated module should not affect another entirely unrelated module with a different schema from running.

### Actual Behavior

When using the same postgres database for multiple states, if any state is new and needs to be created this will block any other module that uses the same backend from running, even if it has a different schema name. This is incredibly painful when working with ephemeral environment type of setups, as a single new environment creates a new schema and can cause cascading failures of any other modules that request a lock during that time. This also has significant impacts with terragrunt run-all for example when planning or applying multiple fresh modules.

This behaviour makes the pg backend basically unusable in any kind of scaled setup that creates new ephemeral environments with new states. A worst-case example would be a PR gets created and spins up a new ephemeral environment which creates a new state in the postgres backend. At the same moment, a production deployment tries to check the advisory lock is available and fails, causing the production deployment to fail. These components may be entirely unrelated to eachother, the only thing they share is the same postgres database.

### Steps to Reproduce

1. Acquire advisory lock -1 in postgres `SELECT pg_try_advisory_lock(-1)`
2. tofu apply - will receive the error "Cannot lock workspace; already locked for workspace creation: default"
3. Release advisory lock in postgres
4. tofu apply - will work as expected

### Additional Context

We are running our tofu via terragrunt run-all which exagerates this problem, but even without that it's still a problem in a scaled environment. We would like to avoid moving to another state provider as postgres works well for us except for this problem.

### References

https://github.com/opentofu/opentofu/blob/main/internal/backend/remote-state/pg/client.go#L94
https://github.com/opentofu/opentofu/blob/main/internal/backend/remote-state/pg/client.go#L101

ollevche:
Resolves #2218

This PR makes `pg` backend to take schema-based locks for workspace creation.

During the implementation I also noticed pg tests are not cleaning up properly, so I reviewed the tests and the implementation to ensure we quote schema name properly everywhere (it only affected test code).

[Rendered version of the RFC](https://github.com/opentofu/opentofu/blob/pg-locks/rfc/20250127-PostgreSQL-Locking.md)

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/backend/remote-state/pg/client.go
```diff
@@ -90,15 +91,19 @@ func (c *RemoteClient) Lock(info *statemgr.LockInfo) (string, error) {
...
+		statesTableName, statesTableName, creationLockID, c.SchemaName, statesTableName, statesTableName)
+
+	row := c.Client.QueryRow(query, c.Name)
 	var pgLockId, didLock, didLockForCreate []byte
 	err = row.Scan(&pgLockId, &didLock, &didLockForCreate)
 	switch {
 	case err == sql.ErrNoRows:
 		// No rows means we're creating the workspace. Take the creation lock.
-		innerRow := c.Client.QueryRow(`SELECT pg_try_advisory_lock(-1)`)
+		innerRow := c.Client.QueryRow(fmt.Sprintf(`SELECT pg_try_advisory_lock(%s)`, creationLockID))
```
cam72cam, line 106: It's an integer we compute and is probably safe regardless, but could we use $1 here?

cam72cam:
LGTM

internal/backend/remote-state/pg/client.go
ollevche: This follows existing `pg` backend code style / approach. All the tofu generated values are formatted and user passed values are parameterized. I would either change it everywhere or leave it as is.

Given that we control the value, it is safe either way (from the SQL injection perspective).

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 UPGRADE NOTES:
 
 * Using the `ghcr.io/opentofu/opentofu` image as a base image for custom images is no longer supported. Please see https://opentofu.org/docs/intro/install/docker/ for instructions on building your own image.
+* OpenTofu 1.10 with `pg` backend must not be used in paralel with older versions. It may lead to unsafe state writes, when the database is shared across multiple projects.
```
cam72cam, line 6: parallel

CHANGELOG.md
```diff
@@ -22,6 +23,7 @@ BUG FIXES:
 - When assigning an empty map to a variable that is declared as a map of an object type with at least one optional attribute, OpenTofu will no longer create a subtly-broken value. ([#2371](https://github.com/opentofu/opentofu/pull/2371))
 - The `format` and `formatlist` functions can now accept `null` as one of the arguments without causing problems during the apply phase. Previously these functions would incorrectly return an unknown value when given `null` and so could cause a failure during the apply phase where no unknown values are allowed. ([#2371](https://github.com/opentofu/opentofu/pull/2371))
 - Provider used in import is correctly identified. ([#2336](https://github.com/opentofu/opentofu/pull/2336))
+- `pg` backend doesn't fail on workspace creation for paralel runs, when the database is shared across multiple projects. ([#2411](https://github.com/opentofu/opentofu/pull/2411))
```
cam72cam, line 26: parallel

ollevche: [b81cd1b](https://github.com/opentofu/opentofu/pull/2411/commits/b81cd1b23af1a7cd30423d34cff39b15ec92aa84)

rfc/20250127-PostgreSQL-Locking.md
```diff
@@ -0,0 +1,57 @@
...
+option. Additional attention needs to be paid to overflows, since `pg_advisory_lock` uses 64-bit signed integers as IDs
+and we want to generate a hash, which fits into that range.
+
+Since `pg_advisory_lock` is session scoped we don't need to worry about the migration or breaking changes. Problem may arise
+if users run paralel `tofu apply` for the same backend with different versions of OpenTofu (i.e. one run acquires `-1` lock
+and the other one acquires the hash of the same schema name). In this scenario, we end up with unsafe database writes.
+
+### Open Questions
+
+* Do we need to put an additional safe guard for unsafe writes described previously? Is the release notes warning enough?
```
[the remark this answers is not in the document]
ollevche: Put a warning in [3f82850](https://github.com/opentofu/opentofu/pull/2411/commits/3f82850055487a63a7fb880bea4dec033c88b800)

cam72cam (dismissed):

website/docs/language/settings/backends/pg.mdx
```diff
@@ -9,6 +9,12 @@ Stores the state in a [Postgres database](https://www.postgresql.org) version 10
 
 This backend supports [state locking](../../../language/state/locking.mdx).
 
+:::warning
+OpenTofu 1.10 changed how workspace creation locking works. It is unsafe to
+use `pg` backend with multiple OpenTofu versions (under 1.10) in parallel if those
+are sharing the same underlying database.
+:::
```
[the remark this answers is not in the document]
ollevche: [57a9c36](https://github.com/opentofu/opentofu/pull/2411/commits/57a9c364fa09cb8cd81c589f5953dc97ef13588d)

cam72cam (dismissed):

cam72cam (approved):

Issue #2162, opened 2024-11-13
PG Backend: Allow customizing name of created table or allow setting an index pre-/suffix


mlhynfield:
### OpenTofu Version

```shell
OpenTofu v1.8.5
on darwin_arm64
```

### The problem in your OpenTofu project

As stated in the referenced issue from hashicorp/terraform, there is currently no clean solution (that I have found) for handling multiple projects with the `pg` backend in OpenTofu either.

### Attempted Solutions

The referenced issue also mentions using unique workspaces for **every** distinct project, which can get rather tedious and may even be impossible if #2160 deprecates/removes workspaces.

### Proposal

I agree with the OP from the referenced issue that the best solution would be to add an option to customize the table name via attribute and environment variable (and possibly even make this a required configuration similar to `key` in the [s3 backend](https://opentofu.org/docs/language/settings/backends/s3/#s3-state-storage)).

### References

- [hashicorp/terraform #35320](https://github.com/hashicorp/terraform/issues/35320)

mlhynfield:
Closing because I realized the `pg` backend is maintained by other contributors.

ollevche:
Hey @mlhynfield! `pg` backend is a part of the OpenTofu codebase so it is maintained by the OpenTofu core team.

Feel free to re-open the issue if you think it is still relevant.

PR #2465, opened 2025-02-01
Make table name and index name a config value for PG backend

taliesins:
Make it possible to specify the table name and index name when using PG backend. 

This makes it easier to use PG backend provider for scenarios such as using Terragrunt with convention based state storage per a `terragrunt.hcl` file by providing an additional layer of grouping.

Having many tables vs many schemas could be preferred by some users and the tools they use too.

### DIFF
--- a/internal/backend/remote-state/pg/backend.go
+++ b/internal/backend/remote-state/pg/backend.go
@@ -9,18 +9,13 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"github.com/lib/pq"
 	"os"
 	"strconv"
 
-	"github.com/lib/pq"
 	"github.com/opentofu/opentofu/internal/backend"
 	"github.com/opentofu/opentofu/internal/encryption"
 	"github.com/opentofu/opentofu/internal/legacy/helper/schema"
-)
-
-const (
-	statesTableName = "states"
-	statesIndexName = "states_by_name"
 )
 
 func defaultBoolFunc(k string, dv bool) schema.SchemaDefaultFunc {
@@ -58,11 +53,25 @@ func New(enc encryption.StateEncryption) backend.Backend {
 				DefaultFunc: defaultBoolFunc("PG_SKIP_SCHEMA_CREATION", false),
 			},
 
+			"table_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Name of the automatically managed Postgres table to store state",
+				DefaultFunc: schema.EnvDefaultFunc("PG_TABLE_NAME", "states"),
+			},
+
 			"skip_table_creation": {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Description: "If set to `true`, OpenTofu won't try to create the Postgres table",
 				DefaultFunc: defaultBoolFunc("PG_SKIP_TABLE_CREATION", false),
+			},
+
+			"index_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Name of the automatically managed Postgres index used for stored state",
+				DefaultFunc: schema.EnvDefaultFunc("PG_INDEX_NAME", "states_by_name"),
 			},
 
 			"skip_index_creation": {
@@ -88,6 +97,8 @@ type Backend struct {
 	configData *schema.ResourceData
 	connStr    string
 	schemaName string
+	tableName  string
+	indexName  string
 }
 
 func (b *Backend) configure(ctx context.Context) error {
@@ -96,7 +107,12 @@ func (b *Backend) configure(ctx context.Context) error {
 	data := b.configData
 
 	b.connStr = data.Get("conn_str").(string)
-	b.schemaName = pq.QuoteIdentifier(data.Get("schema_name").(string))
+	b.schemaName = data.Get("schema_name").(string)
+	b.tableName = data.Get("table_name").(string)
+	b.indexName = data.Get("index_name").(string)
+	skipSchemaCreation := data.Get("skip_schema_creation").(bool)
+	skipTableCreation := data.Get("skip_table_creation").(bool)
+	skipIndexCreation := data.Get("skip_index_creation").(bool)
 
 	db, err := sql.Open("postgres", b.connStr)
 	if err != nil {
@@ -106,11 +122,11 @@ func (b *Backend) configure(ctx context.Context) error {
 	// Prepare database schema, tables, & indexes.
 	var query string
 
-	if !data.Get("skip_schema_creation").(bool) {
+	if !skipSchemaCreation {
 		// list all schemas to see if it exists
 		var count int
 		query = `select count(1) from information_schema.schemata where schema_name = $1`
-		if err := db.QueryRow(query, data.Get("schema_name").(string)).Scan(&count); err != nil {
+		if err = db.QueryRow(query, b.schemaName).Scan(&count); err != nil {
 			return err
 		}
 
@@ -119,31 +135,33 @@ func (b *Backend) configure(ctx context.Context) error {
 		// a user hasn't been granted the `CREATE SCHEMA` privilege
 		if count < 1 {
 			// tries to create the schema
-			query = `CREATE SCHEMA IF NOT EXISTS %s`
-			if _, err := db.Exec(fmt.Sprintf(query, b.schemaName)); err != nil {
+			query = fmt.Sprintf(`CREATE SCHEMA IF NOT EXISTS %s`, pq.QuoteIdentifier(b.schemaName))
+			if _, err = db.Exec(query); err != nil {
 				return err
 			}
 		}
 	}
 
-	if !data.Get("skip_table_creation").(bool) {
-		if _, err := db.Exec("CREATE SEQUENCE IF NOT EXISTS public.global_states_id_seq AS bigint"); err != nil {
+	if !skipTableCreation {
+		query = "CREATE SEQUENCE IF NOT EXISTS public.global_states_id_seq AS bigint"
+		if _, err = db.Exec(query); err != nil {
 			return err
 		}
 
-		query = `CREATE TABLE IF NOT EXISTS %s.%s (
+		query = fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s.%s (
 			id bigint NOT NULL DEFAULT nextval('public.global_states_id_seq') PRIMARY KEY,
 			name text UNIQUE,
 			data text
-			)`
-		if _, err := db.Exec(fmt.Sprintf(query, b.schemaName, statesTableName)); err != nil {
+			)`, pq.QuoteIdentifier(b.schemaName), pq.QuoteIdentifier(b.tableName))
+
+		if _, err = db.Exec(query); err != nil {
 			return err
 		}
 	}
 
-	if !data.Get("skip_index_creation").(bool) {
-		query = `CREATE UNIQUE INDEX IF NOT EXISTS %s ON %s.%s (name)`
-		if _, err := db.Exec(fmt.Sprintf(query, statesIndexName, b.schemaName, statesTableName)); err != nil {
+	if !skipIndexCreation {
+		query = fmt.Sprintf(`CREATE UNIQUE INDEX IF NOT EXISTS %s ON %s.%s (name)`, pq.QuoteIdentifier(b.indexName), pq.QuoteIdentifier(b.schemaName), pq.QuoteIdentifier(b.tableName))
+		if _, err = db.Exec(query); err != nil {
 			return err
 		}
 	}
--- a/internal/backend/remote-state/pg/backend_test.go
+++ b/internal/backend/remote-state/pg/backend_test.go
@@ -77,6 +77,31 @@ func TestBackendConfig(t *testing.T) {
 			},
 		},
 		{
+			Name: "valid-config-with-table-name",
+			Config: map[string]interface{}{
+				"conn_str":    connStr,
+				"schema_name": fmt.Sprintf("terraform_%s", t.Name()),
+				"table_name":  fmt.Sprintf("terraform_%s", t.Name()),
+			},
+		},
+		{
+			Name: "valid-config-with-index-name",
+			Config: map[string]interface{}{
+				"conn_str":    connStr,
+				"schema_name": fmt.Sprintf("terraform_%s", t.Name()),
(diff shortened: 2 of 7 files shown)


taliesins:
This solves #2162

ollevche (commented):
Looks good overall, thank you for the contribution! I left a comment regarding the hashing of schema + table combination.

ollevche:
Also, `pg` tests are not being run automatically so I checked them locally and they are failing. Here is part of the output I receive:
```
client_test.go:130: Unsupported attribute: An attribute named "stable_name" is not expected here.
FAIL: TestConcurrentCreationLocksInDifferentSchemas (0.00s)
...
client_test.go:226: Unsupported attribute: An attribute named "stable_name" is not expected here.
FAIL: TestConcurrentCreationLocksInDifferentTables (0.00s)
```

internal/backend/remote-state/pg/client.go
```diff
@@ -150,6 +152,6 @@ func (c *RemoteClient) Unlock(id string) error {
 
 func (c *RemoteClient) composeCreationLockID() string {
 	hash := fnv.New32()
-	hash.Write([]byte(c.SchemaName))
+	hash.Write([]byte(c.SchemaName + "|" + c.TableName))
```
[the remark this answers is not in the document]
taliesins: I selected the pipe "|" character as it can't be used in schema or table names in Postgres. This means we will never land up with the scenario outlined.

[Allowed characters in identifiers](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)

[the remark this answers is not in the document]
taliesins: The only quirk I could think of is collision of index names across tables. So it will require the user to configure the table and the index appropriately.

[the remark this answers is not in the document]
ollevche: `pg` backend uses quoted identifiers, so it could contain any character, including the pipe. From the docs you linked to:

> Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands. The length limitation still applies.

ollevche:
After thinking about the proposal for a while, I believe we are introducing a less consistent UX here. 

Previously, it was possible to set only a schema, locking would work just fine in that case. You can't break it from the user's perspective, even when the underlying database is shared.

Currently, if you set a different index and reuse the underlying database, locking will have lots of per-state collisions, since it uses database-level advisory locking with potentially multiple sets of state IDs (from different indexes).

Users **must** keep a single index per database, otherwise it will not work as intended.

taliesins:
@ollevche I have fixed the tests. Please could you pull and try to run them locally. Most of the awkwardness in fixing the tests was around literal vs identifier values, so some minor changes to not pass them around and apply escaping at the point of usage.

internal/backend/remote-state/pg/client.go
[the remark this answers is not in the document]
taliesins: @ollevche you are correct. Can you suggest another string to use as a separator or are we willing to accept the chance of a collision being highly unlikely?

taliesins:
> After thinking about the proposal for a while, I believe we are introducing a less consistent UX here.

We will only run into this if people explicitly configure the new configuration fields. They default to what was before. So this challenge would only present itself to people looking to explicitly do this.

> Currently, if you set a different index and reuse the underlying database, locking will have lots of per-state collisions, since it uses database-level advisory locking with potentially multiple sets of state IDs (from different indexes).

Are we talking about the usage of `pg_try_advisory_lock()`?

If we are, don't advisory locks exist per database and not a schema level? This means that multiple schemas vs multiple tables will make no difference as multiple schema's is technically multiple databases? So the number of collisions will be the same as before?

Then if we look at the values that are being used to lock on, changing the value we lock on doesn't result in extra lock collisions if choose to store state in 2 different places? 
- current main is `schemaA` vs `schemaB` collisions
- this pr is `schemaA|tableA` vs `schemaA|tableB` collisions

> Users must keep a single index per database, otherwise it will not work as intended.

Are you thinking about the usage of `nextval('public.global_states_id_seq')` to generate ids? If we are using a static namespaces to generate ids across schemas then we make the problem no worse?

From a clients perspective we generate an id for them using `uuid.GenerateUUID()` and generate the lock name idempotently via `composeCreationLockID` so it should make no difference to the client.

ollevche:
The problem I was describing lies in the usage of advisory lock with the state ID produced by the index. The following query does exactly that:

https://github.com/opentofu/opentofu/blob/60fdd359d59915249e04155f6c02806a0206667e/internal/backend/remote-state/pg/client.go#L98-L99

Imagine, a user sets up two different configurations with two different backends, which share the same underlying database, but have distinct schemas / tables and indexes. It is then possible to have multiple states with the same ID from different configurations (backends) since the indexes (sequences) are different.

Let's say we have state A with ID = 1 from the first configuration and state B with ID = 1 from the second configuration. States A and B are distinct, however share the same ID, which will be used to acquire the advisory lock (which is shared across all the schemas, on the database level). This is where additional collisions occur.

> We will only run into this if people explicitly configure the new configuration fields. They default to what was before. So this challenge would only present itself to people looking to explicitly do this.

Yep, I absolutely agree with that statement. What I meant is that now it is easier for user to break the setup by just setting an index name (without knowing about internal locking behavior). In my opinion, it is unsafe to allow users to control this. Also, I am not sure I understand the use-case under that request.

In the original issue, it was requested to reduce the number of schemas (which could be solved by a table name param). I don't see any problem with this, however custom index names could be more harmful, than useful.

taliesins:
@ollevche  I think because we are using a sequence that is shared between all the schemas a change to multiple tables won't effect this. 

The only two paths that set `.Path` are for locking when there isn't a workspace record yet (creation lock) or we want to update an existing workspace (workspace lock). 

I think there is the potential to simplify to a single lock by using "schema"+"|" + "table" + "|" + "workspace" instead, which is what id is basically an alias of. 

What do you think?

pg

ollevche:
Hey @taliesins thanks for the additional context. 

You are totally right. I initially thought the idea was to allow users to specify a sequence as well as the index in the configuration. After re-reading the PR I understand that's not the case. Sorry for the confusion.

> I think there is the potential to simplify to a single lock by using "schema"+"|" + "table" + "|" + "workspace" instead, which is what id is basically an alias of.

Creation locks must be scoped to the entire table so we shouldn't map it to a specific row (workspace), so I believe the current approach is fine.

internal/backend/remote-state/pg/client.go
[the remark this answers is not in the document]
ollevche: I think we can use some weird sequence (e.g. "|<.") to lower the potential of collisions, but the pipe is already something unpopular I believe.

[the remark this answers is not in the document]
taliesins: I changed the seperator to use "\x00" which is the character with code zero. This way there can never be a conflict.

ollevche:
I started the checks and also ensured the tests were passing. We are moving toward automated acceptance tests, including automatically runnable pg tests, so we will eventually fix that.

internal/backend/remote-state/pg/backend.go
```diff
@@ -96,7 +107,42 @@ func (b *Backend) configure(ctx context.Context) error {
 	data := b.configData
 
 	b.connStr = data.Get("conn_str").(string)
-	b.schemaName = pq.QuoteIdentifier(data.Get("schema_name").(string))
+
+	schemaName, ok := data.Get("schema_name").(string)
```
ollevche, line 111: Why did we move quoting from here to every other SQL query?

internal/backend/remote-state/pg/backend.go
```diff
@@ -96,7 +107,42 @@ func (b *Backend) configure(ctx context.Context) error {
...
+	if !ok {
```
ollevche, line 112: I believe it should be already set at this time. The priority list for getting that kind of values is the following:

* configuration
* env variable
* static default value

I think it is safe to leave this code as is (without type conversion checks).

internal/backend/remote-state/pg/backend_test.go
```diff
@@ -205,9 +248,48 @@ func TestBackendConfig(t *testing.T) {
...
 			}
 
-			_, err = b.db.Query(fmt.Sprintf("SELECT name, data FROM %s.%s LIMIT 1", schemaName, statesTableName))
-			if err != nil {
-				t.Fatal(err)
+			// Make sure everything has been created
+			if !skipSchemaCreation {
+				// Make sure schema exists
+				var count int
+				query := fmt.Sprintf(`select count(*) from information_schema.schemata where schema_name=$1`)
```
ollevche, line 255: from the linter: fmt.Sprintf is unnecessary here and a few other places down below

internal/backend/remote-state/pg/client.go
```diff
@@ -150,6 +154,6 @@ func (c *RemoteClient) Unlock(id string) error {
 
 func (c *RemoteClient) composeCreationLockID() string {
 	hash := fnv.New32()
-	hash.Write([]byte(c.SchemaName))
+	hash.Write([]byte(c.SchemaName + "\x00" + c.TableName))
```
ollevche, line 157: nice! it's a good idea to pick that char

website/docs/language/settings/backends/pg.mdx
```diff
@@ -97,7 +97,9 @@ The following configuration options or environment variables are supported:
...
 - `skip_schema_creation` - If set to `true`, the Postgres schema must already exist. Can also be set using the `PG_SKIP_SCHEMA_CREATION` environment variable. OpenTofu won't try to create the schema, this is useful when it has already been created by a database administrator.
+- `table_name` - Name of the automatically-managed Postgres table, default to `states`. Can also be set using the `PG_TABLE_NAME` environment variable.
 - `skip_table_creation` - If set to `true`, the Postgres table must already exist. Can also be set using the `PG_SKIP_TABLE_CREATION` environment variable. OpenTofu won't try to create the table, this is useful when it has already been created by a database administrator.
+- `index_name` - Name of the automatically-managed Postgres index, default to `states_by_name`. Can also be set using the `PG_INDEX_NAME` environment variable.
 - `skip_index_creation` - If set to `true`, the Postgres index must already exist. Can also be set using the `PG_SKIP_INDEX_CREATION` environment variable. OpenTofu won't try to create the index, this is useful when it has already been created by a database administrator.
 
 ## Technical Design
```
ollevche, line 105: Let's put a warning before the technical design section to remind users OpenTofu is not going to do any migrations:

> Please, keep in mind, that if the data-related configuration changes (such as `table_name` or `schema_name`), you would need to manually migrate the existing data.

internal/backend/remote-state/pg/backend_test.go
```diff
@@ -131,6 +141,8 @@ func TestBackendConfig(t *testing.T) {
 			Config: map[string]interface{}{
 				"conn_str":    connStr,
 				"schema_name": fmt.Sprintf("terraform_%s", t.Name()),
+				"table_name":  fmt.Sprintf("terraform_%s", t.Name()),
+				"index_name":  fmt.Sprintf("terraform_%s", t.Name()),
```
ollevche, line 145: Let's leave a few tests, where we don't set `table_name` and `index_name`

ollevche (commented):
Hey, @taliesins! Sorry for delay from our side on this one.

The PR looks great, I left a few questions and suggestions. Could you please also fix a failed linting check?

internal/backend/remote-state/pg/backend.go
taliesins: Keeping track of variables that may or may not be quoted become hard to track and in some instances I needed the unquoted value. Originally I started a parallel set of variables, but it looked odd to pass in the same variable with a quoted and non quoted version. 

Moving it closer to sql query made it much easier to understand. The only drawback being we might have to quote the same variable more then once.

website/docs/language/settings/backends/pg.mdx
taliesins: If the value is not applied it will fall back to the previous hard coded version, which means it should be 100% backward compatible. Am I missing something?

internal/backend/remote-state/pg/backend_test.go
taliesins: Updated. It should pass golangci-lint now.

internal/backend/remote-state/pg/backend.go
taliesins: Fixed

internal/backend/remote-state/pg/backend_test.go
taliesins: Done! This required me to rework the way test gets config values.

I have added extra test cases:
- valid-config (nothing set)
- valid-config-with-table-name
- valid-config-with-index-name
- valid-config-with-table-name-and-index-name

internal/backend/remote-state/pg/backend.go
taliesins: I am going to close this for now. Feel free to re-open.

cam72cam:
@taliesins sorry for the delay, running now

internal/backend/remote-state/pg/backend_test.go
```diff
@@ -307,47 +410,77 @@ func TestBackendConfigSkipOptions(t *testing.T) {
...
+
+				query := `select count(*) from pg_indexes where schemaname=$1 and tablename=$2 and indexname=$3;`
+				err = b.db.QueryRow(query, schemaName, tableName, indexName+"_name_key").Scan(&count)
+				if err != nil {
 					t.Fatal(err)
 				}
+
 				if count != 1 {
-					t.Fatalf("The index has not been created (%d)", count)
+					t.Fatalf("The index has not been c reated (%d)", count)
```
cam72cam, line 483: small typo `created`

cam72cam:
@taliesins can you add a changelog entry?  Once that's in, I think we are good to merge.

internal/backend/remote-state/pg/backend_test.go
taliesins: Done! I also fixed a typo in the change log.

taliesins:
@cam72cam if you could approve the workflow.

cam72cam (approved):

cam72cam:
Thanks for the submission and your patience @taliesins !

PR #2471, opened 2025-02-04
update upload and download artifacts gha

ollevche:
Half of the checks on main branch are failing because of the [deprecated GitHub actions: upload-artifact and download-artifact](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). This PR updates their versions.

## Target Release

1.10.0

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
           version: ${{ needs.get-product-version.outputs.product-version }}
           product: ${{ env.PKG_NAME }}
 
-      - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+      - uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: metadata.json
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -253,7 +253,7 @@ jobs:
           fail-on-cache-miss: true
           enableCrossOsArchive: true
       - name: "Download OpenTofu CLI package"
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
```
yottta, line 256: `nitpicking`
Why not using directly `@v4.1.8?`
https://github.com/actions/download-artifact/releases/tag/v4.1.8

Same for the others

🤔

yottta (approved):

.github/workflows/build.yml
ollevche: Honestly, I just followed an existing approach. Theoretically, it could be re-tagged to point to some other commit and then break the CI, but I believe this is pretty low risk. If you don't mind, I will merge this as is :)

PR #2479, opened 2025-02-05
Add a new warning when a provider cannot be downloaded and it was requested by an implicit usage

### LINKED ISSUE #2084: `tofu init` should warn if a provider exists with a prefix
### OpenTofu Version

```shell
All
```

We are getting a large number of reports that `registry.opentofu.org/hashicorp/X` doesn't exist, where `X` is a non-HashiCorp provider. This is often a configuration error where a `required_providers` entry exists, but isn't used by defining the correct alias.

### Attempted Solutions

Help users on a case-by-case basis.

### Proposal

OpenTofu should warn users if it can't find a provider on `tofu init` which exists in the configuration under a different namespace.

### References

_No response_

yottta:
Fixes #2084

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/command/init.go
```diff
@@ -1039,6 +1042,34 @@ version control system if they represent changes you intended to make.`))
...
+	if _, ok := qualifs.Explicit[provider]; !ok {
+		var refFromHint string
+		if refs, ok := qualifs.Implicit[provider]; ok && len(refs) > 0 {
+			refFromHint = fmt.Sprintf("Implicitly referenced from %s.", refs[0].String())
+			if noOfTotalRes := len(refs) - 1; noOfTotalRes > 0 {
+				refFromHint += fmt.Sprintf(" And is referenced from %d more resource(s).", noOfTotalRes)
+			}
+		}
+		details := fmt.Sprintf(`The provider %s could not be found in any required_providers blocks. Ensure that the root or any used module is having the required_providers block properly configured. 
+Or if the definition is already there, ensure that the resources are having 'provider' configured properly.`,
```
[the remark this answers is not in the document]
yottta: Thanks for the comment.
I am planning on attaching some screenshots to the PR after this is done. I still have some things to clear out before doing that.

internal/command/e2etest/init_test.go
```diff
@@ -502,6 +502,39 @@ func TestInitProviderNotFound(t *testing.T) {
 			t.Errorf("wrong output:\n%s", cmp.Diff(stripAnsi(stderr), expectedErr))
 		}
 	})
+
+	t.Run("implicit provider not found", func(t *testing.T) {
```
yottta, line 506: To help with the reviewing process, here is a screenshot of the expected output of this test:
image

What I tried here is to gather the in-file references and use those for an easier navigation.
For sure there could be arguments against this, and I would be more than happy to trigger a conversation around this subject.

internal/command/init.go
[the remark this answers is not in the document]
yottta: Please refer to [this](https://github.com/opentofu/opentofu/pull/2479/files#r1944686883) comment.

yottta:
Some questions:
* Should we collect the providers from the tests as well? [This](https://github.com/opentofu/opentofu/blob/657334c2f8542b66d3144bcbab6bcda95e5e7965/internal/configs/config.go#L530-L547) part.
* From my knowledge and general understanding of the `provider` block, we shouldn't consider that. But maybe there is a legacy behaviour for which we should [collect the providers as implicit references too](https://github.com/opentofu/opentofu/blob/657334c2f8542b66d3144bcbab6bcda95e5e7965/internal/configs/config.go#L523-L526)?

internal/configs/config_test.go
```diff
@@ -195,7 +224,7 @@ func TestConfigProviderRequirementsInclTests(t *testing.T) {
 	terraformProvider := addrs.NewBuiltInProvider("terraform")
 	configuredProvider := addrs.NewDefaultProvider("configured")
 
-	got, diags := cfg.ProviderRequirements()
+	got, qualifs, diags := cfg.ProviderRequirements() // TODO: Question: should this handle also the test providers?
```
ollevche, line 227: I don't think we need to gather providers defined in test files. If we have an invalid configuration (from the perspective of this particular issue), then test files don't matter much, otherwise - they must not be able to "break" provider references.

ollevche (commented):
Looks good to me, thanks!

internal/command/e2etest/testdata/provider-implicit-ref-not-found/mod/main.tf
```diff
@@ -0,0 +1,2 @@
+resource "nonexistingProv_res" "test2" {
+}
```
apparentlymart, line 2: This test fixture does seem to cover the most common case, and I'd agree that's what we should focus on here.

I think it would also be worth having tests that this warning _does not_ get activated in two (far less common!) situations that are similar but for which the guidance in the warning message could be misleading:

- When there is an explicit `provider` argument that refers to a provider local name that hasn't been declared in `required_providers` but doesn't match any provider in the default namespace.

    The most likely reason for this one is a typo, like typing `provider = asw.foo` instead of `provider = aws.foo`, which is another situation we _could_ generate a warning for but in my experience doesn't come up often enough to be worth the implementation/maintenance cost.
- When there is a `provider` block in a module whose label is a local name that isn't in `required_providers` but there are no resource blocks referring to it.

    This one could also be caused by a typo in the provider block -- `provider "asw"` instead of "provider "aws"` -- but it can also arise if someone is still working on their module and has so far only written the `provider` blocks, and hasn't yet written any `resource` or `data` blocks. Again, we _could_ generate a warning for that case, but it's relatively uncommon.

In both of these cases my main motivation is just to make sure we avoid showing a warning message that isn't exactly relevant to the situation we've detected, since the warning we're generating here is already covering three possible causes and so I think we need to make sure that we only return that warning when we think it's very likely to be at least one of those three problems.

internal/getproviders/types.go
```diff
@@ -50,6 +51,35 @@ type Warnings = []string
...
 type Requirements map[addrs.Provider]VersionConstraints
 
+// ProvidersQualification is storing the implicit/explicit reference qualification of the providers.
+// This is necessary to be able to warn the user when the resources are referencing a provider that
+// is not specifically defined in a required_providers block. When the implicitly referenced
+// provider is tried to be downloaded without a specific provider requirement, it will be tried
+// from the default namespace (hashicorp), failing to download it when it does not exist in the default namespace.
+// Therefore, we want to let the user know what resources are generating this situation.
+type ProvidersQualification struct {
+	Implicit map[addrs.Provider][]hcl.Range
```
apparentlymart, line 61: I would only care about this if there isn't already a dependency from this package on HCL, but if this package is _only_ depending on HCL to get `hcl.Range` then I might suggest using [`tfdiags.SourceRange`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/tfdiags#SourceRange) instead, which is what we tend to use to represent source ranges in situations where we're not otherwise working _directly_ with HCL's types.

There are conversion functions in both directions between `hcl.Range` and `tfdiags.SourceRange`, so using it tends to just cause a few extra conversion calls. A common place we've drawn this line in the past is that `package configs` primarily uses HCL's types (since its overall job is to map generic HCL stuff to OpenTofu-domain-specific-concepts) while most other packages prefer to use `tfdiags` types. That's more a guideline than a rule, though.

apparentlymart (commented):
Overall this approach makes sense to me. I left a few notes inline.

Unfortunately the one about changing the diagnostic message would, if you agree with it, require retaining more details in the `ProvidersQualification` data structure, but I think it should all be information that we know at the point of calling the `AddImplicitProvider` method.

I wanted to send this early feedback now just to avoid keeping you waiting. I think it would be a good exercise to try various real situations from [the log of old questions I made in the issue](https://github.com/opentofu/opentofu/issues/2084#issuecomment-2581046767) and see if we think the error message we're returning communicates the similar information that helped each of those folks solve their problem, and I'd be happy to help with that testing but my focus is somewhere else today so I can't do that immediately.

internal/command/init.go
```diff
@@ -1039,6 +1042,33 @@ version control system if they represent changes you intended to make.`))
...
+		var refFromHint string
+		if refs, ok := qualifs.Implicit[provider]; ok && len(refs) > 0 {
+			refFromHint = fmt.Sprintf("Implicitly referenced from %s.", refs[0].String())
+			if noOfTotalRes := len(refs) - 1; noOfTotalRes > 0 {
+				refFromHint += fmt.Sprintf(" And is referenced from %d more resource(s).", noOfTotalRes)
+			}
+		}
+		details := fmt.Sprintf(`The provider %s doesn't exist and is referenced by a resource implicitly without a 'required_providers' blocks. This typically happens when you are missing the 'required_providers' block or you are not referencing it correctly using the 'provider=' parameter in your resource or data source block.`,
+			provider.ForDisplay(),
+		)
```
apparentlymart, line 1059: Hmm... GitHub somehow lost this comment the first time, so I'm going to try again. :confounded: 

I'd suggest the following error message text both to try to write this in the same "voice" as other messages OpenTofu produces and to hopefully help users understand how OpenTofu infers provider requirements automatically when none are specified.

```suggestion
		details := fmt.Sprintf(
			"Due to the prefix of the resource type name OpenTofu guessed that you intended to associate %s with a provider whose local name is %q, but that name is not declared in this module's required_providers block. OpenTofu therefore guessed that you intended to use %s, but that provider does not exist.\n\nMake at least one of the following changes to tell OpenTofu which provider to use:\n - Add a declaration for local name %q to this module's required_providers block, specifying the full source address for the provider you intended to use.\n - Verify that %q is the correct resource type name to use. Did you omit a prefix which would imply the correct provider?\n - Use a \"provider\" argument within this resource block to override OpenTofu's automatic selection of the local name %q.",
			/* (..this message needs some more information than we currently have in scope here..) */
		)
```

The main extra thing I want to add here is the information that, for example, OpenTofu guesses that `aws_s3_object` belongs to provider local name `aws` based on the prefix. It's hard to explain that behavior concisely -- making this message longer than I would prefer -- but I hope we can make it understandable by including the actual resource address and inferred local provider name directly in the error message.

Of course, to do that we'd need to track more detailed information in the ProvidersQualification object. I personally think it's worth tracking the full set of resource addresses (`addrs.ConfigResource` values) that had implicit references to each distinct provider address, rather than just the source address as this is currently doing, so that we can say exactly which resource and provider local name we're talking about. Here's how I would expect the above error message to render, just in case it isn't clear what information each of those printf verbs might be representing:

```
Warning: Automatically-inferred provider dependency

  on example.tf line 2:
   2: resource "s3_bucket" "example" {

Due to the prefix of the resource type name OpenTofu guessed that
you intended to associate s3_bucket.example with a provider whose
local name is "s3", but that name is not declared in this module's
required_providers block. OpenTofu therefore guessed that you
intended to use hashicorp/s3, but that provider does not exist.

Make at least one of the following changes to tell OpenTofu which
provider to use:

- Add a declaration for local name "s3" to this module's
required_providers block, specifying the full source address
for the provider you intended to use.
- Verify that "s3_bucket" is the correct resource type name to use.
Did you omit a prefix which would imply the correct provider?
- Use a "provider" argument within this resource block to override
OpenTofu's automatic selection of the local name "s3".
```

Although I very much do not like putting bullet lists in diagnostic messages, I've found from helping people with variations of this problem over the years that all three of these are potential causes and I don't think we really have enough information here to automatically make a strong guess as to which one is most likely. :confounded: 

This rendering also assumes that we'd place the source address of the resource declaration into the diagnostic's "Subject", rather than just inserting the string representation as part of the error message, which isn't necesarily required but would make that information more accessible to tools that are consuming the JSON representation of diagnostics, rather than the human-readable form.

internal/command/init.go
```diff
@@ -1039,6 +1042,33 @@ version control system if they represent changes you intended to make.`))
 	return true, false, diags
 }
 
+// warnOnFailedImplicitProvReference returns a warn diagnostic when the downloader fails to fetch a provider that is implicitly referenced.
+// In other words, if the failed to download provider is having no required_providers entry, this function is trying to give to the user
+// more information on the source of the issue and gives also instructions on how to fix it.
+func warnOnFailedImplicitProvReference(provider addrs.Provider, qualifs *getproviders.ProvidersQualification) tfdiags.Diagnostic {
+	if _, ok := qualifs.Explicit[provider]; !ok {
```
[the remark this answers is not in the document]
yottta: True, I was planning to rework this, but happy to see others agreeing with this.

internal/configs/config.go
```diff
@@ -418,17 +421,29 @@ func (c *Config) addProviderRequirements(reqs getproviders.Requirements, recurse
 	for _, rc := range c.Module.ManagedResources {
 		fqn := rc.Provider
 		if _, exists := reqs[fqn]; exists {
+			// If this is called for a child module, and the provider was added from another implicit reference and not
+			// from a top level required_provider, we need to collect the reference of this resource as well as implicit provider.
+			if _, ok := qualifs.Explicit[fqn]; !ok {
```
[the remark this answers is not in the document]
yottta: Done.

internal/configs/config_test.go
yottta: I think [Martin's comment](https://github.com/opentofu/opentofu/pull/2479#discussion_r1945255473) is already tackling that.
I will rework this while working on his comment.

internal/command/init.go
yottta: Thanks very much for the input. That's the main reason why I kept this PR as draft till now. I am half satisfied by the solution and I was envisioning more of an output like the one you mentioned:
```
Warning: Automatically-inferred provider dependency

  on example.tf line 2:
   2: resource "s3_bucket" "example" {
```

Thanks for the tips and guidance.

internal/getproviders/types.go
yottta: Great. Didn't know about it. Thanks! Applied.

internal/command/init.go
yottta: Changed in this commit: 077d720ad1d28889adb8d1e0f2fa9fa311d88b4d
Feel free to take a look.

internal/command/e2etest/testdata/provider-implicit-ref-not-found/mod/main.tf
yottta: Thanks a lot for the detailed explanation here.
I did some small changes for the first point, in this commit: 6aec456ff76c5ca871015c9bd4c9703cea413888

For the second point, not needed to be tested since that is ensured by the actual implementation: `provider` blocks [are not collected](https://github.com/opentofu/opentofu/blob/6aec456ff76c5ca871015c9bd4c9703cea413888/internal/configs/config.go#L541-L544) into the qualifications.

internal/command/init.go
```diff
@@ -1039,6 +1042,33 @@ version control system if they represent changes you intended to make.`))
...
+		var refFromHint string
+		if refs, ok := qualifs.Implicit[provider]; ok && len(refs) > 0 {
+			refFromHint = fmt.Sprintf("Implicitly referenced from %s.", refs[0].String())
+			if noOfTotalRes := len(refs) - 1; noOfTotalRes > 0 {
+				refFromHint += fmt.Sprintf(" And is referenced from %d more resource(s).", noOfTotalRes)
+			}
+		}
+		details := fmt.Sprintf(`The provider %s doesn't exist and is referenced by a resource implicitly without a 'required_providers' blocks. This typically happens when you are missing the 'required_providers' block or you are not referencing it correctly using the 'provider=' parameter in your resource or data source block.`,
```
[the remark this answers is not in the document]
yottta: I think [Martin's comment](https://github.com/opentofu/opentofu/pull/2479#discussion_r1945255473) is already tackling that.
I will rework this while working on his comment.

internal/command/e2etest/init_test.go
apparentlymart: Although this warning message is pretty verbose, I think it's a good place to start to cover the three main mistakes that can lead to this problem -- which we don't currently have any good way to heuristically prioritize -- and I expect we will improve on this in future once we start to see how people react to this new messaging, if e.g. one or more of the suggestions seems to confuse people.

---

This is not super relevant to what you're sharing here so sorry for the diversion, but: is the red text in the Error diagnostic something you've configured locally in your terminal? e.g. perhaps you've configured it to show stderr in red by default?

I ask because OpenTofu's output typically tries to avoid printing large blocks of colored text for accessibility/readability reasons -- folks are much more likely to have configured good contrast for their default foreground and background color than for other colors -- and so I want to make sure that red text isn't caused by control codes emitted by OpenTofu _itself_, which would suggest some sort of regression. :grinning:

yottta: Regarding the second question, sorry, that's just the output IntelliJ is having by default on their run console. From a regular terminal, this looks the way it needs to be.
image

internal/command/e2etest/testdata/provider-implicit-ref-not-found/mod/main.tf
yottta: Really good point with the negative test. Instead of even more comments, I prefer adding a test. 
Will let you know when that is done.

internal/command/init.go
```diff
@@ -1379,3 +1423,25 @@ The current .terraform.lock.hcl file only includes checksums for %s, so OpenTofu
...
+provider to use:
+
+- Add a declaration for local name "%s" to this module's
+required_providers block, specifying the full source address
+for the provider you intended to use.
+- Verify that "%s" is the correct resource type name to use.
+Did you omit a prefix which would imply the correct provider?
+- Use a "provider" argument within this resource block to override
+OpenTofu's automatic selection of the local name "%s".
+`
```
apparentlymart, line 1447: The diagnostic renderer does its own line-wrapping, so in order for this to be displayed correctly at all screen sizes you'd need to remove all of the "soft-wrapping" -- that is, all of the newlines that aren't separating paragraphs or list items.

That's particularly important in this case where we're substituting user-provided symbols into the message and so we can't actually predict how long the lines will be as part of this constant. This problem is already visible in one of the test cases above, for example:

https://github.com/opentofu/opentofu/blob/c12c63eaf5a81fd4fa95089c3e9d5cf79ffef07d/internal/command/e2etest/init_test.go#L519-L545

Notice that "module's" got wrapped onto a line of its own because "nonexistingprov" made that line too long.

yottta: Thanks a lot. Applied 6237d518c20e419f4321dad86f1d741a1724a335

internal/command/init.go
```diff
@@ -1379,3 +1423,16 @@ The current .terraform.lock.hcl file only includes checksums for %s, so OpenTofu
 To calculate additional checksums for another platform, run:
   tofu providers lock -platform=linux_amd64
 (where linux_amd64 is the platform to generate)`
+
+const implicitProviderReferenceHead = `Automatically-inferred provider dependency
+	on %s line %d:
+		%d: %s "%s" "%s"`
```
apparentlymart, line 1429: Hi again! Sorry I didn't catch this on my first read, but I noticed it once I started actually testing. :confounded: 

Instead of generating this manually, you can use a `&hcl.Diagnostic` instead of a `tfdiags.Sourceless` and set the `Subject` of the diagnostic to [the resource block's `DeclRange`](https://github.com/opentofu/opentofu/blob/5ff40ae5056416568b55f2b1475b51362e4c1e60/internal/configs/resource.go#L62), which is the range that covers its entire header.

Doing it that way will allow the diagnostic renderer to present it consistently with how we show source snippets in other diagnostics, and will also make it machine-readable in the `tofu init -json` mode that wrapping automation tools rely on.

apparentlymart:
Not all of the cases that I collected in [my list of examples of the usability problem](https://github.com/opentofu/opentofu/issues/2084#issuecomment-2581046767) had code examples I could adapt for testing, but I tried a selection of those that did to see if it seems like OpenTofu would generate a helpful/relevant warning that could plausibly have helped the person solve the problem.

---

## [Jan 7 2021](https://discuss.hashicorp.com/t/terraform-init-failed-to-query-available-provider-packages-looking-for-registry-terraform-io-hashicorp-fastly/19486): Missing `required_providers` entry in a child module

The original writeup for this problem suggested that there was only a `provider` block referring to the `fastly/fastly` provider, and that situation isn't covered by our change here.

However, later discussion suggests that there _was_ actually a resource for this provider in some other module in the configuration after all, and so I tested that case instead...

```hcl
# test.tf

terraform {
  required_providers {
    fastly = {
      source = "fastly/fastly"
      version = "0.21.3"
    }
  }
}

module "child" {
  source = "./child"
}

provider "fastly" {

}
```

```hcl
# child/test.tf

resource "fastly_domain_v1" "test" {

}
```

However, this actually succeeded because apparently [we are publishing a `hashicorp/fastly` provider in the OpenTofu registry](https://search.opentofu.org/provider/hashicorp/fastly/latest). It seems like we've built and published from [an archived repository for a long-obsolete provider](https://github.com/hashicorp/terraform-provider-fastly), which is a little worrying -- it means that OpenTofu ends up selecting the "wrong" provider rather than returning an error in this case -- but that's a separate problem from what this PR was about.

```
Initializing provider plugins...
- Finding fastly/fastly versions matching "0.21.3"...
- Finding latest version of hashicorp/fastly...
- Installing fastly/fastly v0.21.3...
- Installed fastly/fastly v0.21.3. Signature validation was skipped due to the registry not containing GPG keys for this provider
- Installing hashicorp/fastly v5.16.0...
- Installed hashicorp/fastly v5.16.0. Signature validation was skipped due to the registry not containing GPG keys for this provider
```

(This also seems to be true for `hashicorp/datadog`, whose upstream repository is also archived.)

---

## [Jul 28 2021](https://discuss.hashicorp.com/t/failed-to-query-available-provider-packages/27419): Missing `aws_` prefix on a resource declaration.

```hcl
terraform {
  required_providers {
    aws = {
      source = "hashicorp/aws"
      version = "~> 3.51.0"
    }
  }
}

provider "aws" {
  region = "eu-west-1"
}

resource "nat_gateway" "test" {
}
```

```
Initializing provider plugins...
- Finding hashicorp/aws versions matching "~> 3.51.0"...
- Finding latest version of hashicorp/nat...
- Installing hashicorp/aws v3.51.0...
- Installed hashicorp/aws v3.51.0 (signed, key ID 0C0AF313E5FD9F80)

Providers are signed by their developers.
If you'd like to know more about provider signing, you can read about it here:
https://opentofu.org/docs/cli/plugins/signing/
╷
│ Warning: Automatically-inferred provider dependency
│       on provider-selection-misunderstandings.tf line 19:
│               19: resource "nat_gateway" "test"
│ 
│ Due to the prefix of the resource type name OpenTofu guessed that you
│ intended to associate nat_gateway.test with a provider whose local
│ name is "nat", but that name is not declared in this module's
│ required_providers block. OpenTofu therefore guessed that you
│ intended to use hashicorp/nat, but that provider does not exist.
│ 
│ Make at least one of the following changes to tell OpenTofu which
│ provider to use:
│ 
│ - Add a declaration for local name "nat" to this module's
│ required_providers block, specifying the full source address for the
│ provider you intended to use.
│ - Verify that "nat_gateway" is the correct resource type name to use.
│ Did you omit a prefix which would imply the correct provider?
│ - Use a "provider" argument within this resource block to override
│ OpenTofu's automatic selection of the local name "nat".
│ 
╵

╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/nat: provider registry registry.opentofu.org does not have
│ a provider named registry.opentofu.org/hashicorp/nat
│ 
│ All modules should specify their required_providers so that external
│ consumers will get the correct providers when using a module. To see
│ which modules are currently depending on hashicorp/nat, run the
│ following command:
│     tofu providers
│ 
│ If you believe this provider is missing from the registry, please
│ submit a issue on the OpenTofu Registry
│ https://github.com/opentofu/registry/issues/new/choose
╵
```

The second option in the warning has correctly diagnosed the problem with this example, and the opening paragraph explains how OpenTofu concluded that "nat" was the local provider name for this resource.

---

## [May 29 2021](https://www.reddit.com/r/Terraform/comments/no567t/error_could_not_load_plugin/): Typoed `aws_` prefix on a resource declaration.

From OpenTofu's perspective this is essentially the same as the previous one, but the person that wrote it had a different experience of the problem: instead of omitting the prefix, they typed it incorrectly.

```hcl
provider "aws" {
}

resource "as_vpc" "First-TF-vpc" {
}


resource "aws_subnet" "subnet-1" {
}
```

```
Initializing provider plugins...
- Finding latest version of hashicorp/aws...
- Finding latest version of hashicorp/as...
- Installing hashicorp/aws v5.86.1...
- Installed hashicorp/aws v5.86.1 (signed, key ID 0C0AF313E5FD9F80)

Providers are signed by their developers.
If you'd like to know more about provider signing, you can read about it here:
https://opentofu.org/docs/cli/plugins/signing/
╷
│ Warning: Automatically-inferred provider dependency
│       on provider-selection-misunderstandings.tf line 25:
│               25: resource "as_vpc" "First-TF-vpc"
│ 
│ Due to the prefix of the resource type name OpenTofu guessed that you
│ intended to associate as_vpc.First-TF-vpc with a provider whose local
│ name is "as", but that name is not declared in this module's
│ required_providers block. OpenTofu therefore guessed that you
│ intended to use hashicorp/as, but that provider does not exist.
│ 
│ Make at least one of the following changes to tell OpenTofu which
│ provider to use:
│ 
│ - Add a declaration for local name "as" to this module's
│ required_providers block, specifying the full source address for the
│ provider you intended to use.
│ - Verify that "as_vpc" is the correct resource type name to use. Did
│ you omit a prefix which would imply the correct provider?
│ - Use a "provider" argument within this resource block to override
│ OpenTofu's automatic selection of the local name "as".
│ 
╵

╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/as: provider registry registry.opentofu.org does not have a
│ provider named registry.opentofu.org/hashicorp/as
│ 
│ All modules should specify their required_providers so that external
│ consumers will get the correct providers when using a module. To see
│ which modules are currently depending on hashicorp/as, run the
│ following command:
│     tofu providers
│ 
│ If you believe this provider is missing from the registry, please
│ submit a issue on the OpenTofu Registry
│ https://github.com/opentofu/registry/issues/new/choose
╵
```

In this case the second suggestion talks only about _omitting_ the prefix and not about mistyping it, which makes this suggestion not 100% relevant but hopefully close enough that (especially when combined with the other context in the opening paragraph) the reader would still notice their mistake and understand how to correct it.

If we wanted to get this closer then we could change that second suggestion to say "Did you omit _or mistype_ a prefix...", but the message is already quite long so I'm personally comfortable with sticking with the current text until we see some evidence that someone found it confusing.

---

## [Dec 15 2021](https://www.reddit.com/r/Terraform/comments/rgzmsl/error_failed_to_query_available_provider_packages/): Dash instead of underscore

This one is an interesting variation of the previous two where the prefix is present and typed correctly but the _delimiter_ isn't: they used `-` instead of `_`.

They didn't include a full code example here but I can guess a suitable approximation of it:

```hcl
resource "aws-subnet" "test" {
}
```

```
Initializing provider plugins...
- Finding latest version of hashicorp/aws-subnet...
╷
│ Warning: Automatically-inferred provider dependency
│       on provider-selection-misunderstandings.tf line 21:
│               21: resource "aws-subnet" "test"
│ 
│ Due to the prefix of the resource type name OpenTofu guessed that you
│ intended to associate aws-subnet.test with a provider whose local
│ name is "aws-subnet", but that name is not declared in this module's
│ required_providers block. OpenTofu therefore guessed that you
│ intended to use hashicorp/aws-subnet, but that provider does not
│ exist.
│ 
│ Make at least one of the following changes to tell OpenTofu which
│ provider to use:
│ 
│ - Add a declaration for local name "aws-subnet" to this module's
│ required_providers block, specifying the full source address for the
│ provider you intended to use.
│ - Verify that "aws-subnet" is the correct resource type name to use.
│ Did you omit a prefix which would imply the correct provider?
│ - Use a "provider" argument within this resource block to override
│ OpenTofu's automatic selection of the local name "aws-subnet".
│ 
╵

╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/aws-subnet: provider registry registry.opentofu.org does
│ not have a provider named registry.opentofu.org/hashicorp/aws-subnet
│ 
│ All modules should specify their required_providers so that external
│ consumers will get the correct providers when using a module. To see
│ which modules are currently depending on hashicorp/aws-subnet, run
│ the following command:
│     tofu providers
│ 
│ If you believe this provider is missing from the registry, please
│ submit a issue on the OpenTofu Registry
│ https://github.com/opentofu/registry/issues/new/choose
╵
```

This one is a little more tenous because I expect a reader would find it clear only if they already know that OpenTofu is expecting `_` as the delimiter marking the end of the "prefix" it mentions in the error message. Combined with the visual similarity of `-` and `_`, I think there's some risk that someone in this situation would fail to correctly diagnose it even with the help of this error message.

However, I don't think we'd want to complicate this long message even more by talking directly about the prefix delimiter, and so again I think we should wait for some evidence that someone was confused by this before we try to address it, and then we can hopefully use the details of how they ask their question to help us hone the message.

There was also [a situation where someone wrote `azure_` instead of `azurerm_`](https://stackoverflow.com/questions/71307245/terraform-source-error-error-failed-to-query-available-provider-packages), which is again pretty subtle and not _obviously_ incorrect with the information in the error message, but the second suggestion does still at least prompt the reader to directly compare what they wrote against the provider's documented examples.

---

## [Jul 29 2022](https://www.reddit.com/r/Terraform/comments/wb4n9e/terraform_not_usingfinding_databricks_provider/): Missing `required_providers` entry in a module, again

This is essentially the same as the first example I tried above, but thankfully in this case there isn't a rogue `hashicorp/databricks` provider in our registry to confuse the issue, so let's try that again...

```hcl
# main.tf

terraform {
  required_providers {
    databricks = {
      source = "databricks/databricks"
      version = "1.1.0"
    }
  }
}

module "child" {
  source = "./child"
}

provider "databricks" {
}
```

```hcl
# child/main.tf

data "databricks_group" "admins" {
}
```

```
Initializing provider plugins...
- Finding databricks/databricks versions matching "1.1.0"...
- Finding latest version of hashicorp/databricks...
- Installing databricks/databricks v1.1.0...
- Installed databricks/databricks v1.1.0. Signature validation was skipped due to the registry not containing GPG keys for this provider
╷
│ Warning: Automatically-inferred provider dependency
│       on child/provider-selection-misunderstandings-child.tf line 1:
│               1: data "databricks_group" "admins"
│ 
│ Due to the prefix of the resource type name OpenTofu guessed that you
│ intended to associate module.child.data.databricks_group.admins with
│ a provider whose local name is "databricks", but that name is not
│ declared in this module's required_providers block. OpenTofu
│ therefore guessed that you intended to use hashicorp/databricks, but
│ that provider does not exist.
│ 
│ Make at least one of the following changes to tell OpenTofu which
│ provider to use:
│ 
│ - Add a declaration for local name "databricks" to this module's
│ required_providers block, specifying the full source address for the
│ provider you intended to use.
│ - Verify that "databricks_group" is the correct resource type name to
│ use. Did you omit a prefix which would imply the correct provider?
│ - Use a "provider" argument within this resource block to override
│ OpenTofu's automatic selection of the local name "databricks".
│ 
╵

╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/databricks: provider registry registry.opentofu.org does
│ not have a provider named registry.opentofu.org/hashicorp/databricks
│ 
│ Did you intend to use databricks/databricks? If so, you must specify
│ that source address in each module which requires that provider. To
│ see which modules are currently depending on hashicorp/databricks,
│ run the following command:
│     tofu providers
│ 
│ If you believe this provider is missing from the registry, please
│ submit a issue on the OpenTofu Registry
│ https://github.com/opentofu/registry/issues/new/choose
╵
```

This time we did actually get an error and thus also see the warning! 🎉

The first suggestion diagnoses this problem. It mentions to add the declaration to "this module's" `required_providers` block, which is relying on the reader to be familiar enough with OpenTofu modules to understand that "this module" refers to the child module rather than to the whole configuration, and so we might not reach everyone with this exact text.

If we _do_ find that folks are still confused by this even with this warning in place, I'd suggest that we revise it by including the actual module path in the error message so that we're not relying on the reader's understanding of what "this module" means, but again I think we can wait and see how folks react to this message before we make it even more complicated.

---

## [Nov 15 2022](https://www.reddit.com/r/Terraform/comments/yw3e37/error_using_aliased_providers/): An interesting misunderstanding of provider local names vs. provider configuration aliases

This one was unusual (compared to the usual shape of these questions) in that it started with a misunderstanding of the difference between provider local names and the concept of "aliases" assigned to additional provider configurations: the author tried to add a `required_providers` entry matching the alias they chose for one of their provider configurations:

```hcl
terraform {
  required_providers {
    azurerm = {
      source = "hashicorp/azurerm"
      version = "3.27.0"
    }
    core = {
      source = "hashicorp/azurerm"
      version = "3.27.0"
    }
  }
}

provider "azurerm" {
}

provider "azurerm" {
  alias = "core"
}
```

This PR does not change anything about OpenTofu's handling of _this_ situation, because it doesn't involve any resource blocks. Since I've only seen this shape of problem happen once I think we can safely ignore this for now and revisit it later if we see variations on this question appear more in future.

---

The list I made over in the issue did not include an example of something leading to the third suggestion in the error message: missing a `provider` argument for a provider whose name disagrees with its typical prefix.

Here's one way that folks tend to get that wrong:

```hcl
terraform {
  required_providers {
    google-beta = {
      source = "hashicorp/google-beta"
    }
  }
}

resource "google_compute_region_instance_group_manager" "test" {
  # Since this provider's local name doesn't match the prefix from
  # its resource type names, an explicit provider argument is needed:
  #    provider = google-beta
}
```

```
Initializing provider plugins...
- Finding latest version of hashicorp/google-beta...
- Finding hashicorp/azurerm versions matching "3.27.0"...
- Finding latest version of hashicorp/google...
- Installing hashicorp/google-beta v6.20.0...
- Installed hashicorp/google-beta v6.20.0 (signed, key ID 0C0AF313E5FD9F80)
- Installing hashicorp/azurerm v3.27.0...
- Installed hashicorp/azurerm v3.27.0 (signed, key ID 0C0AF313E5FD9F80)
- Installing hashicorp/google v6.20.0...
- Installed hashicorp/google v6.20.0 (signed, key ID 0C0AF313E5FD9F80)

Providers are signed by their developers.
If you'd like to know more about provider signing, you can read about it here:
https://opentofu.org/docs/cli/plugins/signing/

OpenTofu has created a lock file .terraform.lock.hcl to record the provider
selections it made above. Include this file in your version control repository
so that OpenTofu can guarantee to make the same selections by default when
you run "tofu init" in the future.
```

Unfortunately there's no direct diagnosis at all for this case, because there _is_ a provider named `hashicorp/google`. 😖

I don't really think there's any good way for us to diagnose this case unless we hard-code this particular pair of providers for a special message, which I don't think we should do. It's unfortunate that these two providers interact in this way, but it's a logical consequence of intentional decisions made by their developers to make it easier for users to switch between the two providers when needed.

Our third suggestion would work for any variant of this situation where there isn't also another provider that OpenTofu can implicitly select, but I'm not aware of any specific provider other than `hashicorp/google-beta` that follows this unusual design of using resource type names that disagree with the resource type prefixes.

For what it's worth, the following is actually the more recommended way to make the above example work, since it avoids a separate declaration for each resource. I wrote the example above the way I did only because I've helped people who wrote it that way before.

```hcl
terraform {
  required_providers {
    # The local name is not required to match the last segment of
    # the provider's source address, and so we can choose a name
    # that agrees with the provider's resource type name prefixes.
    google = {
      source = "hashicorp/google-beta"
    }
  }
}

resource "google_compute_region_instance_group_manager" "test" {
  # ...and so no explicit provider argument is needed here anymore
}
```

---

Overall I think what we have in this PR is a good start toward helping with the most common mistakes, and at least _hinting_ for some less common problems.

There are some other quirks in the overall system that mean this warning doesn't appear in all of the cases where it might be helpful -- particularly the fact that the OpenTofu Registry contains some copies of archived providers that no longer exist upstream -- but those are beyond the scope of this pull request.

internal/command/init.go
yottta: Thanks a lot Martin! Having you on this PR made me to learn a lot about the way the whole tool is working!
Lovely how this part of the diagnostics is handled! 🙌

yottta:
Thanks a lot Martin for the thorough testing. Good lesson that I got here.

Regarding `Missing required_providers entry in a module, again`, the fact that we are having the whole path of the file in the diagnostic, that also includes the module directory. Therefore, referencing "this module's" in the actions below, should help with that.
Indeed, it could create some confusion regarding the actual module that we are referring to:
* is it the module from which we are running the `tofu` command?
* is it the module referenced in the diagnostic above?

Maybe we can add a small indication there about it, but will make the message even larger.

Nonetheless, would be good to go with this and wait for more feedback from the users.

cam72cam (approved):

apparentlymart (approved):

PR #2492, opened 2025-02-06
Clarify which functions exist on configured providers only

cam72cam:
This PR explicitly groups what functions can be called on configured vs unconfigured providers.  This has been a bit of a "implied" concept for quite a while, and I figure it's time to make it explicit.

There is a lot more refactoring we could do to the codebase to clarify when a provider is configured or not, but for now splitting the interface is documentation / a starting point.

Follow up to #2370 

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/providers/provider.go
+++ b/internal/providers/provider.go
@@ -13,12 +13,12 @@ import (
 	"github.com/opentofu/opentofu/internal/tfdiags"
 )
 
-// Interface represents the set of methods required for a complete resource
-// provider plugin.
-type Interface interface {
-	// GetMetadata is not yet implemented or used at this time.  It may
+// Unconfigured represents a provider plugin that has not yet been configured. It has
+// limited functionality that must not depend on ConfigureProvider having been called.
+type Unconfigured interface {
+	// GetMetadata is not yet implemented or used at this time. It may
 	// be used in the future to avoid loading a provider's full schema
-	// for initial validation.  This could result in some potential
+	// for initial validation. This could result in some potential
 	// memory savings.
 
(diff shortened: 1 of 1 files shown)


yottta (dismissed):

internal/providers/provider.go
```diff
@@ -38,15 +38,19 @@ type Interface interface {
...
-	// instance state whose schema version is less than the one reported by the
-	// currently-used version of the corresponding provider, and the upgraded
-	// result is used for any further processing.
-	UpgradeResourceState(UpgradeResourceStateRequest) UpgradeResourceStateResponse
+	// MoveResourceState requests that the given resource data be moved from one
+	// type to another, potentially between providers as well.
+	MoveResourceState(MoveResourceStateRequest) MoveResourceStateResponse
+
+	// CallFunction requests that the given function is called and response returned.
+	CallFunction(CallFunctionRequest) CallFunctionResponse
```
ollevche, line 46: I don't remember the details, but we do allow custom functions to be called. I assume they are part of the configuration process so they will be unavailable here?

ollevche: I am also confused that we can `CallFunction` but not `GetFunctions` with `Unconfigured`. Comment would be helpful here.

cam72cam: CallFunction is a weird one, it can call both unconfigured and configured functions in OpenTofu-land.  I can update the documentation to clarify this quirk.

Gogotchuri (approved):

yottta (approved):

Issue #1148, opened 2024-01-19
Certifications such as soc2 / iso21007


nitrocode:
### OpenTofu Version

```shell
Latest
```

### Use Cases

If opentofu was soc2 and iso27001 certified, more orgs would be more likely to use it

### Attempted Solutions

N/A

### Proposal

- Get certified
- Show certification on a public page
- If certs can be retrieved, show on a public page how security is a first class customer to aleviate risks from more corporate customers

### References

N/A

janosdebugs:
Hi @nitrocode thank you for this issue. For both SOC2 and ISO27001 what could be in scope would be how we manage the GitHub organization and our release channels, who has access to what. We don't handle customer data or anything of the sort, so a lot of these certification criteria don't really apply to us. I also feel that showing off the founding member's certifications could be misleading since their audits, as far as I know, currently don't have OpenTofu in scope. This may also be easier to do once #346 is complete since, as far as I know, being part of the CNCF gives us access to GitHub Enterprise, which allows for org-wide policy enforcement.

It's also worth noting that these certifications say nothing about how secure OpenTofu itself is, only that we comply with the practices set out for managing access. I have to wonder if the considerable amount of money and time could be better spent on performing an actual security audit on the OpenTofu code in order to make sure that the code is secure. Do you know of any certifications to that effect?

There are also several other things we could potentially do to improve security before going for a purely administrative measure, such as starting to ship an SBOM, or these security-related issues: #307, #315, #673, #428, #297 (currently work in progress).

Do you know of any open source project that has performed these certifications? Have you encountered companies that have refused to use OpenTofu or similar open source tools (not services, but the tools themselves) due to a lack of these certifications?

nitrocode:
Thanks @janosdebugs for the reply and the related tickets to track.

The core of the ask from these customers is basically "What does the vendor do to ensure that their product and processes are secure" so if a cert was there that would be sufficient, if not, then a security audit should be sufficient as well. Some orgs do not publicize these but make them available upon request which is also sufficient for approval for some orgs.

Issue #315, opened 2023-09-06
Generate SLSA provenance for OpenTF release builds


developer-guy:
### Summary

All reports published by security companies prove that Software Supply Chain Attacks are on the rise. There is no doubt that they will continue to increase in the coming years. With this undeniable increase, securing our software supply chains is becoming more important for everyone.

Protecting our software from such attacks can be quite challenging as there will be a lot of moving parts in our software delivery pipelines, but a great starting point would be to create non-falsifiable, verifiable evidence logs about the software we build, including all the details such as steps, build environment, environment variables, etc. from start to finish when we release our software. This is where [Supply Chain Levels for Software Artifacts or SLSA](https://slsa.dev) for short comes into the picture.

SLSA is a set of standards and technical controls you can adopt to improve artifact integrity and build toward completely resilient systems. It’s not a single tool, but a step-by-step outline to prevent artifacts being tampered with and tampered artifacts from being used, and at the higher levels, hardening up the platforms that make up a supply chain.

### Problem Statement

As @dlorenc said in one of his talks at [Open Source Summit](https://dlorenc.medium.com/zero-trust-supply-chain-security-e3fb8b6973b8) (which is my favorite quote from his presentation), `Every artifact can be verifiably traced to Source Code And Hardware` because otherwise, we couldn't be sure the software we are running on our platforms even our local workstations. I roughly liken this approach to how it would be no different than picking up a USB Drive on the road and plugging it into our computer, as in one of my favorite scenes from Mr. Robot.

### User-facing description

This feature can be used by users of the OpenTF project to make sure that the software they download has not been tampered with. Users can even know everything about the software, including when it was created, who released it, how it was created, etc. As I said above, this document must be verifiable, which means that you have to sign it before you store it, so that you verify that you are who you say you are. 

### Technical Description

One of the ways of generating the SLSA provenance on the GitHub Actions platform is to use [SLSA GitHub Generators](https://github.com/slsa-framework/slsa-github-generator) project which provides language-agnostic SLSA provenance generation for your releases. `slsa-github-generator` is a set of tools for the generation of SLSA3+ provenance for native GitHub projects. It allows projects to generate [SLSA provenance](https://slsa.dev/provenance/) safely and accurately using [GitHub Actions](https://github.com/features/actions). 

This project provides a bunch of trusted builders such as Generic, Container, Docker, Go, and many more. We can use the `Generic Builder` since we are using the GoReleaser to release artifacts so it'd be pretty easy to integrate.

[...]

Issue #307, opened 2023-09-06
Support keyless provider signing/verification using Sigstore


imjasonh:
### Summary

Hashicorp Terraform requires provider authors use GPG keys to sign provider binaries, which it then verifies upon installing them. The typical workflow is that the GPG private key is generated, stored as a GitHub secret, and the public key is uploaded to the TF registry for verification. Keys are long-lived and can leak. When keys are lost, they are generally regenerated by a GH user with access to update the private secret and public key on TFR.

I'd like to propose that opentf registry and CLI support signing provider binaries with [Sigstore](https://www.sigstore.dev/), which enables keyless signing and verification, and a public immutable transparency log of previous signings.

### Problem Statement

GPG is brittle, hard to use, and requires using long-lived secrets, which can leak.

Sigstore provides smooth integration with GitHub Actions (and any OIDC provider), and provides a stronger guarantee that the artifact was built by the GitHub Actions workflow, and not an attacker.

### User-facing description

Provider authors can more easily integrate with the opentf registry to provide artifacts that can be more easily and securely verified.

Provider users can have better assurance that their providers were built from source available in GitHub.

Provider authors may have more choice about where they host the source for their providers, and sign them with other OIDC providers (Gitlab, AWS CodeBuild, etc.). This largely depends on opentf registry support; signing is just OIDC + Sigstore.

### Technical Description

OpenTF registry would need to support verifying providers using keyless verification.

OpenTF provider scaffolding (e.g., GitHub repo templates) should document -- ideally recommend -- keyless signing using built-in OIDC integration.

Providers would be able to migrate from signing with GPG keys to using keyless Sigstore signing. If a provider had previously signed artifacts with GPG but switched to keyless signing, those old releases should still be verifiable and installable by users. A provider that uses keyless signing should be able to change their keyless signing params (e.g., rename their GH repo) and the OpenTF registry should continue to be able to verify old releases.

### Rationale and alternatives

Keyless signing is more secure and more ergonomic than using GPG.

Alternatives include:
- keep using GPG (which folks certainly can) and recommend GH secrets
- use another signing scheme, possibly some future one that doesn't exist yet

### Downsides

Work would be involved.

Relying on the public transparency log might not be preferrable for providers that are intended to be private (a record of the name of the repo and digest of the artifact could be exposed publicly; not the provider binary itself). Care would need to be taken for private providers, or they could keep using GPG.

[...]

PR #2494, opened 2025-02-07
RFC: Strengthening supply chain security: SBOMs and more

ghost:
This RFC addresses a component issue from #2163, which deals with allowing SBOM artifacts in the OpenTofu registry.

Rendered version: https://github.com/opentofu/opentofu/blob/rfc/sbom/rfc/20250207-sbom.md

References:

- #2163
- #1148
- #315
- #307

## Target Release

1.10.0

## Checklist

### DIFF
--- a/rfc/20250207-sbom.md
+++ b/rfc/20250207-sbom.md
@@ -22,7 +22,7 @@ However, Software Bill of Materials artifact formats, such as the [SPDX JSON](ht
 This RFC addresses the compliance concern in the following ways:
 
 1. OpenTofu should publish a Software Bill of Materials (SBOM) artifact for the main OpenTofu binary that users can rely on to list OpenTofu's direct dependencies.
-2. OpenTofu should make it possible to publish SBOM and SLSA attestation artifacts for providers and expose them in the OpenTofu Registry. 
+2. OpenTofu should make it possible to publish SBOM artifacts for providers and expose them in the OpenTofu Registry. 
 3. OpenTofu should publish SBOM artifacts for the providers mirrored from the HashiCorp repositories (AWS, etc.)
 
 For the current RFC, we consider out of scope:
@@ -138,8 +138,8 @@ On the implementation side, this change will require a change in the [OpenTofu R
       ],
       "attachments": [
         {
-          "media_type": "application/vnd.in-toto+json",
-          "filename": "slsa.in-toto.jsonl",
+          "media_type": "application/spdx+json",
+          "filename": "sbom.spdx.json",
           "url": "..."
         }
       ]
@@ -153,7 +153,7 @@ On the implementation side, this change will require a change in the [OpenTofu R
 OpenTofu should offer the ability to download all attestations of a certain type into a local folder. For example:
 
 ```
-tofu providers download-attachments --types=application/spdx+json,application/vnd.in-toto+json --destination=./attestations
+tofu providers download-attachments --types=application/spdx+json,application/vnd.some.other.type --destination=./attestations
 ```
 
 This command would iterate over all providers in the current project and download attachments related to the providers used in the current project into the destination folder, grouped by provider name.


rfc/20250207-sbom.md
```diff
@@ -0,0 +1,184 @@
...
+
+- `*.spdx.xml`, `*.spdx.yml`, and `*.spdx.yaml` (SPDX) will *not* be supported until [there is an approved MIME type for them](https://github.com/spdx/spdx-spec/issues/577#issuecomment-960295523)
+- `bom.xml` and `bom.json` (CycloneDX) will *not* be supported [until there is an approved MIME type for it](https://github.com/CycloneDX/specification/issues/210)
+
+These rules may be amended later if there is sufficient proof that a correct and unique MIME type has been established by the format vendor.
+
+Since the response is a per-platform response, the OpenTofu registry should follow the naming rules for the provider binaries to determine if an attestation/SBOM artifact is related to a single platform or to all platforms. In the registry response both kinds of responses should be included. For example:
+
+- `terraform-provider-aws_5.86.0_linux_arm.spdx.json` relates to the `linux/arm` platform
+- `slsa.in-toto.jsonl` relates to all platforms
```
apparentlymart, line 97: The rest of this proposal is focused only on the SPDX-format SBOM and doesn't include in-toto attestations, so this appearing here is perhaps a little confusing, but I understand that we don't actually _have_ any non-platform-specific files right now.

For the sake of this RFC in particular, could we just state that any file without a recognizable platform identifier is ignored? A future RFC can define other naming schemes that don't include platform identifiers if needed.

rfc/20250207-sbom.md
```diff
@@ -0,0 +1,184 @@
...
+}
+```
+
+### CLI changes
+
+OpenTofu should offer the ability to download all attestations of a certain type into a local folder. For example:
+
+```
+tofu providers download-attachments --types=application/spdx+json,application/vnd.some.other.type --destination=./attestations
+```
```
apparentlymart, line 157: While I do agree that downloading all of the individual attachments (and, presumably, verifying their associated signatures) is useful, I think there might be another use-case here too:

An OpenTofu module _itself_ also has a "bill of materials" that consists of all of the providers and other modules it directly depends on. Since OpenTofu modules are written in OpenTofu's own language, we can potentially have OpenTofu _itself_ generate shallow SBOMs for them, rather than relying on external tools to do it.

We could also potentially have the registry generate an SBOM for each module at the same time as it runs (a custom build of) OpenTofu to collect the module API information. One annoyance with this is that, as usual, our unit of module distribution is the _module package_ rather than an individual module, and so a particular module package can have zero or more modules inside it. However, the registry presumably already relies on [the so-called "Standard Module Structure"](https://opentofu.org/docs/v1.8/language/modules/develop/structure/) to find modules and examples, so we could decide that the registry's SBOM generation is limited only to modules that are discoverable by following those conventions.

We would also probably need to just skip any `module` block whose `source` argument isn't a constant, since a module whose dependencies depend on input variables effectively has an unbounded set of possible dependencies which we therefore cannot represent in the SPDX format.

But those quirks aside: if we could do that then we'd be able to provide the data needed to traverse the full dependency graph starting at a root module, including the detail about which nested modules depend on which providers so that e.g. a detected provider vulnerability can be reported to the maintainer of the individual module that depends on the provider, rather than just to the top-level configuration owner.

(This is _also_ all stuff that can be considered a stretch goal I think, since it would only increase the level of detail in our SBOM reporting and would not conflict with anything else that's proposed here.)

apparentlymart: I have no strong opinions about whether this would be all in one RFC or across many, but I don't think we can say that we've fully addressed the problem until it's also possible to take inventory of modules that are being used, because modules themselves can have license concerns and security vulnerabilities just as providers can. 

For the sake of this particular RFC though, I'm going to resolve this thread. I have no particular interest in solving this for anything other than OCI registry distribution right now, since that is our current focus. Any other problems we can solve with relatively low effort at the same time are fine, but I agree that SBOMs for modules don't fit into that category.

rfc/20250207-sbom.md
```diff
@@ -0,0 +1,184 @@
...
+
+The OpenTofu Registry should perform a *narrowly tailored* scan for SBOM and other attestation artifacts and include them in the `attachments` section according to the following rules:
+
+- `*.spdx.json` and `*.spdx.sbom.json` (SPDX) will be identified as [`application/spdx+json`](https://www.iana.org/assignments/media-types/application/spdx+json)
+
+However, we are currently excluding the following:
+
+- `*.spdx.xml`, `*.spdx.yml`, and `*.spdx.yaml` (SPDX) will *not* be supported until [there is an approved MIME type for them](https://github.com/spdx/spdx-spec/issues/577#issuecomment-960295523)
+- `bom.xml` and `bom.json` (CycloneDX) will *not* be supported [until there is an approved MIME type for it](https://github.com/CycloneDX/specification/issues/210)
```
shoeffner, line 90: I know this topic is right now on hold (and I think it's fine to focus on one format), but would like to mention that CycloneDX has two registered MIME types (JSON and XML). I believe the linked issue is about "multiple JSON SBOMs in a single file".

https://www.iana.org/assignments/media-types/application/vnd.cyclonedx+json
https://www.iana.org/assignments/media-types/application/vnd.cyclonedx+xml

cam72cam:
Closing as this has not been directly requested as a feature for the OCI work.  Please open an issue to continue the discussion if this is a feature you desire in OpenTofu.

PR #2511, opened 2025-02-12
RFC: s3 locking based on conditional writes

yottta:
Part of #599 
[Rendered version](https://github.com/opentofu/opentofu/blob/s3-locking-rfc/rfc/20250211-s3-locking-with-conditional-writes.md)

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/rfc/20250211-s3-locking-with-conditional-writes.md
+++ b/rfc/20250211-s3-locking-with-conditional-writes.md
@@ -59,6 +59,8 @@ terraform {
   * Release the lock from the S3 bucket;
   * Release the lock from DynamoDB table.
 
+The usage of [workspaces](https://opentofu.org/docs/language/state/workspaces/) will not impact this new way of locking. The locking object will always be stored right next to its related state object.
+
 > [!NOTE]
 >
 > OpenTofu [recommends](https://opentofu.org/docs/language/settings/backends/s3/) to have versioning enabled for the S3 buckets used to store state objects.
--- a/rfc/20250211-s3-locking-with-conditional-writes.md
+++ b/rfc/20250211-s3-locking-with-conditional-writes.md
@@ -2,11 +2,11 @@
 
 Issue: https://github.com/opentofu/opentofu/issues/599
 
-Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on another instance of DynamoDB.
+Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on DynamoDB.
 
 The main reasons for such a change could be summarized as follows:
 * Less resources to maintain.
-* Removing costs that now are totally useless.
+* Potentially reducing costs by eliminating usage of dynamo db.
 * One less point of failure (removing DynamoDB from the state management)
 
 The most important things that need to be handled during this implementation:
@@ -17,7 +17,7 @@ The most important things that need to be handled during this implementation:
 ## Proposed Solution
 
 Until recently, most of the approaches that could have been taken for this implementation, could have been prone to data races.
-But AWS released a new functionality to the S3, and now it supports conditional writes on any given S3 bucket.
+But AWS has released new functionality for S3, supporting conditional writes on objects in any S3 bucket.
 
 For more details on the AWS S3 feature and the way it works, you can read more on the [official docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html).
 
--- a/rfc/20250211-s3-locking-with-conditional-writes.md
+++ b/rfc/20250211-s3-locking-with-conditional-writes.md
@@ -141,7 +141,7 @@ if err {
   ```
   Asking because this could also lead to other issues in case it doesn't match. I would suggest to have this only for showing an error in case the digest is different.
 * Right now, since the state digest is stored in DynamoDB, this is written only when the `dynamodb_table` is specified.
-  * Should we do the same in the new implementation? Write the state digest only if the `use_lockfile=true`?
+  * Should we do the same in the new implementation? Write the state digest only if `use_lockfile=true`?
     * Considering the other conditional write option, with the header `IfMatch=<ETag of <state-object-path-before-applying>-md5>`, in case the user(s) is having no locking enabled, we could actually be able to warn the user if the digest was updated concurrently and suggesting enabling the locking mechanism to avoid concurrent writes on the state object.
       * If needed, I could provide a small POC for it.
 ### Future Considerations
--- a/rfc/20250211-s3-locking-with-conditional-writes.md
+++ b/rfc/20250211-s3-locking-with-conditional-writes.md
@@ -125,7 +125,9 @@ But if both locks are enabled (`use_lockfile=true` and `dynamodb_table=<actual_t
 * Do we want to provide the option to have the lock objects into another bucket? This will break the feature parity.
 
 ### Future Considerations
-If this feature will prove to have a high adoption rate, later, we might consider to deprecate the DynamoDB locking mechanism. 
+Later, DynamoDB based locking might be considered for deprecation once (at least) the following are true:
+* Conditional writes are implemented in the majority of the S3 compatible services.
+* The adoption rate for the new S3 based locking will be high enough to not affect the existing users.
 
 ## Potential Alternatives
 Since this new feature relies on the S3 conditional writes, there is hardly other reliable alternative to implement this. 
--- a/rfc/20250211-s3-locking-with-conditional-writes.md
+++ b/rfc/20250211-s3-locking-with-conditional-writes.md
@@ -1,0 +1,119 @@
+# S3 Backend: Locking by using the recently released feature of conditional writes
+
+Issue: https://github.com/opentofu/opentofu/issues/599
+
+Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on another instance of DynamoDB.
+
+The main reasons for such a change could be summarized as follows:
+* Less resources to maintain.
+* Removing costs that now are totally useless.
+* One less point of failure (removing DynamoDB from the state management)
+
+The most important things that need to be handled during this implementation:
+* A simple way to enable/disable locking by using S3 conditional writes.
+* A path forward to migrate the locking and state digest from DynamoDB to S3.
+* The default behavior should be untouched as long as the `backend` block configuration contains no attributes related to the new locking option.
+
+## Proposed Solution
+
+Until recently, most of the approaches that could have been taken for this implementation, could have been prone to data races.
+But AWS released a new functionality to the S3, and now it supports conditional writes on any given S3 bucket.
+
+For more details on the AWS S3 feature and the way it works, you can read more on the [official docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html).
+
+> By using conditional writes, you can add an additional header to your WRITE requests to specify preconditions for your Amazon S3 operation. To conditionally write objects, add the HTTP `If-None-Match` or `If-Match` header.
+> 
+> The `If-None-Match` header prevents overwrites of existing data by validating that there's not an object with the same key name already in your bucket.
+>
+> Alternatively, you can add the `If-Match` header to check an object's entity tag (`ETag`) before writing an object. With this header, Amazon S3 compares the provided `ETag` value with the `ETag` value of the object in S3. If the ETag values don't match, the operation fails.
+
+
+To allow this in OpenTofu, the `backend "s3"` will receive one new attribute:
+* `use_lockfile` `bool` (Default: `false`) - Flag to indicate if the locking should be performed by using strictly the S3 bucket.
+
+> [!NOTE]
+>
+> The name `use_lockfile` was selected this way to keep the [feature parity with Terraform](https://developer.hashicorp.com/terraform/language/backend/s3#state-locking).
+
+When OpenTofu will find the above attribute in the backend configuration, it will handle the state locking and the digest of the state file in the same bucket as configured for the state file.
+
+### User Documentation
+
+In order to make use of this new feature, the users will have to add the attribute in the `backend` block:
+```terraform
+terraform {
+  backend "s3" {
+    bucket = "tofu-state-backend"
+    key = "statefile"
+    region = "us-east-1"
+    dynamodb_table = "tofu_locking"
+    use_lockfile = true
+  }
+}
+```
+
+* When the new attribute `use_lockfile` exists and `dynamodb_table` is missing, OpenTofu will behave as normal:
+  * The only thing that will have to do extra, is to ensure that the MD5 file is there and if not, to create it.
+    * The generation of the MD5 checksum is already part of the [implementation](https://github.com/opentofu/opentofu/blob/6614782/internal/backend/remote-state/s3/client.go#L178).
+    * We just need to store it in a new file in the configured S3 bucket (file name: <state-file-path>-md5)
+* When the new attribute `use_lockfile` will exist **alongside** `dynamodb_table`, OpenTofu will:
+  * Acquire the lock in the S3 bucket;
+  * Get the digest of the statefile from S3. If missing:
+    * Acquire the lock in DynamoDB table; 
+    * Get the digest of the statefile from DynamoDB and migrate it to the S3 bucket;
+    * Release the lock from DynamoDB table;
+  * Perform the requested sub-command
+  * Release the lock from the S3 bucket;
+
+> [!NOTE]
+>
+> OpenTofu [recommends](https://opentofu.org/docs/language/settings/backends/s3/) to have versioning enabled for the S3 buckets used to store state files.
+>
+> Acquiring and releasing locks will add a good amount of writes and reads to the bucket. Therefore, for a versioning-enabled S3 bucket, the number of versions for that file could grow significantly.
+> Even though the cost should be negligible for the locking files, any user using this feature could consider configuring the lifecycle of the S3 bucket to limit the number of versions of a file.
+
+### Technical Approach
+
+In order to achieve and ensure a proper state locking via S3 bucket, the implementation will do a call to `s3client.PutObject` with the property `IfNoneMatch: "*"`.
+For more information, please check the [official documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html#conditional-write-key-names).
+
+But the simplified implementation should look like this:
+```go
+input := &s3.PutObjectInput{
+    Bucket:            aws.String(bucket),
+    Key:               aws.String(key),
+    Body:              bytes.NewReader([]byte(lockInfo)),
+    IfNoneMatch:       aws.String("*"),
+}
+_, err := actor.S3Client.PutObject(ctx, input)
+```
+
+### Open Questions
+
+* Do we want to provide the option to have the lock files into another bucket? This will break the feature parity.
+* When writing the digest (*-md5) file, should we consider conditional writing by using the `IfMatch` argument?
+  ```go
+  func (actor S3Actions) UpdateDigest(ctx context.Context, bucket string, key string, beforeApplyDigest string) error {
+      newDigest := uuid.NewString()
+      input := &s3.PutObjectInput{
+          Bucket:            aws.String(bucket),
+          Key:               aws.String(key),
+          Body:              bytes.NewReader([]byte(newDigest)),
+          ChecksumAlgorithm: types.ChecksumAlgorithmSha256,
+          IfMatch:           aws.String(beforeApplyDigest),
+      }
+      _, err := actor.S3Client.PutObject(ctx, input)
+      return err
+  }
+  ```
+  Asking because this could also lead to other issues in case it doesn't match. I would suggest to have this only for showing an error in case the digest is different.
+* Right now, since the MD5 digest is stored in DynamoDB, this is written only when the `dynamodb_table` is specified.
+  * Should we do the same in the new implementation? Write the MD5 digest only if the `use_lockfile=true`?
+    * Considering the other conditional write option, with the header `IfMatch=<ETag of <state-file-path-before-applying>-md5>`, in case the user(s) is having no locking enabled, we could actually be able to warn the user if the digest was updated concurrently and suggesting enabling the locking mechanism to avoid concurrent writes on the state file.
+      * If needed, I could provide a small POC for it.
+### Future Considerations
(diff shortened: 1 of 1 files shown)


rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,119 @@
+# S3 Backend: Locking by using the recently released feature of conditional writes
+
+Issue: https://github.com/opentofu/opentofu/issues/599
+
+Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on another instance of DynamoDB.
+
+The main reasons for such a change could be summarized as follows:
+* Less resources to maintain.
+* Removing costs that now are totally useless.
```
cam72cam, line 9: * Potentially reducing costs by eliminating usage of dynamo db.

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,119 @@
+# S3 Backend: Locking by using the recently released feature of conditional writes
+
+Issue: https://github.com/opentofu/opentofu/issues/599
+
+Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on another instance of DynamoDB.
```
cam72cam, line 5: remove "another instance of"

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,119 @@
...
+
+The most important things that need to be handled during this implementation:
+* A simple way to enable/disable locking by using S3 conditional writes.
+* A path forward to migrate the locking and state digest from DynamoDB to S3.
+* The default behavior should be untouched as long as the `backend` block configuration contains no attributes related to the new locking option.
+
+## Proposed Solution
+
+Until recently, most of the approaches that could have been taken for this implementation, could have been prone to data races.
+But AWS released a new functionality to the S3, and now it supports conditional writes on any given S3 bucket.
```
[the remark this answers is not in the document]
yottta: Thanks. Fixed 711eaf70265ee42a67c97c178b4cb21326460354

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Thanks. Fixed 711eaf70265ee42a67c97c178b4cb21326460354

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,148 @@
...
+_, err := actor.S3Client.PutObject(ctx, input)
+```
+
+The `err` returned above should be handled accordingly with the [behaviour defined](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html) in the official docs:
+* HTTP 200 (OK) - Means that the locking file was not existing. Therefore, the lock can be considered as acquired.
+  * For buckets with versioning enabled, if there's no current object version with the same name, or if the current object version is a delete marker, the write operation succeeds.
+* HTTP 412 (Precondition Failed) - Means that the locking file is already there. Therefore, the whole process should exit because the lock couldn't be acquired.
+* HTTP 409 (Conflict) - Means that the there was a conflict of concurrent requests. AWS recommends to retry the request in such cases, but we could just handle this similarly to the 412 case.
+
+#### Digest updates
```
ollevche, line 98: I may be missing something, but do we need to update the digest? If I understand correctly, we could go with a `IfNoneMatch` lock creation so it would fail for any other tofu run, while the file exists. On Unlock, we could just delete it altogether. As far as I can tell, this matches the dynamodb implementation and eliminates data races via `IfNoneMatch` option. 

What do you think?

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,148 @@
...
+          ChecksumAlgorithm: types.ChecksumAlgorithmSha256,
+          IfMatch:           aws.String(beforeApplyDigest),
+      }
+      _, err := actor.S3Client.PutObject(ctx, input)
+      return err
+  }
+  ```
+  Asking because this could also lead to other issues in case it doesn't match. I would suggest to have this only for showing an error in case the digest is different.
+* Right now, since the MD5 digest is stored in DynamoDB, this is written only when the `dynamodb_table` is specified.
+  * Should we do the same in the new implementation? Write the MD5 digest only if the `use_lockfile=true`?
```
ollevche, line 141: I think we should only write lock files if the locking is enabled, so I would say yes.

ollevche (commented):
Looks good overall, thanks! I left a few questions and suggestions.

nit: I would also suggest keeping wording consistent and using a single term for the lock file. Curently, it is referred to as lock file, md5 file, lock object, etc.

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: I used terms like "digest", "md5" or "digest file" for the part where OpenTofu is writing an md5 hash of the statefile.
For example, looking into the dynamoDB, there is the md5 sum of the state file. Can be easily checked like this:
image

ollevche: Got you, thanks! I missed that part, I will re-read it and leave another comment.

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,148 @@
...
+Issue: https://github.com/opentofu/opentofu/issues/599
+
+Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on DynamoDB.
+
+The main reasons for such a change could be summarized as follows:
+* Less resources to maintain.
+* Potentially reducing costs by eliminating usage of dynamo db.
+* One less point of failure (removing DynamoDB from the state management)
+
+The most important things that need to be handled during this implementation:
```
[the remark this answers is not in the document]
yottta: Really good point!
From a quick check, I see that some already implemented it but there are still some that didn't yet.
Added a warning about it e22385a328c24a07904fa422aad15849b57ff047

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,148 @@
...
+    * The generation of the MD5 checksum is already part of the [implementation](https://github.com/opentofu/opentofu/blob/6614782/internal/backend/remote-state/s3/client.go#L178).
+    * We just need to store it in a new file in the configured S3 bucket (file name: <state-file-path>-md5)
+* When the new attribute `use_lockfile` will exist **alongside** `dynamodb_table`, OpenTofu will:
+  * Acquire the lock in the S3 bucket;
+  * Get the digest of the statefile from S3. If missing:
+    * Acquire the lock in DynamoDB table; 
+    * Get the digest of the statefile from DynamoDB and migrate it to the S3 bucket;
+    * Release the lock from DynamoDB table;
+  * Perform the requested sub-command
+  * Release the lock from the S3 bucket;
```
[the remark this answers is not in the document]
yottta: Thanks. Added 7b88c6f84ddb90024fff88eec0fa7f8f3fa7ad6f

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,148 @@
...
+    dynamodb_table = "tofu_locking"
+    use_lockfile = true
+  }
+}
+```
+
+* When the new attribute `use_lockfile` exists and `dynamodb_table` is missing, OpenTofu will behave as normal:
+  * The only thing that will have to do extra, is to ensure that the MD5 file is there and if not, to create it.
+    * The generation of the MD5 checksum is already part of the [implementation](https://github.com/opentofu/opentofu/blob/6614782/internal/backend/remote-state/s3/client.go#L178).
+    * We just need to store it in a new file in the configured S3 bucket (file name: <state-file-path>-md5)
```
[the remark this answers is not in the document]
yottta: Checked. [This](https://github.com/opentofu/opentofu/blob/6614782e6da121c1e1464a9471d943c7cc939a6b/internal/backend/remote-state/s3/backend_state.go#L229-L235) part is handling that.
Also, in the [OpenTofu S3 docs](https://opentofu.org/docs/language/settings/backends/s3/), that's clearly stated:
image

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Updated this section due to not needed to handle this digest file anymore.
This should be handled only when the DynamoDB locking is enabled. That's due to having two sources of information when it comes to locking and it was used to ensure that the state object from the S3 bucket was not altered separately, without acquiring a lock.

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Removed some questions since are not needed anymore. More details in the [above comment](https://github.com/opentofu/opentofu/pull/2511#discussion_r1955828819).

rfc/20250211-s3-locking-with-conditional-writes.md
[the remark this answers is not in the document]
ollevche: We need not to forget to update the user-facing docs as well.

rfc/20250211-s3-locking-with-conditional-writes.md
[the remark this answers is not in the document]
ollevche: Yep, I meant we should add a comment to the RFC stating that it will not cause any problem when using multiple workspaces.

rfc/20250211-s3-locking-with-conditional-writes.md
ollevche: After some digging, I think this option is unnecessary even for dynamodb locking. This feature was introduced to workaround the eventual consistency nature of s3 (before 2020). It is checking if the read state file is the latest written one and we no longer need to check this since s3 become strongly consistent ([blogpost](https://aws.amazon.com/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/)) in 2020.

ollevche: This is something only slightly relevant for this RFC, so I think we can tackle it separately. Alternatively, we can leave it as is if we plan to deprecate dynamodb locking completely. Otherwise, removing md5 write / read from dynamodb locking will positively affect the performance.

rfc/20250211-s3-locking-with-conditional-writes.md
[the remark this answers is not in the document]
yottta: Added in 8ff0b16733c70782c209c3f5a023190a35f93bab

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Added in 38f1eb921775d9ac783fffe8c8cb3d2b11a67419

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,131 @@
...
+> [!WARNING]
+> 
+> By enabling the S3 locking and disabling the DynamoDB one, the digest from DynamoDB will become stale. This means that if it is desired to go back to the DynamoDB locking, the digest needs to be cleaned up or updated in order to allow the content integrity check to work.
+
+### Open Questions
+
+* Do we want to provide the option to have the lock objects into another bucket? This will break the feature parity.
+
+### Future Considerations
+If this feature will prove to have a high adoption rate, later, we might consider to deprecate the DynamoDB locking mechanism.
```
Gogotchuri, line 128: This reads as adoption rate by the OpenTofu users.
When in actuality to deprecate the DynamoDB locking we would need the adoption of the locking feature by the majority of S3 compatible storage services.
We should highlight that explicitly.

Gogotchuri (approved):
Other than the comment I left everything looks great!

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Good point! Thanks for the input. Fixed in 12adf6200ff60dfe12597285173e08990cf5b3cb

Gogotchuri (approved):

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+> [!WARNING]
+> 
+> When OpenTofu S3 backend is used with an S3 compatible provider, it needs to be checked that the provider is supporting conditional writes in the same way AWS S3 is offering.
+
+#### Migration paths
+##### I have no locking enabled
+In this case, the user can just add the new `use_lockfile=true` and run `tofu init -reconfigure`.
+
+##### I have DynamoDB locking enabled
+In case the user is having DynamoDB enabled, there are two paths forward:
```
ollevche, line 80:
```suggestion
In case the user has DynamoDB enabled, there are two paths forward:
```

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+1. Add the new attribute `use_lockfile=true` and run `tofu init -reconfigure`
+   * Later, remove the `dynamodb_table` attribute and run `tofu init -reconfigure` again
```
ollevche, line 82: > Later, remove the `dynamodb_table` attribute and run `tofu init -reconfigure` again

Not sure if I understand what "later" exactly means here. It may be also unclear for end-users. Could you please elaborate more on that one?

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+2. Add the new attribute `use_lockfile=true`, remove the `dynamodb_table` one and run `tofu init -reconfigure`
+
+OpenTofu recommends to have both locking mechanisms enabled for a limited amount of time and later remove the DynamoDB locking.
```
ollevche, line 85: The same question as above, why do we recommend having both locking mechanisms enabled for a limited amount of time?

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Migration paths
+##### I have no locking enabled
+In this case, the user can just add the new `use_lockfile=true` and run `tofu init -reconfigure`.
+
+##### I have DynamoDB locking enabled
+In case the user is having DynamoDB enabled, there are two paths forward:
+1. Add the new attribute `use_lockfile=true` and run `tofu init -reconfigure`
+   * Later, remove the `dynamodb_table` attribute and run `tofu init -reconfigure` again
+2. Add the new attribute `use_lockfile=true`, remove the `dynamodb_table` one and run `tofu init -reconfigure`
```
ollevche, line 83: Also, I am not sure what is the difference between these two approaches. I would prefer to stick to a single one or describe more when using the first or the second one is better.

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+  * For buckets with versioning enabled, if there's no current object version with the same name, or if the current object version is a delete marker, the write operation succeeds.
+* HTTP 412 (Precondition Failed) - Means that the locking object is already there. Therefore, the whole process should exit because the lock couldn't be acquired.
+* HTTP 409 (Conflict) - Means that the there was a conflict of concurrent requests. AWS recommends to retry the request in such cases, but we could just handle this similarly to the 412 case.
+
+#### Digest updates
+> [!NOTE]
+> Right now, when locking is enabled on DynamoDB, at the moment of updating the state object content, OpenTofu also writes an entry in DynamoDB with the MD5 sum of the state object.
+> The reason is to be able to check the integrity of the state object from the S3 bucket in a future run. This is done by reading the digest from DynamoDB and comparing it with the ETag attribute of the state object from S3. 
+
+By moving to the S3 based locking, OpenTofu will hold no other file for the digest of the state object, since the digest is kept in `ETag` header of the object.
```
ollevche, line 114: > since the digest is kept in `ETag` header of the object

I believe digest check was introduced to tackle the eventual consistency nature of s3 before 2020, so we can clearly state this is deprecated behavior and we don't need that at all. Unless, we also want to check if that object hasn't been updated, but that part we handle via a separate lock file. So I am not sure if checksum validation is something we want to keep.

WDYT?

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+> 
+> By enabling the S3 locking and disabling the DynamoDB one, the digest from DynamoDB will become stale. This means that if it is desired to go back to the DynamoDB locking, the digest needs to be cleaned up or updated in order to allow the content integrity check to work.
+
+### Open Questions
+
+* Do we want to provide the option to have the lock objects into another bucket? This will break the feature parity.
+
+### Future Considerations
+Later, DynamoDB based locking might be considered for deprecation once (at least) the following are true:
+* Conditional writes are implemented in the majority of the S3 compatible services.
```
ollevche, line 129: That's an interesting question. How do people handle locking with non-AWS S3 compatible storage? Do they use DynamoDB? I believe it is easier to switch to AWS S3 if you are already in use with DynamoDB service

ollevche (commented):
Left a few clarification questions to understand the proposal better

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: This would be a question for the community TBH.

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Reworked in 6c43f94d46a40ebf22b652cb08d42bd5ba7f4179.
Yes, this is the idea. Once only the S3 locking is in use, the digest of the state will not be handled anymore by OpenTofu.

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Added more details in 6c43f94d46a40ebf22b652cb08d42bd5ba7f4179.
Please let me know if that is making clearer what the pros and cons of each is are.

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Added more details in 6c43f94d46a40ebf22b652cb08d42bd5ba7f4179.
Please let me know if that is making clearer what the pros and cons of each is are.

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Added more details in 6c43f94d46a40ebf22b652cb08d42bd5ba7f4179.
Please let me know if that is making clearer what the pros and cons of each is are.

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,133 @@
...
+> [!NOTE]
+>
+> OpenTofu [recommends](https://opentofu.org/docs/language/settings/backends/s3/) to have versioning enabled for the S3 buckets used to store state objects.
+>
+> Acquiring and releasing locks will add a good amount of writes and reads to the bucket. Therefore, for a versioning-enabled S3 bucket, the number of versions for that object could grow significantly.
+> Even though the cost should be negligible for the locking objects, any user using this feature could consider configuring the lifecycle of the S3 bucket to limit the number of versions of an object.
+
+> [!WARNING]
+> 
+> When OpenTofu S3 backend is used with an S3 compatible provider, it needs to be checked that the provider is supporting conditional writes in the same way AWS S3 is offering.
```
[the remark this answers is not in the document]
yottta: Applied in 6c43f94d46a40ebf22b652cb08d42bd5ba7f4179

rfc/20250211-s3-locking-with-conditional-writes.md
yottta: Applied in 6c43f94d46a40ebf22b652cb08d42bd5ba7f4179

rfc/20250211-s3-locking-with-conditional-writes.md
ollevche: Yep, agree. Just wanted to share a thought

cam72cam: My understanding is that the only real option is to run with locking disabled.  AFAIK there's no way to separate the credentials for S3 from the credentials from Dynamo

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,136 @@
+# S3 Backend: Locking by using the recently released feature of conditional writes
+
+Issue: https://github.com/opentofu/opentofu/issues/599
+
+Considering the request from the ticket above, and the newly AWS released S3 feature, we can now have state locking without relying on DynamoDB.
+
+The main reasons for such a change could be summarized as follows:
+* Less resources to maintain.
+* Potentially reducing costs by eliminating usage of dynamo db.
+* One less point of failure (removing DynamoDB from the state management)
```
cam72cam, line 10: Additonal point:
* Easy for other S3 compatible services to implement and enable locking.

yottta: Added in a3139cf174b05fdcd85dab98b3726874d9a10c4f

ollevche (approved):

Gogotchuri (approved):

rfc/20250211-s3-locking-with-conditional-writes.md
```diff
@@ -0,0 +1,137 @@
...
+   * Letting both locking mechanisms enabled, ensures that nobody will acquire the lock regardless of having or not the latest configuration.
+2. Add the new attribute `use_lockfile=true`, remove the `dynamodb_table` one and run `tofu init -reconfigure` 
+   * **Caution:** when the configuration updated is executed from multiple places (multiple machines, pipelines on PRs, etc), you might get into issues where one outdated copy of the configuration is using DynamoDB locking and the one updated is using S3 locking. This could create concurrent access on the same state.
+   * Once the state is updated by using this approach, the state digest that OpenTofu was storing in DynamoDB (for data consistency checks) will get stale. If it is wanted to go back to DynamoDB locking, the old digest needs to be cleaned up manually.
+
+OpenTofu recommends to have both locking mechanisms enabled for a limited amount of time and later remove the DynamoDB locking. This is to ensure that the changes pushed upstream will be propagated into all of your places that the configuration could be executed from.
```
diofeher, line 89: Ok, now I understand the reason :)

PR #2517, opened 2025-02-13
TofuPipes experiment

abstractionfactory:
This is an experiment related to #2221 and #931. It simplifies wiring together multiple OpenTofu projects by allowing a project to consume the outputs of another projects like this:

```
tofu apply -var-upstream=/path/to/upstream/project
```

Internally, this works by executing `tofu -chdir /path/to/upstream/project output -json` and applying the received output the same way as `-var-file` would. The main limitation of this approach is that the upstream project must declare the exact same output names as the downstream project declares variables and any marks, such as sensitive, will not translate through the project boundary.

As a side-effect, it also adds `-var-command`, which serves as the implementation for `-var-upstream`.

## Alternatives

- The source code makes mention of a potential `-var-file=<(./scripts/vars.sh)` possibility, but this doesn't actually work.
- One could create Unix pipes and read from them, or write the contents (including sensitive values) to a file, but that's clunky and suboptimal.
- We could implement this as full-blown block in OpenTofu.

## Target Release

N/A, this is an experiment.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/extended.go
+++ b/internal/command/arguments/extended.go
@@ -228,8 +228,10 @@ func (o *Operation) Parse() tfdiags.Diagnostics {
 // desirable for the arguments package to handle the gathering of variables
 // directly, returning a map of variable values.
 type Vars struct {
-	vars     *flagNameValueSlice
-	varFiles *flagNameValueSlice
+	vars        *flagNameValueSlice
+	varFiles    *flagNameValueSlice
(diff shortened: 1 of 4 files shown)


cam72cam:
Can you describe why you would use this over `data "terraform_remote_state" {}`?  It seems like an expensive and potentially error prone method of accomplishing the same thing?

PR #2528, opened 2025-02-18
Upgrade aws sdk to a version that will give us access to s3 conditional writes arguments

yottta:
Part of #599.
This is necessary to get access to the `IfNotMatch` and `IfMatch` properties in `s3.PutObjectInput` introduced in [aws-sdk-go-v2/service/s3@v1.69.0](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/CHANGELOG.md#v1690-2024-11-25)

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/encryption/keyprovider/aws_kms/config.go
```diff
@@ -92,6 +92,10 @@ func (c Config) asAWSBase() (*awsbase.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+	var roles []awsbase.AssumeRole
```
yottta, line 95: `note`
Upgraded aws/s3 lib => rendered some tests in backend_complete_test.go to fail => forced upgrading hashicorp/aws-sdk-go-base/v2 => forced upgrading the way assumed role(s) are passed over.

go.mod
```diff
@@ -17,12 +17,12 @@ require (
 	github.com/apparentlymart/go-userdirs v0.0.0-20200915174352-b0c018a67c13
 	github.com/apparentlymart/go-versions v1.0.2
 	github.com/armon/circbuf v0.0.0-20190214190532-5111143e8da2
-	github.com/aws/aws-sdk-go-v2 v1.23.2
-	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.14.6
-	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.25.5
-	github.com/aws/aws-sdk-go-v2/service/kms v1.26.5
-	github.com/aws/aws-sdk-go-v2/service/s3 v1.46.0
-	github.com/aws/smithy-go v1.17.0
+	github.com/aws/aws-sdk-go-v2 v1.32.7
```
yottta, line 20: `note`
The new conditional writes feature was added in s3/1.69.0.
Due to a [breaking change](https://github.com/aws/aws-sdk-go-v2/issues/2370#issuecomment-1953308268) that was added in github.com/aws/aws-sdk-go-v2@1.24.0, all deps under the `aws-sdk-go-v2` namespace needs to be updated.

Since the s3 version we need is from 2024.11.25, All the other versions are the versions released right before the s3 one. Most of the versions are from 2024.11.18.

diofeher (dismissed):
LGTM!

Did not check the source code but wanted to ask: do we have tests for backend-remote-state/s3? I'm assuming your change didn't break them.

yottta:
> LGTM!
> 
> Did not check the source code but wanted to ask: do we have tests for backend-remote-state/s3? I'm assuming your change didn't break them.

Good point. I ran some tests, but only some specific tests. Your suggestion made me realise this and actually ran all of them. There was indeed one test that required to be fixed.
Now everything works well.
Screenshot 2025-02-18 at 13 27 52

internal/backend/remote-state/s3/backend.go
```diff
@@ -772,9 +772,9 @@ func (b *Backend) Configure(obj cty.Value) tfdiags.Diagnostics {
 	}
 
 	if value := obj.GetAttr("assume_role"); !value.IsNull() {
-		cfg.AssumeRole = configureNestedAssumeRole(obj)
+		cfg.AssumeRole = append(cfg.AssumeRole, configureNestedAssumeRole(obj))
```
cam72cam, line 775: The append here is a bit strange, as we know the field is not set.

IMO it would be more clear to create a new list inline.

yottta: Just to be clear, you mean something like
```
	var roles []awsbase.AssumeRole
	if value := obj.GetAttr("assume_role"); !value.IsNull() {
		roles = append(roles, configureNestedAssumeRole(obj))
	} else if value := obj.GetAttr("role_arn"); !value.IsNull() {
		roles = append(roles, configureAssumeRole(obj))
	}
	cfg.AssumeRole = roles
```
If yes, this was my first approach, but functionally, it's the same, but more verbose.

Or you mean something like this?
```
	if value := obj.GetAttr("assume_role"); !value.IsNull() {
		cfg.AssumeRole = []awsbase.AssumeRole{configureNestedAssumeRole(obj)}
	} else if value := obj.GetAttr("role_arn"); !value.IsNull() {
		cfg.AssumeRole = []awsbase.AssumeRole{configureAssumeRole(obj)}
	}
```

cam72cam: Something like your second example.  The code as written will only honor the first one it sees.  If we want to change that, we should address that in it's own PR as it will also require some documentation changes.

yottta: Done

yottta: Isn't it done this way because `role_arn` is deprecated? Meaning that `assume_role` is having higher priority in processing here?

Screenshot 2025-02-18 at 15 02 46

cam72cam: That's correct, I should have double checked the docs :)

cam72cam (approved):
Assuming all of the acceptance tests pass with real credentials, this LGTM!

internal/backend/remote-state/s3/backend.go
yottta: Glad that this is cleared out. Better to leave a clean trace behind :D

yottta:
> Assuming all of the acceptance tests pass with real credentials, this LGTM!

Yes, that was against AWS directly

PR #2543, opened 2025-02-25
Adds support for run block outputs in the test provider block

### LINKED ISSUE #1185: `tofu test` - allow provider blocks in tests to access variables and run outputs

Gogotchuri:
# Support for run block outputs in provider block.

### About the latest approach
We have iterated on this solution a few times, and the main goal was to ensure that the least amount of test module implementation details were leaked inside the core codebase.
The current solution:
- Prepares the `evalCtx` with pre-evaluated variables and built-in functions; 
- Embeds the prepared context inside the `testPoviderConfig` struct (`hcl.Body`) 
- And evaluates the expressions containing the `run` block references in the original body (using the prepared `evalCtx`). 
 
Attributes containing the `run` block references are replaced by attributes containing the evaluated value inside the `hcl.LiteralValueExpr`.

Resolves #1185 

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/command/test.go
```diff
@@ -507,6 +508,38 @@ func (runner *TestFileRunner) ExecuteTestFile(ctx context.Context, file *modulet
 	}
 }
 
+// transformTestFileProviderConfigs replaces run block variable references with their literal values
+func transformTestFileProviderConfigs(file *moduletest.File, evalCtx *hcl.EvalContext) tfdiags.Diagnostics {
```
cam72cam, line 512: I think this is an interesting approach, and avoids modifying the opentofu core for a testing specific workaround.  I'm curious if state access in the provider test block should be allowed.  I'm guessing not, but we should look into it.

internal/command/test.go
```diff
@@ -544,6 +577,16 @@ func (runner *TestFileRunner) ExecuteTestRun(ctx context.Context, run *moduletes
 		return state, false
 	}
 
+	evalCtx, evalDiags := getEvalContextForTest(runner.States, config, runner.Suite.GlobalVariables)
+	run.Diagnostics = run.Diagnostics.Append(evalDiags)
+
+	diags := transformTestFileProviderConfigs(file, evalCtx)
```
ollevche, line 583: I think this part of the code should be executed on a file level instead of run level. We would have the same results regardless of what run is being executed.

internal/command/test.go
```diff
@@ -507,6 +508,38 @@ func (runner *TestFileRunner) ExecuteTestFile(ctx context.Context, file *modulet
...
+	var diags tfdiags.Diagnostics
+	if !evalCtx.Variables["run"].IsNull() && len(evalCtx.Variables["run"].AsValueMap()) > 0 {
+		for _, provider := range file.Config.Providers {
+			body, ok := provider.Config.(*hclsyntax.Body)
+			if !ok {
+				break
+			}
+			attrs, diag := provider.Config.JustAttributes()
+			diags = diags.Append(diag)
+			for name, attr := range attrs {
```
ollevche, line 522: I agree with the downside you mentioned in the description regarding the duplication between the tofu test and tofu core code. 

I think we can tweak that part to defer the actual evaluation to a later phase. Instead of pre-evaluation of the expression here, we could introduce a wrapper around `hcl.Expression` which will have an additional `hcl.EvalContext` for tests internally. Then, we can override the `hcl.Expression.Value` function to include that test eval context as a parent (or vice versa) to extend the available values for this exact expression. This approach makes the evaluation process more similar to what we already have in the core codebase. 

What do you think?

Gogotchuri: That sounds interesting, I will look into it and let you know, Thanks a lot!

cam72cam: That's clever @ollevche.  You'd also have to filter the references for the Variables() impl.

ollevche: This looks like a good solution from the stability perspective, however the leakage of test code into the core codebase concerns me.

@Gogotchuri and I talked a bit about the wrapper idea in Slack. If I understood correctly, the main issue was with the validation of expression traversals, which needs to know about `run` references. However, as @cam72cam mentioned previously, run references should be filtered out from `Expression.Variables` call result. I am wondering if that part was handled in the implementation. I may be wrong, but I think if we do filter out those special traversals, they shouldn't be validated later. Could you please confirm filtering part was addressed or share the implementation?

Gogotchuri: @ollevche I've thought more about the implementation. In my opinion, the test code now included in the core codebase, should not cause any issues or add a maintenance burden because the logic is isolated from the core command flows. 
There are only two points of contact, where the conditional branching occurs for the test and neither can be entered from other commands (take a look at #parse_ref.go and #eval.go).
You can take a look at the same two files, and see that run references are handled there, so no need for additional pre-filtering.

In the long run, extending the testing framework functionality will require including some of the testing parts in the core codebase either way.

internal/command/test.go
Gogotchuri: This comment is obsolete for the new approach. Resolving.

internal/command/test.go
ollevche: Agree with @cam72cam here. I am okay with the approach, however, it would be nice to still separate those "contexts" (core and test executions).

> In my opinion, the test code now included in the core codebase, should not cause any issues or add a maintenance burden because the logic is isolated from the core command flows.

I think this is more about simplicity, because to me it is not obvious from the first read whether or not those commands are using newly provided functions of the interface. Separation would make it obvious.

Gogotchuri: This thread is obsolete now that I have implemented yet another approach. Resolving.

internal/command/test.go
```diff
@@ -1054,6 +1065,40 @@ func (runner *TestFileRunner) Cleanup(ctx context.Context, file *moduletest.File
...
+	vars, varDiags := buildInputVariablesForTest(run, file, config, globals, evalCtx)
+	diags = diags.Append(varDiags)
+	if diags.HasErrors() {
+		return evalCtx, diags
+	}
+	for name, val := range vars {
+		if val == nil {
+			continue
+		}
+		varMap := evalCtx.Variables["var"].AsValueMap()
```
cam72cam, line 1089: The way I typically approach this is constructing a map[string]cty.Value outside of the loop, filling it within the loop, and then constructing the ObjectVal after the loop.

internal/configs/test_provider_body.go
```diff
@@ -0,0 +1,114 @@
...
+	evalCtx      *hcl.EvalContext
+}
+
+func (c testProviderBody) Content(schema *hcl.BodySchema) (*hcl.BodyContent, hcl.Diagnostics) {
+	content, diags := c.originalBody.Content(schema)
+	if diags.HasErrors() {
+		return nil, diags
+	}
+	attrs := content.Attributes
+	for name, attr := range attrs {
```
cam72cam, line 27: Looks like this shares a lot of code between Content and PartialContent.  Perhaps a common function for transforming hcl.BodyContent would help.

internal/configs/test_provider_body.go
```diff
@@ -0,0 +1,114 @@
...
+		vars := attr.Expr.Variables()
+		containsRunRef := false
```
cam72cam, line 29: Given what we are trying to do here, I think we should try to evaluate and replace all attribute expressions with computed values.  All the variables and run data should be available at this point.  This would also simplify some of the other code in this PR.

internal/configs/test_provider_body.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		attr, valueDiags := c.getLiteralAttr(attr)
+		diags = append(diags, valueDiags...)
+		if diags.HasErrors() {
+			continue
+		}
+		attrs[name] = attr
+	}
+	return content, diags
```
cam72cam, line 47: It looks like we are not iterating through the "blocks" within the BodyContent, are those allowed in the test providers?

internal/command/test.go
Gogotchuri: It's a mistake, the map should have been constructed outside. Thanks for catching it.

internal/configs/test_provider_body.go
Gogotchuri: I thought not, I'm not quite sure.
Should we implement block evaluation additionally just in case?

cam72cam: Lots of providers do support and expect configuration blocks, we should support them here.

Gogotchuri: Done

ollevche (approved):
Looks great, thank you! I like how clear the final solution is. The only thing I was confused by - I didn't expect `EvalContext.Variables` to already include `run`s.

Do you plan to extend the docs here or as a separate PR?

Gogotchuri:
I'll change the naming.
About the variables including "run" that part was written by Christian in older PR.
`getEvalContextForTest` is doing that part. (Which was previously only used to prepare variables block defined inside run block)

ollevche:
Nope, I don't think we need to do anything about this. This is more of a comment for future reviewers :)

ollevche (approved):

PR #2565, opened 2025-03-05
fix: remove elapsed_seconds check on the golden reference

diofeher:
In slower environment where our tests are run, like `linux_arm`, this golden reference test can break on the "elapsed_seconds" field. To make sure we don't have flaky tests, I'm removing this field from the checks.

https://github.com/opentofu/opentofu/actions/runs/13627724097/job/38088463026

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/command/command_test.go
```diff
@@ -1153,21 +1153,38 @@ func checkGoldenReference(t *testing.T, output *terminal.TestOutput, fixturePath
...
 	}
+
 	if diff := cmp.Diff(wantLineMaps, gotLineMaps); diff != "" {
 		t.Errorf("wrong output lines\n%s\n"+
 			"NOTE: This failure may indicate a UI change affecting the behavior of structured run output on TFC.\n"+
 			"Please communicate with Terraform Cloud team before resolving", diff)
 	}
 }
+
+func deleteTimestampField(fieldMap map[string]interface{}, rootField, field string) map[string]interface{} {
```
yottta, line 1179: `suggestion`
If we really want to go this path, recommend doing it cleaner, with an early return instead of of this level of nested code.

diofeher: Good one, didn't notice haha thanks again for the review!

yottta (approved):

yottta (approved):

PR #2566, opened 2025-03-05
Add variable deprecation impl

yottta:
Resolves # 

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/tofu/eval_variable.go
```diff
@@ -477,3 +477,33 @@ You can correct this by removing references to sensitive values, or by carefully
...
+	if val.IsNull() {
+		log.Printf("[TRACE] evalVariableDeprecation: variable %s is marked as deprecated null value given", addr)
+		return nil
+	}
+	log.Printf("[TRACE] evalVariableDeprecation: usage of deprecated variable %q detected", addr)
+	// Any deprecated
+	return diags.Append(&hcl.Diagnostic{
+		Severity: hcl.DiagWarning,
+		Summary:  fmt.Sprintf(`The variable %q is marked as deprecated by module author`, config.Name),
+		Detail:   config.Deprecated,
```
ollevche, line 506: I missed that part in the RFC example of variable deprecation, however it is slightly different for module outputs:

```
│ This value is derived from module.mod.this_is_my_output, which is
│ deprecated with the following message:
|
| This output will be removed on 2024-12-31. Use another_output instead.
```

I believe explicitly saying this message is from the deprecation configuration (and not from OpenTofu) is a better approach. So I suggest to tweak the messaging so it is more similar to module outputs:

```
│ This variable is marked as deprecated with the following message:
|
| This variable will be removed on 2024-12-31. Use another_variable instead.
```

WDYT?

internal/tofu/node_variable_reference.go
```diff
@@ -146,6 +146,7 @@ func (n *nodeVariableReferenceInstance) ModulePath() addrs.Module {
 func (n *nodeVariableReferenceInstance) Execute(ctx EvalContext, op walkOperation) tfdiags.Diagnostics {
 	log.Printf("[TRACE] nodeVariableReferenceInstance: evaluating %s", n.Addr)
 	diags := evalVariableValidations(n.Addr, n.Config, n.Expr, ctx)
+	diags = diags.Append(evalVariableDeprecation(n.Addr, n.Config, n.Expr, ctx, false)) // TODO provide this bool from cli arg
```
ollevche, line 149: I believe it should work for `tofu validate` as well. Could you please confirm with a manual or auto test?

ollevche (commented):
Thank you for implementing this! I was thinking about the same approach, it looks pretty reasonable. I left a few questions / ideas, but the PR already looks good.

internal/configs/named_values.go
```diff
@@ -38,9 +38,11 @@ type Variable struct {
 	ParsingMode VariableParsingMode
 	Validations []*CheckRule
 	Sensitive   bool
+	Deprecated  string
```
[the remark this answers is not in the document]
yottta: Thanks a lot for the input! Good catches with these 2 prompt parts.
When it comes to the prompts, I would go with the first approach. The second option, where we could skip the deprecated variables, could block some users from using the configuration correctly, so I am not sure about it.

As for the `tofu show --json planfile`, yeah, totally agree.

internal/tofu/eval_variable.go
```diff
@@ -477,3 +477,33 @@ You can correct this by removing references to sensitive values, or by carefully
...
 	}, diags
 }
+
+// evalVariableDeprecation checks a variable "deprecated" message
+func evalVariableDeprecation(addr addrs.AbsInputVariableInstance, config *configs.Variable, expr hcl.Expression, ctx EvalContext, onlyLocal bool) (diags tfdiags.Diagnostics) {
+	if !config.DeprecatedSet {
+		log.Printf("[TRACE] evalVariableDeprecation: variable %s is having no deprecation configured", addr)
+		return nil
+	}
+	if onlyLocal && strings.HasPrefix(config.DeclRange.Filename, ".terraform") {
```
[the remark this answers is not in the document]
cam72cam: I'd look at the internal/configs/Config.SourceAddr field to see if it's a local address or not.

internal/configs/named_values.go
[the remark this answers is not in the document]
yottta: When it comes to the output, is this what are you thinking about?
```
"root_module": {
      "module_calls": {
        "modcall": {
          "source": "./mod1",
          "expressions": {
            "this_is_my_variable": {
              "constant_value": "given value from modcall"
            }
          },
          "module": {
            "variables": {
              "this_is_my_variable": {
                "default": "default value",
                "description": "This is a variable for the old way of configuring things.",
                "deprecated": "This variable will be removed on 2024-12-31. Use another_variable instead."
              }
            }
          }
        }
      }
    }
```

internal/tofu/eval_variable.go
```diff
@@ -477,3 +477,33 @@ You can correct this by removing references to sensitive values, or by carefully
 		FailureMessage: errorMessage,
 	}, diags
 }
+
+// evalVariableDeprecation checks a variable "deprecated" message
+func evalVariableDeprecation(addr addrs.AbsInputVariableInstance, config *configs.Variable, expr hcl.Expression, ctx EvalContext, onlyLocal bool) (diags tfdiags.Diagnostics) {
```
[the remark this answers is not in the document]
yottta: Thanks a lot for the suggestion. Wanted to use enums, but quite confused about the package where to define such an enum.
Removed the changes in 4a7e2a201b5937bcfe16b03c8cc8ffa78a9a9f1d.
Will be included in the `deprecation-control-flag` branch.

internal/tofu/eval_variable.go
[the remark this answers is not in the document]
yottta: Removed. Please refer to [this comment](https://github.com/opentofu/opentofu/pull/2566#discussion_r1984696984).

internal/tofu/eval_variable.go
yottta: Yep, we missed this specificity from the RFC. I totally agree that it's better to give better context to the users.
Changed in 000434c5cde89696960449ece0160595ac64c5c5
Screenshot 2025-03-07 at 11 18 44

internal/tofu/node_variable_reference.go
yottta: Yes, it works well.
Screenshot 2025-03-07 at 11 23 57

internal/tofu/eval_variable.go
```diff
@@ -477,3 +477,33 @@ You can correct this by removing references to sensitive values, or by carefully
...
+	// if the variable is not given in the module call, do not show a warning
+	if expr == nil {
+		log.Printf("[TRACE] evalVariableDeprecation: variable %s is marked as deprecated but is not used", addr)
+		return nil
+	}
+	val := ctx.GetVariableValue(addr)
+	if val.IsNull() {
+		log.Printf("[TRACE] evalVariableDeprecation: variable %s is marked as deprecated null value given", addr)
+		return nil
+	}
```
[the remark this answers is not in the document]
ollevche: Yes, that's exactly what I meant. This will be inconsistent with other parts of tofu (including module output and resource fields deprecation, for example), however, this feature will be more useful. It is possible to set a ternary expression, which could result in `null` and silence the warning. This will be an issue once the variable is actually removed.

internal/tofu/node_variable_reference.go
ollevche: Thanks for checking it!

internal/tofu/eval_variable.go
ollevche: Looks good, thanks!

internal/configs/named_values.go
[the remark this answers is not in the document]
ollevche: Thanks! JSON outputs look good to me. However, I am not sure if I understand how the deprecated message is shown on interactive input. I would use similar approach as in the warning, where we say "Variable is marked as deprecated with the following message: <MESSAGE>".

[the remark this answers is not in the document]
yottta: Thanks. Was not sure about it either.
I applied that in a4c9d758df57f524f11d288127096a6353e8d5db.
Now, it looks like this.
Screenshot 2025-03-07 at 18 17 38

internal/command/meta_config.go
```diff
@@ -163,11 +163,19 @@ func (m *Meta) getInput(ctx context.Context, variable *configs.Variable) (string
...
 		return "", fmt.Errorf("input is disabled")
 	}
+
+	varDesc := variable.Description
+	switch {
+	case variable.Description != "" && variable.DeprecatedSet:
+		varDesc = fmt.Sprintf("%s.\nVariable is marked as deprecated with the following message: %s", variable.Description, variable.Deprecated)
+	case variable.DeprecatedSet:
+		varDesc = fmt.Sprintf("Variable is marked as deprecated with the following message: %s", variable.Deprecated)
+	}
```
ollevche, line 173: nit: it seems like a duplicate of the function from `backend_local.go`. We could move it to `configs.Variable` and reuse in both places.

internal/configs/named_values.go
```diff
@@ -38,9 +38,11 @@ type Variable struct {
...
 
 	DescriptionSet bool
 	SensitiveSet   bool
+	DeprecatedSet  bool
```
ollevche, line 45: Does it make sense to set `deprecated` with an empty string? I am not sure if we need `DeprecatedSet` property.

ollevche (commented):
Looks good, thank you! I left a few nit comments and I would like one more review before merging this one.

internal/configs/named_values.go
yottta: To be honest, didn't want to add `DeprecatedSet` either. Didn't really see the benefit of it, but I wanted to have it here just to spark the conversation about the standards here.
This seems more of a standards than an actual functional approach. I was looking at the `SensitiveSet` too and couldn't find any usage of that other than setting it again later to another attribute when merged.

I would like to remove the `DeprecatedSet` too and since you suggest the same, I will do that happily.

ollevche: It makes sense if we want to distinguish between the empty value and unset value. For example, we could differentiate whether sensitive was absent completely or set to false. Not sure if we use it in practice, but it still makes sense. For the deprecated field, I am not sure if we should allow `deprecated = ""`.

internal/tofu/eval_variable.go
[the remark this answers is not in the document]
yottta: Looking into this more carefully, not sure that this part makes sense to me:
> It is possible to set a ternary expression, which could result in null and silence the warning. This will be an issue once the variable is actually removed.

You mean that the variable value can be a ternary expression that could result in "null"? If so, shouldn't that result in a go-cty type of the type defined for the variable then `val.IsNull()` would actually be more appropriate?

Also, digging more into go-cty to get a better understanding of the `cty.NilVal` and `val.IsNull()`, if I understand this correctly, `cty.NilValue` shouldn't be really able to reach this part since it's mainly meant [to be used to be returned together with errors/diags](https://github.com/zclconf/go-cty/blob/main/cty/value.go#L55-L78), which should return way before getting in here.

Maybe I am misunderstanding its usage, so feel free to correct me.

Meanwhile, changed the condition to guard against a situation where the val is indeed `cty.NilVal` and we call `IsNull()` on it.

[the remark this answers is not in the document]
yottta: Now, the problem is that when running `tofu apply` the warning will be shown twice: once for the `plan` phase and one for the `apply` phase.
Screenshot 2025-04-07 at 09 02 35

I am not sure that I like it 🤔 
What do you think about this?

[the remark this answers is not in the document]
cam72cam: Given that folks split plan and apply in many CI/CD pipelines, I think that the "extra" warning is not an issue IMO.

internal/configs/named_values.go
```diff
@@ -123,6 +125,19 @@ func decodeVariableBlock(block *hcl.Block, override bool) (*Variable, hcl.Diagno
 		v.SensitiveSet = true
 	}
 
+	if attr, exists := content.Attributes["deprecated"]; exists {
+		valDiags := gohcl.DecodeExpression(attr.Expr, nil, &v.Deprecated)
+		if strings.TrimSpace(v.Deprecated) == "" {
```
cam72cam, line 130: Missing a check for valDiags.HasErrors()

yottta: Good catch. Thanks

internal/configs/named_values.go
```diff
@@ -123,6 +125,19 @@ func decodeVariableBlock(block *hcl.Block, override bool) (*Variable, hcl.Diagno
...
+	if attr, exists := content.Attributes["deprecated"]; exists {
+		valDiags := gohcl.DecodeExpression(attr.Expr, nil, &v.Deprecated)
+		diags = append(diags, valDiags...)
+		if !valDiags.HasErrors() && strings.TrimSpace(v.Deprecated) == "" {
+			diags = append(diags, &hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid deprecated value",
+				Detail:   fmt.Sprintf("The variable name %q 'deprecated' field is defined but is empty.", v.Name),
+				Subject:  &block.LabelRanges[0],
+			})
```
ollevche, line 137: We had a small discussion around the messaging in https://github.com/opentofu/opentofu/pull/2633#discussion_r2056387265

I would suggest to have similar messages in both cases (variables and outputs). This is from the deprecation of outputs:

```
Summary: "Invalid `deprecated` attribute"
Detail: The "deprecated" argument must not be empty, and should provide instructions on how to migrate away from usage of this deprecated output value.`
```

internal/configs/named_values.go
```diff
@@ -123,6 +125,19 @@ func decodeVariableBlock(block *hcl.Block, override bool) (*Variable, hcl.Diagno
 		v.SensitiveSet = true
 	}
 
+	if attr, exists := content.Attributes["deprecated"]; exists {
+		valDiags := gohcl.DecodeExpression(attr.Expr, nil, &v.Deprecated)
+		diags = append(diags, valDiags...)
+		if !valDiags.HasErrors() && strings.TrimSpace(v.Deprecated) == "" {
```
ollevche, line 131: FYI: tweaked module output deprecation check to be the same as in here, thanks!

ollevche (approved):

internal/configs/named_values.go
yottta: Applied in 68df3ff06c44a207e60b6ccf6c31af26594b6cfe.
Thanks for the suggestion.

internal/tofu/eval_variable.go
```diff
@@ -477,3 +477,29 @@ You can correct this by removing references to sensitive values, or by carefully
 		FailureMessage: errorMessage,
 	}, diags
 }
+
+// evalVariableDeprecation checks if a variable is deprecated and if so it returns a warning diagnostic to be shown to the user
+func evalVariableDeprecation(addr addrs.AbsInputVariableInstance, config *configs.Variable, expr hcl.Expression, ctx EvalContext) tfdiags.Diagnostics {
+	if config.Deprecated == "" {
+		log.Printf("[TRACE] evalVariableDeprecation: variable %s is having no deprecation configured", addr)
```
cam72cam, line 484: Nit: "is having no" is not conventional English.  "does not have" is more standard here.

internal/tofu/eval_variable.go
```diff
@@ -477,3 +477,29 @@ You can correct this by removing references to sensitive values, or by carefully
...
+		return nil
+	}
+	// if the variable is not given in the module call, do not show a warning
+	if expr == nil {
+		log.Printf("[TRACE] evalVariableDeprecation: variable %s is marked as deprecated but is not used", addr)
+		return nil
+	}
+	val := ctx.GetVariableValue(addr)
+	if val == cty.NilVal {
```
cam72cam, line 493: I wonder if this check makes sense.

```hcl
module "foo" {
  field = null
}```

still represents a value and usage (of null) for the input

cam72cam:
Overall this approach looks good.  I've left a few small questions nits.

My main concern is that we don't have at least one context plan or context apply test in the PR.  There are a lot of places where we could be accidentally dropping warning diagnostics and should make sure at least a few examples work as expected.

internal/tofu/eval_variable.go
yottta: This is based on the information from the [RFC](https://github.com/opentofu/opentofu/pull/2180/files#diff-f75ee43ad75dd61d95c52383a325c31bee32a3b22df43e8a523dc6b2dba6f4d7R65).
Maybe we could change that if we really want to warn on these cases.

yottta:
> Overall this approach looks good. I've left a few small questions nits.
> 
> My main concern is that we don't have at least one context plan or context apply test in the PR. There are a lot of places where we could be accidentally dropping warning diagnostics and should make sure at least a few examples work as expected.

Added a new test in cf9a7d23760a7b5c2af120efd82d532454842094. Please let me know if that is enough or if I could cover this better.

PR #1070, opened 2024-01-05
Updating install instructions to include the installer script

janosdebugs:
Resolves #1035.

To be merged when opentofu/get.opentofu.org#1 is merged.

## Target Release

1.6.0

PR #2590, opened 2025-03-13
website/docs/intro/install/examples/verify-cosign: Fix quotes

travier:
Do not quote the '*' glob pattern so that the shell can interpret it.

Fixes: https://github.com/opentofu/opentofu/pull/1070

## Target Release

1.10.0

## Checklist

  - I have not done that yet but this is a fairly small change. Will do later if needed


website/docs/intro/install/examples/verify-cosign.sh
```diff
@@ -4,7 +4,7 @@ IDENTITY="https://github.com/opentofu/opentofu/.github/workflows/release.yml@ref
 cosign \
     verify-blob \
     --certificate-identity "${IDENTITY}" \
-    --signature "tofu_*.sig" \
-    --certificate "tofu_*.pem" \
+    --signature "tofu_"*".sig" \
```
cam72cam, line 7: Should the quotes be removed all-together?

travier: That's also an option :). I can do it if you prefer.

travier: yes, we have the `OPENTOFU_VERSION_MAJORMINOR=` variable above but it's missing the patch number for we can't use it directly and we would need to add another variable for this.

travier:
Updated

apparentlymart (approved):
This looks good to me. Thanks! :tada:

PR #2595, opened 2025-03-14
Encryption improve error messages

### LINKED ISSUE #2546: Improve the error messages from the encryption setup
Several topics to handle here:
* Message when multiple methods are tried for decryption and all are failing. The wrongly formatted error can be seen also in [this discussion](https://github.com/orgs/opentofu/discussions/2539).
    Image
* Handle the `TODO`s from `internal/encryption` that specifically point to poorly handled error messages.

Before starting with this, we need to compile a list with all the existing places from the `internal/encryption` package that are pointing to better error messages.

Klopklopi:
Resolves #2546 

## Target Release

1.10.0

## Checklist

### Go checklist

 

Made a mistake on the previous PR, sorry for the duplicate


internal/encryption/keyprovider.go
```diff
@@ -147,7 +147,8 @@ func setupKeyProvider(enc *config.EncryptionConfig, cfg config.KeyProviderConfig
 				Severity: hcl.DiagError,
 				Summary:  "Circular reference detected",
 				// TODO add the stack trace to the detail message
-				Detail: fmt.Sprintf("Cannot load %s due to circular reference between key providers.", addr),
+				// not sure at all, couldn't test the circular dep stack was always empty
+				Detail: fmt.Sprintf("Cannot load %s due to circular reference between key providers. Stack trace %s", addr, stack),
```
yottta, line 151: Indeed, that will not work as expected.
You could make use of a function like this:
```

func keyProvidersStack(stack []config.KeyProviderConfig) ([]string, hcl.Diagnostics) {
	res := make([]string, len(stack))
	var diags hcl.Diagnostics
	for i, cfg := range stack {
		addr, diag := cfg.Addr()
		diags = diags.Extend(diag)
		if diag.HasErrors() {
			res[i] = "<unknown>"
			continue
		}
		res[i] = string(addr)
	}
	return res, diags
}
```

And then, before this `return` statement, you can do this:
```
			addr, diags := keyprovider.NewAddr(cfg.Type, cfg.Name)
			stackAddrs, diag := keyProvidersStack(append(stack, cfg))
			diags = diags.Extend(diag)
```
And this line will become
```
				Detail: fmt.Sprintf("Cannot load %s due to circular reference between key providers. Stack trace %s", addr, strings.Join(stackAddrs, " -> ")),
```

Ending up in something similar to this. _Note: the message presented is not a real use case, that cannot happen._
Screenshot 2025-03-17 at 12 13 17

Klopklopi: Thank you for the feedback ! I was wondering if it was a problem if we couldn't properly test the circular reference ?

yottta: We could have done this, but right now the key provider ([XOR](https://github.com/opentofu/opentofu/tree/main/internal/encryption/keyprovider/xor)) that could have been used to reproduce this is not included in the default allowed providers.
I tested this with some slight temp modifications to the source code.

Klopklopi:
Made some changes on the tests, but I'm not sure if the type change on the compliance_test was the best to get around my problem

yottta:
> Made some changes on the tests, but I'm not sure if the type change on the compliance_test was the best to get around my problem

Great! Try adding a new test case in the `aesgcm/complience_test.go#TestCompliance` test to get that covered.

Klopklopi:
When you mean to add a test case for the `aesgcm/complience_test.go` file, do you mean a test to verify the type of the error ?

yottta:
> When you mean to add a test case for the `aesgcm/complience_test.go` file, do you mean a test to verify the type of the error ?

That structure used in `aesgcm/complience_test.go#TestCompliance` is supporting configuring that the test is expecting an error or not.
If the error returned is not a `hcl.Diagnostics` the test will fail in [here](https://github.com/opentofu/opentofu/pull/2595/files#diff-b6cd88494caef5758baa3c33346d5b1e6b0c9f3efd4ce3d9a1700193f9969aa4L314), where the error type assertions is performed.

Hopefully I understood correctly your concern here:

> Made some changes on the tests, but I'm not sure if the type change on the compliance_test was the best to get around my problem

If not, please let me know.

Klopklopi:
Maybe my statement was a bit confusing I'm sorry... 

I was talking about [this line](https://github.com/opentofu/opentofu/pull/2595/commits/d65bf375edd85bd1398b0a40fb45b77278353546#diff-b6cd88494caef5758baa3c33346d5b1e6b0c9f3efd4ce3d9a1700193f9969aa4R314), I wasn't sure if this type change (`*method.ErrInvalidConfiguration` to `hcl.Diagnostics`) was the good way to make sure the compliance test in `aesgcm` were passing correctly.

yottta:
> Maybe my statement was a bit confusing I'm sorry...
> 
> I was talking about [this line](https://github.com/opentofu/opentofu/pull/2595/commits/d65bf375edd85bd1398b0a40fb45b77278353546#diff-b6cd88494caef5758baa3c33346d5b1e6b0c9f3efd4ce3d9a1700193f9969aa4R314), I wasn't sure if this type change (`*method.ErrInvalidConfiguration` to `hcl.Diagnostics`) was the good way to make sure the compliance test in `aesgcm` were passing correctly.

Now I am thinking that my explanation was confusing, sorry. We are talking about the same thing.
That line change is affecting multiple tests, including the ones in `aesgcm/complience_test.go#TestCompliance`. You can check if there is at least one test that is actually reaching that path and if the validation is working well there. If there is no test, i would suggest to add one, where you can just configure a wrong length of the encryption/decryption key, rendering the `Build()` call to fail and return a `hcl.Diagnostics`. This way, you can verify that the change that you did there is actually compliant with the actual validation that it's in there.

Klopklopi:
Ok I think I understood now, thank you for the explanation ! Also, there is a e2e test failing because of the stderr is formated differently than before, should I try to tackle this problem too ?

yottta:
> Ok I think I understood now, thank you for the explanation ! Also, there is a e2e test failing because of the stderr is formatted differently than before, should I try to tackle this problem too ?

If it's because of that, yes, please. If you need help with investigating that test, please let me know. I will try to find some time for it.

Klopklopi:
Added a sanitize function in `command/e2etest/encryption_test.go` to clean the stderr output. Waiting for your feedbacks, thanks  in advance :p.

internal/encryption/keyprovider.go
```diff
@@ -124,6 +125,21 @@ func setupKeyProviders(enc *config.EncryptionConfig, cfgs []config.KeyProviderCo
 	return kpData.hclEvalContext("key_provider"), diags
 }
 
+func keyProvidersStack(stack []config.KeyProviderConfig) ([]string, hcl.Diagnostics) {
```
yottta, line 128: `suggestion`
Would be better for the readability of the code to have this function moved at the bottom of the file. This is more of a personal preference, so that's why I tagged this as `suggestion`.

internal/encryption/base.go
```diff
@@ -230,11 +231,15 @@ func (base *baseEncryption) decrypt(data []byte, validator func([]byte) error) (
...
-	for _, err := range errs {
-		errMessage += err.Error() + sep
-		sep = "\n"
-	}
+	//errMessage := "decryption failed for all provided methods: "
+	//sep := ""
+	//for _, err := range errs {
+	//	errMessage += err.Error() + sep
+	//	sep = "\n"
+	//}
```
yottta, line 239:
```suggestion
```

The comment above can be removed too.

internal/command/e2etest/encryption_test.go
```diff
@@ -53,6 +54,17 @@ func (r tofuResult) StderrContains(msg string) tofuResult {
 	return r
 }
 
+func SanitizeStderr(result *tofuResult) {
```
yottta, line 57: `improvement`
This is indeed one way to do it. Though, not really agree with changing the value of the `*tofuResult` in the process.

One more thing, would be better to move this function `strip_ansi.go`, and reuse the already existing function in there (`stripAnsi`).

internal/command/e2etest/encryption_test.go
```diff
@@ -203,7 +215,10 @@ func TestEncryptionFlow(t *testing.T) {
 		apply().Failure().StderrContains("decryption failed for state")
 		requireEncryptedState()
 
-		applyPlan(encryptedPlan).Failure().StderrContains("decryption failed: cipher: message authentication failed")
+		tofuResultError := applyPlan(encryptedPlan).Failure()
+		SanitizeStderr(&tofuResultError)
+		tofuResultError.StderrContains("decryption failed: cipher: message authentication failed")
```
yottta, line 220: `suggestion`
As a continuation of the comment on `SanitizeStderr`, once you have a new function created to handle the sanitisation of the stderr, would be better to call that func inside `StderrContains`. That way, all the testing structure in here will be able to make use of it.

yottta (commented):
Thanks for your effort and time put into this. Looks great! 👏 
Left some comments and there is one more thing that should be taken care of in this PR.
Please check [ErrInvalidConfiguration](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/errors.go#L74) usage. It seems that after the refactor, its usage is limited to tests only. This is raising another question: [why those tests](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/aesgcm/config_test.go#L18) are not failing since the `Build` return type was changed?

internal/command/e2etest/encryption_test.go
Klopklopi: Oh I didn't check if a strip ansi method was already implemented (though it was in the same folder...) Thank you 😅.

Klopklopi:
I assume that [the tests](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/aesgcm/config_test.go#L18) are not failing because the [error type is any](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/aesgcm/config_test.go#L22) ?

yottta:
> I assume that [the tests](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/aesgcm/config_test.go#L18) are not failing because the [error type is any](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/aesgcm/config_test.go#L22) ? Or because the type changes in [this condition](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/aesgcm/config_test.go#L133C1-L134C1), the As function changes the type of the error.

Yeah, you're right. Didn't look close into the test function. `any` can get everything so yeah, `errors.As` always returns `true` because of it.

internal/encryption/methods.go
```diff
@@ -77,14 +77,12 @@ func setupMethod(enc *config.EncryptionConfig, cfg config.MethodConfig, meta key
...
-		return nil, diags.Append(&hcl.Diagnostic{
-			Severity: hcl.DiagError,
-			Summary:  "Encryption method configuration failed",
-			Detail:   err.Error(),
-		})
+	m, methodConfigDiags := methodConfig.Build()
+	diags = diags.Extend(methodConfigDiags)
+
+
+	if diags.HasErrors() {
```
yottta, line 84:
```suggestion
	diags = diags.Extend(methodConfigDiags)
	if diags.HasErrors() {
```

internal/encryption/method/aesgcm/config_test.go
```diff
@@ -8,18 +8,17 @@ package aesgcm
 import (
 	"bytes"
 	"errors"
+	"github.com/hashicorp/hcl/v2"
```
yottta, line 11: `goimports -w internal/encryption/method/aesgcm/config_test.go`

yottta: And for other changed files as well please.

internal/encryption/base.go
```diff
@@ -230,11 +231,9 @@ func (base *baseEncryption) decrypt(data []byte, validator func([]byte) error) (
 	}
 
 	// This is good enough for now until we have better/distinct errors
-	errMessage := "decryption failed for all provided methods: "
-	sep := ""
-	for _, err := range errs {
-		errMessage += err.Error() + sep
-		sep = "\n"
-	}
+
```
yottta, line 234:
```suggestion
```

yottta: And please remove the comment above too.

yottta (commented):
Small nitpicks added.
Also, [ErrorInvalidConfiguration](https://github.com/opentofu/opentofu/blob/d437ae0427c55a5a77acff7a630f9dee60b6d555/internal/encryption/method/errors.go#L74-L87) can be removed altogether now.

yottta (approved):
Great! Thanks a lot for the quick actions. This looks good now FMPOV.
This is ok from my side, but let's wait for another review from one more core maintainer before merging this.

internal/encryption/methods.go
```diff
@@ -77,14 +77,10 @@ func setupMethod(enc *config.EncryptionConfig, cfg config.MethodConfig, meta key
...
 
-	m, err := methodConfig.Build()
-	if err != nil {
-		// TODO this error handling could use some work
-		return nil, diags.Append(&hcl.Diagnostic{
-			Severity: hcl.DiagError,
-			Summary:  "Encryption method configuration failed",
-			Detail:   err.Error(),
-		})
+	m, methodConfigDiags := methodConfig.Build()
```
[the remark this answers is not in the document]
Klopklopi: I will try to implement this when I have time, thanks for the feedback !

cam72cam:
@Klopklopi just reaching out, do you think you will have time to work on this?

Klopklopi:
@cam72cam hey, yeah I forgot about this so yes I will have some free time to work on the issue this week :)

internal/encryption/methods.go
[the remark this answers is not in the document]
Klopklopi: I tried to implement a "smart" way to convert error to diags. Hope this will cover all the errors that can be raised in this package.

internal/encryption/base.go
```diff
@@ -229,12 +230,8 @@ func (base *baseEncryption) decrypt(data []byte, validator func([]byte) error) (
...
 
-	// This is good enough for now until we have better/distinct errors
-	errMessage := "decryption failed for all provided methods: "
-	sep := ""
-	for _, err := range errs {
-		errMessage += err.Error() + sep
-		sep = "\n"
-	}
+	errs = append([]error{fmt.Errorf("decryption failed for all provided methods")}, errs...)
+	errMessage := errors.Join(errs...).Error()
```
cam72cam, line 234: We should probably just return the result of errors.Join() here instead of creating a new error below.

internal/encryption/method/aesgcm/example_test.go
```diff
@@ -111,9 +111,9 @@ func Example_config_json() {
 	}
 
 	// Now you can build a method:
-	method, err := config.Build()
-	if err != nil {
-		panic(err)
+	method, diags := config.Build()
```
cam72cam, line 114: Was this missed in the revert?

internal/encryption/method/external/config.go
```diff
@@ -36,14 +36,14 @@ func (c *Config) Build() (method.Method, error) {
 	if len(c.DecryptCommand) < 1 {
 		return nil, &method.ErrInvalidConfiguration{
 			Cause: &method.ErrCryptoFailure{
-				Message: "the decrypt_command option is required",
+				Message: "the first entry of encrypt_command must not be empty",
```
cam72cam, line 39: This is not "encrypt_command"

cam72cam:
@Klopklopi Thanks!  I've left a few more comments, but I think we are close to wrapping up this open-ended issue :)

internal/encryption/method/external/config.go
Klopklopi: Oh I'm sorry I think I got mixed up somewhere here 😅

PR #2597, opened 2025-03-14
refactoring for_each evaluation tests + bug fix for #2532

diofeher:
This is the first part for solving #2532.

This PR covers:

- Refactoring of test suite to cover the validate phase;
- Add one more test + bug fix for #2532;

I'm going to open another PR for refactoring `eval_for_each_test.go`.

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +496,62 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+			"provided a value of type set of object",
+			true, false,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			allowUnknown := true
+			allowTuple := false
+			_, diags := EvaluateForEachExpressionValue(test.Expr, mockRefsFunc(), allowUnknown, allowTuple, test.ExcludableAddr)
```
cam72cam, line 524: Let's make sure we check the return type

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +496,62 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+			nil,
+			"Invalid for_each set argument",
+			"provided a value of type set of object",
+			true, false,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			allowUnknown := true
```
cam72cam, line 522: Let's move move this to the test structure so we can see the differing results/errors

diofeher: Fixed on https://github.com/opentofu/opentofu/pull/2597/commits/dcf4e90f5b952ed3860b018859be8cc8783f4451

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Fixed on https://github.com/opentofu/opentofu/pull/2597/commits/dcf4e90f5b952ed3860b018859be8cc8783f4451

internal/lang/evalchecks/eval_for_each.go
```diff
@@ -130,7 +130,7 @@ func EvaluateForEachExpressionValue(expr hcl.Expression, ctx ContextFunc, allowU
 }
 
 // performForEachValueChecks ensures the for_each argument is valid
-func performForEachValueChecks(expr hcl.Expression, hclCtx *hcl.EvalContext, allowUnknown bool, forEachVal cty.Value, allowedTypesMessage string, excludableAddr addrs.Targetable) (cty.Value, tfdiags.Diagnostics) {
+func performForEachValueChecks(expr hcl.Expression, hclCtx *hcl.EvalContext, allowUnknown bool, forEachVal cty.Value, isAllowedType bool, allowedTypesMessage string, excludableAddr addrs.Targetable) (cty.Value, tfdiags.Diagnostics) {
```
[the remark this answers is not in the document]
yottta: This is not used inside the function.

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,767 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
 }
+
+type expectedErr struct {
+	Summary           string
+	Detail            string
+	CausedByUnknown   bool
+	CausedBySensitive bool
+}
+
+func evaluateErrors(t *testing.T, diags []tfdiags.Diagnostic, testDiags []expectedErr, phase string) {
```
apparentlymart, line 263: It seems like the goal of this function is to determine whether the actual diagnostics returned (in `diags`) matches a set of _expected_ diagnostics (in `testDiags`). Is that right?

In some other parts of this codebase we have a naming scheme where a function that takes a `testing.T` and marks it as failed when something is wrong has the prefix `assert`, so for example this could be called `assertErrorsMatch`.

Here's one example of that naming scheme, with quite a similar helper function in `package tofu`:

https://github.com/opentofu/opentofu/blob/43a6fc2eda3a0a9729ee345d0ffce5b593904f56/internal/tofu/context_test.go#L975-L983

I don't feel strongly about this; if you have a good reason for the different naming then that's fine.

yottta (commented):
Left one comment. Reviewing the tests tomorrow.
Also tested this, and during `tofu validate` this still is passing. The problem seems to come from [here](https://github.com/opentofu/opentofu/blob/main/internal/tofu/node_resource_validate.go#L574-L576).
I will dig further into this tomorrow.

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,767 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+	} else {
+		t.Errorf("diagnostic does not support FromExpr\ngot: %s", spew.Sdump(diag))
+	}
+}
+
+func TestEvaluateForEach(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		Expr                 hcl.Expression
```
apparentlymart, line 312: I don't feel strongly about this, but it seems that every value assigned to this field is the result of `hcltest.MockExprLiteral`, so I wonder if it would be easier to maintain to change this field to just `Input cty.Value` and then make the testing loop below call `hcltest.MockExprLiteral(test.Value)` in just one place.

Supporting different kinds of expression here could be useful if the logic behaved differently depending on the expression type, but I believe the evaluation function takes an expression really just so that it can get the information required to produce helpful diagnostic messages, and so it's not important exactly which expression implementation is passed in, right?

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,767 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+		ValidateExpectedErrs []expectedErr
+		ValidateReturnValue  cty.Value
+		PlanExpectedErrs     []expectedErr
+		PlanReturnValue      map[string]cty.Value
+		ExcludableAddr       addrs.Targetable
```
apparentlymart, line 317: It seems like this field is always `nil` in every case right now, right? If so, could we omit this until we have a reason to use it?

If we _do_ keep it, I think it's helpful to group the fields of a table test struct so that all of the "input" fields are first and all of the "expected output" fields come after, and this seems to be an input field (passed in at the same time as as `Expr`) so I would've expected it to be declared after `Expr` and before `ValidateExpectedErrs`.

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,767 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+					CausedByUnknown:   true,
+					CausedBySensitive: false,
+				},
+			},
+			PlanReturnValue: map[string]cty.Value{},
+			ExcludableAddr:  nil,
+		},
+		// # Tuples
+		"empty_tuple": {
+			Expr:                 hcltest.MockExprLiteral(cty.SetValEmpty(cty.String)),
```
apparentlymart, line 411: This seems to be a duplicate of the "empty_set" test above. Was this intended to be `cty.EmptyTupleVal`?

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,767 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+					Detail:            "must be a map, or set of strings, and you have provided a value of type number",
+					CausedByUnknown:   false,
+					CausedBySensitive: false,
+				},
+			},
+			PlanReturnValue: map[string]cty.Value{},
+			ExcludableAddr:  nil,
+		},
+		"bool": {
+			Expr: hcltest.MockExprLiteral(cty.BoolVal(true)),
```
apparentlymart, line 748: This really doesn't matter :grinning: but FWIW you can write `cty.True` as a shorthand for this boolean value.

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,767 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+					Summary:           "Invalid for_each argument",
+					Detail:            "set includes values derived from resource attributes that cannot be determined until apply",
+					CausedByUnknown:   true,
+					CausedBySensitive: false,
+				},
+			},
+			PlanReturnValue: map[string]cty.Value{},
+			ExcludableAddr:  nil,
+		},
+		"unknown_empty_map": {
```
apparentlymart, line 884: One additional case to test would be the entire expression evaluating to `cty.DynamicVal` -- unknown value of unknown type -- whose behavior I think should be the same as for an unknown map or an unknown set of unknown element type.

(For context, it might help to know that `cty.DynamicVal` is really just `cty.UnknownVal(cty.DynamicPseudoType)`, but it's available directly as a variable because it arises quite often as a way to represent "we know nothing at all about this value".)

apparentlymart (commented):
Overall this looks good to me! Thanks for putting together this big table of test cases.

Much of my feedback inline is subjective stuff that I don't feel strongly about but mentioned in case it was useful. I think only the one about the `empty_tuple` test case and the `cty.DynamicVal` test case are important to deal with somehow before we merge this.

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Oh, I did miss this pattern from other files, let me change it! Thanks

internal/lang/evalchecks/eval_for_each.go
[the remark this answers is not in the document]
diofeher: Addressed on https://github.com/opentofu/opentofu/pull/2597/commits/d5216e2d153cb915dd6ecbe18be456314ccfe073

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Do you think it makes sense to use a tfdiags.Diagnostics instead of the custom struct that I've used (expectedErr)?

I can see some pros and cons when doing that, since I would need to match the whole structure to be able to use with `cmp.Diff`.

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Makes sense! I used the hcl.Expression since it was being used already, but it makes sense what you've said.

diofeher: Addressed by https://github.com/opentofu/opentofu/pull/2597/commits/64237a4dd520f504a50f89d116134153eea216b2

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Thanks for the catch, I've missed when reviewing, addressed by https://github.com/opentofu/opentofu/pull/2597/commits/fbb3eba099bf4c576c7db50a0333d2bdeac7c2a0

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Do you think this test should be removed?

apparentlymart: What I meant is that `cty.True` is equivalent to `cty.BoolVal(true)`, and the shorter version is the more common way to write it when a true value is being hard-coded instead of being chosen based on dynamic user input.

I think the test itself is fine!

diofeher: Oh, I see! Addressed at https://github.com/opentofu/opentofu/pull/2597/commits/29ff2aee161b43dafde515c1b4130482822b0297

internal/lang/evalchecks/eval_for_each.go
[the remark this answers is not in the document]
diofeher: @yottta Thanks for catching this one! I may have changed something that didn't catch the test and thought I was addressing because I was testing another one. Gonna work on it again.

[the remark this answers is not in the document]
diofeher: I copied on the wrong place from the other branch :( But now it's fixed, thanks! 

https://github.com/opentofu/opentofu/pull/2597/commits/1fbdc62941d15d74b1ba943b1d5ff2983ef293ce

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Makes sense! Addressed on https://github.com/opentofu/opentofu/pull/2597/commits/14b5326baea49499abe382b382d424a2220b9dca

internal/lang/evalchecks/eval_for_each_test.go
diofeher: Oh, that's great, didn't know! Addressed on https://github.com/opentofu/opentofu/pull/2597/commits/4c707b81e536ac371c9762c27ddc6cafda34ffde

internal/lang/evalchecks/eval_for_each_test.go
diofeher: You're right! I was planning to convert the excludableaddr tests on this big one, but I think it makes sense to have them separated. So I didn't touch these tests and removed this attribute on https://github.com/opentofu/opentofu/pull/2597/commits/9f4830d5cf4cbf23c990ee9f74c11a61e5ae0774

Thanks for the suggestion!

apparentlymart (approved):

yottta:
> Left one comment. Reviewing the tests tomorrow. Also tested this, and during `tofu validate` this still is passing. The problem seems to come from [here](https://github.com/opentofu/opentofu/blob/f1358f9fe8ada76e2a65e1bd033596b0d82c85d9/internal/tofu/node_resource_validate.go#L574-L576). I will dig further into this tomorrow.

Building on top of my last comment and considering the source code linked there, I was wondering why the value is unknown, which makes that validation linked above to skip returning the diagnostics.
I think that [here](https://github.com/opentofu/opentofu/blob/f1358f9fe8ada76e2a65e1bd033596b0d82c85d9/internal/tofu/eval_variable.go#L79) might be the cause, but not sure.
I played around a little bit with different ways to change the way default values are used instead of the actual value, but didn't manage to get acceptable results.

@apparentlymart, when a variable is having a default value, shouldn't that value be used also during the `tofu validate`? 🤔 
Right now, from my checks, the value is not the default one, but unknown.
image

diofeher:
I think your first comment was actually spot on: https://github.com/opentofu/opentofu/blob/f1358f9fe8ada76e2a65e1bd033596b0d82c85d9/internal/tofu/node_resource_validate.go#L574-L576

When the value returned is unknown, we do not append the errors to the diagnostics. That's why the bug wasn't fixed because I was returning UnknownVal, making the errors not being added, then no errors were returned on validate. By returning it NullVal, this check is skipped and then we can see the errors at the end.

yottta:
> @yottta I think your first comment was actually spot on:
> 
> https://github.com/opentofu/opentofu/blob/f1358f9fe8ada76e2a65e1bd033596b0d82c85d9/internal/tofu/node_resource_validate.go#L574-L576
> 
> When the value returned is unknown, we do not append the errors to the diagnostics. That's why the bug wasn't fixed because I was returning UnknownVal, so errors were not being added, so validate passed successfully.
> 
> https://github.com/opentofu/opentofu/blob/1fbdc62941d15d74b1ba943b1d5ff2983ef293ce/internal/lang/evalchecks/eval_for_each.go#L220-L224
> 
> By returning it NullVal, [this](https://github.com/opentofu/opentofu/blob/f1358f9fe8ada76e2a65e1bd033596b0d82c85d9/internal/tofu/node_resource_validate.go#L574-L576) check is skipped and then we can see the errors at the end.

Great! Just tested since I have the setup ready. Yes, works just fine! It yields the same error even when the default value is not given.
Great catch! Apparently I went sideways really bad with my last comment 😵 
I would like to have also @apparentlymart's take on returning `cty.NullVal` there. Just to be sure that we are no hitting other side-effects with this. 😅

internal/lang/evalchecks/eval_for_each_test.go
```diff
@@ -496,3 +252,752 @@ func TestForEachCommandLineExcludeSuggestion(t *testing.T) {
...
+	if len(diags) != len(testDiags) {
+		t.Fatalf("got %d errors in %s phase; expected %d", len(diags), phase, len(testDiags))
+	}
+
+	for i := range diags {
+		assertDiagnosticMatch(t, diags[i], testDiags[i])
+	}
+}
+
+func assertDiagnosticMatch(t *testing.T, diag tfdiags.Diagnostic, testDiags expectedErr) {
```
yottta, line 274: `nitpick`
```suggestion
func assertDiagnosticMatch(t *testing.T, wantDiag tfdiags.Diagnostic, gotDiag expectedErr) {
```

Just for the ease of reading

yottta (approved):
Managed to review the tests as well. Looks really good! 🙌  
Left just one small nitpick that you can just ignore if you do not agree.

internal/lang/evalchecks/eval_for_each_test.go
diofeher: I agree with you that's a good idea, thanks!

apparentlymart:
The historical purpose of `tofu validate`, and the validate walk _in general_, was to check whether the configuration alone is valid, ignoring anything about the input variables, the prior state, etc. For that reason, root module input variables are always treated as unknown during the validation phase, by the special case here:

https://github.com/opentofu/opentofu/blob/dee6c567d0462ebe409836d04f7e55178c8e19c1/internal/tofu/evaluate.go#L254-L273

Only during the plan phase do we check whether the expressions are valid for the _actual_ values of those input variables. At that point, we already know whether each one is set or not, and so we can know whether to use the default value or to use the caller-provided value, which is handled by some logic in the graph node type that represents a root module input variable:

https://github.com/opentofu/opentofu/blob/0211040452f41bad793a24e2235393c35410a258/internal/tofu/eval_variable.go#L78-L100

The addition of "early evaluation" as an entirely separate mechanism inside `package configs` has made this situation confusing, because now `tofu validate` _does_ use the input variables when loading the configuration and dealing with the early evaluation steps, but that feature didn't make any changes to the main language runtime so it continues to treat them all as unknown for anything that gets resolved dynamically.

I hope that one day we'll find a way to reunite these two paths so that we don't have two separate implementations of expression evaluation that are inconsistent with one another, but for the sake of _this_ PR I suggest that we just accept it as intended behavior that root module input variables are always unknown during the validation phase, since changing that assumption could be very impactful on the behavior of existing configurations.

PR #2609, opened 2025-03-18
Adding missing CHANGELOG entries

yottta:
Resolves #2596, #2606
Adding changelog entries for the aforementioned issues.

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ ENHANCEMENTS:
 * The test `run` outputs can now be used in the test `provider` blocks defined in test files. ([#2543](https://github.com/opentofu/opentofu/pull/2543))
 * Provider instance keys now automatically converted to string ([#2378](https://github.com/opentofu/opentofu/issues/2378))
 * Remove progress messages from commands using -concise argument ([#2549](https://github.com/opentofu/opentofu/issues/2549))
-* Upgrade aws-sdk version to include `mx-central-1` region. Also, `skip_s3_checksum=true` now blocks the [aws-sdk new default S3 integrity checks](https://github.com/aws/aws-sdk-go-v2/discussions/2960)([#2596](https://github.com/opentofu/opentofu/pull/2596))
+* Upgrade aws-sdk version to include `mx-central-1` region. ([#2596](https://github.com/opentofu/opentofu/pull/2596))
 
 
 BUG FIXES:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ ENHANCEMENTS:
 * The test `run` outputs can now be used in the test `provider` blocks defined in test files. ([#2543](https://github.com/opentofu/opentofu/pull/2543))
 * Provider instance keys now automatically converted to string ([#2378](https://github.com/opentofu/opentofu/issues/2378))
 * Remove progress messages from commands using -concise argument ([#2549](https://github.com/opentofu/opentofu/issues/2549))
+* Upgrade aws-sdk version to include `mx-central-1` region. Also, `skip_s3_checksum=true` now blocks the [aws-sdk new default S3 integrity checks](https://github.com/aws/aws-sdk-go-v2/discussions/2960)([#2596](https://github.com/opentofu/opentofu/pull/2596))
```
ollevche, line 33: > Also, `skip_s3_checksum=true` now blocks the aws-sdk new default S3 integrity checks

I believe we should put here actual user-facing changes. What does it do from the user perspective?

CHANGELOG.md
```diff
@@ -48,6 +49,7 @@ BUG FIXES:
 - `base64gunzip` now doesn't expose sensitive values if it fails during the base64 decoding. ([#2503](https://github.com/opentofu/opentofu/pull/2503))
 - Fix loading only the necessary encryption key providers and methods for better `terraform_remote_state` support. ([2551](https://github.com/opentofu/opentofu/issues/2551))
 - Better error messages when using `null` in invalid positions in the argument to the `transpose` function. ([#2553](https://github.com/opentofu/opentofu/pull/2553))
+- Do not allow aws-sdk to control the checksum headers for the lock object but instead calculate it before doing the request. ([#2606](https://github.com/opentofu/opentofu/pull/2606))
```
ollevche, line 52: The same question here, what has changed from the UX perspective?

CHANGELOG.md
yottta: I wanted to highlight that the behaviour for skipping checksums suffered a change. Just in case somebody is stumbling into an issue with it. This way would be easier to find what changed.
From the user perspective, it is meant to work the same, nothing changes.
If we keep only the user facing things in here, let me remove that second statement.

ollevche: I believe we could mention it as an internal change. There is an example for 1.9 release: https://github.com/opentofu/opentofu/blob/v1.9/CHANGELOG.md (`INTERNAL CHANGES` block)

CHANGELOG.md
```diff
@@ -48,7 +49,10 @@ BUG FIXES:
 - `base64gunzip` now doesn't expose sensitive values if it fails during the base64 decoding. ([#2503](https://github.com/opentofu/opentofu/pull/2503))
 - Fix loading only the necessary encryption key providers and methods for better `terraform_remote_state` support. ([2551](https://github.com/opentofu/opentofu/issues/2551))
 - Better error messages when using `null` in invalid positions in the argument to the `transpose` function. ([#2553](https://github.com/opentofu/opentofu/pull/2553))
+- S3 based state locking now works also with 3rd party providers that support SHA256 checksums. ([#2606](https://github.com/opentofu/opentofu/pull/2606))
```
ollevche, line 52: We introduced a new feature (line #19) and fixed a bug within it before the actual release, so I believe we shouldn't include that bug fix message (it didn't actually exist for end users).

Additional note: 3rd party s3 compatible storage implementations must also support conditional writes as well as SHA256 checksums.

yottta: Good point. 
About your note, that's coming from the actual implementation. Nonetheless, I just removed this. I was having second thoughts about adding this or not too. Thanks for helping with the decision here.

ollevche (approved):

PR #2613, opened 2025-03-18
Enhance native S3 lock for providers where putObject does not support…

brano543:
I have reviewed the locking mechanism introduced by https://github.com/opentofu/opentofu/issues/599 and found two providers (Telnyx, Mega) which support If-None-Match and If-Match headers only within headObject and getObject calls.

I have modified the s3Lock function to first call headObject function and bail early if the lock already exists.

S3Unlock function relies on GetObject and it seems that both above mentioned providers to support the required headers here as well (https://github.com/meganz/s4-specs?tab=readme-ov-file#getobject, https://developers.telnyx.com/api/cloud-storage/object-operations/get-object). 

Question: Are we not using If-Match header because we want to be sure all locks are always cleared?

Related to 

Resolves # 

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/s3/client.go
+++ b/internal/backend/remote-state/s3/client.go
@@ -341,6 +341,23 @@ func (c *RemoteClient) s3Lock(info *statemgr.LockInfo) error {
 		return nil
 	}
 
+	ctx := context.TODO()
+	ctx, _ = attachLoggerToContext(ctx)
+
+	headParams := &s3.HeadObjectInput{
+		Bucket:  aws.String(c.bucketName),
+		Key:     aws.String(c.lockFilePath()),
+		IfMatch: aws.String("*"),
+	}
+	log.Printf("[DEBUG] Check existence of s3 locking object: %#v", headParams)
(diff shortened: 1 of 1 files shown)


yottta:
Hi @brano543 and thanks for the submission.

There are several things to mention here regarding the s3 backend:
* First and foremost, the backend is meant to work with the AWS S3. Therefore, this backend can work with other 3rd party services that are as compatible as possible with AWS S3 API.
* As long as a 3rd party service is starting to diverge in functionality from the actual S3 API, does not make a good candidate to be used in here.
* We don't have an official list of 3rd party S3 APIs that we support, and right now is not a priority to do so.

Regarding the implementation, having the acquiring of the lock separated from the actual check, can introduce unwanted behaviours and concurrent access to the lock object (and implicitly to the state object). 
Just think of the same configuration being executed in the same time. Head can succeed for both processes and put then overwrites the lock, allowing both processes to write on the state.

Having all the above as arguments, I don't think that OpenTofu should accept these changes.

brano543:
Hello @yottta !

Oh, you are actually correct. The lock operation needs to be atomic in order for it to work concurrently.

It's my miss.

PR #2660, opened 2025-04-14
Fix #2584: Accept null value for version if module source is local

### LINKED ISSUE #2584: disregard module version if null
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.9.0
on linux_amd64
```

Unlike terraform, OpenTofu allows to use variables do define de module source, which is great, but I'd like to go even further and be able to dynamically switch between local source and registry.

```
locals {
  module_source =  var.is_local ? "../my-module" : "app.terraform.io/ORG/my-module"
  module_version = var.is_local ? null : "~> 1.0.0"
}
```

is_local would be defined as TF_VAR_is_local environment variable

With this working I would not need change the script again when running it in a CI/CD

### Attempted Solutions

Setting module version as null

### Proposal

Currently when version is null the following error happens:
OpenTofu assumed that you intended a module registry source address because you also set the argument "version", which applies only to registry modules.

Disregard the version attribute if it is null

### References

_No response_

baa-ableton:
Resolves #2584 

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/configs/module_call.go
```diff
@@ -72,7 +72,10 @@ func decodeModuleBlock(block *hcl.Block, override bool) (*ModuleCall, hcl.Diagno
 	}
 
 	if attr, exists := content.Attributes["version"]; exists {
-		mc.VersionAttr = attr
+		val, _ := attr.Expr.Value(nil)
```
cam72cam, line 75: Unfortunately, this does not work with static (early) evaluation as requested in the issue.  Take a look at the decodeStaticVersion function further down in this file.

cam72cam:
@baa-ableton feel free to reach out on the community slack if you need any assistance with this PR!

baa-ableton:
@cam72cam 

Hey, could you please review once again? I checked the function you mentioned, and noticed the nil check there, which the previous commit did not cover. 

Did not mention it but I think `decodeModuleBlock` is the correct place to be making changes to service the issue - is that correct? I just started learning the relevant codebase and it seems `decodeModuleBlock` runs during file parsing which is to say even before static evaluation AFA I can tell. LMK!

baa-ableton:
@cam72cam bump

Please let me know if I'm missing something here, then I'll go reach out on the community slack

cam72cam:
@baa-ableton I my comment above is still unresolved.  The evaluation of the Value of the expression must use the static context.  If you start adding some tests based on the issue, it will become more clear.

  Also make sure you get your DCO sorted out :+1:

baa-ableton:
hey @cam72cam sorry my bad, work's been busy this past week but I'll be working on this on Friday and will reach out

baa-ableton:
hey @cam72cam ,

I force pushed a new commit, so my previous work has been lifted but:

* Instead of checking values during config file parsing, null value or static expressions evaluating to null value are being checked during `decodeStaticSource` in order to avoid `if err != nil ` condition in the same function.
* `decodeStaticVersion` now has a case to handle null values.
* Wrote a new test providing a `mod` as well as a `call` to `NewStaticEvaluator`.
* The new test reads from a new test `tf` file. Please let me know if there is a better place for the tf file since I just put it where I thought it was relevant.
* Fixed DCO

baa-ableton:
Unit tests had failed because I didn't have a good understanding of the test file/directory structure at the time. The test file I wrote, `null-version-module.tf` is indeed a valid file within the scope of the new feature, so I put it from `testdata/invalid-files/` to `testdata/valid-files/`.

internal/configs/module_call.go
```diff
@@ -169,7 +169,18 @@ func (mc *ModuleCall) decodeStaticSource(eval *StaticEvaluator) hcl.Diagnostics
 		// NOTE: This code was originally executed as part of decodeModuleBlock and is now deferred until we have the config merged and static context built
 		var err error
 		if mc.VersionAttr != nil {
-			mc.SourceAddr, err = addrs.ParseModuleSourceRegistry(mc.SourceAddrRaw)
+			// Check if version is null before forcing registry module parsing
+			val, valDiags := eval.Evaluate(mc.VersionAttr.Expr, StaticIdentifier{
```
cam72cam, line 173: If you re-order the decode functions above in decodeStaticFields, we could remove this additional logic as you already set mc.VersionAttr to nil.

internal/configs/testdata/valid-files/null-version-module.tf
```diff
@@ -0,0 +1,9 @@
+variable "module_version" {
+  type    = string
+  default = null
+}
+
+module "foo" {
+  source  = "./foo"
+  version = var.module_version
```
cam72cam, line 8: Instead of creating NullVersionModuleCallForTesting and passing that in, we could instead use a local instead of a variable in here.

cam72cam:
@baa-ableton this looks much better, thanks!  I've left a few comments where I think we could simplify the changes / reduce the diff.

internal/configs/module_call.go
```diff
@@ -151,8 +151,8 @@ func (mc *ModuleCall) decodeStaticFields(eval *StaticEvaluator) hcl.Diagnostics
 	mc.decodeStaticVariables(eval)
 
 	var diags hcl.Diagnostics
-	diags = diags.Extend(mc.decodeStaticSource(eval))
 	diags = diags.Extend(mc.decodeStaticVersion(eval))
+	diags = diags.Extend(mc.decodeStaticSource(eval))
```
cam72cam, line 155: Can you add a comment that decodeStaticSource depends on the results of decodeStaticVersion?

cam72cam (approved):

internal/configs/module_call.go
baa-ableton: @cam72cam we good?

cam72cam: Perfect, thanks!

baa-ableton:
hey @yottta thanks for the feedback i can implement the exact same test case but I will need to write a new function that returns a NewStaticModuleCall instead of RootModuleCallForTesting() since that one cannot evaluate Expressions. This in turn will raise the diff size again. What do you think, @cam72cam ?

cam72cam:
@baa-ableton I think the easiest way to solve this is to add a case to your existing test file with a local that is set to some valid version number.

Adding something like this to your test file:
```hcl
locals {
  module_version_set = "1.0.0"
}
module "foo_remote" {
  source  = "remote-module"
  version = local.module_version_set
} 
```

yottta (approved):
Looks good! Thanks for reworking the unit test.

cam72cam (approved):

cam72cam (approved):

PR #2665, opened 2025-04-15
[OpenTelemetry] Add traces to `init` command

Yantrio:
## Description

> Part of the OTEL Tracing implementation tracked here #2664

This PR attempts to solve the initial parts of adding opentelemetry traces to the opentofu codebase.

It does this by defining the standards of how we want to be adding otel traces to opentofu in a developer friendly way, and then providing tracing around the initialization phase of the project as an example that can be used for discussion.

**Please use this PR as a point of discussion for the approach and conventions, however please remember that this is experimental and can be subject to change in the future**

## Creation of tracers across the project
The creation of the tracers used to be handled by creating a single file inside the package and defining a package-wide variable to store the tracer. This became a point of friction whenever moving around files because the developer would have to ensure that the file is created for every package in which tracing can occur! This file also contained hard-coded strings for the scope name, which could cause room for human error.

To simplify this I have introduced a utility method (`tracing.Tracer()`) which will provide the developer with a tracer with a correctly set scope based on the package name. The small downside of this is that we will be using the `runtime` package to reflect and understand the package name. This adds a minor overhead and I have ensured that this overhead is only incurred if opentelemetry tracing is enabled. For the amount of tracing that this project will do I am quite certain that this overhead is very acceptable.

This now means that to begin a span in any part of the codebase, the developer only needs to call `tracing.Tracer().Start(ctx, "<spanname>")`.

## Naming convention of spans
There is currently no convention of naming spans in opentelemetry as far as I am aware, however there is conventions for naming attributes which can be found [here](https://opentelemetry.io/docs/specs/semconv/general/naming/). 

By using similar conventions of attributes such as namespacing, we can define a set of span names that help end users and consumers of the telemetry to understand what is happening.

Please provide any/all feedback and ideas you have about this area.

## Span errors
It is recommended that whenever a method returns an error, that we set that span to the error status and record the error in the span. This can be achieved by doing the following:

```go
if err != nil {
    span.SetStatus(codes.Error, "failed to do XYZ")
    span.RecordError(err)
}
```

## Example traces of installing 2 providers
image

## Remaining work

This PR only covers the basics of the intialization phase (Installation of providers, etc) and is not exhaustive.

- Module Installation
- OCI provider + module tracing
- Backend Initialization

## Target Release

1.10.0 / 1.11.0

## Checklist

### Go checklist

 


cmd/tofu/main.go
```diff
@@ -71,23 +72,21 @@ func realMain() int {
...
-		// At minimum we emit a span covering the entire command execution.
-		_, displayArgs := shquot.POSIXShellSplit(os.Args)
-		ctx, otelSpan = tracer.Start(context.Background(), fmt.Sprintf("tofu %s", displayArgs))
-		defer otelSpan.End()
-	}
+	ctx := context.Background()
+	var span trace.Span
+
+	// At minimum, we emit a span covering the entire command execution.
+	ctx, span = tracing.Tracer().Start(ctx, "opentofu")
```
diofeher, line 88: ctx 
```suggestion
	// At minimum, we emit a span covering the entire command execution.
	ctx, span := tracing.Tracer().Start(ctx, "opentofu")
```

internal/getproviders/package_location_local_archive.go
```diff
@@ -67,6 +81,8 @@ func (p PackageLocalArchive) InstallProviderPackage(_ context.Context, meta Pack
 	//nolint:mnd // magic number predates us using this linter
 	err := unzip.Decompress(targetDir, filename, true, 0000)
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
```
diofeher, line 84: I was thinking if it would be a good idea to include a function on the internal/tracing library for calling these error recordings, like:

tracing.SetError(err)

```
SetError(span trace.Span, err error) {
  span.SetStatus(codes.Error, err.Error());
  span.RecordError(err)
}
```

Just so we could remove some duplication of the code. Let me know what you think about it.

internal/providercache/installer.go
```diff
@@ -460,6 +473,7 @@ func (i *Installer) ensureProviderVersionsInstall(
 					if cb := evts.ProviderAlreadyInstalled; cb != nil {
 						cb(provider, version)
 					}
+					span.End()
```
diofeher, line 476: Quick question: why in this case are you using `span.End()` in different code paths and not using `defer span.End()` after the assignment?

diofeher:
Really good to see this work moving forward! I've left a few comments, thanks @Yantrio !

internal/getproviders/package_location_local_archive.go
Yantrio: I was thikning about this too, but then I also realised that there are some situations where we want to report the error from a diag or something else, so i decided to just go with this method. but I'm 50/50 on this.

internal/command/cliconfig/cliconfig.go
```diff
@@ -116,7 +117,10 @@ func DataDirs() ([]string, error) {
 // LoadConfig reads the CLI configuration from the various filesystem locations
 // and from the environment, returning a merged configuration along with any
 // diagnostics (errors and warnings) encountered along the way.
-func LoadConfig() (*Config, tfdiags.Diagnostics) {
+func LoadConfig(ctx context.Context) (*Config, tfdiags.Diagnostics) {
+	_, span := tracing.Tracer().Start(ctx, "opentofu.config.load")
```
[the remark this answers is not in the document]
Yantrio: I agree! But also after some consideration, i dont think this is worth tracing right now. It's a minor code path that doesn't take too long.

internal/command/init.go
```diff
@@ -132,6 +134,9 @@ func (c *InitCommand) Run(args []string) int {
 	ctx, done := c.InterruptibleContext(c.CommandContext())
 	defer done()
 
+	ctx, span := tracing.Tracer().Start(ctx, "opentofu.init")
```
yottta, line 137: Similar with the comment left by @apparentlymart, I don't think that we should `opentofu` prefix everywhere.
Here, I would suggest something like `command.init` instead.
Also, the addition of that empty line at the beginning of the method hints that this line here was there initially. If so, why moved it from there? 🤔 
I generally like to have the traces at the beginning of the functions to be able to trace back from where the flow got interrupted. Of course, this is not a hard rule, but something that I generally do.

internal/command/init.go
```diff
@@ -159,9 +164,10 @@ func (c *InitCommand) Run(args []string) int {
 			ShowLocalPaths: false, // since they are in a weird location for init
 		}
 
-		ctx, span := tracer.Start(ctx, "-from-module=...", trace.WithAttributes(
-			attribute.String("module_source", src),
+		ctx, span := tracing.Tracer().Start(ctx, "opentofu.init.from_module", trace.WithAttributes(
+			attribute.String("opentofu.module_source", src),
 		))
+		defer span.End()
```
yottta, line 170: This call here should be removed. Check that `span.End()` is called a little bit down the road.

internal/getproviders/registry_client.go
```diff
@@ -181,6 +196,8 @@ func (c *registryClient) PackageMeta(ctx context.Context, provider addrs.Provide
 		target.OS,
 		target.Arch,
 	))
+	ctx, span := tracing.Tracer().Start(ctx, "opentofu.provider.meta")
```
yottta, line 199: `suggestion`
Maybe moving this after the definition of `endpointURL` and add that as attribute? 🤔

internal/providercache/installer.go
yottta: Agree, but really error prone this part. Isn't there a way to extract the content of the `for` loop into a new function?
If somebody will ever come and add a new conditional `continue` or worse, a `break` or a `return`, that span will never be closed and will create a havoc in the traces.

internal/getproviders/package_location_local_archive.go
yottta: I agree with Dio here. If we want to do the diags thing, we could go with a similar implementation of the `Diagnostics.Append` where it checks the type and handles that internally.

yottta (commented):
Nice! Well done! This is clean and tidy. Like it.
Just a few comments.

internal/command/init.go
```diff
@@ -159,9 +164,10 @@ func (c *InitCommand) Run(args []string) int {
 			ShowLocalPaths: false, // since they are in a weird location for init
 		}
 
-		ctx, span := tracer.Start(ctx, "-from-module=...", trace.WithAttributes(
-			attribute.String("module_source", src),
+		ctx, span := tracing.Tracer().Start(ctx, "opentofu.init.from_module", trace.WithAttributes(
```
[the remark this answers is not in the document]
Yantrio: So, after some thought I strongly agree with not using the heirarchical-shaped trace names here. I feel as if they add very little value when exploring the traces. 

What I am currently unsure about is the disparity between our UX for end-users and the ability to trace it back to the codebase. If we are following the intention of the RFC that was accepted for opentelemetry then I believe that these traces are not to be consumed fully by opentofu developers, but by end-users who may not be familiar with the codebase.

With your comment in mind I propose we do the following:

- Name traces in a human readable way
- Traces should not be hierarchical (The hierrachy is already created by the tracing it'sself)
- Trace names should be comparible to the UX, and not to the codebase.

With these things in mind, I believe the name of this trace quickly becomes `Init from module`. WDYT?

internal/providercache/installer.go
Yantrio: I didn't want to do large refactorings of existing code in this codebase, plus i feel like this sets a good example of how we should be handling spans in loops anyway.

internal/tracing/utils.go
```diff
@@ -0,0 +1,50 @@
...
+//	github.com/you/pkg.(*SomeType).Method-fm
+//	github.com/you/pkg.functionName
+func extractImportPath(fullName string) string {
+	lastSlash := strings.LastIndex(fullName, "/")
+	if lastSlash == -1 {
+		// When there is no slash, then use everything before the first dot
+		if dot := strings.Index(fullName, "."); dot != -1 {
+			return fullName[:dot]
+		}
+		log.Printf("[WARN] unable to extract import path from function name: %q`. Tracing may be incomplete. This is a bug in OpenTofu, please report it.", fullName)
```
yottta, line 39:
```suggestion
		log.Printf("[WARN] unable to extract import path from function name: %q. Tracing may be incomplete. This is a bug in OpenTofu, please report it.", fullName)
```

Typo?

yottta: Same for the other log statement.

internal/getproviders/package_location_local_archive.go
Yantrio: I've added this method in the utils file here: https://github.com/opentofu/opentofu/pull/2665/files#diff-338d0067ba5fd1e76a34f61066eecaed4b9688f08e3869682fa04c0c05305c3cR40-R70

Let me know what you guys think @yottta / @diofeher

cmd/tofu/main.go
```diff
@@ -117,6 +114,7 @@ func realMain() int {
 	streams, err := terminal.Init()
 	if err != nil {
 		Ui.Error(fmt.Sprintf("Failed to configure the terminal: %s", err))
+		tracing.ForceFlush(5 * time.Second)
```
cam72cam, line 117: What's so special about the ForceFlush and why don't we do it in a defer call as it's part of all the error paths?

internal/providercache/installer.go
```diff
@@ -440,7 +444,16 @@ func (i *Installer) ensureProviderVersionsInstall(
...
+		traceCtx, span := tracing.Tracer().Start(ctx,
+			"Install Provider",
+			trace.WithAttributes(
+				attribute.String(tracing.ProviderAddressAttributeName, provider.String()),
+				attribute.String(tracing.ProviderVersionAttributeName, version.String()),
+			),
+		)
+		if err := traceCtx.Err(); err != nil {
+			span.SetStatus(codes.Error, "context cancelled")
+			span.RecordError(err)
```
cam72cam, line 456: Do you still need to span.End() here?

internal/providercache/installer.go
cam72cam: Another approach that could be considered is something like https://github.com/opentofu/opentofu/pull/1878/files#diff-385ff28969dc6410989718822b4c24e033e3a10d679bdd792fd2d55148c49471R456

It's a similar situation where I didn't want to heavily refactor the loop.  It mixes defer and iterative "actions" in a way that's comprehensive

cam72cam: Alternatively, this is now the second time we are running into this and should probably take the time to refactor it :D

internal/tracing/init.go
```diff
@@ -0,0 +1,128 @@
...
+	"go.opentelemetry.io/otel/sdk"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+
+	"github.com/opentofu/opentofu/version"
+)
+
+/*
+BEWARE! This is not a committed external interface.
```
cam72cam, line 27: Why not make it live in internal then?

cmd/tofu/main.go
Yantrio: Because I wrote this early in the day and forgot that defer is a thing 🤦 please forgive me! I'll re-work this.

internal/providercache/installer.go
Yantrio: I'll rebase ontop of this PR once it's in: https://github.com/opentofu/opentofu/pull/2695

internal/providercache/installer.go
Yantrio: Yup! I missed this one. thanks.

Gogotchuri (approved):

cam72cam (approved):

internal/tracing/init.go
cam72cam: I'm blind, nevermind!

apparentlymart: This comment was originally from https://github.com/opentofu/opentofu/commit/503896af54d6eacd685873488494e6f14067a2de, in which I was intending it to mean that anything new that begins happening when you run with the `OTEL_TRACES_EXPORTER` environment variable is not protected by compatibility promises.

This new placement of the comment does perhaps leave it a little ambiguous which "this" it is talking about, and so perhaps we could clarify that by replacing "this" with something like "OpenTofu's tracing output". :man_shrugging:

internal/providercache/installer.go
```diff
@@ -439,20 +442,31 @@ func (i *Installer) ensureProviderVersionsInstall(
 	authResults := map[addrs.Provider]*getproviders.PackageAuthenticationResult{} // record auth results for all successfully fetched providers
 
 	for provider, version := range need {
-		if err := ctx.Err(); err != nil {
+		traceCtx, span := tracing.Tracer().Start(ctx,
+			"Install Provider",
+			trace.WithAttributes(
+				attribute.String(tracing.ProviderAddressAttributeName, provider.String()),
+				attribute.String(tracing.ProviderVersionAttributeName, version.String()),
```
apparentlymart, line 449: I guess I reviewed these in the wrong order :facepalm: so I should really have left https://github.com/opentofu/opentofu/pull/2694#discussion_r2056333802 here instead.

apparentlymart (approved):
This looks great!

I do have some general concerns that this might be a little too granular for our intended audience (users of OpenTofu, rather than developers of OpenTofu) but as long as we're not committing to preserving any specific tracing hierarchy as a machine-readable interface I'm happy to just run with this and see how it goes.

internal/tracing/utils.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	case tfdiags.Diagnostics: // Assuming Diagnostics is a custom type you have defined elsewhere
+		if v.HasErrors() { // Assuming IsEmpty() checks if the diagnostics object has content
+			span.SetStatus(codes.Error, v.Err().Error())
+			span.RecordError(v.Err())
+		}
+	default:
+		// Handle unsupported types gracefully
+		// TODO: Discuss if this should panic?
+		span.SetStatus(codes.Error, "ERROR: unsupported input type for SetSpanError.")
```
diofeher, line 66: Do you think it could be helpful to add the `input.(type)` in the error message to help debugging?

internal/tracing/utils.go
```diff
@@ -0,0 +1,121 @@
...
+
+	switch v := input.(type) {
+	case error:
+		if v != nil {
+			span.SetStatus(codes.Error, v.Error())
+			span.RecordError(v)
+		}
+	case string:
+		if v != "" {
+			span.SetStatus(codes.Error, v)
```
diofeher, line 55: `question`

Why in these functions are you choosing to not returns the errors to the caller and just log them in here?

Yantrio: This method is to add the error to the trace, it shouldn't do anything else.

cam72cam (approved):

PR #2673, opened 2025-04-16
refactor a few more places for proper sensitive marks treatment

ollevche:
This PR refactors a few more places where any mark is treated as sensitive. It introduces `sensitiveMarksEqual` to properly compare only sensitive marks in order to determine whether or not the sensitivity of the value has changed. Also, the PR makes root module outputs to preserve their pathes during sensitive marks removal.

Complimentary to https://github.com/opentofu/opentofu/pull/2503

Found those cases during the work on https://github.com/opentofu/opentofu/pull/2633

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/marks.go
+++ b/internal/tofu/marks.go
@@ -9,8 +9,38 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/opentofu/opentofu/internal/lang/marks"
 	"github.com/zclconf/go-cty/cty"
 )
+
+func sensitiveMarksEqual(a, b []cty.PathValueMarks) bool {
+	a = filterNonTargetMarks(a, marks.Sensitive)
+	b = filterNonTargetMarks(b, marks.Sensitive)
+	return marksEqual(a, b)
+}
+
+func filterNonTargetMarks(pvms []cty.PathValueMarks, target interface{}) []cty.PathValueMarks {
+	pvmsCopy := make([]cty.PathValueMarks, 0, len(pvms))
+	for _, pvm := range pvms {
+		pvmsCopy = append(pvmsCopy, copyPathValueMarks(pvm))
+	}
+
+	for i := range pvms {
+		for k := range pvmsCopy[i].Marks {
+			if k != target {
+				delete(pvmsCopy[i].Marks, k)
+			}
+		}
+
+		// Remove PathValueMarks from the slice if there is no
+		// marks anymore.
+		if len(pvmsCopy[i].Marks) == 0 {
+			pvmsCopy = append(pvmsCopy[:i], pvmsCopy[i+1:]...)
+		}
+	}
+
+	return pvmsCopy
+}
 
 // marksEqual compares 2 unordered sets of PathValue marks for equality, with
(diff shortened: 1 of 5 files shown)


internal/tofu/marks.go
```diff
@@ -9,9 +9,43 @@ import (
...
+
+		// Remove PathValueMarks from the slice if there is no
+		// marks anymore.
+		if len(pvmsCopy[i].Marks) == 0 {
+			pvmsCopy = append(pvmsCopy[:i], pvmsCopy[i+1:]...)
+		}
+	}
+
+	return pvmsCopy
+}
```
Gogotchuri, line 47: I think we could join the two loops and simplify the logic.
```go
	for _, pvm := range pvms {
		cp := copyPathValueMarks(pvm)
		for k := range cp.Marks {
			if k != target {
				delete(cp.Marks, k)
			}
		}
		// Remove PathValueMarks from the slice if there is no
		// marks anymore.
		if len(cp.Marks) == 0 {
			pvmsCopy = append(pvmsCopy, cp)
		}
	}
```

Gogotchuri (approved):
Everything looks good! Left a small simplification suggestion.

internal/tofu/marks.go
```diff
@@ -9,9 +9,43 @@ import (
...
 
+// sensitiveMarksEqual filters out any non-sensitive marks and then uses
+// marksEqual to compare the resulting lists.
+func sensitiveMarksEqual(a, b []cty.PathValueMarks) bool {
+	a = filterNonTargetMarks(a, marks.Sensitive)
+	b = filterNonTargetMarks(b, marks.Sensitive)
+	return marksEqual(a, b)
+}
+
+// filterNonTargetMarks makes a copy of pvms and then filters out every
```
diofeher, line 24: I think using `PathValueMarks` instead of `pvms` would make this comment clearer.

internal/tofu/marks_test.go
```diff
@@ -204,3 +204,106 @@ func TestCombinePathValueMarks(t *testing.T) {
 		})
 	}
 }
+
+func TestSensitiveMarksEqual(t *testing.T) {
+	for i, tc := range []struct {
```
diofeher, line 209: Do you think it could make sense to add a name to the test so we know we're testing in each struct?

diofeher:
Looking great! I've left a few suggestions that should be treated as nitpicks :)

internal/tofu/marks_test.go
ollevche: Makes sense, I added names for test cases in [9d345f3](https://github.com/opentofu/opentofu/pull/2673/commits/9d345f3385a809b204485d667b1c7b5b0d77a370)

internal/tofu/marks.go
ollevche: Totally makes sense, thank you! Simplified in [bf6f01e](https://github.com/opentofu/opentofu/pull/2673/commits/bf6f01e3ce82da03b930329f28f83858f39e742a)

cam72cam (approved):

PR #2679, opened 2025-04-17
add deprecation warnings support for terraform_remote_state

ollevche:
Part of https://github.com/opentofu/opentofu/issues/1005

This PR adds support for deprecation warnings for module outputs from `terraform_remote_state` data sources. Since `terraform_remote_state` data source is not configuration based, we need to propagate deprecation marks into the state file. Since those values are in the state, it makes sense to show them on `tofu show -json` as well.

Tests are added in #2687 and documentation for the whole feature is going to be a separate PR as well.

This PR will not cover what is described in https://github.com/opentofu/opentofu/issues/2680.

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/states/statefile/version4.go
```diff
@@ -764,6 +767,8 @@ type outputStateV4 struct {
 	ValueRaw     json.RawMessage `json:"value"`
 	ValueTypeRaw json.RawMessage `json:"type"`
 	Sensitive    bool            `json:"sensitive,omitempty"`
+	// TODO: Do we need to update the state schema or state documentation somewhere?
```
cam72cam, line 770: We typically don't suggest folks look directly at the statefile, I don't think we have these fields documented anywhere concrete.

internal/command/jsonstate/state.go
```diff
@@ -48,9 +48,10 @@ type StateValues struct {
 }
 
 type Output struct {
-	Sensitive bool            `json:"sensitive"`
-	Value     json.RawMessage `json:"value,omitempty"`
-	Type      json.RawMessage `json:"type,omitempty"`
+	Sensitive  bool            `json:"sensitive"`
+	Deprecated string          `json:"deprecated,omitempty"`
```
cam72cam, line 52: This is the field we may want to document somewhere

ollevche: I will extend [the json output format page](https://opentofu.org/docs/internals/json-format/), thanks!

ollevche: This is covered in https://github.com/opentofu/opentofu/pull/2706

internal/states/instance_object_src.go
```diff
@@ -58,6 +58,11 @@ type ResourceInstanceObjectSrc struct {
 	// state, or to save as sensitive paths when saving state
 	AttrSensitivePaths []cty.PathValueMarks
 
+	// TransientPathValueMarks helps propagate non-sensitive marks through the
+	// internal representation of a state, without being serialized into the final
+	// state file.
+	TransientPathValueMarks []cty.PathValueMarks
```
cam72cam, line 64: The code in instance_object.go does not filter out sensitive marks.  I don't think that's a problem functionally, but it's made a bit confusing by the comment.

cam72cam (approved):

internal/states/instance_object_src.go
ollevche: That's indeed confusing, thanks for catching!

[dd8e250](https://github.com/opentofu/opentofu/pull/2679/commits/dd8e2509e3a3f88750b4ba86d4b68e4266bf842e)

yottta (approved):

internal/states/instance_object_src.go
yottta: Thanks for fixing it. I was already having some comments to add related to `TransientPathValueMarks`.

PR #2695, opened 2025-04-23
Refactor the provider installer loop

cam72cam:
Both the OTEL and Provider Cache work both ran up against some intersting workarounds needed for this function.

This refactoring does slightly change the events emitted by the installer, but looking at the usages it should be safe.

Recommended to review with whitespace ignored.

See #1878 and #2665

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -927,7 +927,7 @@ func (c *InitCommand) getProviders(ctx context.Context, config *configs.Config, 
 			// Either way, this is bad. Let's complain/warn.
 			incompleteProviders = append(incompleteProviders, provider.ForDisplay())
 		},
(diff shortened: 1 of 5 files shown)


apparentlymart (approved):
This looks good to me!

I agree that the slight changes to the event log don't seem like a big deal: our use of the `ProvidersFetched` event is mainly just so that the `tofu init` UI can know to emit the warning about third-party providers, which only applies if there is a package authentication result anyway.

With that said, the event name does now seem a little misleading since it seems like it only reports packages that were authenticated, rather than all packages that were fetched regardless of their authentication result. So maybe we should change the event name to `PackagesAuthenticated` to communicate more clearly that providers whose packages were not authenticated will not be reported?

Yantrio (approved):
I like the new event naming 👍

PR #2705, opened 2025-04-24
Add new CLI arg to control what warnings should be shown for deprecated outputs/variables

yottta:
## Context
As part of the deprecation work, this PR is the last big part.
When it comes to deprecated outputs and variables, there is this concept of "deprecated value from local and remote module".
This is to make a clear separation between the modules maintained by the user that is using a deprecated value and a module that is not.

More specifically, a remote module is actually a call to a module that is having the source pointing to a remote source, like:
```hcl
module "call" {
    source = "registry.opentofu.org/foo/bar/baz"
}
// or
module "call" {
    source = "git::https://github.com.com/foo/bar.git//baz"
}
// etc
```
while a local module is a call to a locally referenced module, like:
```hcl
module "call" {
    source = "./baz"
}
// or 
module "call" {
    source = "../baz"
}
```

A really important note is that a module "A" imported relatively by a remote module "B" is still considered a remote module because the root module is importing the module B remotely.
This is also caught in this PR by the functions [isCallFromRemote](https://github.com/opentofu/opentofu/pull/2705/files#diff-abfc3097bacebcfd62d6dd0531297ab21a8285f9b4990c5d26eac340d57b817eR358) and [moduleImportedFromRemote](https://github.com/opentofu/opentofu/pull/2705/files#diff-50f4be59e0ee4a8cf2d08ecb9fdf0c6a73f9b8a8ce088dd28288a7d564724af4R143).

## Deprecation warnings filtering argument
Having these notions, this PR is adding a way to control the what kind of deprecation warnings a user can see:
* Can see all of the warnings, from all types of modules (local and remote)
* Can see only the warnings from the local modules
* Can see no warnings about deprecation values at all

Since only `plan` and `apply` commands can generate this kind of warnings, these command received a new CLI argument, `-deprecation-warns` that can be used by the caller to say what type of warning it wants to see (all, local, none).

## How does it work?
For being able to control what level (all, local, none) of deprecation warnings are shown, we need 2 pieces of information:
1. how the module with the deprecated output is invoked from the root module
2. the CLI flag the user will be able to configure in order to say if it wants to see warnings only from the module(s) called from local source(s), from remote source or if it wants no warns at all.

For the information related to how the deprecated value relates to the root module (remote or local):
* For outputs:
  * This is handled during creation of the deprecationMark. This is created by iterating backwards, from the module that contains the output, to the root module. If it is finding any moduleCall with source=remote, it is labelling the mark as being from a remote source.
* For variables:
  * This is handled during validation of the variable. It is using a similar mechanism with the outputs, where it is iterating backwards, from the module that contains the variable, to the root module. If it is finding any moduleCall with source=remote, it is labelling accordingly the graph node responsible with the validation of the deprecated variable.

And later, when the deprecation diagnostics are generated, this information about the type of the module is given into the diagnostic.

When it comes to the information related to what value the user gave in the CLI arg, this is used to configure the view that is responsible with printing the diagnostics back to the user.
Therefore, when diagnostics are sent for being printed, we are checking in the view if the diagnostic is a deprecation one and if it refers to a local or remote variable/output.

Part of #2555

## Target Release

1.10.0

## Checklist

### Go checklist

 


internal/tofu/context_apply_test.go
```diff
@@ -5358,7 +5358,7 @@ func TestContext2Apply_provisionerMultiSelfRef(t *testing.T) {
 	plan, diags := ctx.Plan(context.Background(), m, states.NewState(), DefaultPlanOpts)
 	assertNoErrors(t, diags)
 
-	state, diags := ctx.Apply(context.Background(), plan, m)
+	state, diags := ctx.Apply(context.Background(), plan, m, DeprecationWarningLevelAll)
```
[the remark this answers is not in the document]
yottta: Thanks for the suggestion. I managed to add the field in the `tofu.ContextOpts` and `tofu.Context`.
I like it more this way.
Please check it and let me know if the approach is ok.
You can find these changes in 6f1803e12500f24cd69e075b46917d8264c2b06d

internal/command/arguments/apply_test.go
```diff
@@ -26,12 +26,13 @@ func TestParseApply_basicValid(t *testing.T) {
...
-				ViewType:     ViewHuman,
-				State:        &State{Lock: true},
-				Vars:         &Vars{},
+				AutoApprove:               false,
+				InputEnabled:              true,
+				PlanPath:                  "",
+				ViewType:                  ViewHuman,
+				State:                     &State{Lock: true},
+				Vars:                      &Vars{},
+				ModuleDeprecationWarnings: "all",
```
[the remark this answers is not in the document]
yottta: Yes, that was easy to do. Thanks.
Please check it in f1cb5965badd07b7c91ad0fbd8a6a9705b435421 and let me know if we could do it better.

internal/command/arguments/apply.go
```diff
@@ -51,6 +54,7 @@ func ParseApply(args []string) (*Apply, tfdiags.Diagnostics) {
 	cmdFlags.BoolVar(&apply.AutoApprove, "auto-approve", false, "auto-approve")
 	cmdFlags.BoolVar(&apply.InputEnabled, "input", true, "input")
 	cmdFlags.BoolVar(&apply.ShowSensitive, "show-sensitive", false, "displays sensitive values")
+	cmdFlags.StringVar(&apply.ModuleDeprecationWarnings, "deprecation-warn", "all", "control the level of deprecation warnings")
```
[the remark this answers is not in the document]
yottta: Changed in fc41d218d2e4e08344c83dd21a3a3cced549dfc7.
The reason why I had it before as `deprecation-warn` was due to the commands help text. With this new version, `module-deprecation-warnings`, the whole help text blocks needed an update on the indentation. Personally, I have no idea if that will have any impact on the way OpenTofu works on some devices. Do you have any idea why those help text blocks are formatted the way they are? Seems to be a specific width of the text that was in there.

internal/command/arguments/validate.go
```diff
@@ -45,6 +48,7 @@ func ParseValidate(args []string) (*Validate, tfdiags.Diagnostics) {
 	cmdFlags.BoolVar(&jsonOutput, "json", false, "json")
 	cmdFlags.StringVar(&validate.TestDirectory, "test-directory", "tests", "test-directory")
 	cmdFlags.BoolVar(&validate.NoTests, "no-tests", false, "no-tests")
+	cmdFlags.StringVar(&validate.ModuleDeprecationWarnLevel, "deprecation-warn", "all", "control the level of deprecation warnings")
```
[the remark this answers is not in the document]
ollevche: Currently, module call is treated as a whole without separation for its outputs. The resulting value is an `UnknownVal` of `Object` / `List` / `Map` type (based on `for_each`). The module won't be expanded during `tofu validate` so this value is the final one we can handle. In order to mark it we either need to mark the whole `UnknownVal` which leads to false positive deprecation warnings (raised warning for any module output if the module has at least one output marked as deprecated). Or we should make it "known" with unknowns inside. I didn't dig too deep into the second option, however it looks like a relatively large change for this type of problem.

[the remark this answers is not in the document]
yottta: Then it is rising the next question: should we remove the variables warnings from the validate command then? Or should we leave it as it is and maybe update docs for validate (and possible other places) to mention this thing?

internal/command/graph.go
```diff
@@ -32,6 +32,7 @@ func (c *GraphCommand) Run(args []string) int {
 	var moduleDepth int
 	var verbose bool
 	var planPath string
+	var moduleDeprecationWarnLevel string // TODO andrei update the help section
```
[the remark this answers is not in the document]
yottta: You are right. Is not even walking the graph. I will remove it.

internal/tofu/node_variable_reference.go
```diff
@@ -118,6 +127,12 @@ type nodeVariableReferenceInstance struct {
 	Addr   addrs.AbsInputVariableInstance
 	Config *configs.Variable // Config is the var in the config
 	Expr   hcl.Expression    // Used for diagnostics only
+
+	// ModuleDeprecationWarnLevel controls if the deprecation warning should be shown for this variable.
+	ModuleDeprecationWarnLevel DeprecationWarningLevel
+	// ModuleSource is the source module where this variable is coming from. When a variable is from a module call whose source is a relative
+	// path, then the source is pointing to an addrs.ModuleSourceLocal.
+	ModuleSource addrs.ModuleSource
```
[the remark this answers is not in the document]
yottta: Thanks for this. Indeed, cleaner this way.
Changed in 4f79e98156be1a1dca73290959ee55fe7e51a108.

internal/command/arguments/validate.go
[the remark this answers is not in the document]
ollevche: As per our discussion in Slack, we decided to leave it as is for now. `tofu validate` is meant to only validate syntax, but deprecation checks are based on the evaluated values.

Regarding the docs - it is a good idea to mention that somewhere. I believe module refactoring is the right place (this is a single piece of doc on the deprecation warnings). I will tweak that page in a separate PR: https://opentofu.org/docs/main/language/modules/develop/refactoring/#module-variables-and-outputs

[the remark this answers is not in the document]
ollevche: Updated docs in https://github.com/opentofu/opentofu/pull/2737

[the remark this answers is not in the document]
yottta: Thanks a lot for this! 🙌

internal/backend/local/backend_local.go
```diff
@@ -80,6 +80,8 @@ func (b *Local) localRun(ctx context.Context, op *backend.Operation) (*backend.L
 	coreOpts.UIInput = op.UIIn
 	coreOpts.Hooks = op.Hooks
 	coreOpts.Encryption = op.Encryption
+	// TODO andrei should we do this in remote/backend too?
```
ollevche, line 83: I would not extend those due to our limited ability to test those backends out. Also, implementations of remote / cloud backend APIs might potentially need to support this flag as well. I would treat it as a separate issue if someone from the community asks for it.

internal/command/console.go
```diff
@@ -30,9 +30,13 @@ type ConsoleCommand struct {
 func (c *ConsoleCommand) Run(args []string) int {
 	ctx := c.CommandContext()
 
+	var (
+		moduleDeprecatedWarning string
+	)
```
ollevche, line 35: nit:

```go
var moduleDeprecatedWarning string
```

internal/command/arguments/apply.go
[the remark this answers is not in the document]
ollevche: I don't think the formatting is critical, changes look good to me

internal/tofu/evaluate.go
```diff
@@ -352,9 +357,33 @@ func (d *evaluationStateData) GetLocalValue(addr addrs.LocalValue, rng tfdiags.S
...
+	}
+	calledModuleName := moduleCall.Name
+	parent := parentCfg.Parent
+	for parent != nil {
+		refCallCfg, ok := parent.Module.ModuleCalls[calledModuleName]
+		if !ok {
+			log.Printf("[ERROR] no module call found in %q for %q. This should not happen. Please report this to OpenTofu", parent.Path, calledModuleName)
+			return false
+		}
+		if _, ok := refCallCfg.SourceAddr.(addrs.ModuleSourceRemote); ok {
```
apparentlymart, line 372: I think this condition is essentially the same as [`refCallCfg.EntersNewPackage()`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/configs#ModuleCall.EntersNewPackage). "Module Package" is our terminology for something OpenTofu can fetch from somewhere else and extract into a local directory containing zero or more modules, and so "enters new package" means "the child module is in a different package than the parent module", which is just a different way to say "the call is to a non-local module".

That method is actually defined slightly differently: it returns `true` also for _module registry_ package addresses ([`addrs.ModuleSourceRegistry`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/addrs#ModuleSourceRegistry)), but I think that probably matches what this rule was intending to achieve, right?

If we _can_ use that existing method then I'd prefer that because it means we can keep the rule for deciding whether a source address is local vs. remote in only one central place.

internal/tofu/evaluate.go
```diff
@@ -352,9 +357,33 @@ func (d *evaluationStateData) GetLocalValue(addr addrs.LocalValue, rng tfdiags.S
...
+func isCallFromRemote(parentCfg *configs.Config, moduleCall addrs.ModuleCall) bool {
+	if _, ok := parentCfg.SourceAddr.(addrs.ModuleSourceRemote); ok {
+		return true
+	}
+	calledModuleName := moduleCall.Name
+	parent := parentCfg.Parent
+	for parent != nil {
+		refCallCfg, ok := parent.Module.ModuleCalls[calledModuleName]
+		if !ok {
+			log.Printf("[ERROR] no module call found in %q for %q. This should not happen. Please report this to OpenTofu", parent.Path, calledModuleName)
```
apparentlymart, line 369: We don't typically expect end-users to notice messages in the logs, so if we want folks to report this as a bug then it should probably be either a normal error diagnostic or a panic.

However, we do also ask folks to send us the trace logs when they report a strange behavior, so it would be a reasonable compromise to just remove the "This should not happen..." part of this message and rework this to be something we can notice in the trace logs that folks send us:

```
[ERROR] tofu.isCallFromRemote: no module call found in foo for bar
```

internal/tofu/evaluate_test.go
```diff
@@ -648,3 +648,63 @@ func evaluatorForModule(stateSync *states.SyncState, changesSync *plans.ChangesS
...
+		s, err := addrs.ParseModuleSource(source)
+		if err != nil {
+			t.Fatalf("failed to parse module source %q: %s", source, err)
+		}
+		return s
+	}
+	root := &configs.Config{
+		Module: &configs.Module{
+			ModuleCalls: map[string]*configs.ModuleCall{
+				childName: {SourceAddr: parseModuleSource(t, "git::https://github.com/user/repo//child")},
```
apparentlymart, line 674: Continuing from my earlier comment about using the `EntersNewPackage` method, a good extra test case to notice the difference between those two approaches would be:

`parseModuleSource(t, "registry.example.com/foo/bar/baz")`

...which should be parsed as an `addrs.ModuleSourceRegistry` for a module in the module registry at `registry.example.com`.

apparentlymart (commented):
I think this approach seems okay to me and my comments inline are minor.

---

I think there's at least one other way we could solve this that might be less invasive to existing code, though:

Our diagnostics model has a special idea of "extra data" that can travel along with a diagnostic, exposed through the `ExtraInfo` method of [`tfdiags.Diagnostic`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/tfdiags#Diagnostic). We use this in a few different situations where we need to treat certain diagnostics in a special way in the UI. For example:

- [We only mention that there are unknown values or sensitive values in scope when we're reporting an error that could potentially have been caused by those special situations being present](https://github.com/opentofu/opentofu/blob/1d3881630a39fb5882f6b3ae530d29a9f6b56a21/internal/command/views/json/diagnostic.go#L360-L361), because otherwise readers would often think that OpenTofu stating that e.g. `var.foo is unknown` is the _cause_ of the error being reported, rather than just some context reported alongside that error.
- [Certain messages caused by `check` blocks get filtered out during the apply phase](https://github.com/opentofu/opentofu/blob/1d3881630a39fb5882f6b3ae530d29a9f6b56a21/internal/backend/local/backend_apply.go#L236-L238).

That second example of totally discarding certain diagnostics seems quite similar to what you were aiming for in this commit, and so I wonder if a similar strategy would work here and would thus avoid the need for us to pass the "module deprecation warning level" value through all of these intermediate functions: instead, the code that generates the deprecation warnings and the code that renders the diagnostics could cooperate using `ExtraInfo` without any of the intermediate functions needing to know about it.

In the other examples you can see that we tend to include a function that can take a single `tfdiags.Diagnostic` and return a value depending on whether it has a particular kind of extra info, and that function usually belongs to the same package that's responsible for generating the diagnostics that function recognizes, which means that therefore the special types used for the extra info value can be unexported implementation details.

In this case since it's `package marks` that's responsible for generating the deprecation warning diagnostics, I guess following that model would mean that `package marks` gets a new function like this:

```go
package marks

// DiagnosticDeprecationCause checks whether the given diagnostic is
// a deprecation warning, and if so returns the deprecation cause and
// true. If not, returns the zero value of DeprecationCause and false.
func DiagnosticDeprecationCause(diag tfdiags.Diagnostic) (DeprecationCause, bool) {
    maybe := tfdiags.ExtraInfo[diagnosticExtraDeprecationCause](diag)
    if maybe == nil {
        return DeprecationCause{}, false
    }
    return maybe.diagnosticDeprecationCause(), true
}

type diagnosticExtraDeprecationCause interface {
    diagnosticDeprecationCause() DeprecationCause
}

// diagnosticDeprecationCause implements diagnosticExtraDeprecationCause
func (c DeprecationCause) diagnosticDeprecationCause() DeprecationCause {
    return c
}
```

The `DeprecationCause` from the mark could then be used directly as the "extra" of the diagnostic, as an implementation detail, since it would implement this unexported interface that the `DiagnosticDeprecationCause` function checks for. Code in a higher layer that knows the author's preference for warning level could then cause `marks.DiagnosticDeprecationCause` and filter out any unwanted diagnostics before they get rendered in the UI.

One annoyance with this today is that not all of our concrete implementations of `tfdiags.Diagnostic` can carry "ExtraInfo" values. :confounded: I notice that `marks.ExtractDeprecatedDiagnosticsWithExpr` uses `*hcl.Diagnostic`, which is okay because that _does_ support "extra info", but `ExtractDeprecationDiagnosticsWithBody` uses `tfdiags.AttributeValue` which currently does not.

We have [`tfdiags.Override`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/tfdiags#Override) that allows wrapping any arbitrary diagnostic value with an overridden severity and "extra info" value, which I believe is how the extra info for `check` blocks is implemented today, but it doesn't currently cooperate properly with `tfdiags.AttributeValue`. I previously ran into this same limitation in a draft PR, so there's a fix for it in https://github.com/opentofu/opentofu/pull/2643/commits/4a9f10a726ef6a925393528336db51ecf77e75f0 that we could potentially pull out into a separate PR to merge it.

internal/command/views/view.go
```diff
@@ -100,6 +105,21 @@ func (v *View) Diagnostics(diags tfdiags.Diagnostics) {
...
+	// Filter the deprecation warnings based on the cli arg
+	if v.ModuleDeprecationWarnLvl != tofu.DeprecationWarningLevelAll {
+		var newDiags tfdiags.Diagnostics
+		for _, diag := range diags {
+			if !tofu.DeprecatedOutputDiagAllowed(v.ModuleDeprecationWarnLvl, diag) {
+				continue
+			}
+			if !tofu.DeprecatedVariableDiagAllowed(v.ModuleDeprecationWarnLvl, diag) {
+				continue
+			}
```
apparentlymart, line 117: Do we need to distinguish these two cases with separate functions? It seems like both of these functions are doing essentially the same logic, so I wonder if it would work to have a single function `DeprecationDiagAllowed` which handles both input variable and output value deprecation warnings together. :thinking:

internal/command/apply.go
```diff
@@ -350,55 +350,63 @@ Usage: tofu [global options] apply [options] [PLAN]
...
+                                    
+  -show-sensitive                   If specified, sensitive values will be displayed.
+
+  -module-deprecation-warnings=all  Specify what type of warnings are shown. Accepted
+                                    values: all, local, none. When "all" is selected, OpenTofu
+                                    will show the deprecation warnings for all modules.
+                                    When "local" is selected, the warns will be shown
+                                    only for the modules that are imported with a relative path.
+                                    When "none" is selected, all the deprecation warnings
+                                    will be dismissed.
```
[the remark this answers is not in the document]
yottta: Omg! Thanks a lot! I asked before about this as it seemed to be indeed limited to a specific line length. I will change it back to what it was initially: deprecation-warns.

internal/command/views/view.go
yottta: Added in 5582eddc3354359f89d8731fce21827a4b109ac7. Thanks for the suggestion.

internal/command/apply.go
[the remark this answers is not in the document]
yottta: [Renamed](https://github.com/opentofu/opentofu/pull/2705/commits/26c311915d2b4b5f045f86f6680d7617193d9b96) and also added a comment in [alias.go](https://github.com/opentofu/opentofu/pull/2705/commits/01b720acfef5e2e3f2922914f8b8bcfc8f9f14fc)

internal/tofu/evaluate_test.go
yottta: Reworked in af24f0e5a28b560e26070247d80945985e1e1b2e

internal/tofu/transform_module_variable.go
```diff
@@ -133,3 +136,31 @@ func (t *ModuleVariableTransformer) transformSingle(g *Graph, parent, c *configs
 
 	return nil
 }
+
+// moduleImportedFromRemote is traversing upwards from the given module call to the root module and is looking for any
+// module on the path for which configs.Module.EntersNewPackage=true.
+// This is needed to know if a variable is referenced from a module imported from a remote source or from a local one.
+func moduleImportedFromRemote(c *configs.Config, call *configs.ModuleCall) bool {
```
cam72cam, line 143: This appears to be a duplicate function from evaluate.go:isCallFromRemote

Could we move this common functionality to the configs package as it's only operating on objects that live there?

`func (c *configs.Config) IsModuleCallFromRemoteModule(call *configs.ModuleCall) bool`

internal/command/apply.go
```diff
@@ -397,9 +397,18 @@ Options:
   -show-sensitive        If specified, sensitive values will be displayed.
 
   -json                  Produce output in a machine-readable JSON format,
-                         suitable for use in text editor integrations and 
+                         suitable for use in text editor integrations and
                          other automated systems. Always disables color.
 
+  -deprecation-warns     Specify what type of warnings are shown. Accepted
```
cam72cam, line 403: This is a common comment to all of the similar changes below:
The format is usually -$arg=$default, otherwise it's seen as a boolean flag

internal/tofu/transform_module_variable.go
yottta: There is one important difference: one is using configs.ModuleCall and the other is using addrs.ModuleCall.
If you know a way to generate a addrs.ModuleCall from a configs.ModuleCall, I would gladly get rid of one of those.

internal/command/apply.go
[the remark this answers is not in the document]
cam72cam: Being frank, I'm not a huge fan of -deprecation-warns.  It does not match -consolidate-warnings, we don't typically abbreviate CLI args.

I think we should use either `-deprecation=level` or `-deprecation-warnings=level`, with a preference for the first option.  Warnings feels like an implementation detail leak.  We should also consider `-deprecation-level=level` or similar.

internal/tofu/transform_module_variable.go
cam72cam: Ah, thanks.  Let me take a look

internal/command/apply.go
yottta: Fixed in https://github.com/opentofu/opentofu/pull/2705/commits/71850ba3be3c15cb1fe310619e94a3fc27a94b6e

internal/tofu/transform_module_variable.go
cam72cam: I think that other than the `if _, ok := c.SourceAddr.(addrs.ModuleSourceRemote); ok {` check, the rest of the code is identical.  We could just pass the calledModuleName in to the common function instead of a more complex string.

yottta: Funny how I missed that opportunity. Thanks.
Changed in 9782eb2495d09b5869fe8ede4e57447e4aceeb22

internal/command/apply.go
apparentlymart: The `-arg=default` shape is actually more of a legacy thing on docs that haven't been updated for a long time. This changed because folks found it confusing that the left column effectively described an option that would do nothing at all (select the same behavior as default) and then in the case of boolean options the description would then also tend to describe the _opposite_ of the effect of setting the option.

The more modern style for non-boolean options is to use a generic placeholder noun for the argument and then use the same terminology in the descriptive text. There isn't an obviously-useful placeholder now to use here but I might suggest something like "mode" or "type" and then use that same noun in the text on the right when describing the option.

(Not relevant to this case, but for completeness: the modern convention for _boolean_ options is to show them with no `=value` at all if `false` is the default, or to show them with `=false` if `true` is the default, so that copying exactly what's shown in the left column will produce the effect described in the right column, rather than the opposite of that effect.)

cam72cam: I'd prefer using "all" as it's the default value

internal/command/e2etest/automation_test.go
```diff
@@ -243,3 +243,42 @@ func TestPlanOnlyInAutomation(t *testing.T) {
...
 	}
 }
+
+func TestPlanOnDeprecated(t *testing.T) {
+	t.Parallel()
+
+	// This test reaches out to registry.opentofu.org to download the
+	// template and null providers, so it can only run if network access is
+	// allowed.
+	skipIfCannotAccessNetwork(t)
```
apparentlymart, line 253: It doesn't seem like the `deprecated-values` test fixture actually uses either of the providers mentioned in the comment above, or anything else that would be loaded over the network, so I think we could remove this guard to allow this test to run unconditionally.

internal/lang/marks/diag.go
```diff
@@ -0,0 +1,53 @@
...
+	return c
+}
+
+// DeprecatedOutputDiagnosticExtra is a container for the DeprecationCause used to decide later if the diagnostic
+// needs to be shown or not
+type DeprecatedOutputDiagnosticExtra struct {
+	Cause DeprecationCause
+
+	wrapped interface{}
+}
```
apparentlymart, line 36: Does this type need to be exported?

I see that it's used in some of the tests in `package tofu` but I wonder if we could encapsulate this a little better so that only `package marks` needs to know how this works internally. For example, we could export from _this_ package a function that produces a function that matches the signature used by the last argument of `tfdiags.Override`, and then use that function directly as the argument:

```go
func DeprecatedOutputDiagnosticOverride(cause DeprecationCause) func() tfdiags.DiagnosticExtraWrapper {
    return func () tfdiags.DiagnosticExtraWrapper {
        return &DeprecatedOutputDiagnosticExtra{
            Cause: cause,
         }
    }
}
```

Honestly the design of `tfdiags.Override` is kinda complicated and maybe one day we can find a way to simplify it a little so this extra function isn't needed, but hopefully something like the above could at least keep most of the details encapsulated in `package marks`?

(This is not super important, so if this is particularly hard to do then I'm okay with not bothering. Just trying to think about ways to minimize how much exported API surface we have in each package.)

apparentlymart (approved):
I know you're currently still considering different ways to name the command line option and so perhaps this will stay open until that discussion is resolved, but overall I think this looks great! Thanks for trying the alternative approach.

I wonder if we could merge this with `-deprecation-warns` as a placeholder name for now and then adjust that in a separate PR later if we choose a different approach... though whether that's worth doing probably depends on how large the subsequent renaming change would be. If the changes would be limited only to the `command/arguments` package -- or at least, _mostly_ just that -- then it would be nice to have the rest of the functionality landed just so that we can avoid this potentially becoming conflicted with other work in the meantime.

internal/lang/marks/diag.go
yottta: Thanks! Really good point! Really kind of you to provide some code that actually worked flawlessly.
So applied in 538dec6f1ad081f9b78bcbb614cf595869b21a6f.

internal/command/e2etest/automation_test.go
yottta: Yep, good catch. Copy-paste leftover.
Fixed in 538dec6f1ad081f9b78bcbb614cf595869b21a6f

cam72cam (approved):

internal/command/apply.go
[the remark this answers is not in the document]
yottta: Based on the discussion from dev-general in the community slack workspace, we will provide a new PR soon for updating the cli argument based on the decision there.

PR #2726, opened 2025-04-28
Remove providercache.Dir.metaCache

cam72cam:
This is in preparation for parallelizing the package installation process.  Most of the "locate package(s)" functions on the Dir struct are only currently used by tests.  The only one that is used by non-test code is Dir.ProviderVersion, which is really just a path construction and `os.Stat()` call.

On the other hand, we had the metaCache structure on Dir which needed to be wiped in preparation for a a "rescan" every time we think the dir on disk was modified.  As it's only current utility was exercised during tests, it made sense to remove the cache and do the re-scan for those cases.

As these changes also reduce the number of times we scan the filesystem looking for providers, this could have a mild performance improvement for situations with a massive cache directory or slow networked filesystem.

## Target Release

1.10.0

## Checklist

### Go checklist

 

### DIFF
--- a/internal/providercache/dir.go
+++ b/internal/providercache/dir.go
@@ -37,23 +37,6 @@ import (
 type Dir struct {
 	baseDir        string
 	targetPlatform getproviders.Platform
-
-	// metaCache is a cache of the metadata of relevant packages available in
-	// the cache directory last time we scanned it. This can be nil to indicate
-	// that the cache is cold. The cache will be invalidated (set back to nil)
-	// by any operation that modifies the contents of the cache directory.
-	//
-	// We intentionally don't make effort to detect modifications to the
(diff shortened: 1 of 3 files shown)


Yantrio (approved):
LGTM, however I think we should read in depth all the commits here: https://github.com/hashicorp/terraform/pull/24447 to check that we're making the correct assumptions around the intentions of this metaCache.

apparentlymart (approved):
This makes sense to me!

Perhaps some of this functionality will come back later as we add new features, but I agree that as we currently stand really the only questions we need to be able to answer are "do we have this version of this provider in the cache?" and "where is the package directory for this version of this provider in the cache?" and we don't need to be enumerating the entire directory structure to answer either of those questions.

PR #2793, opened 2025-05-12
[RFC] Ephemeral resources and write-only attributes

yottta:
Part of #1996
[Rendered view](https://github.com/opentofu/opentofu/blob/ephemeral-rfc/rfc/20250317-ephemeral-resources.md)

## Target Release

1.11.0

## Checklist

### Go checklist

 

### DIFF
--- a/rfc/20250317-ephemeral-resources.md
+++ b/rfc/20250317-ephemeral-resources.md
@@ -661,12 +661,12 @@ This is useful when user wants to configure different properties between write o
 
 > [!NOTE]
 >
-> This keyword is related to the `apply` command and not to the `apply` phase, meaning that when running `tofu apply`, `terraform.applying` should still be `true` also during the `plan` phase of the `apply` command.
+> This keyword is related to the `apply` command and not to the `apply` phase, meaning that when running `tofu apply`, `terraform.applying` should still be `true` also during the `plan` phase of the `apply` command. This is `true` also when running a destroy operation.
 
 This is an ephemeral value that should be handled correctly and ensure that its value or any other value generate from it will not end up in a plan/state file.
 
 #### `ephemeralasnull` function
-`ephemeralsnull` function is useful when an object that was built by referencing an ephemeral value wants to be used into a non-ephemeral context.
+`ephemeralasnull` function is useful when an object that was built by referencing an ephemeral value wants to be used into a non-ephemeral context.
 This is getting a dynamic value and by traversing it, is looking for any ephemeral value and is nullifying it, but it does not nullify any non-ephemeral value within the object.
 
 For example:
@@ -698,6 +698,11 @@ test = {
 ```
 
 This function should work perfectly fine also with a non-ephemeral value.
+
+> [!NOTE]
+>
+> When we encounter an output in the root module that is referencing an ephemeral value, we could recommend to use `ephemeralasnull` to be able to store that information in the state.
+> This would be a warning that will come together with the error diagnostic discussed in the ephemeral outputs section.
 
 ## Open Questions
 
--- a/rfc/20250317-ephemeral-resources.md
+++ b/rfc/20250317-ephemeral-resources.md
@@ -2,11 +2,23 @@
 
 Issue: https://github.com/opentofu/opentofu/issues/1996
 
-Right now, OpenTofu information for resources and outputs are written to state as it is. This is presenting a security risk
-as some of the information from the stored objects can contain sensitive bits that can become visible to whoever is having access to the state file.
+Right now, OpenTofu information for resources and outputs are written to state and plan files as it is. This is presenting a security risk
+as some of the information from the stored objects can contain sensitive bits that can become visible to whoever is having access to the state or plan files.
 
-In order to provide a better solution for the aforementioned situation, OpenTofu introduces the concept of "ephemerality".
-Any new feature under this new concept will provide ways to skip values from being written to the state and plan files.
+In order to provide a better solution for the aforementioned situation, OpenTofu introduces the concept of "ephemerality" which is meant to make use of the already existing functionality in terraform-plugin-framework.
+Any new feature under this new concept should provide ways to skip values from being written to the state and plan files.
+
+This new concept is going to offer another way to tackle the aforementioned issue, adding one more option in OpenTofu to choose for securing the plan and state files.
+Here are the other existing options, providing different levels of safety:
+* sensitive marked outputs and variables
+  * The values marked this way ensures only that the information is sanitized from the user interface, but these are still stored in plaintext in the state and plan files.
+* state encryption (recommended way)
+  * This is meant to provide in transit and at rest encryption of the state and plan files, regardless of what providers/modules offer.
+  * By using this you don't need to choose what to store and what not, but everything is safely stored.
+  * This is also preventing state tempering and [privilege escalation](https://www.plerion.com/blog/hacking-terraform-state-for-privilege-escalation).
+
+
+
 
 To make this work seamlessly with most of the blocks that OpenTofu supports, the following functionalities need to be able to work with the ephemeral concept:
 * `resource`'s `write-only` attributes
@@ -17,6 +29,10 @@ To make this work seamlessly with most of the blocks that OpenTofu supports, the
 * providers
 * provisioners
 * `connection` block
+
+> [!NOTE]
+>
+> In order to provide a similar and familiar UX for the users, the proposal in this RFC is heavily inspired by the Terraform public documentation and available blog posts on the matter.
 
 ## Proposed Solution
 
--- a/rfc/20250317-ephemeral-resources.md
+++ b/rfc/20250317-ephemeral-resources.md
@@ -244,6 +244,17 @@ Beside the attributes in a schema of an ephemeral resource, the block should sup
 * `provider`
 * `lifecycle`
 
+The only `lifecycle` content that ephemerals should support are `precondition` and `postcondition`. In case OpenTofu will find other known attributes of the `lifecycle` block, it should show an error similar to the following:
+```
+│ Error: Invalid lifecycle configuration for ephemeral resource
+│                                                           
+│   on ../mod/main.tf line 44, in ephemeral "aws_secretsmanager_secret_version" "secret_retrieval":                     
+│   44:     create_before_destroy = true                    
+│                                                           
+│ The lifecycle argument "create_before_destroy" cannot be used in ephemeral resources. This is meant 
+│ to be used strictly in "resource" blocks.
+```
+
 The meta-argument `provisioner` should not be supported.
 #### Providers
 `provider` block is ephemeral by nature, meaning that the configuration of this is never stored into state/plan file.
@@ -717,6 +728,7 @@ In order to make that available to be used, [providers.Interface](https://github
 #### Configuration model
 Beside the attributes that are defined by the provider for an ephemeral resource, the following meta-arguments needs to be supported by any ephemeral block:
 * lifecycle
+  * The only attributes supported by the `lifecycle` in `ephemeral` blocks context are `precondition` and `postcondition`. If any found, an error will be raised.
 * count
 * for_each
 * depends_on
--- a/rfc/20250317-ephemeral-resources.md
+++ b/rfc/20250317-ephemeral-resources.md
@@ -75,9 +75,9 @@ resource "aws_db_instance" "example" {
 
 As seen in this particular change of the [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework/commit/ecd80f67daed0b92b243ae59bb1ee2077f8077c7), the write-only attribute cannot be configured for set attributes, set nested attributes and set nested blocks.
 
-Write-only attributes cannot generate a plan diff. 
-This is because the prior state does not contain a value that OpenTofu can use to compare the new value against and 
-also the value provider returns during planning of a write-only argument should always be null. 
+Write-only attributes cannot generate a plan diff.
+This is because the prior state does not contain a value that OpenTofu can use to compare the new value against and
+also the value provider returns during planning of a write-only argument should always be null.
 This means that there could be inconsistencies between plan and apply for the write-only arguments.
 
 #### Variables
@@ -110,7 +110,7 @@ Usage in any other place should raise an error:
 ╵
 ```
 
-For being able to use ephemeral variables, the module's authors need to mark those as so. 
+For being able to use ephemeral variables, the module's authors need to mark those as so.
 If OpenTofu finds an ephemeral value given to a non-ephemeral variable in a module call, an error will be shown:
 ```hcl
 │ Error: Invalid usage of ephemeral value
@@ -123,8 +123,8 @@ If OpenTofu finds an ephemeral value given to a non-ephemeral variable in a modu
 ```
 
 
-OpenTofu should not store ephemeral variable(s) in plan files. 
-If a plan is generated from a configuration that is having at least one ephemeral variable, 
+OpenTofu should not store ephemeral variable(s) in plan files.
+If a plan is generated from a configuration that is having at least one ephemeral variable,
 when the planfile will be applied, the value(s) for the ephemeral variable(s) needs to be provided again.
 
 #### Outputs
@@ -146,20 +146,20 @@ The ephemeral outputs are available during plan and apply phase and can be acces
 * `provisioner` block
 * `connection` block
 
-An `output` block from a root module cannot be marked as ephemeral. 
-This limitation is natural since ephemeral outputs are meant to be skipped from the state file. 
+An `output` block from a root module cannot be marked as ephemeral.
+This limitation is natural since ephemeral outputs are meant to be skipped from the state file.
 Therefore, there is no usage of such a defined output block in a root module.
 When encountering an ephemeral output in a root module, an error similar to this one should be shown:
 ```hcl
 │ Error: Unallowed ephemeral output
-│ 
+│
 │   on main.tf line 36:
 │   36: output "write_only_out" {
-│ 
+│
 │ Root module is not allowed to have ephemeral outputs
 ```
 
-Ephemeral outputs are useful when a child module returns sensitive data, 
+Ephemeral outputs are useful when a child module returns sensitive data,
 allowing the caller to use the value of that output in other ephemeral contexts.
 When using outputs in non-ephemeral contexts, OpenTofu should show an error similar to the following:
 ```hcl
@@ -173,7 +173,7 @@ When using outputs in non-ephemeral contexts, OpenTofu should show an error simi
 ╵
 ```
 
-Any output that wants to use an ephemeral value, it needs to be marked as ephemeral too. 
+Any output that wants to use an ephemeral value, it needs to be marked as ephemeral too.
 Otherwise, it needs to show an error:
 ```hcl
 │ Error: Output not marked as ephemeral
@@ -258,12 +258,14 @@ Whenever doing so, the output of the provisioner execution should be suppressed:
 (local-exec): (output suppressed due to ephemeral value in config)
 ```
 #### `connection` block
-When the `connection` block is configured, this should be allowed to use ephemeral values from variables, outputs, locals and values from ephemeral resources. 
+When the `connection` block is configured, this should be allowed to use ephemeral values from variables, outputs, locals and values from ephemeral resources.
 
 ### Example on how the new changes will work together
 To better understand how all of this should work in OpenTofu, let's take a look at a comprehensive example.
 #### Configuration
-##### `./mod/main.tf`
+<details>
+<summary>./mod/main.tf</summary>
+
 ```hcl
 terraform {
   required_providers {
@@ -275,10 +277,8 @@ terraform {
 }
 
 variable "secret_map" {
-  type      = map(string)
-  default   = {
-    "key_from_module": "default value from module"
-  }
+  type = map(string)
+  default = {}
   ephemeral = true # (1)
 }
 
@@ -287,30 +287,63 @@ variable "secret_version" { # (2)
   default = 1
 }
 
+variable "secret_manager_arn" {
+  type    = string
+  default = ""
+}
+
 resource "aws_secretsmanager_secret" "manager" {
-  name = "ephemeral-rfc-example"
+  count = var.secret_version > 1 ? 1 : 0
+  name  = "ephemeral-rfc-example"
 }
 
 resource "aws_secretsmanager_secret_version" "secret_creation" {
-  secret_id                = aws_secretsmanager_secret.manager.arn
-  secret_string_wo         = jsonencode(var.secret_map)
+  count                    = var.secret_version > 1 ? 1 : 0
+  secret_id                = aws_secretsmanager_secret.manager[0].arn
+  secret_string_wo = jsonencode(var.secret_map) # (3)
   secret_string_wo_version = var.secret_version
 }
 
-ephemeral "aws_secretsmanager_secret_version" "secret_retrieval" { # (3)
-  secret_id  = aws_secretsmanager_secret.manager.arn
+ephemeral "aws_secretsmanager_secret_version" "secret_retrieval" { # (4)
+  count     = var.secret_version > 1 ? 1 : 0
+  secret_id = aws_secretsmanager_secret.manager[0].arn
   depends_on = [
     aws_secretsmanager_secret_version.secret_creation
   ]
 }
 
+ephemeral "aws_secretsmanager_secret_version" "secret_retrieval_direct" {
+  count     = var.secret_version > 1 ? 0 : 1
+  secret_id = var.secret_manager_arn
+}
+
 output "secrets" {
-  ephemeral = true # (4)
-  value     = jsondecode(ephemeral.aws_secretsmanager_secret_version.secret_retrieval.secret_string)
+  value = "${var.secret_version > 1 ?
+    jsondecode(ephemeral.aws_secretsmanager_secret_version.secret_retrieval[0].secret_string) :
+    jsondecode(ephemeral.aws_secretsmanager_secret_version.secret_retrieval_direct[0].secret_string)}"
+  ephemeral = true # (5)
+}
+
+output "secret_manager_arn" {
+  value = var.secret_version > 1 ? aws_secretsmanager_secret.manager[0].arn : null
 }
 ```
+This module can be used for two separate operations:
+1. When `secret_version` is having a value greater than 1, it will add the given `secret` into a secret manager.
+2. When `secret_version` is not given, it will use the `secret_manager_arn` to read the secret by using an ephemeral resource.
 
-##### `./main.tf`
+Details:
+- (1) Variables should be able to be marked as ephemeral. By doing so, those should be able to be used only in ephemeral contexts.
+- (2) Version field that is going together with the actual write-only argument to be able to update the value of it. To upgrade the secret, the version field needs to be updated, otherwise OpenTofu should generate no diff for it.
+- (3) `aws_secretsmanager_secret_version.secret_creation.secret_string_wo` is the write-only attribute that is receiving the `secret` variable which is ephemeral (even though a write-only argument can use also a non-ephemeral value).
+- (4) Using ephemeral resource to retrieve the secret. Maybe looks a little bit weird, because right above we are having the resource of the same type that is looking like it should be able to be used to get the secret. In reality, because that `resource` is using the write-only attribute `secret_string_wo` to store the information, that field is going to be null when referenced.
+- (5) Module output that is referencing an ephemeral value, it needs to be marked as ephemeral too. Otherwise, OpenTofu should generate an error.
+  - This is similar to the behavior that is already present for `sensitive` values.
+</details>
+
+<details>
+<summary>./store/main.tf</summary>
+
 ```hcl
 terraform {
   required_providers {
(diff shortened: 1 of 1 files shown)


rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+* Based on the marks, ensure that the variable cannot be used in other contexts than the ephemeral ones (see the User Documentation section for more details on where this is allowed).
+* Check the state of [#1998](https://github.com/opentofu/opentofu/pull/1998). If that is merged, in the changes where variables from plan are verified against the configuration ones, we also need to add a validation on the ephemerality of variables. If the variable is marked as ephemeral, then the plan value is allowed (expected) to be missing. 
+* Ensure that when the prompt is shown for an ephemeral variable there is an indication of that:
+  ```hcl
+  var.password (ephemeral)
+     Enter a value:
+  ````
+* If a module is having an ephemeral variable declared, that variable can get values from any source, even another non-ephemeral variable.
+
+We should use boolean marks, as no additional information is required to be carried. When introducing the marks for these, extra care should be taken in *all* the places marks are handled and ensure that the existing implementation around marks is not affected.
```
Yantrio, line 435: do we also need to introduce some checks/tests to ensure that functions such as `sensitive()` do not accidentally strip the ephemeral mark too?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+* Mark the outputs with a new mark and ensure that the marks are propagated correctly.
+  * We should use boolean marks, as no additional information is required to be carried. When introducing the marks for these, extra care should be taken in *all* the places marks are handled and ensure that the existing implementation around marks is not affected.
+* Based on the marks, ensure that the output cannot be used in other contexts than the ephemeral ones (see the User Documentation section for more details on where this is allowed).
+
+> [!TIP]
+>
+> For an example on how to properly introduce a new mark in the outputs, you can check the [PR](https://github.com/opentofu/opentofu/pull/2633) for the deprecated outputs.
+
+Strict rules:
+* A root module cannot define ephemeral outputs since are useless.
```
Yantrio, line 450:
```suggestion
* A root module cannot define ephemeral outputs.
```

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+  eg2 = var.var2 // not ephemeral
+  eg3 = var.var3 == "" ? var.var2 : var.var1 // ephemeral because of var3 conditional
+  eg4 = var.var1 == "" ? var.var2 : var.var3 // ephemeral because of var3 usage
+  eg5 = "${var.var3}-${var.var1}" // ephemeral because of var3 usage
+  eg6 = local.eg4 // ephemeral because of eg4 is ephemeral
+}
+```
+
+Once a local is marked as ephemeral, this can be used only in other ephemeral contexts. Check the `User Documentation` section for more details on the allowed contexts.
+
```
Yantrio, line 500: let's not forget to document here that variables that take in ephemeral marked values should also be implicitly marked as ephemeral

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+In order to provide a better solution for the aforementioned situation, OpenTofu introduces the concept of "ephemerality".
+Any new feature under this new concept will provide ways to skip values from being written to the state and plan files.
+
+To make this work seamlessly with most of the blocks that OpenTofu supports, the following functionalities need to be able to work with the ephemeral concept:
+* `resource`'s `write-only` attributes
+* variables
+* outputs
+* locals
+* `ephemeral` resources
```
Yantrio, line 16: I just want to go on record that I don't like how they're called `ephemeral resources` because this is confusing. Mainly because `resource` and `ephemeral` here are 2 different types of blocks.

I would prefer a different name, such as "ephemeral entities" etc, but I fear it may be too late.

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+  eg5 = "${var.var3}-${var.var1}" // ephemeral because of var3 usage
+  eg6 = local.eg4 // ephemeral because of eg4 is ephemeral
+}
+```
+
+Once a local is marked as ephemeral, this can be used only in other ephemeral contexts. Check the `User Documentation` section for more details on the allowed contexts.
+
+### Ephemeral resources
+Due to the fact ephemeral resources are not stored in the state/plan file, this block is not creating a diff in the OpenTofu's UI.
+Instead, OpenTofu will notify the user of opening/renewing/closing an ephemeral resource with messages similar to the following:
```
Yantrio, line 503: Can we discuss what these actions mean? `opening` `renewing` and `closing`?  As someone approaching this fresh I think I can assume what it means but I think we should declare it explicitly so we can have a conversation about it. 

It feels like a strange set of verbs to show up in the ui to me (Opening.. Closing... etc)

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+ephemeral.playground_random.password: Closing...
+ephemeral.playground_random.password: Closing succeeded after 0s
+```
+
+Ephemeral resources lifecycle is similar with the data blocks:
+* Both basic implementations require the same methods (`Metadata` and `Schema`) while datasource is defining `Read` compared with the ephemeral resource that is defining `Open`. When talking about the basic functionality of the ephemeral resources, the `Read` method will behave similarly with the `Read` on a datasource, where it reads the data.
+* Also, both blocks support `Configure`, `ConfigValidators` and `ValidateConfig` as extensions of the basic definition.
+* Ephemeral resources do support two more operations in contrast with datasources:
+  * `Renew`
+    * Together with the data returned by the `Open` method call, the provider can also specify a `RenewAt` which will be a specific moment in time when OpenTofu should call the `Renew` method to get an updated information from the ephemeral resource. OpenTofu will have to check for `RenewAt` value anytime it intends to use the value returned by the ephemeral resource.
```
Yantrio, line 516: This renewat functionality isn't super clear to me based on the doc above, I wasn't expecting this. 

Can you describe it in abit more detail and why it's needed? Is it so that providers can determine if/when opentofu should refetch rather than it only being a human (version) driven re-fetch?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+  * `Close`
+    * When an ephemeral resource is having this method defined, it is expecting it to be called in order to release a possible held resource. A good example of this is with a Vault provider that could provide a secret by obtaining a lease, and when the secret is done being used, OpenTofu should call `Close` on that ephemeral resource to instruct on releasing the lease and revoking the secret.
```
Yantrio, line 518:
```suggestion
    * When an ephemeral resource is having this method defined, it is expecting it to be called in order to release a possible held resource. A good example of this is with a Vault/OpenBao provider that could provide a secret by obtaining a lease, and when the secret is done being used, OpenTofu should call `Close` on that ephemeral resource to instruct on releasing the lease and revoking the secret.
```

Yantrio, line 518: >it to be called in order to release a possible held resource.

When, how?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+* `Result` that will contain the actual ephemeral information. This is what OpenTofu needs to handle to make it available to other ephemeral contexts to reference.
+* `RenewAt` being an optional timestamp indicating when OpenTofu will have to call `Renew` method on the provider before using again the data from the `Result`.
+
+Observations:
+* In the `Result`, OpenTofu is epecting to find any non-computed given values in the request, otherwise will return an error.
+* In the `Result`, the fields marked as computed can be either null or have an actual value. If an unknown if found, OpenTofu will return an error.
+
+> [!NOTE]
+>
+> If any information in the configuration of an ephemeral resource is unknown during the `plan` phase, OpenTofu will defer the provisioning of the resource for the `apply` phase.
```
Yantrio, line 556:
```suggestion
> If any information in the configuration of an ephemeral resource is unknown during the `plan` phase, OpenTofu will defer the provisioning of the resource for the `apply` phase. This means that inconsistency can occur between the plan and the apply phase.
```

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+#### `Renew` method details
+The `Renew` method is called only if the response from `Open` or another `Renew` call is containing a `RenewAt`.
+When `RenewAt` is present, OpenTofu, before using the `Result` from the `Open` method response, will check if the current timestamp is at or over `RenewAt` and will call the `Renew` method by providing the previously returned `Private` information.
+
+> [!NOTE]
+> 
+> `Renew` does not return a *new* information meant to replace the initial `Result` returned by the `Open` call.
+> Due to this, `Renew` is only useful for system similar to Vault where the lease can be renewed without generating new data.
```
Yantrio, line 565:
```suggestion
> Due to this, `Renew` is only useful for systems where an entity can be renewed without generating new data.
```

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+#### `Close` method details
+When OpenTofu is done using an ephemeral resource, it needs to call its `Close` method to ensure that any remote data associated with the data returned in `OpenResponse.Result` is released and/or cleaned up properly.
```
Yantrio, line 568: When you say "Done using", is that
- when the apply phase is finished
- when the value is no longer referenced by a resource/output?
- when the node in the graph has been read completely

Can you elaborate on what `done` is here please?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+The `Renew` request is requiring the latest `Private` data returned by the call to `Open` or `Renew` method.
```
Yantrio, line 569: I think this line needs to be further up, not under the `Close` method details

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+#### `ConfigValidators` and `ValidateConfig` methods details
+There is not much to say here, since this is the same lifecycle that a datasource is having.
+
+#### Testing support
+The testing support will be documented later into a different RFC, or as amendment to this one.
```
Yantrio, line 575: Can you give a reason as to why we're not putting it in this RFC?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+There are already OpenTofu contexts that are not saved in state/plan file:
+* `provider` configuration
+* `provisioner` blocks
+* `connection` blocks
+
+In all of these, referencing an ephemeral value should work as normal.
+
+### Utilities
+#### `terraform.applying`
+The `terraform.applying` needs to be introduced to allow the user to check if the current command that is running is `apply` or not.
```
Yantrio, line 587: why terraform and not tofu?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+This is useful when user wants to configure different properties between write operations and read operations.
+
+`terraform.applying` will be set to `true` when `tofu apply` is executed and `false` in any other command.
+
+> [!NOTE]
+>
+> This keyword is related to the `apply` command and not to the `apply` phase, meaning that when running `tofu apply`, `terraform.applying` will still be `true` also during the `plan` phase of the `apply` command.
```
Yantrio, line 594: what about during destroy?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+This is an ephemeral value that should be handled correctly and ensure that its value or any other value generate from it will not end up in a plan/state file.
+
+#### `ephemeralasnull` function
+`ephemeralsnull` function is useful when an object that was built by referencing an ephemeral value wants to be used into a non-ephemeral context.
```
Yantrio, line 599: should we recommend that people use this with a nice warning in the UI when they try and create root module outputs with ephemeral values ?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
+# Ephemeral resources, variables, outputs, locals and write-only arguments
+
+Issue: https://github.com/opentofu/opentofu/issues/1996
+
+Right now, OpenTofu information for resources and outputs are written to state as it is. This is presenting a security risk
+as some of the information from the stored objects can contain sensitive bits that can become visible to whoever is having access to the state file.
```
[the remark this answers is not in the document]
cam72cam: It also might be worth introducing the idea that state encryption is our currently recommended solution.

IMO they are two overlapping, but not exclusive concepts

Advantages of State Encryption:
* All secrets are protected, regardless of provider or resource/module configuration
* Don't need to chose what is protected and what is exposed
* Also protects against provider injection
  - https://www.plerion.com/blog/hacking-terraform-state-for-privilege-escalation
  
  Advantages of Ephemeral
  * No keys to loose / have leaked, secrets (if identified properly) are not stored in the first place.

cam72cam:
Overall, this is a really good start!  I think the main area I'd like to see expanded is the user documentation.  Your example is quite comprehensive and also quite dense.  I think introducing a few simpler examples, before going with the "Here's the a complex example showing how the simpler building blocks interplay" would be helpful.

rfc/20250317-ephemeral-resources.md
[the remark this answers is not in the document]
Yantrio: That raises another question, we should explain how we shouldn't use this with state encryption (due to the ephemerality) because some people may think "This is great for secrets, i can use it for my passphrase" and have issues.

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+A write-only attribute can accept an ephemeral or a non-ephemeral value, even though it's recommended to use ephemeral values for such attributes.
+
+Because these attributes are not written to the plan file, the update of a write-only attribute it's getting a little bit trickier.
+Provider implementations do generally include also a "version" argument linked to the write-only one.
+For example having a write-only argument called `secret`, providers should also include
+a non-write-only argument called `secret_version`. Every time the user wants to update the value of `secret`, it needs to change the value of `secret_version` to trigger a change.
+The provider implementation is responsible with handling this particular case: because the version attribute is stored also in the state, the provider needs to compare the value from the state with the one from the configuration and in case it differs, it will trigger the update of the `secret` attribute.
+
+The write-only attributes are supported momentarily by a low number of providers and resources.
```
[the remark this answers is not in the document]
yottta: In 91f105241ac49c3c20ba51d52306be62c7ac4c03

rfc/20250317-ephemeral-resources.md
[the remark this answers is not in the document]
yottta: Please check this out: ae712fdaf7a5ba4b69cfcf0e65d960df7d301dac

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
+# Ephemeral resources, variables, outputs, locals and write-only arguments
+
+Issue: https://github.com/opentofu/opentofu/issues/1996
+
+Right now, OpenTofu information for resources and outputs are written to state as it is. This is presenting a security risk
+as some of the information from the stored objects can contain sensitive bits that can become visible to whoever is having access to the state file.
+
+In order to provide a better solution for the aforementioned situation, OpenTofu introduces the concept of "ephemerality".
```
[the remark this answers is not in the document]
yottta: Please check this out ae712fdaf7a5ba4b69cfcf0e65d960df7d301dac

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+* write-only arguments
+* other ephemeral variables
+* ephemeral outputs
+* local values
+* ephemeral resources
+* provisioner blocks
+* connection blocks
+* provider configuration
+
+Usage in any other place will raise an error:
```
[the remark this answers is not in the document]
yottta: Changed in 119b5c9a3c32f5a41a671586e4852515b4ee3b34. Indeed there were a lot of places. Sorry for this, seems a bias on my side. Weird.

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+Locals marked as ephemeral are available during plan and apply phase and can be referenced only in specific contexts:
+* ephemeral variables
+* other ephemeral locals
+* write-only attributes
+* ephemeral resources
+* `provider` blocks configuration
+* `connection` and `provisioner` blocks
+
+### Ephemeral resource
+In contrast with the write-only arguments where only specifically tagged attributes are skipped from the state/plan file, `ephemeral` resources are skipped entirely.
```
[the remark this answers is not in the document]
yottta: e7c684a0fbb6f68fd87bcaeec74e17c7978d76bb

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+  // ...
+  password_wo         = "new-password"
+  password_wo_version = 2
+  // ...
+}
+```
+
+As seen in this particular change of the [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework/commit/ecd80f67daed0b92b243ae59bb1ee2077f8077c7), the write-only attribute cannot be configured for set attributes, set nested attributes and set nested blocks.
+
+Write-only attributes cannot generate a plan diff because the prior state does not contain a value that OpenTofu can use to compare the new value against and also the planned value of a write-only argument will always be empty.
```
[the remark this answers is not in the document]
yottta: b4121ce87a6c902a783827039b5f38ddac6d7bc6

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+│   30:   secret_name = var.password
+│
+│ "secret_name" cannot accept an ephemeral value because it is not a write-only attribute which means that will be written to the state.
+╵
+```
+
+OpenTofu will not store ephemeral variable(s) in plan files. 
+If a plan is generated from a configuration that is having at least one ephemeral variable, 
+when the planfile will be applied, the value(s) for the ephemeral variable(s) needs to passed again by 
+using `-var` or `-var-file` arguments.
```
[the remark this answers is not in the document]
yottta: c47bbb084a80d34a729715b39653b14bc870f3e5

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Outputs
+An `output` block can be configured as ephemeral as long as it's
+not from the root module. 
+This limitation is natural since ephemeral outputs are meant to be skipped from the state file. Therefore, there is no usage of such a defined output block in a root module.
+
+Ephemeral outputs are useful when a child module returns sensitive data, allowing the caller to use the value of that output in other ephemeral contexts.
```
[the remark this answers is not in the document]
yottta: 14bc104ef981f19fe02640e00ee25e76ccab2070

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+OpenTofu will not store ephemeral variable(s) in plan files. 
+If a plan is generated from a configuration that is having at least one ephemeral variable, 
+when the planfile will be applied, the value(s) for the ephemeral variable(s) needs to passed again by 
+using `-var` or `-var-file` arguments.
+
+### Outputs
+An `output` block can be configured as ephemeral as long as it's
+not from the root module. 
+This limitation is natural since ephemeral outputs are meant to be skipped from the state file. Therefore, there is no usage of such a defined output block in a root module.
```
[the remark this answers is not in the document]
yottta: 603979cec2bf7e3c76064ef7515982f3c5d1c703

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+    when    = create
+    command = "echo non-ephemeral variable from root: #${jsonencode(var.secrets)}#"
+  }
+}
+```
+
+- (1) Variables will be able to be marked as ephemeral. By doing so, those will be able to be used only in ephemeral contexts.
+- (2) Version field that is going together with the actual write-only argument to be able to update the value of it. To upgrade the secret, the version field needs to be updated, otherwise OpenTofu will generate no diff for it.
+- (3) Using ephemeral resource to retrieve the secret. Maybe looks a little bit weird, because right above we are having the resource of the same type that is looking like it should be able to be used to get the secret. In reality, because that `resource` is using `secret_string_wo` to store the information, that field is going to be null when referenced. Check (9) for more details.
+- (4) Module output that is referencing an ephemeral value, it needs to be marked as ephemeral too. Otherwise, OpenTofu will generate an error.
```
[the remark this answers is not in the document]
yottta: 4d02f23b7bca34f689e30afa7290ed9cdd355c7d

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+- (14).  This is how a provisioner output should look like when an ephemeral value is used inside.
+- (15).  This will be visible in the action logs, while an ephemeral resource will be closed.
+
+## Technical Approach
+In this section, as in the "Proposed Solution" section, we'll go over each concept, but this time with a more technical focus.
+
+### Write-only arguments
+Most of the write-only arguments logic is already in the [provider-framework](https://github.com/hashicorp/terraform-plugin-framework):
+* [Initial implementation](https://github.com/hashicorp/terraform-plugin-framework/pull/1044)
+* [Sets comparisson enhancement](https://github.com/hashicorp/terraform-plugin-framework/pull/1064)
```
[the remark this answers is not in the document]
yottta: 11730c4361e2d3137791ceeb79a6a3cce5444d44

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+- (4) Module output that is referencing an ephemeral value, it needs to be marked as ephemeral too. Otherwise, OpenTofu will generate an error.
+- (5) The variable that is going to be used in an ephemeral variable, is not required to be ephemeral. The value can also be a hardcoded value without being ephemeral.
+- (6) Here we used another `aws_secretsmanager_secret` just to have an easier example on how ephemeral/write-only/(ephemeral variables)/(ephemeral outputs) are working together. But there will be more and more resources that will allow to work with this new concept.
+- (7) Referencing a module ephemeral output to ensure that the ephemeral information is passed correctly between two modules.
+- (8) Creating an output that is referencing a write-only argument *requires* the output to be marked as sensitive.
```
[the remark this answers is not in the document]
yottta: Changed here: 88a7826941e9048742cd732405ac2cf6f3417da6
I agree that would make more sense on having this also for the child modules, but do we want to create this diff between TF and OpenTofu?

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+For enabling ephemeral variables, these are the basic steps that need to be taken:
+* Update config to support the `ephemeral` attribute.
+* Mark the variables with a new mark ensure that the marks are propagated correctly.
+* Based on the marks, ensure that the variable cannot be used in other contexts than the ephemeral ones (see the User Documentation section for more details on where this is allowed).
+* Check the state of [#1998](https://github.com/opentofu/opentofu/pull/1998). If that is merged, in the changes where variables from plan are verified against the configuration ones, we also need to add a validation on the ephemerality of variables. If the variable is marked as ephemeral, then the plan value is allowed (expected) to be missing. 
+* Ensure that when the prompt is shown for an ephemeral variable there is an indication of that:
+  ```hcl
+  var.password (ephemeral)
+     Enter a value:
+  ````
```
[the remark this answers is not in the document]
yottta: Thanks. Fixed in 4d02f23b7bca34f689e30afa7290ed9cdd355c7d

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,646 @@
...
+
+> [!NOTE]
+>
+> Write-only attributes will be presented in the OpenTofu's UI as `(write-only attribute)` instead of the actual value.
+
+### Variables
+
+For enabling ephemeral variables, these are the basic steps that need to be taken:
+* Update config to support the `ephemeral` attribute.
+* Mark the variables with a new mark ensure that the marks are propagated correctly.
```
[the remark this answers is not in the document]
yottta: I am not aware of this. Could you expand it a little bit please?

rfc/20250317-ephemeral-resources.md
yottta: 3fe9247caedeef0099138111dece7a18065cda2b

rfc/20250317-ephemeral-resources.md
yottta: ce97918f1ad6429e17d88f7c250530742926f3f0

rfc/20250317-ephemeral-resources.md
yottta: Added information in 11a29cb050fd95ab55b5f6d5df01bba3c73af854.
A variable will not become ephemeral strictly from referencing an ephemeral value. In order for it to be able to work with ephemeral values, it needs to be configured specifically.

rfc/20250317-ephemeral-resources.md
yottta: I don't think that it's too late and I agree that it can become confusing. If others agree with this, it's an easy change:
* "%s/ephemeral resource/ephemeral entity/g"
* "%s/ephemeral resources/ephemeral entities/g"

Hitting enter? 😅

rfc/20250317-ephemeral-resources.md
yottta: da777bb65a5a2162f48c80bca72301d6412b937f

rfc/20250317-ephemeral-resources.md
yottta: Changed a little in 24b57e7c7b16816ae8e5d98b57658781656179c9.
But please check the specific chapter dedicated to the `Renew` [method](https://github.com/opentofu/opentofu/blob/6b81cb76c65cb76311db9ecf9d67a94fbb0185cc/rfc/20250317-ephemeral-resources.md#renew-method-details).

rfc/20250317-ephemeral-resources.md
yottta: Is the `Renew` [method] (https://github.com/opentofu/opentofu/blob/6b81cb76c65cb76311db9ecf9d67a94fbb0185cc/rfc/20250317-ephemeral-resources.md#renew-method-details) dedicated section enough?

rfc/20250317-ephemeral-resources.md
yottta: Added in https://github.com/opentofu/opentofu/commit/24b57e7c7b16816ae8e5d98b57658781656179c9#diff-1b793b731c0b3e0929632ddfa87429e5d0d2ddbe95bd3f9a09dea465ac1d406dR587-R597

Do you think that we should go more in-depth? Most probably the work of this should be done in `graphNodeCloseProvider`, but I think that this kind of details should be left for the implementor to decide.

rfc/20250317-ephemeral-resources.md
yottta: f603db886aa3d830534b5cc68f7e415e58e3d343

rfc/20250317-ephemeral-resources.md
yottta: f603db886aa3d830534b5cc68f7e415e58e3d343

rfc/20250317-ephemeral-resources.md
yottta: Yeah, I am having a rough time wording that correctly. "done using" in this context means that the apply phase finished and now it executes the close nodes. Right before executing `graphNodeCloseProvider` nodes, it should call `Close` on the ephemeral resources.
If you can help with a good wording around this, would greatly appreciate it.

rfc/20250317-ephemeral-resources.md
yottta: I didn't dig into this part yet and I barely touched the testing feature of this project.
If you think that this should be added, I can try my best but might take some time until I get a good understanding of the things in there.

rfc/20250317-ephemeral-resources.md
yottta: Because of the feature parity. Maybe having both just to be able to migrate? And in case we encounter `terraform.applying` write a warning that we also support `tofu.applying`? 🤔

rfc/20250317-ephemeral-resources.md
yottta: Fixed in 2c6587907eae9c404931ce789091dfd192f5628b

rfc/20250317-ephemeral-resources.md
yottta: That's a good point 🤔 I like it! Thanks
Added in 2c6587907eae9c404931ce789091dfd192f5628b

yottta:
@Yantrio, @cam72cam. Thanks a lot for the reviews.



> Overall, this is a really good start! I think the main area I'd like to see expanded is the user documentation. Your example is quite comprehensive and also quite dense. I think introducing a few simpler examples, before going with the "Here's the a complex example showing how the simpler building blocks interplay" would be helpful.

For this, I reworked the RFC and moved all the information that was initially directly under the `Proposed solution`, into the `User Documentation` and made `User documentation` as a subchapter of `Proposed solution`.

Also, moved all the suggested errors to be shown in the `User documentation` and `Technical approach` is just stating what it needs to be done.

rfc/20250317-ephemeral-resources.md
cam72cam: Internally, they both map to the same structure for now.  Let's at least mention tofu.applying here.

rfc/20250317-ephemeral-resources.md
cam72cam: Given the limited usage of the testing feature, I agree with @yottta here.  We should focus on the main impact of this feature.

rfc/20250317-ephemeral-resources.md
yottta: Done e8c189287f194d9ea8f2319658605ad22d618f45

rfc/20250317-ephemeral-resources.md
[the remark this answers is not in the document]
Yantrio: I think we should consider if it's useful or not. I don't have much of an opinion either way here really. I feel like root level only is fine.

rfc/20250317-ephemeral-resources.md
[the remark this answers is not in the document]
Yantrio: https://github.com/opentofu/opentofu/pull/2503

rfc/20250317-ephemeral-resources.md
Yantrio: I think I'm just looking for some reasoning to be added to the RFC. I'm fine with not doing it now, we should just document why.

rfc/20250317-ephemeral-resources.md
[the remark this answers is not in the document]
yottta: Follow up on this: After further investigation on this one, this was having **nothing** to do with write-only argument being referenced in an output. The provider field I was using, was marked as `sensitive` in the provider schema. Therefore, [this](https://github.com/opentofu/opentofu/blob/d1f0999aed2b171f62ae802c23b9674ca0afda6c/internal/tofu/node_output.go#L377-L389) is the code that was executed.
Still, the rule with only the root output still stands but it's in OpenTofu for a long time now.

rfc/20250317-ephemeral-resources.md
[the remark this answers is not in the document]
yottta: Oh, this is what you meant! Ok. I misread your initial comment.
Updated in 2695d6c58dc739ceae192242733ddea987cdf450

rfc/20250317-ephemeral-resources.md
yottta: Quick fix: dd4ca4729bfb730c406cd06ceaa52e2d6fcfea25

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+> In order to provide a similar and familiar UX for the users, the proposal in this RFC is heavily inspired by the Terraform public documentation and available blog posts on the matter.
+
+In the attempt to provide to the reader an in-depth understanding of the ephemerality implications in OpenTofu,
+this section will try to explain the functional approach of the new concept in each existing feature.
+
+### User Documentation
+#### Write-only attributes
+This is a new concept that allows any existing `resource` to define attributes in its schema that can be only written without the ability to retrieve the value afterwards.
+
+By not being readable, this also means that an attribute configured by a provider this way, should not be written to the state or plan file either.
```
Gogotchuri, line 44: Correctness @grammarly
```suggestion
By not being readable, this also means that an attribute configured by a provider this way should not be written to the state or plan file either.
```

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+Therefore, these attributes are suitable for configuring specific resources with sensitive data, like passwords, access keys, etc.
+
+A write-only attribute can accept an ephemeral or a non-ephemeral value, even though it's recommended to use ephemeral values for such attributes.
+
+Because these attributes are not written to the plan file, the update of a write-only attribute it's getting a little bit trickier.
```
Gogotchuri, line 49: Correctness And Clarity @grammarly
```suggestion
Because these attributes are not written to the plan file, updating a write-only attribute is getting a bit trickier.
```

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+* ephemeral variables
+* other ephemeral outputs
+* write-only attributes
+* ephemeral resources
+* `provisioner` block
+* `connection` block
+
+An `output` block from a root module cannot be marked as ephemeral.
+This limitation is natural since ephemeral outputs are meant to be skipped from the state file.
+Therefore, there is no usage of such a defined output block in a root module.
```
Gogotchuri, line 151: Correctness @grammarly
```suggestion
Therefore, there is no use for such a defined output block in a root module.
```

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+allowing the caller to use the value of that output in other ephemeral contexts.
+When using outputs in non-ephemeral contexts, OpenTofu should show an error similar to the following:
+```hcl
+│ Error: Invalid use of an ephemeral value
+│
+│   with aws_secretsmanager_secret_version.store_from_ephemeral_output,
+│   on main.tf line 31, in resource "aws_secretsmanager_secret_version" "store_from_ephemeral_output":
+│   31:   secret_string = module.secret_management.secrets
+│
+│ "secret_string" cannot accept an ephemeral value because it is not a write-only attribute which means that will be written to the state.
```
Gogotchuri, line 172: Correctness @grammarly
```suggestion
│ "secret_string" cannot accept an ephemeral value because it is not a write-only attribute, meaning it will be written to the state.
```

Gogotchuri (approved):
Looks Great! Thanks @yottta 
I've used Grammarly, which flagged a few grammar issues that might improve the document's readability; I'm adding them as suggestions.

rfc/20250317-ephemeral-resources.md
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+* locals
+* to configure `provider` blocks
+* in `provisioner` and `connection` blocks
+* in write-only arguments
+
+For example, you can have an ephemeral resource that is retrieving the password from a secret manager, password that can be passed later into a write-only attribute of another normal `resource`.
+To do so, the flow of an ephemeral resource should look similar to the following:
+* Requests the information from the provider.
+* Is passing it into the evaluation context that will be used to evaluate expressions that are referencing it.
+  * Is not storing the value in plan or state file.
```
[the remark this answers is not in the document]
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+* When accessing the value, OpenTofu will have to check for the presence of the `RenewAt`:
+  * If the timestamp is having a value and the current timestamp is at or over the timestamp indicated by `RenewAt`, call the provider `Renew` method.
+* When a reference is found to an ephemeral resource, OpenTofu should double check that the attribute referencing it is allowed to do so.
+  * See the contexts above where an ephemeral resource can be referenced.
+* At the end, the ephemeral resource needs to be closed. OpenTofu should call `Close` on the provider for all opened ephemeral resources.
+
+Beside the attributes in a schema of an ephemeral resource, the block should support also the meta-arguments existing in OpenTofu:
```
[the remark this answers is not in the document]
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+  * Metadata - returns the type name of the ephemeral resource.
+  * Schema - should be called on the provider to get the schema defined for the ephemeral resource.
+* Optional:
+  * Renew - if the response from `Open` contains a valid `RenewAt`, OpenTofu should call this method in order to instruct the provider to renew any possible remote information related to the secret returned from the `Open` call.
+  * Close - should be called on the provider to clean any possible remote information related to the secret returned in the response from `Open`.
+  * ValidateConfig - validates the configuration provided for the ephemeral resource.
+
+Ephemeral resources lifecycle is similar with the data blocks:
+* Both basic implementations require the same methods (`Metadata` and `Schema`) while datasource is defining `Read` compared with the ephemeral resource that is defining `Open`. When talking about the basic functionality of the ephemeral resources, the `Open` method should behave similarly with the `Read` on a datasource, where it asks the provider for the data associated with that particular ephemeral resource.
```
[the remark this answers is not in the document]
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+
+Which after running `tofu apply` should show an output like this:
+```hcl
+test = {
+  "ephemeral" = tostring(null)
+  "non-ephemeral" = "non-ephemeral-value"
+}
+```
+
+This function should work perfectly fine also with a non-ephemeral value.
```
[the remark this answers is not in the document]
yottta: Wanted to add the suggestions individually but the signoff of the commit is not working correctly from the GH UI, so I added all your suggestions in one commit: 4ee1ec72ed946b32d50637bb893c4c754d1b6e4e

rfc/20250317-ephemeral-resources.md
```diff
@@ -0,0 +1,859 @@
...
+}
+```
+
+The ephemeral outputs are available during plan and apply phase and can be accessed only in specific contexts:
+* ephemeral variables
+* other ephemeral outputs
+* write-only attributes
+* ephemeral resources
+* `provisioner` block
+* `connection` block
```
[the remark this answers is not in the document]
yottta: Added in 5c34b67c8d5e71d22fe4116af6bc9929ea2ee97d

PR #2799, opened 2025-05-13
Fix potential loss of local state

cam72cam:
In 6162cde6ffde8436ff82aa42b50f7fe7d44fb9e1, a bug was introduced where creating a new local statefile would ignore file creation errors.

## Target Release

1.10.0

## Checklist

### Go checklist

 


apparentlymart (approved):
This seems reasonable to me. I have to assume this was an authoring error when this code was originally written, since all of the other similar constructs elsewhere in the file already say `return err`.

I guess this is technically a user-facing change that could have a changelog update, but in practice I think it's pretty hard to hit because in normal use one of the many other methods would already have called `createStateFiles` and actually checked the error, so if there _is_ some way to end up in here with `s.stateFileOut` still set to nil then it's presumably an edge case. Therefore it might not be significant enough to be worth mentioning.

PR #2800, opened 2025-05-13
RFC creation - 20250514-locals-unused-detection

jgalais:
RFC creation for "20250514-variables-locals-unused-detection" link with [2583](https://github.com/opentofu/opentofu/issues/2583) issue.

### DIFF
--- a/rfc/20250514-locals-unused-detection.md
+++ b/rfc/20250514-locals-unused-detection.md
@@ -37,7 +37,7 @@ The validation process must perform static analysis of the configuration to dete
 
 Regarding [20241118-module-vars-and-outputs-deprecation RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20241118-module-vars-and-outputs-deprecation.md#silencing-deprecation-warnings-for-dependencies) we can use a dedicated flag "deprecation" to activate the detection feature to avoid weird interactions for users that don't really care about this.
 
-The "deprecation" Flag can be define the warning type that we want to have (module:all, module:local, module:none) according with the RFC.
+The "deprecation" flag can be define the warning type that we want to have (module:all, module:local, module:none) according with the RFC.
 
 ### Future Considerations
 
--- a/rfc/20250514-locals-unused-detection.md
+++ b/rfc/20250514-locals-unused-detection.md
@@ -37,7 +37,7 @@ The validation process must perform static analysis of the configuration to dete
 
 Regarding [20241118-module-vars-and-outputs-deprecation RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20241118-module-vars-and-outputs-deprecation.md#silencing-deprecation-warnings-for-dependencies) we can use a dedicated flag "deprecation" to activate the detection feature to avoid weird interactions for users that don't really care about this.
 
-The "deprecation" flag can be define the warning type that we want to have (module:all, module:local, module:none) according with the RFC.
+The "deprecation" flag can be define the warning scope that we want to have (module:all, module:local, module:none) according with the RFC.
 
 ### Future Considerations
 
--- a/rfc/20250514-locals-unused-detection.md
+++ b/rfc/20250514-locals-unused-detection.md
@@ -1,0 +1,53 @@
+# Locals unused detection
+
+Issue: [2583](https://github.com/opentofu/opentofu/issues/2583)
+
+## Summary
+
+This proposal introduces warnings during validation when a local value (`local`) is declared in a module but never referenced. The aim is to improve readability, maintainability, and code quality by identifying unused definitions.
+
+Currently, OpenTofu does not emit any warning when a local value is declared but unused. This can lead to configuration bloat, making codebases harder to read and maintain. By warning on unused values, developers are encouraged to clean up stale or redundant code, enhancing the overall quality of OpenTofu configurations.
+
+## Proposed Solution
+
+- **Local Values**:
+  If a value declared inside a `locals` block is not referenced elsewhere in the module, a warning is emitted during validation.
+
+### User Documentation
+
+```bash
+tofu validate
+
+│ Warning: The local "this_is_my_local" is unused.
+│
+│   on local.tf line 9:
+│    9:     this_is_my_local="local_value"
+
+│ Warning: The local "this_is_my_local" is unused.
+│
+│   on mod1/mod2/local.tf line 9:
+│    9:     this_is_my_local="local_value"
+
+Success! The configuration is valid.
+```
+
+### Technical Approach
+
+The validation process must perform static analysis of the configuration to determine whether each declared local value is actually used. This will require traversing the module's dependency graph and checking references.
+
+Regarding [20241118-module-vars-and-outputs-deprecation RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20241118-module-vars-and-outputs-deprecation.md#silencing-deprecation-warnings-for-dependencies) we can use a dedicated flag "deprecation" to activate the detection feature to avoid weird interactions for users that don't really care about this.
+
+The "deprecation" Flag can be define the warning type that we want to have (module:all, module:local, module:none) according with the RFC.
+
+### Future Considerations
+
+- **Unused Provider Configurations**:
+  These are out of scope for this RFC. Providers are propagated automatically and require a separate design consideration.[2537](https://github.com/opentofu/opentofu/issues/2537)
+
+## Potential Alternatives
+
+- **Emit errors instead of warnings**:
+  This was deemed too aggressive and could introduce friction for existing users.
+
+- **Custom scripts**:
+  Currently we must to create custom script to detect unused locals. I think it's not a good solution to use custom script in addition of OpenTofu.


rfc/20250514-variables-locals-unused-detection.md
```diff
@@ -0,0 +1,44 @@
...
+  During validation, if an input variable is declared in a module but never referenced, a warning is emitted—except when the variable is marked as `deprecated`. In that case, the warning is only emitted if a value is assigned to the variable by a caller.
+
+* **Local Values**:
+  If a value declared inside a `locals` block is not referenced elsewhere in the module, a warning is emitted during validation.
+
+### User Documentation
+
+```bash
+tofu validate
+[Warning]  variable xxx defined but not used inside code
```
Yantrio, line 23: If a variable is used deep within a nest of modules, how do you feel is best to help the end user running something like `tofu validate` find that variable in the code? Do we link to the file, do we try and construct an address for this variable?

jgalais: I think that we must to check the variable in a level where she must defined.

Idea:
- Check variable.tf in a root module (Check unused variable in this level)
- Check unused locals in a root module

- Check variable.tf in a sub-module (Check unused variable in the sub-module level)
- Check unused locals in a sub module

What do you think about that ?

cam72cam:
My main concern with this is that you could be using modules in a "hot-swappable" fashion.

Ex:
```
module "foo" {
   source = tofu.workspace == "prod" ? "./prod-foo" : "./dev-foo"
   only-used-in-prod = "some-value"
}
```

Both modules would have to have the same "interface", but would only use a subset of the passed variables each.  In this scenario, users would have to choose between an "unused" warning and a "deprecated" warning.

Detecting unused locals does not suffer from this as far as I know.

jgalais:
> @jgalais This is missing a DCO signoff, please fix your commit.

I didn't see the DCO check.
It's okay now I removed my gpg signature.

Thank you for your remark.

Regards,
Jgalais

jgalais:
> My main concern with this is that you could be using modules in a "hot-swappable" fashion.
> 
> Ex:
> 
> ```
> module "foo" {
>    source = tofu.workspace == "prod" ? "./prod-foo" : "./dev-foo"
>    only-used-in-prod = "some-value"
> }
> ```
> 
> Both modules would have to have the same "interface", but would only use a subset of the passed variables each. In this scenario, users would have to choose between an "unused" warning and a "deprecated" warning.
> 
> Detecting unused locals does not suffer from this as far as I know.

@cam72cam 
It's a good question but i suppose that if you done that you will have an error.

Tofu plan:
- In Production: It will be ok because this variable is defined
- In Dev: You will have an error because the variable is not defined inside the module
 
You must declare the variable inside your dev module with default value "null" for example to avoid this error.
At the end the variable will be defined but not used inside dev module.

 The message can be "Unused value <VALUE> inside module <module_name>/<module_path>".
 With this information you can easily know why you have this warning message.
 
 What do you think about that ?

cam72cam:
I think that it would be annoying from a user's perspective when they have intentionally unused variables.  Do you have any thoughts on @apparentlymart's comment from the issue: https://github.com/opentofu/opentofu/issues/2583#issuecomment-2737400669?

jgalais:
@cam72cam @apparentlymart ,

For the comment of @apparentlymart it's completely justify and i forgot to answer.
I'm really sorry about that.

I have two ideas:

- We can add flag example "--ignore-unused-variable" for example if the user doesn't want to see unused variables errors during opentofu execution.

- We accept that this feature can be a good new feature for the user and a great argument to use Opentofu compared to Terraform because this mechanism doesn't exist on Terraform.
It's like the dynamic backend feature present on Opentofu, I suppose that a lot of users use Opentofu because this feature is not present in Terraform.

For the last point, I have not measure about the number of module with unused variable but I suppose that in 99% of case the module owner will fix the module present on Terraform/Opentofu registry to drop unused variables.

I don't see a case where keep unused variables is needed.

Another good point is that if the Owner drop unused variables there are no incidence on Opentofu or Terraform runs.

jgalais:
@cam72cam @apparentlymart ,

Please can you review this RFC to implement this feature in OpenTofu ?

if I can help you on something don't hesitate.

Thank you for your work.

rfc/20250514-variables-locals-unused-detection.md
yottta: @jgalais I agree on that part. Would be interesting to understand what would be the most useful way to point the user to the variable/local in question..

For example, if we want to have it indicated in a filesystem way, we could have it something like this in a root module:
```hcl
│ Warning: The variable "this_is_my_variable" is unused.
│ 
│   on variables.tf line 9: // <- this is the thing in question
│    9:     variable "this_is_my_variable"
```
and in a sub module (2 levels deep), would be something like this:
```hcl
│ Warning: The variable "this_is_my_variable" is unused.
│ 
│   on mod1/mod2/variables.tf line 9: // <- this is the thing in question
│    9:     variable "this_is_my_variable"
```


Or if we want to have it referenced in an HCL way, it would look something like this:
```hcl
│ Warning: The variable "this_is_my_variable" is unused.
│ 
│   on <root> module line 9: // <- this is the thing in question. This case is debatable because we need to find a way to reference root module here. But this conversation more about the preference on what you think would be useful instead.
│    9:     variable "this_is_my_variable"
```
and in a sub module (2 levels deep), would be something like this:
```hcl
│ Warning: The variable "this_is_my_variable" is unused.
│ 
│   on module.call1.module.call2 line 9: // <- this is the thing in question
│    9:     variable "this_is_my_variable"
```

yottta:
Hi @jgalais, sorry for the slow answer.

This looks good. 

Though, I would suggest to take a look at the excelent RFC for the vars and outputs deprecation since in there I feel that are some points that we should take care of in this RFC too. For example, we should be clear in the intent of the modules that we want to scan for unused variables/locals. We should scan only the local modules and not the remote ones that the user have no access to update. There is some information about this in [this section](https://github.com/opentofu/opentofu/blob/541fc48a7bf5e7239ecf78979ce2dd9cd59b1173/rfc/20241118-module-vars-and-outputs-deprecation.md#silencing-deprecation-warnings-for-dependencies).

Another thing that I would like to throw out there, is that by reading these comments ([this](https://github.com/opentofu/opentofu/issues/2583#issuecomment-2737400669) and [this](https://github.com/opentofu/opentofu/pull/2800#issuecomment-2912175052)), it makes me strongly lean towards having this as an opt-in functionality of the `tofu validate` command. Otherwise, could create weird interactions for users that don't really care about this. Maybe would worth adding a section in the RFC to explore the pros and cons of such an idea 🤷

rfc/20250514-variables-locals-unused-detection.md
jgalais: I think that he first proposition is cool.
In a first time i will change the RFC to reduce the perimeter of "locals" only but this warning output.

I will add your proposition inside new RFC version.

jgalais:
@yottta I find that your comments are very interesting.

I added the "flag" mechanism inside technical approach with different values to define which warning scope you want to have according the [20241118-module-vars-and-outputs-deprecation RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20241118-module-vars-and-outputs-deprecation.md#silencing-deprecation-warnings-for-dependencies).

Your right it's better to activate feature to avoid noise when users don't want to use that.

rfc/20250514-locals-unused-detection.md
```diff
@@ -0,0 +1,53 @@
...
+│    9:     this_is_my_local="local_value"
+
+Success! The configuration is valid.
+```
+
+### Technical Approach
+
+The validation process must perform static analysis of the configuration to determine whether each declared local value is actually used. This will require traversing the module's dependency graph and checking references.
+
+Regarding [20241118-module-vars-and-outputs-deprecation RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20241118-module-vars-and-outputs-deprecation.md#silencing-deprecation-warnings-for-dependencies) we can use a dedicated flag "deprecation" to activate the detection feature to avoid weird interactions for users that don't really care about this.
```
yottta, line 38: I suggest on creating a separate section for this part under the `Technical Approach`, and don't tie the deprecation flag specifically to this new feature.

We should introduce a new flag, maybe called `-unused-values` and for the moment, it could get values like `locals` or `none`.
Obviously, I am interested in takes, yours and of maintainers, on what values we could have this flag configured with to allow evolution of it in the future if we will add other concepts to have their usage detected and reported.

diofeher: I agree with @yottta here. Since `deprecated` has a different meaning in our systems, I wouldn't like to tie to this feature, but I like the idea of having the argument similarly designed to it:

```
  -deprecation=module:m        Specify what type of warnings are shown. Accepted
                               values for "m": all, local, none. Default: all.
                               When "all" is selected, OpenTofu will show the
                               deprecation warnings for all modules. When "local"
                               is selected, the warns will be shown only for the
                               modules that are imported with a relative path.
                               When "none" is selected, all the deprecation
                               warnings will be dropped.
```

In my opinion it would make more sense if the API would be:

```
-unused=locals:all|none
```

So this could be extendable to other unused things.

diofeher:
Overall, the RFC is good, but having unused errors as optional (not the default!) would be valuable as well. I'm not sure if the other maintainers agree with that, but it would be helpful to have that in a CI pipeline, for example.

yottta:
@jgalais, I see no activity on this for a while. Would you have time to tackle the comments?

jgalais:
> @jgalais, I see no activity on this for a while. Would you have time to tackle the comments?

Hello @yottta,

I forgot to make this change.
I will make that this week to change RFC.

ilmax:
Will there also be an option to fail `tofu validate` when warnings are present? Looking at the current 1.20.0 I don't seem to find that option

yottta:
> Will there also be an option to fail `tofu validate` when warnings are present? Looking at the current 1.20.0 I don't seem to find that option

The whole deprecation feature is about warning about those, not returning errors.
For such an option, please check https://github.com/opentofu/opentofu/issues/757, this looks to be the place for this request.

> Looking at the current 1.20.0 I don't seem to find that option

You meant 1.12.0?

ilmax:
Yup, #757 is what I was looking for! Thanks

> You meant 1.12.0?

Indeed, that was a typo, apologies

yottta:
Now that #3999 has been merged and we started working on it, as suggested in [this comment](https://github.com/opentofu/opentofu/issues/2583#issuecomment-4781069672), we will include this functionality in the linting functionality which makes this PR and the associated issue to be absorbed into that.

Therefore, we will close this and will implement this functionality as a linting rule.

Thanks @jgalais for driving this. We greatly appreciate it! 🙌 
If you have other ideas on how to make OpenTofu better, we would be more than happy to hear back from you.

Issue #2834, opened 2025-05-21
RFC tracker: Ephemeral resources and write-only arguments


yottta:
This issue is an [RFC Tracker](https://github.com/opentofu/opentofu/blob/main/rfc/20241023-rfc-tracking-issues.md) of the [Ephemeral Resources RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20250317-ephemeral-resources.md) ([PR](https://github.com/opentofu/opentofu/pull/2793)) and it covers all the blockers and the development needed to deliver this feature as requested in #1996.

---

The implementation of the ephemeral resources, write-only arguments, ephemeral variables, ephemeral outputs and other integrations is an extensive work whose impact on the current code base is quite high so this tracker will be updated as we go with additional information besides the one that initially was provided.

### Blockers
- [x] RFC #2793
  - Add more details regarding the decision of storing minimal information of ephemeral resources in the plan file: #2996 
  - Rework the details about `terraform.applying` utility: #3155

### Feature branches
* All the branches below marked as merged can be found in #2852 and #3122.
  * These PRs and the associated branches will collect all the PRs until will be able to get merged into main. For that, OpenTofu 1.10-final will have to be released first. 
  * #2852 has been merged since it grew up quite heavily and became hard to maintain. One additional reason was to have it merged to have it tested by everybody to ensure that the work added will not affect any existing functionality of OpenTofu. 
    * Before merging, we had a review of the feature branch which resulted in #3077 where the suggestions have been applied and things refined. 
  * Meanwhile, the rest of the work will be gathered in a separate branch (ephemeral-feature-part2) that will have its own PR: #3122

### Implementation for ephemeral resources:
- [x] Update the provider plugin layer with the new methods required to work with ephemeral resources (#2828)
- [x] Update the base packages (addrs, lang, etc) with the necessary information to make those work with ephemeral resources
  - #2833
  - #2844
  - #2846
  - #2851 
  - #2855
  - #2866
  - #2878
- [x] Add the `ephemeral` block support into the configs package (#2840)
- [x] Implement ephemeral resources during plan phase (#2916)
- [x] Implement ephemeral resources during apply phase (#2985)
- [x] Ensure that ephemeral resources work correctly during destroy (#3032)
- [x] Mark the values returned by the ephemeral resources with a new `ephemeral` mark (#3049)
  - [x] Ensure that ephemeral resources can be referenced only in ephemeral contexts
[...]

PR #2852, opened 2025-05-26
Feature branch: Ephemeral resources

yottta:
Part of #2834 

This PR is only to have an overview of the changes already ready to be merged in `main`.

⚠️ Do not commit directly to `ephemeral-feature` branch.

This branch is intended to gather all the other PRs related to ephemeral resources.
By doing so, is wanted to provide a history of PRs with detailed descriptions for others to review when needed.

PRs included till now:
* #2846
* #2828 
* #2833
* #2840
* #2851
* #2844
* #2853 
* #2865
* #2855 (together with #2867)
* #2866
* #2877
* #2878
* #2876
* #2946
* #2916
* #2838
  * This has been merged just to enable easier testing of the ephemeral resources. The job on write-only arguments it's not done yet. 
* #2985
* #3032
* #3049
* #3077

## Checklist

### Go checklist

 


internal/plugin6/grpc_provider.go
```diff
@@ -748,6 +788,103 @@ func (p *GRPCProvider) ReadDataSource(ctx context.Context, r providers.ReadDataS
...
+
+	schema := p.GetProviderSchema(ctx)
+	if schema.Diagnostics.HasErrors() {
+		resp.Diagnostics = schema.Diagnostics
+		return resp
+	}
+
+	ephemeralResourceSchema, ok := schema.EphemeralResources[r.TypeName]
+	if !ok {
+		schema.Diagnostics = schema.Diagnostics.Append(fmt.Errorf("unknown ephemeral resource %q", r.TypeName))
```
cam72cam, line 802: Why not set resp.Diagnostics = ... instead of temporarily storing the result in schema.Diagnostics?

internal/plugin6/grpc_provider.go
```diff
@@ -748,6 +788,103 @@ func (p *GRPCProvider) ReadDataSource(ctx context.Context, r providers.ReadDataS
...
+	if protoResp.RenewAt != nil {
+		renewAt := protoResp.RenewAt.AsTime()
+		resp.RenewAt = &renewAt
+	}
+
+	return resp
+}
+
+func (p *GRPCProvider) RenewEphemeralResource(ctx context.Context, r providers.RenewEphemeralResourceRequest) (resp providers.RenewEphemeralResourceResponse) {
+	logger.Trace("GRPCProvider.v6: RenewEphemeralResource")
```
cam72cam, line 848: Should we bother checking the schema for the ephemeral resource type here and in Close? The provider should throw an error with bad input, but maybe we should be consistent?

internal/configs/config.go
```diff
@@ -460,6 +460,27 @@ func (c *Config) addProviderRequirements(reqs getproviders.Requirements, qualifs
 		})
 		reqs[fqn] = nil
 	}
+	for _, rc := range c.Module.EphemeralResources {
```
cam72cam, line 463: We might want to consider breaking this block into it's own function, am I reading it right that it's the same for resource/data/ephemeral?

internal/addrs/move_endpoint.go
```diff
@@ -299,3 +299,18 @@ func (e *MoveEndpoint) internalAddrType() TargetableAddrType {
 		panic(fmt.Sprintf("unsupported address type %T", addr))
 	}
 }
+
+// SubjectAllowed is validating what types of resource can be used with the moved block.
+// At the time of writing, it was only ensuring that the moved blocks cannot be used against ephemeral resources.
+// This can later be expanded with more rules
+func (e *MoveEndpoint) SubjectAllowed() bool {
```
cam72cam, line 306: This could definitely be more strict, afaik it should also not function on data blocks?

yottta: It is allowed 🤔 
Screenshot 2025-05-29 at 10 59 23

Am I wrong here?

internal/configs/module.go
```diff
@@ -404,6 +409,14 @@ func (m *Module) appendFile(file *File) hcl.Diagnostics {
 				Subject:  &o.DeclRange,
 			})
 		}
+		if o.Ephemeral && m.SourceDir == "." {
```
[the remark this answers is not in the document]
yottta: I spent too much on this one trying to find a better place to add it.
I feel quite confident with this approach after I dug deep into how config paths are handled in OpenTofu. Tested this with plain `tofu plan` and also with `tofu -chdir=<dir> plan` and it works the same.
I also seen a statement in [2820](https://github.com/opentofu/opentofu/pull/2820#discussion_r2096528388) that is supporting my checks:
Screenshot 2025-05-29 at 11 03 06

But if you know a better way of doing this, please let me know, I would gladly change it.

internal/plugin6/grpc_provider.go
yottta: Good point. Reworked this to have better diagnostics collection.
Changes are in fb86ede8275ce6edc116778b829cc9abd08cf6a4

internal/configs/config.go
yottta: Thanks a lot! Reworked in #2876 (bbf8c18c0e56baf7041c71dc2b96590bec399b79)

internal/configs/module.go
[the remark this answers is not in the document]
cam72cam: I think a better place for this is in the execution of the OutputNode during the validate phase.

internal/plugin6/grpc_provider.go
cam72cam: Sounds well reasoned to me :+1:

internal/addrs/move_endpoint.go
cam72cam: I think with ephemeral on the brain I got mixed up about what data is saved in a statefile

/facepalm

internal/configs/module.go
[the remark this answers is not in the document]
yottta: Fixed in https://github.com/opentofu/opentofu/pull/3006/commits/0cdcf1207075b0ae9fd691c8d3655ef78f767874. Will be merged together with that PR.

internal/command/e2etest/primary_test.go
```diff
@@ -232,3 +238,279 @@ func TestPrimaryChdirOption(t *testing.T) {
...
+			stateResources := state.RootModule().Resources
+			if len(stateResources) != 0 {
+				t.Errorf("wrong resources in state after destroy; want none, but still have:%s", spew.Sdump(stateResources))
+			}
+		}
+	}
+
+	// Run the test above for both supported plugin versions
+	t.Run("proto version 5", func(t *testing.T) {
+		pluginVersionRunner(t, "testdata/ephemeral-workflow/proto5", buildV5TestProvider)
```
cam72cam: Is there any reason that the configurations need to be different for proto5 vs proto6?

cam72cam:
As we are considering merging this into main, perhaps we should make a record of all of the TODOs and record them there?

internal/command/jsonplan/plan.go
```diff
@@ -868,6 +880,8 @@ func actionString(action string) []string {
 		return []string{"delete", "create"}
 	case "Forget":
 		return []string{"forget"}
+	case "Open":
+		return []string{"open"}
```
cam72cam, line 903: What do we want to do if Renew or Close makes it's way into here?

internal/plans/action.go
```diff
@@ -16,6 +16,9 @@ const (
 	CreateThenDelete Action = '±'
 	Delete           Action = '-'
 	Forget           Action = '.'
+	Open             Action = '⁐'
+	Renew            Action = '◌'
```
cam72cam: I spoke with @Yantrio earlier today and we discussed the hooks.  In practice, the only one of these "actions" used outside of Hooks is Open (which is well documented below).

Do we want to get rid of Renew and Close here and instead create Ephemeral specific hooks?  There is some precedent for that today with the non-resource hooks.

internal/command/state_mv.go
```diff
@@ -493,8 +493,19 @@ func (c *StateMvCommand) sourceObjectAddrs(state *states.State, matched addrs.Ta
 
 func (c *StateMvCommand) validateResourceMove(addrFrom, addrTo addrs.AbsResource) tfdiags.Diagnostics {
 	const msgInvalidRequest = "Invalid state move request"
-
 	var diags tfdiags.Diagnostics
+
+	if addrFrom.Resource.Mode == addrs.EphemeralResourceMode || addrTo.Resource.Mode == addrs.EphemeralResourceMode {
```
cam72cam, line 498: Can we even hit this case given the validation in internal/configs/moved.go, which uses the new SubjectAllowed function?

internal/command/state_mv.go
```diff
@@ -493,8 +493,19 @@ func (c *StateMvCommand) sourceObjectAddrs(state *states.State, matched addrs.Ta
...
+		diags = diags.Append(
+			tfdiags.Sourceless(
+				tfdiags.Error,
+				msgInvalidRequest,
+				"Ephemeral resources cannot be used as sources or targets for the move action. Just update your configuration accordingly.",
```
cam72cam, line 503: If we decide to keep this error message, "Just" is not a word we typically use in them.

internal/configs/config.go
```diff
@@ -419,47 +419,9 @@ func (c *Config) addProviderRequirements(reqs getproviders.Requirements, qualifs
...
-		qualifs.AddImplicitProvider(fqn, getproviders.ResourceRef{
-			CfgRes:            rc.Addr().InModule(c.Path),
-			Ref:               tfdiags.SourceRangeFromHCL(rc.DeclRange),
-			ProviderAttribute: rc.ProviderConfigRef != nil,
-		})
-		reqs[fqn] = nil
-	}
+	c.collectImplicitProviders(c.Module.ManagedResources, reqs, qualifs)
+	c.collectImplicitProviders(c.Module.DataResources, reqs, qualifs)
+	c.collectImplicitProviders(c.Module.EphemeralResources, reqs, qualifs)
```
cam72cam, line 424: Thanks for cleaning this up :+1:

internal/configs/configschema/path.go
```diff
@@ -58,3 +58,20 @@ func (o *Object) AttributeByPath(path cty.Path) *Attribute {
 	}
 	return nil
 }
+
+// PathName is trying to get the name of the attribute referenced by the cty.Path.
+// Due to the implementation difference between cty.IndexStep and cty.GetAttrStep,
+// this method is trying to find the most right-sided cty.GetAttrStep and returns the
+// name of that.
+// If the path contains no cty.GetAttrStep, it returns <unknown>.
+// TODO ephemeral - is there a better way to do this?
```
cam72cam: Only other option I can think of is a recursive call, which does not improve over what you have written here.

cam72cam: Is there any reason in lang/eval.go that we can't word the error message to format the entire cty.Path as a string?  I'm sure we already have functions for that.

internal/configs/resource.go
```diff
@@ -547,6 +547,176 @@ func decodeDataBlock(block *hcl.Block, override, nested bool) (*Resource, hcl.Di
 	return r, diags
 }
 
+func decodeEphemeralBlock(block *hcl.Block, override bool) (*Resource, hcl.Diagnostics) {
```
cam72cam, line 550: At first glance, it would seem not that hard to merge this and the data function above into decodeDataOrEphemeralBlock.  A lot of this is probably shared with resource as well.

Maybe we put a pin in this comment and come back to it later as a cleanup PR.

internal/lang/eval.go
```diff
@@ -619,3 +626,54 @@ func normalizeRefValue(val cty.Value, diags tfdiags.Diagnostics) (cty.Value, tfd
...
+// The schema argument is used to figure out if the value is for an ephemeral
+// context. If it is, then we don't even validate ephemeral marks.
+// If val contains any ephemeral mark, we check if the attribute containing
+// an ephemeral value is a write-only one. If not, we generate a diagnostic.
+// The diagnostics returned by this method need to go through InConfigBody
+// by the caller of the evaluator to append additional context to the diagnostic
+// for an enhanced feedback to the user.
+func validEphemeralReferences(schema *configschema.Block, val cty.Value) (diags tfdiags.Diagnostics) {
+	// Ephemeral resources can reference values with any mark, so ignore this validation for ephemeral blocks
+	if schema == nil || schema.Ephemeral {
```
cam72cam: I'm curious about this nil check here.  If the schema is nil, shouldn't that imply that we don't know if ephemeral values can be used in here or not?

internal/legacy/helper/schema/schema.go
```diff
@@ -252,6 +252,8 @@ type Schema struct {
 	// secret fields. Future versions of OpenTofu may encrypt these
 	// values.
 	Sensitive bool
+
+	WriteOnly bool
```
cam72cam: Given that this is in the legacy package, do we need to add WriteOnly here?

internal/plans/changes.go
```diff
@@ -37,8 +37,24 @@ func NewChanges() *Changes {
...
+	return c
+}
+
 func (c *Changes) Empty() bool {
 	for _, res := range c.Resources {
+		// We ignore Open actions which are specific to ephemeral resources.
+		// A configuration containing ephemeral resources will always have changes planned,
+		// but if there is no other change recorded, there is no need for a prompt
+		// on the user.
+		if res.Action == Open {
```
cam72cam, line 55: We don't handle Refresh or Close, see above comment about if those should be considered "plan actions"

internal/tofu/evaluate.go
```diff
@@ -743,6 +750,18 @@ func (d *evaluationStateData) GetResource(ctx context.Context, addr addrs.Resour
 			// We should only end up here during the validate walk,
 			// since later walks should have at least partial states populated
 			// for all resources in the configuration.
+			if schema.Ephemeral {
+				// If the block that it's evaluated is an ephemeral one, we want to mark
+				// the cty.DynamicVal as ephemeral to ensure that the ephemeral references
+				// check is working properly during walkValidate.
+				// TODO ephemeral - for the reviewers: do you see any unwanted consequences in
```
cam72cam: I think this is reasonable and don't see an issue with it.

I do wonder if we should also handle sensitive here?

cam72cam: We could potentially re-use schema.ValueMarks

internal/tofu/node_resource_closeable.go
```diff
@@ -0,0 +1,43 @@
...
+var (
+	_ GraphNodeCloseableResource = (*nodeCloseableResource)(nil)
+)
+
+func (n *nodeCloseableResource) Name() string {
+	return n.Addr.String() + " (close)"
+}
+
+func (n *nodeCloseableResource) Execute(_ context.Context, _ EvalContext, _ walkOperation) (diags tfdiags.Diagnostics) {
+	for _, cb := range n.cbs {
```
cam72cam, line 42: Should we consider doing this in parallel?  If each close operation takes multiple seconds (worst case), this could balloon out pretty quickly.

internal/tofu/node_resource_plan.go
```diff
@@ -438,3 +450,15 @@ func (n *nodeExpandPlannableResource) resourceInstanceSubgraph(ctx context.Conte
...
 }
+
+// Close implements closableResource
+func (n *nodeExpandPlannableResource) Close() (diags tfdiags.Diagnostics) {
+	if n.Addr.Resource.Mode != addrs.EphemeralResourceMode {
+		return diags
+	}
+
+	for _, c := range n.closers {
+		diags = diags.Append(c())
```
cam72cam: Should we consider doing this in parallel? If each close operation takes multiple seconds (worst case), this could balloon out pretty quickly.

internal/addrs/parse_target.go
```diff
@@ -135,9 +139,14 @@ func parseResourceUnderModule(moduleAddr Module, remain hcl.Traversal) (ConfigRe
...
 	mode := ManagedResourceMode
-	if remain.RootName() == "data" {
+
+	switch remain.RootName() {
+	case "data":
 		mode = DataResourceMode
 		remain = remain[1:]
+	case "ephemeral":
+		mode = EphemeralResourceMode
+		remain = remain[1:]
```
cam72cam, line 149: Github won't let me comment below, but I think there's an ephemeral case missing from parseResourceTypeAndName down below.

internal/configs/configschema/marks.go
```diff
@@ -27,6 +27,16 @@ func copyAndExtendPath(path cty.Path, nextSteps ...cty.PathStep) cty.Path {
 func (b *Block) ValueMarks(val cty.Value, path cty.Path) []cty.PathValueMarks {
 	var pvm []cty.PathValueMarks
 
+	// When the block is marked as ephemeral, the whole value needs to be marked accordingly.
```
cam72cam, line 30: I think we should look closely at where ValueMarks is used.  In most cases, it relies on ContainsSensitive() to opt-in to that functionality.

I think that the easy solution is to change ContainsSensitive to HasAdditionalMarks (or similar).

There's also the matter of removing nested ephemeral marks when present.  Not sure we want to include that functionality in the same concern.

internal/command/e2etest/primary_test.go
yottta: The reason was that we built the providers with different names. The HCL configuration just had a different `required_providers` content.
To be able to keep just one configuration, I changed a little bit the function that builds the provider binary to have it renamed the same for both v5 and v6 binaries.

Fixed in c9129955b55eec4d9a372bfee782ba1ef6e8cafd

yottta:
> As we are considering merging this into main, perhaps we should make a record of all of the TODOs and record them there?

Not sure what do you mean by this. There was one TODO that was off, but fixed it in f73d3613b211b71703fe81bcf2a58744ec0a6796.
All the TODOs related to ephemeral work start with `TODO ephemeral`, so we can identify those later.
Is there something else more that you suggest on doing with these?

internal/command/jsonplan/plan.go
yottta: I didn't add those, on purpose. Those two are never meant to reach this point. The only one that is meant to reach this point is `Open`.
For the other 2, we can either have a comment or add a panic. What do you think?

cam72cam:
I missed that it was only one that was not correctly marked, that grep sounds good to me @yottta

internal/plans/action.go
yottta: Removed the Renew and Close action and added specific hook methods for those in fa44c594c8b8122a1f78f48b598e451c315c5c01

internal/command/state_mv.go
yottta: You're right!
The config is always loaded in the `state mv` cmd. Therefore, this cannot be reached.
Cleaned up in 08bb3a9a25a96c93bfc368cc3321cbc7801d4ca9

internal/command/state_mv.go
yottta: Removed the message in 08bb3a9a25a96c93bfc368cc3321cbc7801d4ca9

internal/configs/configschema/path.go
yottta: Great! Thanks for the additional "push". Just a search for `cty.Path) string` returned exactly what you were talking about. Lazy me 🤦 

Removed this function in 6ea41a7752ca668c5cd5302e854ae7b7703cf871 and replace its usage with an existing one.

internal/configs/resource.go
yottta: Yes please! 😅 
I wanted to do so, but there are a lot of overlapping concepts and some apply in some cases and some in others.
I didn't want to do this in a rush. I will take a note and will do a refactor after we are done with the ephemeral resources.
If you want to do this differently, please let me know.

internal/lang/eval.go
yottta: Yeah, thinking more thorough about this, kind of made me realise that my initial assumption was wrong: if there is a block to be evaluated, will have a schema and if that is not marked as ephemeral, to the validation. But if there is no schema, that information will not be saved/printed either way.

From my checks, **most** calls to the `EvalBlock` (the one calling `validEphemeralReferences`), is done from places where if schema is missing an error is returned right before calling this.
But there is one clear place where things can go south:
https://github.com/opentofu/opentofu/blob/3c76c5f419f1aac9f1f30b177fa4cb42f6118eca/internal/tofu/node_provider.go#L156-L168

And seeing that, I wonder on how we can handle the testing support part of the ephemeral resources when it comes to mocking schemas and other blocks 🤔 . As I am not familiar with that part, I will just defer the "fun" for later.

But coming back to your question, I would have another question: what is stored in plan and state files, isn't required to have a schema? Except outputs and variables (for plan). Any other value go one way or another through some kind of schema validation, right? I would check, but only if I get a non-conclusive answer since that's a pretty deep rabbit hole... If so, even though a (possible) schema-less block at some point has been validated as "valid to reference ephemeral" (because of that nil check) and then later referenced in another resource with a schema (not marked as ephemeral), that will fail.

Great question, really good one!

internal/legacy/helper/schema/schema.go
yottta: Definitely not needed. Removed in dd64b8c13c3e5ca5ae1ddfd4b3abea9cb864417f

internal/plans/changes.go
yottta: That's not needed since those actions were not meant to be used in the resource generated changes.
Even more, after fa44c594c8b8122a1f78f48b598e451c315c5c01, a change with Refresh/Close action is not even possible now.

internal/tofu/evaluate.go
yottta: > I think this is reasonable and don't see an issue with it.
> 
> I do wonder if we should also handle sensitive here?

No, we cannot actually do it. The difference between ephemeral and sensitive is that the ephemeral is applied at the root level of a value while sensitive is per attribute of the value.

> We could potentially re-use schema.ValueMarks

Even by using this, the returned slice of `cty.PathValueMarks` will not be "applicable" to the `cty.DynamicVal` since it has no inner attributes to mark at this point.
I would stick with the current implementation instead of using `schema.ValueMarks` since the call the the latter  could be less performant for schema with multiple layers of nested objects.

yottta: Since you already answered to that, I removed that TODO and updated the comment b540b653e53e8d4065f4543addf5fc0c8da4a191.
Please let me know if you have any feedback on the new commit.

internal/tofu/node_resource_plan.go
yottta: Yes, was on my TODOs. Let's handle it now! Done in 5e2af23e07dac405b4229ed68d7fb05973256964.
Please let me know if you would like to do this differently.
Before:
image
After:
image

Plan phase:
```
2025-07-28T17:09:11.193+0300 [TRACE] nodeCloseableResource - scheduling 1 closing operations for of ephemeral resource ephemeral.playground_secret.test
2025-07-28T17:09:11.193+0300 [TRACE] nodeExpandPlannableResource - scheduling 5 closing operations for of ephemeral resource ephemeral.playground_secret.test
```
Apply phase:
```
2025-07-28T17:09:12.282+0300 [TRACE] nodeCloseableResource - scheduling 5 closing operations for of ephemeral resource ephemeral.playground_secret.test
```

internal/addrs/parse_target.go
yottta: Added in 8a7e4a566e9c80cc0ff0c942a044cea8139c87fb. Good catch!

internal/configs/configschema/marks.go
yottta: Really good point!
In d2b7575b9e645fde34363cc4c87d16bc4d899da6 went with your suggestion of having another wrapper on `ContainsSensitive` and replaced its usage with the new one. The only places where `ContainsSensitive` is used in the `configschema` package for internal handling.
By replacing its usage with the new one, especially in the `jsonplan` and `jsonstate` packages and adding a new validation in those packages to ensure that we don't leak ephemeral marked values into plan and/or state, started highlighting other issues: there were indeed ephemeral marks reaching this point without being handled properly:
* Ephemeral resource `schema.ValueMarks(...)`, with the new additions from several commits ago, now gets its ephemeral mark on the root path of the value, but that has no impact so we can ignore the ephemeral marks check for this type of resources.
* One really concerning thing appeared after fixing the above one, where actual managed resources had this reported. My initial assumption was right, but incomplete: attributes of managed resources are not meant to reference ephemeral values and that validation has been implemented in a [previous PR](https://github.com/opentofu/opentofu/pull/3049/commits/3c0277dab9dd304dcfd278e2870ff22a64568516#diff-b54de9c9ff84c7be2610d760a124ca7424c0563448a2f766fc3fa7b5e08ad8e7R639). The thing that I missed was around write-only attributes, which is a new concept that is incompletely introduced. These attributes **can** reference an ephemeral value and when doing so will inherit the mark during evaluation.
  * While the terraform-plugin-go is meant to nullify these attributes before return ([check docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments#general-concepts)), there is no need to carry the ephemeral mark.
  * Therefore, in f85d1c6db6041abe4400261e0d64b35c1928f9e7 started removing the ephemeral mark from values the are configured as write-only in the schema.
  * After doing this, another issue bubbled up. This is due to the incomplete work on write-only arguments. More details can be found in the commit message of ec1d72428984d26f838707df3c05ab02f6e1d20d.

internal/tofu/transform_destroy_edge.go
```diff
@@ -309,6 +315,7 @@ func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) err
 					// root module outputs indicate they are not temporary by
 					// returning false here.
 					if !n.temporaryValue() {
+						log.Printf("[TRACE] pruneUnusedNodes: temporary value vertex %q kept because it's not a temporary value vertex", dag.VertexName(n))
```
Gogotchuri, line 318: I think we need to improve this error message. This is simply stating a contradiction.

Gogotchuri (commented):
Fantastic work, Andrei!
Everything looks great, and Christian has already outlined most of the issues I've found.
In ~8k lines of changes, I have found one nitpick 😅

One thing that I am not quite sure about is the interactions between the `import` block and `ephemeral` resources. What happens if we try to target an ephemeral resource with an import block?

internal/tofu/transform_destroy_edge.go
yottta: I added those just to have a trace log for each possible branch. TBH, I didn't put much effort into the message.
If you have any suggestion, please do, would be appreciated for sure 🙌

yottta:
> Fantastic work, Andrei! Everything looks great, and Christian has already outlined most of the issues I've found. In ~8k lines of changes, I have found one nitpick 😅
> 
> One thing that I am not quite sure about is the interactions between the `import` block and `ephemeral` resources. What happens if we try to target an ephemeral resource with an import block?

Thanks a lot Ilia 🙌 

### About the `import` block
https://github.com/opentofu/opentofu/blob/be433be96f0fc82ea932b0b9ff378f08f813babb/internal/tofu/context_plan.go#L751-L763
This is the func that generates the error in case a **managed resource** is not found in the configuration for the address given in the `import.to` field.
And the actual place where that is called is here:
https://github.com/opentofu/opentofu/blob/be433be96f0fc82ea932b0b9ff378f08f813babb/internal/tofu/context_plan.go#L678-L683

### Import command
This is validated in the `Run` method of the command. Updated the message to be more clear when the addr targets an ephemeral resource. Done in f233ed33b90e61c0891c5d0c5974d15136943e7b

TLDR: An import block will fail if the address is not a **managed resource** in the configuration of the root module.

internal/tofu/transform_destroy_edge.go
Gogotchuri: Yeah, this path will likely never be displayed to the user. You are correct. And I have no idea what a better error message would be—resolving this.

internal/lang/eval.go
cam72cam: I think the safer default is that if there is no schema, there can not be any ephemeral values.

For the code that you linked, no attributes or values can be set there anyways so it's a moot point?

yottta: > For the code that you linked, no attributes or values can be set there anyways so it's a moot point?

Mostly, yes. I am more worried about how the testing support will suffer from this 🤔 

> I think the safer default is that if there is no schema, there can not be any ephemeral values.

Done in 6894f75efd40694e3206652fda7f5c6b8c5d6492

Yantrio (approved):
I'm at a point where I'm happy to merge this as long as we keep track of all of the TODO items and any missing test cases. Its very easy to lose track of those and I would like to avoid littering the codebase with long-lived todo comments if possible.

yottta:
> I'm at a point where I'm happy to merge this as long as we keep track of all of the TODO items and any missing test cases. Its very easy to lose track of those and I would like to avoid littering the codebase with long-lived todo comments if possible.

Totally agree. Every time I encountered a place where ephemeral concept might need a check and/or implementation I added a `TODO ephemeral - ` to not loose the track of it.
Those are meant to be removed one by one once the work has been added to resolve it. I plan to keep adding and remove these until I find no other places to add any. At that moment, I will consider the work done for the current RFC.

internal/configs/resource.go
yottta: Revisited today this and right now this is not feasible anymore.
The [`enabled` attribute](https://github.com/opentofu/opentofu/issues/3247) added a significant amount of differences between the data and ephemeral blocks decoding and refactoring this would add additional complexity and reduce readability.

Will leave those methods as they are and not act on the note that we took back in July.

PR #2903, opened 2025-06-10
Cleanup github workflows

### LINKED ISSUE #2480: chore: revisit the usage of hashicorp/X github actions
We are using a few github actions, which we inherited from the upstream:
* `hashicorp/actions-set-product-version`
* `hashicorp/actions-generate-metadata`
* `hashicorp/actions-docker-build`

Currently, those are the only actions we use with no version tags due to unreleased dependency upgrades.

We should re-evaluate the purpose of those in our workflows and potentially migrate off.

cam72cam:
Summary:
* Removed unneded actions/cache calls (setup-go handles this now)
* Removed non-functional backport tool
* Trimmed build.yml to only binary build and direct e2e tests

checks.yml run: https://github.com/opentofu/opentofu/actions/runs/15570389473?pr=2903
build.yml run: https://github.com/opentofu/opentofu/actions/runs/15572538948

Resolves #2480

## Checklist


apparentlymart (approved):
This looks good to me!

The "setup-go" action's ability to take the toolchain version from `go.mod` seems like it could eventually remove the redundant `.go-version` file which comes from before `go.mod` could represent that, but I think we should do that separately later because folks working in this repository might be relying on version managers that don't understand that [the Go toolchain is now able to version-manage itself](https://go.dev/blog/toolchain), so I expect we'll want to communicate that more loudly and give folks some time to adapt to it.

cam72cam:
~Still need to finish fixing the windows e2e tests before this can be merged~ Done

yottta (approved):

PR #2905, opened 2025-06-11
Allow configuration aliases in root module to impact validation only

### LINKED ISSUE #2862: Support validating modules that use configuration_aliases
### Community note

> [!TIP]

### OpenTofu Version

```shell
v1.9.1
```

I have modules that use configuration_aliases to accept multiple providers  of a given type. For example, because it needs to create resources in multiple regions in AWS. 

Running `tofu validate` for such a module will result in "Provider configuration not present" errors, which makes CI for these modules more difficult, as well as local validation.

### Attempted Solutions

There are some workarounds mentioned below, but they aren't obvious, and add unnecessary complexity to both manual and automated workflows.

### Proposal

`tofu validate` should work for modules that use `configuration_aliases`. 

The way I would expect this to work is that before running the validate, it creates a default provider instance for each alias. 

Or alternatively, creates a single provider, as would be done if there weren't any `configuration_aliases` configuration, and set all the aliases to point to that provider.  So that, for example, if there were aws aliases defined as `configuration_aliases = [aws.a, aws.b]`, the module would be validated as if it had been instantiated with something like:

```terraform
module "some_name" {
  # ...
  providers = {
    aws.a = aws
    aws.b = aws
  }
}
```

I know of two potential workarounds for this:

1. Create a temporary .tf file that defines dummy providers while running tofu validate. However, it is critical that this file is not checked in to git, or included in published versions of the module.
2. Create a separate test module, that instantiates the module to be validated, and defines the necessary providers, and run tofu validate on that project instead. This adds boilerplate, and while less critical, ideally this test module would not be included in published versions of the module either.

Both workarounds require developers to be aware of the problem, and the workarounds, and adds complexity to CI flows.

### References

Related issue from terraform repo: https://github.com/hashicorp/terraform/pull/37172

cam72cam:
Resolves #2862

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ENHANCEMENTS:
 
 * `tofu show` now supports a `-config` option, to be used in conjunction with `-json` to produce a machine-readable summary of the configuration without first creating a plan. ([#2820](https://github.com/opentofu/opentofu/pull/2820))
+* `tofu validate` now supports running in a module that contains provider configuration_aliases. ([#2905](https://github.com/opentofu/opentofu/pull/2905))
 
 ## Previous Releases
 
--- a/internal/tofu/context_validate_test.go
+++ b/internal/tofu/context_validate_test.go
@@ -2506,3 +2506,73 @@ func TestContext2Validate_rangeOverZeroPlanTimestamp(t *testing.T) {
 		t.Fatal(diags.ErrWithWarnings())
 	}
 }
+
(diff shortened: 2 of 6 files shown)


apparentlymart (approved):
It's a shame that the provider-related logic is so sprawled around `package tofu` that we couldn't address this only with a surgical change to a specific validation function, but I agree that this seems like a reasonable way to get this done with the system we have and then we can think a little more about how we might handle this situation more directly in future work.

PR #2916, opened 2025-06-13
Ephemeral during plan

yottta:
WIP

Resolves # 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/node_resource_abstract_instance.go
+++ b/internal/tofu/node_resource_abstract_instance.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
@@ -80,6 +81,19 @@ type NodeAbstractResourceInstance struct {
 	generatedConfigHCL string
 
 	ResolvedProviderKey addrs.InstanceKey
+
+	// These channels are meant to be used strictly in case this struct is representing an ephemeral resource.
+	// These channels are not initialized until the renewal of the ephemeral resource is scheduled.
+	//
+	// closeCh is the channel to signal the ephemeral resource renewal goroutine to shut down and return the data it's having.
+	// This is closed when the NodeAbstractResourceInstance.Close is called.
+	// ephemeralDiags is used by the renewal goroutine to return whatever issues encountered during the process.
+	// This is the channel that NodeAbstractResourceInstance.Close is blocking on, so be sure that when closeCh is called,
+	// something is written into ephemeralDiags.
+	// The same channel is also used by the NodeAbstractResourceInstance.closeEphemeralResource to add the diagnostics
+	// that it encountered, if any.
+	closeCh        chan struct{}
+	ephemeralDiags chan tfdiags.Diagnostics
 }
 
 // NewNodeAbstractResourceInstance creates an abstract resource instance graph
@@ -1795,7 +1809,7 @@ func (n *NodeAbstractResourceInstance) providerMetas(ctx context.Context, evalCt
 	return metaConfigVal, diags
 }
 
-func (n *NodeAbstractResourceInstance) openEphemeralResource(ctx context.Context, evalCtx EvalContext, configVal cty.Value) (cty.Value, tfdiags.Diagnostics) {
+func (n *NodeAbstractResourceInstance) openEphemeralResource(ctx context.Context, evalCtx EvalContext, configVal cty.Value) (cty.Value, []byte, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 	var newVal cty.Value
 
@@ -1804,13 +1818,13 @@ func (n *NodeAbstractResourceInstance) openEphemeralResource(ctx context.Context
 	provider, providerSchema, err := n.getProvider(ctx, evalCtx)
 	diags = diags.Append(err)
 	if diags.HasErrors() {
-		return newVal, diags
+		return newVal, nil, diags
 	}
 	schema, _ := providerSchema.SchemaForResourceAddr(n.Addr.ContainingResource().Resource)
 	if schema == nil {
 		// Should be caught during validation, so we don't bother with a pretty error here
 		diags = diags.Append(fmt.Errorf("provider %q does not support ephemeral resource %q", n.ResolvedProvider.ProviderConfig, n.Addr.ContainingResource().Resource.Type))
-		return newVal, diags
+		return newVal, nil, diags
 	}
 
 	// Unmark before sending to provider, will re-mark before returning
@@ -1827,7 +1841,7 @@ func (n *NodeAbstractResourceInstance) openEphemeralResource(ctx context.Context
 	)
 	diags = diags.Append(validateResp.Diagnostics.InConfigBody(config.Config, n.Addr.String()))
 	if diags.HasErrors() {
-		return newVal, diags
+		return newVal, nil, diags
 	}
 
 	// If we get down here then our configuration is complete and we're ready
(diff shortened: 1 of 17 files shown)


internal/tofu/node_resource_abstract_instance.go
```diff
@@ -1796,6 +1821,127 @@ func (n *NodeAbstractResourceInstance) providerMetas(ctx context.Context, evalCt
...
+	if diags.HasErrors() {
+		return newVal, providers.DeferredReasonUnknown, diags
+	}
+
+	if resp.Deferred != nil {
+		return newVal, resp.Deferred.DeferralReason, diags
+	}
+	newVal = resp.Result
+	if newVal == cty.NilVal {
+		// This can happen with incompletely-configured mocks. We'll allow it
```
cam72cam: I don't love adding code that's designed to work around test setup inconsistencies

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -1796,6 +1821,127 @@ func (n *NodeAbstractResourceInstance) providerMetas(ctx context.Context, evalCt
...
+				n.ResolvedProvider.ProviderConfig.InstanceString(n.ResolvedProviderKey), tfdiags.FormatErrorPrefixed(err, n.Addr.String()),
+			),
+		))
+	}
+	if diags.HasErrors() {
+		return newVal, providers.DeferredReasonUnknown, diags
+	}
+
+	if newVal.IsNull() {
+		diags = diags.Append(tfdiags.Sourceless(
```
cam72cam, line 1900: Any reason not to return diags here?

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -2885,3 +3039,241 @@ func maybeImproveResourceInstanceDiagnostics(diags tfdiags.Diagnostics, excludeA
...
+			// the goroutine that is handling the renewal of the ephemeral resource.
+		}
+	}
+
+	return plannedChange, plannedNewState, keyData, diags
+}
+
+func (n *NodeAbstractResourceInstance) startEphemeralRenew(ctx context.Context, evalContext EvalContext, provider providers.Interface, renewAt *time.Time, privateData []byte) {
+	if n.Addr.Resource.Resource.Mode != addrs.EphemeralResourceMode {
+		return
```
cam72cam: panic?

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -2885,3 +3039,241 @@ func maybeImproveResourceInstanceDiagnostics(diags tfdiags.Diagnostics, excludeA
...
+		diags = diags.Append(evalContext.Hook(func(h Hook) (HookAction, error) {
+			// We are using cty.EmptyObject here because the prior state and the new planned state does not matter
+			// in ephemeral resources context, especially in the context of the renew operation.
+			return h.PreApply(n.Addr, states.CurrentGen, plans.Renew, cty.NullVal(cty.EmptyObject), cty.NullVal(cty.EmptyObject))
+		}))
+		req := providers.RenewEphemeralResourceRequest{
+			TypeName: n.Addr.Resource.Resource.Type,
+			Private:  privateData,
+		}
+		resp := provider.RenewEphemeralResource(ctx, req)
```
cam72cam, line 3205: Are we not tracking the resp.Diagnostics within the diags returned?

internal/tofu/node_resource_plan.go
```diff
@@ -438,3 +450,15 @@ func (n *nodeExpandPlannableResource) resourceInstanceSubgraph(ctx context.Conte
...
 }
+
+// Close implements closableResource
+func (n *nodeExpandPlannableResource) Close() (diags tfdiags.Diagnostics) {
+	if n.Addr.Resource.Mode != addrs.EphemeralResourceMode {
+		return diags
+	}
+
+	for _, c := range n.closers {
+		diags = diags.Append(c())
```
cam72cam, line 461: This is way out of scope and more of a refinement, but as each of these calls will be waiting on a chan/timeout, this could take up to 10 seconds * (count or len(for_each)).

internal/tofu/transform_reference.go
```diff
@@ -218,8 +220,8 @@ func (t attachDataResourceDependsOnTransformer) Transform(_ context.Context, g *
 			res = append(res, d)
 		}
 
-		log.Printf("[TRACE] attachDataDependenciesTransformer: %s depends on %s", depender.ResourceAddr(), res)
-		depender.AttachDataResourceDependsOn(res, fromModule)
+		log.Printf("[TRACE] attachResourceDepenndsOnTransformer: %s depends on %s", depender.ResourceAddr(), res)
```
cam72cam: nit: typo

internal/tofu/transform_reference.go
```diff
@@ -454,10 +461,46 @@ func (m ReferenceMap) dataDependsOn(depender graphNodeDependsOn) []*addrs.Refere
 	return refs
 }
 
-// parentModuleDependsOn returns the set of vertices that a data sources parent
+// ephemeralDependsOn returns extra depends_on references if this is an ephemeral resource.
+// For ephemeral resources we implicitly treat references to managed resources as
+// depends_on entries. If an ephemeral resource references a managed resource, even if
+// that reference is resolvable, it stands to reason that the user intends for
+// the ephemeral resource to require that resource in some way.
+func (m ReferenceMap) ephemeralDependsOn(depender graphNodeDependsOn) []*addrs.Reference {
```
cam72cam: As far as I can tell, this is nearly identical to the above function, is there any reason to split them out?

internal/tofu/node_resource_abstract_instance.go
yottta: You are right! I will remove this and later, when we are going to work on the testing support for ephemerals, we can come back to this kind of changes if really necessary.
Fixed in 28c5484f5ed13719efa75b055f18ac2bb6c4ecf0

internal/tofu/node_resource_abstract_instance.go
yottta: Good catch! Missed the return there. Thanks.
Fixed in 28c5484f5ed13719efa75b055f18ac2bb6c4ecf0

internal/tofu/node_resource_abstract_instance.go
yottta: Would make sense thinking that this code path could be reached only if the setup of nodes is broken in OpenTofu, but I am not eager on throwing panics here and there 😅 
I will digest this tomorrow morning, hopefully with a fresh perspective.

internal/tofu/node_resource_abstract_instance.go
yottta: Uh! Good catch!

Fixed in 28c5484f5ed13719efa75b055f18ac2bb6c4ecf0

internal/tofu/transform_reference.go
yottta: I did it this way just to split the logic, but yeah, at a second check on this, I moved the splitting one level below so I removed this one.
Fixed in 28c5484f5ed13719efa75b055f18ac2bb6c4ecf0

internal/tofu/node_resource_plan.go
yottta: You're right! I will take a note on this and have this checked later.

internal/tofu/transform_reference.go
yottta: Fixed in 28c5484f5ed13719efa75b055f18ac2bb6c4ecf0

internal/tofu/transform_reference.go
yottta: Didn't like the approach pushed in 28c5484f5ed13719efa75b055f18ac2bb6c4ecf0 so reworked `ReferenceMap` in f54202b7819975d9ce0a4a530c8e9addb94b5007. Feel free to check this out.

internal/tofu/node_resource_abstract_instance.go
yottta: Added in fda1322f753ff8e9a5e9a0230346b5e3b4fdca12

cam72cam:
This looks good to me, it will be easier to test in a more comprehensive way when the full PR is ready to go :+1:

PR #2921, opened 2025-06-16
RFC: Alternate Execution Model

cam72cam:
This is a work in progress that I would like some early feedback on!


cam72cam:
Closing this as subsumed by https://github.com/opentofu/opentofu/pull/3078

PR #2985, opened 2025-07-02
Ephemeral during apply

yottta:
Part of #2834

This adds execution of ephemeral resources during the apply phase.
Since the apply execution, in general, acts upon the changes generated by the plan phase, in order to be able to run apply by using a planfile, there were 2 approaches to make this work:
* Create a new subgraph of the ephemeral resources from the config and walk that to add the graph nodes for executing them
* Rely on the already existing DiffTransformer, that is using the plan.changes to generate the execution nodes.

We decided to go with the second option. To do so:
* we had to update the planfile.proto to include one more action (OPEN)
* nullify the values of the ephemeral resource that is going to be stored and instead store only its addr for later use

Since we started storing information for ephemeral resources in the plan file we also had to update the rendering of the changes to ignore ephemeral ones.
This is applicable for both, `OperationJSON` and `OperationHuman`.

## Checklist

### Go checklist

 


internal/plans/planfile/tfplan.go
```diff
@@ -823,6 +821,15 @@ func changeToTfplan(change *plans.ChangeSrc) (*planproto.Change, error) {
...
 		ret.Action = planproto.Action_CREATE_THEN_DELETE
 		ret.Values = []*planproto.DynamicValue{before, after}
+	case plans.Open:
+		ret.Action = planproto.Action_OPEN
+		// We need to write ephemeral resources to the plan file to be able to build
+		// the apply graph on `tofu apply <planfile>`.
+		// The DiffTransformer needs the changes from the plan to be able to generate
+		// executable resource instance graph nodes so we are adding the ephemeral resources too.
+		// Even though we are writing these, the actual values of the ephemeral *must not*
+		// be written to the plan so set nothing here.
```
cam72cam, line 831: Perhaps a comment linking this to changeFromTfplan?

internal/plans/planfile/tfplan_test.go
```diff
@@ -325,10 +325,12 @@ func TestTFPlanRoundTrip(t *testing.T) {
 		t.Fatal(err)
 	}
 	{
-		// remove ephemeral from the plan since that should have been skipped from been written
-		plan.Changes.Resources = slices.DeleteFunc(plan.Changes.Resources, func(src *plans.ResourceInstanceChangeSrc) bool {
+		// nullify the ephemeral values from the initial plan since those must be nil in the plan file
+		i := slices.IndexFunc(plan.Changes.Resources, func(src *plans.ResourceInstanceChangeSrc) bool {
```
cam72cam, line 329: ~Why are we doing this in a _test file?  To replicate other functionality?~

Edit: I see now, I should have double checked what the test function was trying to do and misunderstood.

yottta: Exactly! Because we are sending a change with values to the function that writes the plan, and when we read it, we want to double check that the information didn't end up in the file.

Seems redundant, but I wanted to have another guarding layer while writing the plan file. I know that the nodes are writing the changes with no values, but better to have a sanitisation process in the planfile writing

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3037,6 +3043,83 @@ func maybeImproveResourceInstanceDiagnostics(diags tfdiags.Diagnostics, excludeA
...
+	var diags tfdiags.Diagnostics
+	var keyData instances.RepetitionData
+	var configVal cty.Value
+
+	_, providerSchema, err := n.getProvider(ctx, evalCtx)
+	if err != nil {
+		return nil, keyData, diags.Append(err)
+	}
+
+	config := *n.Config
```
cam72cam: Why deref the config?

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3037,6 +3043,83 @@ func maybeImproveResourceInstanceDiagnostics(diags tfdiags.Diagnostics, excludeA
...
+	schema, _ := providerSchema.SchemaForResourceAddr(n.Addr.ContainingResource().Resource)
+	if schema == nil {
+		// Should be caught during validation, so we don't bother with a pretty error here
+		diags = diags.Append(fmt.Errorf("provider %q does not support ephemeral resource %q", n.ResolvedProvider.ProviderConfig.InstanceString(n.ResolvedProviderKey), n.Addr.ContainingResource().Resource.Type))
+		return nil, keyData, diags
+	}
+
+	forEach, _ := evaluateForEachExpression(ctx, config.ForEach, evalCtx, n.Addr)
```
cam72cam: Can we use the existing InstanceExpander data to retrieve this the instance data?  I know we can't do this for some resources / data sources as the for_each might have changed between runs, but that does not apply to ephemeral?

internal/tofu/node_resource_abstract_instance.go
yottta: TBH, no good reason. Followed the existing approaches around this. From a rough check, doesn't seem to be required. I will remove it, we'll see later if will create unwanted behavior.

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3037,6 +3043,83 @@ func maybeImproveResourceInstanceDiagnostics(diags tfdiags.Diagnostics, excludeA
...
+	configKnown := configVal.IsWhollyKnown()
+	// If our configuration contains any unknown values, or we depend on any
+	// unknown values then we must defer the opening to the apply phase by
+	// producing an "Open" change for this resource, and a placeholder value for
+	// it in the state.
+	if !configKnown {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Incomplete configuration for ephemeral resource",
+			Detail:   fmt.Sprintf("Ephemeral resource %q is having incomplete configuration.", n.Addr.String()),
```
cam72cam: "is having"

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3037,6 +3043,83 @@ func maybeImproveResourceInstanceDiagnostics(diags tfdiags.Diagnostics, excludeA
...
+		})
+		return nil, instances.RepetitionData{}, diags
+	}
+
+	// Now we've loaded the data, and diags tells us whether we were successful
+	// or not, we are going to create our plannedChange and our
+	// proposedNewState.
+	var plannedNewState *states.ResourceInstanceObject
+
+	diags = diags.Append(readDiags)
```
cam72cam, line 3101: I'd move the diags append right after the call.  After the deferall check, do the usual
if diags.HasErrors() {
  return nil, keyData, diags
}
then construct planned new state and return it

internal/plans/planfile/tfplan_test.go
cam72cam: This is good, thanks :+1:

cam72cam (approved):
Lots of great comments here, this is very thorough.

internal/plans/planfile/tfplan.go
yottta: Not sure that I get your point. This comment is already in `changeFromTfplan`. Do you mean to add a comment in DiffTransformer about `changeFromTfplan`? 🤔

internal/tofu/node_resource_abstract_instance.go
yottta: That's a good question, thanks for bringing it up. Nice thing with the repeat data from the expander.

Though, could you expand a little bit on this part? 
> we can't do this for some resources / data sources as the for_each might have changed between runs

I agree with you that for ephemeral resources does not matter what for_each block is it having in the configuration because it will trigger no changes if you change it and run a new `tofu apply`. But looking deeper into this, I have a hard time understanding why data sources and resources do the for_each evaluation again during execution 🤔 . Because the first transformer executed in the apply graph (`ConfigTransformer`) creates `nodeExpandApplyableResource` that is responsible with populating the `InstanceExpander` with the expansion information for each resource. Therefore, on *every new run*, at the time of data/resource execution, the repetition information is already up to date because of this `nodeExpandApplyableResource`, but it's evaluated again during the execution as it's done above too.
What do I miss? I misread into the "runs" word? Do you mean phases and I am lacking some knowledge there?

internal/tofu/node_resource_abstract_instance.go
yottta: Fixed in d8bb0fc63dbf6710503a04e5044cc0aae7feb043

internal/tofu/node_resource_abstract_instance.go
yottta: Switched to the suggested approach in d8bb0fc63dbf6710503a04e5044cc0aae7feb043

internal/tofu/node_resource_abstract_instance.go
yottta: Fixed in d8bb0fc63dbf6710503a04e5044cc0aae7feb043

internal/tofu/node_resource_abstract_instance.go
yottta: Fixed in d8bb0fc63dbf6710503a04e5044cc0aae7feb043

internal/tofu/transform_closeable_resource.go
```diff
@@ -36,67 +36,91 @@ func (t *CloseableResourceTransformer) Transform(_ context.Context, g *Graph) er
...
+	for _, dn := range g.DownEdges(inst) {
+		switch dn.(type) {
+		case GraphNodeProvider:
+			deps = append(deps, dn)
+			//g.Connect(dag.BasicEdge(ncr, dn))
+		}
+
+	}
+
+	// connect all the resource's dependencies to the close node to ensure that we are not executing the closing of the
```
cam72cam: Old comment paired with the commented out connect below?

yottta: Thanks a lot. Ugly slip up on my side. Fixed in a096347897b66a03f32c11392caf8ce8b7a18ba2

PR #3006, opened 2025-07-09
Ephemeral resources mark and references

yottta:
Part of #2834

The main goal of this PR is to add the ephemeral mark and ensure that the values evaluated from expressions pointing to ephemeral resources are marked with the new mark.
To be able to test the ephemeral resources marks, some other blocks had to get their ephemeral mark too:
* provider
* connection
* provisioner (remote and file since local had been tackled previously)
* output
  * here only partial work has been done in this PR. More will be done in another PR, specifically for outputs 
  * moved the output validations from config to graph node
* variables
  * here only partial work has been done in this PR. More will be done in another PR, specifically for variables

The approach took here was to add a new `Ephemeral bool` in the `*configschema.Block` that is true only for the contexts mentioned above.
This way, inside evalCtx, when a block is evaluated, we can validate its schema allow to contain ephemeral values.
[Here](https://github.com/opentofu/opentofu/pull/3006/commits/8acb084af3bdb4227e77fb00c2fc72150c29a90a) is where the validation, the `Ephemeral bool` and some related changes had been added.

Because of the way we generate the ephemeral errors in the evalCtx, some existing evaluation diagnostics now are returned after calling [InConfigBody](https://github.com/opentofu/opentofu/pull/3006/commits/ec35ffd9beebb363a7c5426efc470f0ffd3e62e3).

## Checklist

### Go checklist

 


yottta:
Closing in favor of #3049

Issue #2948, opened 2025-06-24
Tofu planning with lots of depends_on does not output info log


ozeta:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
OpenTofu v1.9.1
on linux_amd64
```

### The problem in your OpenTofu project

We are deploying our infrastructure using [Spacelift](https://spacelift.io/), which provides public runners with a non-configurable timeout of 7 minutes. 

When OpenTofu enters a phase of dependency resolution at planning time, it may produce no INFO output for several minutes, triggering a premature termination from Spacelift:

 

```
…
[01JXMCV46C8DS9MKH07V79RKEW] Still running…
…
[01JXMCV46C8DS9MKH07V79RKEW] Still running...
[01JXMCV46C8DS9MKH07V79RKEW] No logs received for 7m0s, killing the process group
[01JXMCV46C8DS9MKH07V79RKEW] On private workers, this timeout is controlled by the SPACELIFT_LAUNCHER_LOGS_TIMEOUT environment variable

```

 

This is problematic because, although OpenTofu is actively working, the lack of output leads Spacelift to assume the process is blocked.

 

After enabling TRACE log, we were able to inspect the full logs and identify the phase where OpenTofu was still processing for long — specifically in the `attachDataDependenciesTransformer`. It appears that the loop iterating over resources/data sources can be long-running without producing any meaningful log output. these 4 lines like below repeated for 6 minutes:

 

```

2025-04-30T14:44:25.924Z [TRACE] attachDataDependenciesTransformer: module.karpenter_role.data.aws_region.current depends on []
2025-04-30T14:44:27.410Z [TRACE] attachDataDependenciesTransformer: module.studio.module.iam_assumable_role_modeling_service_s3.data […]
2025-04-30T14:44:27.410Z [TRACE] attachDataDependenciesTransformer: module.eks.module.kms.data.aws_caller_identity.current depends on []
2025-04-30T14:44:29.020Z [TRACE] attachDataDependenciesTransformer: module.studio.module.iam_eks_role_ms.data.aws_iam_policy_document.this depends on [modul [… a very long line of log]

```

 

[...]

PR #3016, opened 2025-07-11
RFC: OpenTofu Middleware System - Extensions made easy

Yantrio:
## RFC: OpenTofu Middleware System

This RFC proposes adding a middleware system to OpenTofu that allows external processes to hook into resource operations.

### TL;DR

- **Problem**: Tools like OPA, Checkov, and Infracost run outside OpenTofu, creating integration complexity
- **Solution**: Built-in middleware hooks that run during OpenTofu execution
- **How**: External processes communicate via JSON-RPC over stdio (like MCP servers)
- **When**: Multiple hooks 
	- pre/post for plan, apply, and refresh operations per resource/data-source. 
	- pre/post for each phase: init, plan, apply.
- **Result**: Middleware can fail operations with custom errors or attach metadata to state

### Example

```typescript
## MIDDLEWARE
import { MiddlewareServer, StdioTransport } from '@opentofu/middleware-sdk';

const server = new MiddlewareServer({
  name: "cost-estimator",
  version: "1.0.0"
});

server
  .postPlan(async (params) => {
    const cost = await estimateCost(params.resource_type, params.after);
    return {
      status: cost.monthly > 1000 ? "fail" : "success",
      message: `Estimated cost: $${cost.monthly}/month`,
      metadata: { estimated_cost: cost }
    };
  });

await new StdioTransport().connect(server);
```

```hcl
## OpenTofu Config
middleware "cost_estimator" {
  command = "estimate-costs"
}

provider "aws" {
  middleware = [middleware.cost_estimator]
}
```

If the cost estimator returns `{"status": "fail", "message": "Resource exceeds $1000/month budget"}`, OpenTofu aborts the run and displays the error to the user.

### DIFF
--- a/rfc/20250711-Middleware-For-Enhanced-Operations.md
+++ b/rfc/20250711-Middleware-For-Enhanced-Operations.md
@@ -8,7 +8,7 @@ Today, a rich ecosystem of third-party tools has emerged around Terraform and Op
 - **Complex workflows**: Users must orchestrate multiple tools and handle failures across disconnected systems
 - **No feedback loop**: External tools cannot influence OpenTofu's behavior or provide metadata back to the state
 
-This RFC proposes bringing this extensibility directly into OpenTofu through a middleware system that allows users to intercept and augment operations at key points during execution. By providing official hook points and a well-defined protocol, we can enable the same policy enforcement, cost estimation, and compliance validation use cases while offering several advantages:
+This RFC proposes bringing this extensibility directly into OpenTofu through a middleware system that allows users to intercept operations at key points during execution. By providing official hook points and a read-only, well-defined protocol, we can enable the same policy enforcement, cost estimation, and compliance validation use cases while offering several advantages:
 
 - **Real-time validation**: Catch issues during planning, not after
 - **Rich context**: Access to full resource state, provider details, and operation context
@@ -21,7 +21,9 @@ By offering this extensibility through a well-defined protocol, we aim to enable
 The architecture of this middleware system is inspired by the recent Model Context Protocol (MCP) approach, which has proven successful for extending AI assistants through local server processes communicating via JSON-RPC over stdio (along with other transports). This design pattern offers simplicity, language independence, and process isolation - qualities that align perfectly with OpenTofu's needs for a stable, extensible middleware system without the overhead of producing a full SDK or requiring middleware authors to be locked into a specific languge/ecosystem. You can find more information about this specification [at modelcontextprotocol.io/docs](https://modelcontextprotocol.io/docs/concepts/architecture).
 
 ## Approach
-To provide users with a flexible and powerful way to extend OpenTofu's functionality, I propose implementing a middleware system that communicates with external processes via JSON-RPC 2.0 over STDIO. The middleware will be invoked at specific hook points during OpenTofu operations, allowing users to inspect, validate, and augment the behavior of resource operations without requiring changes to OpenTofu's core.
+To provide users with a flexible and powerful way to extend OpenTofu's functionality, I propose implementing a middleware system that communicates with external processes via JSON-RPC 2.0 over STDIO. The middleware will be invoked at specific hook points during OpenTofu operations, allowing users to inspect, validate, and augment the behavior of resource operations without requiring changes to OpenTofu's core. 
+
+When it comes to augmentation of the resource operations, I propose that initially the middleware cannot alter the state or inputs to a resource/datasource etc. I believe that providing a read only system that can return failures and metadata is the safest way to go and then there is less risk for the end users too. Less "magic" altering things means that the config is the source of truth.
 
 ### Hook Points for Operations
 The middleware system will provide hook points at two levels:
--- a/rfc/20250711-Middleware-For-Enhanced-Operations.md
+++ b/rfc/20250711-Middleware-For-Enhanced-Operations.md
@@ -138,6 +138,13 @@ In this case:
 
 Users should generally avoid this pattern unless the middleware is designed to handle multiple executions.
 
+### Redistributing Middleware
+We may decide to make it possible to allow for the creation and distribution of middleware configuration via modules.
+
+There are situations today where company's may want to create redistributable packs of pre-defined middleware configuration to enforce policies company-wide. To combat this I propose that we make it possible in the future to define middleware blocks inside modules, and also allow for module calls to take middleware as input and provide middleware as outputs.
+
+For more information around this, you can see an example I put together [here](https://gist.github.com/Yantrio/a9bb20fcfeb15f610e11ea6b71c2112d). This example acts as an extreme case of how you can compose middleware inside modules to be used elsewhere in your projects.
+
 ### Communication
 As mentioned above, I propose that to keep it as open as possible we should use JSON-RPC 2.0 over stdio for communication with the possibility to open up to other transports in the future.
 
--- a/rfc/20250711-Middleware-For-Enhanced-Operations.md
+++ b/rfc/20250711-Middleware-For-Enhanced-Operations.md
@@ -1,0 +1,378 @@
+# OpenTofu Middleware For Enhanced Operations
+
+### Introduction
+Today, a rich ecosystem of third-party tools has emerged around Terraform and OpenTofu that analyze plan files, parse HCL configurations, and enforce policies. Tools like OPA (Open Policy Agent), Sentinel, Checkov, Infracost, and others have proven invaluable for organizations implementing governance, compliance, and cost controls. However, these tools operate entirely outside the OpenTofu execution flow, requiring users to integrate them through CI/CD pipelines, wrapper scripts, or manual processes. This separation creates several challenges:
+
+- **Timing gaps**: Policies are evaluated after plans are generated, missing opportunities for early validation
+- **Limited context**: External tools only see what's exported to plan files, missing runtime state and provider-specific details  
+- **Complex workflows**: Users must orchestrate multiple tools and handle failures across disconnected systems
+- **No feedback loop**: External tools cannot influence OpenTofu's behavior or provide metadata back to the state
+
+This RFC proposes bringing this extensibility directly into OpenTofu through a middleware system that allows users to intercept and augment operations at key points during execution. By providing official hook points and a well-defined protocol, we can enable the same policy enforcement, cost estimation, and compliance validation use cases while offering several advantages:
+
+- **Real-time validation**: Catch issues during planning, not after
+- **Rich context**: Access to full resource state, provider details, and operation context
+- **Simplified workflows**: Middleware runs as part of normal OpenTofu operations
+- **Bidirectional communication**: Middleware can fail operations and store metadata in state
+- **Language agnostic**: Write middleware in any language
+
+By offering this extensibility through a well-defined protocol, we aim to enable organizations to bring their existing policy and governance tools into the OpenTofu execution flow while fostering innovation in new use cases that weren't possible with external analysis alone.
+
+The architecture of this middleware system is inspired by the recent Model Context Protocol (MCP) approach, which has proven successful for extending AI assistants through local server processes communicating via JSON-RPC over stdio (along with other transports). This design pattern offers simplicity, language independence, and process isolation - qualities that align perfectly with OpenTofu's needs for a stable, extensible middleware system without the overhead of producing a full SDK or requiring middleware authors to be locked into a specific languge/ecosystem. You can find more information about this specification [at modelcontextprotocol.io/docs](https://modelcontextprotocol.io/docs/concepts/architecture).
+
+## Approach
+To provide users with a flexible and powerful way to extend OpenTofu's functionality, I propose implementing a middleware system that communicates with external processes via JSON-RPC 2.0 over STDIO. The middleware will be invoked at specific hook points during OpenTofu operations, allowing users to inspect, validate, and augment the behavior of resource operations without requiring changes to OpenTofu's core.
+
+### Hook Points for Operations
+The middleware system will provide hook points at two levels:
+
+**Resource-level hooks** (called for each individual resource/data source):
+- **pre-plan**
+- **post-plan**
+- **pre-apply**
+- **post-apply**
+- **pre-refresh**
+- **post-refresh**
+
+**Operation-level hooks** (called once per operation):
+- **init-stage-start**
+- **init-stage-complete**
+- **plan-stage-start**
+- **plan-stage-complete**
+- **apply-stage-start**
+- **apply-stage-complete**
+
+Additional hooks can be added in future versions as use cases emerge. It is possible that we should also consider hooks that handle failures/errors too.
+
+### Middleware Attachments
+
+For now I propose that middleware should be "attachable" to the entire project, or to a specific provider instance. This opens up the prospect of having provider-extension middleware or project-wide enforcement of rules.
+
+### Configuration Design
+Users will configure middleware through HCL blocks in their OpenTofu configuration:
+
+```hcl
+
+middleware "naming_convention_checker" {
+  command = "lint-opentofu-names"
+}
+
+middleware "cost_estimator" {
+  command = "python3"
+  args    = ["cost_estimator.py"]
+  env = {
+    API_KEY = var.cost_api_key
+  }
+}
+
+terraform { 
+  middleware = [middleware.naming_convention_checker] # Attach it across the entire project
+}
+
+provider "aws" {
+  middleware = [middleware.cost_estimator] # Or just attach it to a single provider
+  region = "us-east-1"
+}
+```
+
+### Middleware Execution Order
+
+When multiple middleware are attached to a provider or project, they execute in the order defined in the middleware array. This deterministic ordering allows for:
+
+- **Pipeline Processing**: Earlier middleware can validate/transform data for later middleware
+- **Priority Control**: Critical security checks can run before cost estimation
+- **Dependency Management**: Middleware that depends on others' metadata can run later
+
+Example execution flow:
+```hcl
+provider "aws" {
+  middleware = [
+    middleware.security_scanner,    # Runs first - can fail fast on violations
+    middleware.compliance_checker,  # Runs second - sees security metadata
+    middleware.cost_estimator      # Runs last - only if security/compliance pass
+  ]
+}
+```
+
+For pre-hooks (pre-plan, pre-apply, etc.), if any middleware returns a "fail" status, execution stops immediately and subsequent middleware in the chain are not called. For post-hooks, all middleware execute regardless of individual failures, allowing comprehensive metadata collection.
+
+### Execution Hierarchy
+
+When middleware is attached at both the project level (via `terraform` block) and provider level, the execution order follows a clear hierarchy:
+
+1. **Project-level middleware executes first** - Applied to all resources regardless of provider
+2. **Provider-level middleware executes second** - Applied only to resources from that provider
+
+This hierarchy allows for:
+- Global policies to be enforced before provider-specific checks
+- Provider middleware to see and build upon project middleware metadata
+- Layered validation from general to specific
+
+#### Duplicate Middleware Handling
+
+If the same middleware is attached at both levels, it will execute **twice** for resources from that provider:
+
+```hcl
+middleware "cost_estimator" {
+  command = "cost-estimate"
+}
+
+terraform {
+  middleware = [middleware.cost_estimator]  # First execution
+}
+
+provider "aws" {
+  middleware = [middleware.cost_estimator]  # Second execution - overwrites metadata!
+}
+```
+
+In this case:
+- The middleware runs twice for AWS resources
+- The second execution's metadata **overwrites** the first
+- Each execution has no knowledge it's being run multiple times
+
+Users should generally avoid this pattern unless the middleware is designed to handle multiple executions.
+
+### Communication
+As mentioned above, I propose that to keep it as open as possible we should use JSON-RPC 2.0 over stdio for communication with the possibility to open up to other transports in the future.
+
+#### Initialize Handshake
+```json
+// Request
+{
+  "jsonrpc": "2.0",
+  "method": "initialize",
+  "params": {
+    "version": "1.0",
+    "name": "cost_estimator"
+  },
+  "id": 1
+}
+
+// Response
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "capabilities": ["pre-plan", "post-plan", "pre-apply", "post-apply"]
+  }
+}
+```
+
+#### Hook Execution
+```json
+// Post-Plan Hook Request
+{
+  "jsonrpc": "2.0",
+  "method": "post-plan",
+  "params": {
+    "provider": "registry.opentofu.org/hashicorp/aws",
+    "resource_type": "aws_instance",
+    "resource_name": "example",
+    "resource_mode": "managed",
+    "planned_action": "Create",
+    "before": null,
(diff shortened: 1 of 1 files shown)


rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
+# OpenTofu Middleware For Enhanced Operations
+
+### Introduction
+Today, a rich ecosystem of third-party tools has emerged around Terraform and OpenTofu that analyze plan files, parse HCL configurations, and enforce policies. Tools like OPA (Open Policy Agent), Sentinel, Checkov, Infracost, and others have proven invaluable for organizations implementing governance, compliance, and cost controls. However, these tools operate entirely outside the OpenTofu execution flow, requiring users to integrate them through CI/CD pipelines, wrapper scripts, or manual processes. This separation creates several challenges:
+
+- **Timing gaps**: Policies are evaluated after plans are generated, missing opportunities for early validation
+- **Limited context**: External tools only see what's exported to plan files, missing runtime state and provider-specific details  
+- **Complex workflows**: Users must orchestrate multiple tools and handle failures across disconnected systems
+- **No feedback loop**: External tools cannot influence OpenTofu's behavior or provide metadata back to the state
```
TheOutdoorProgrammer, line 9: > **Limited context**: External tools only see what's exported to plan files, missing runtime state and provider-specific details
> **No feedback loop**: External tools cannot influence OpenTofu's behavior or provide metadata back to the state

Would such a plugin be allowed to change attributes? For example, infracost finds that an instance size is too large/costly. So it changes the user requested `t3.100xl` to a `t3.medium`?

rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
...
+
+**Operation-level hooks** (called once per operation):
+- **init-stage-start**
+- **init-stage-complete**
+- **plan-stage-start**
+- **plan-stage-complete**
+- **apply-stage-start**
+- **apply-stage-complete**
+
+Additional hooks can be added in future versions as use cases emerge. It is possible that we should also consider hooks that handle failures/errors too.
```
TheOutdoorProgrammer, line 47: > It is possible that we should also consider hooks that handle failures/errors too.

This should be a first-class citizen to this RFC IMO. I can see people wanting to trigger other workflows depending on what the failure was, or wanting to aggregate the failures in some way.

rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
...
+middleware "naming_convention_checker" {
+  command = "lint-opentofu-names"
+}
+
+middleware "cost_estimator" {
+  command = "python3"
+  args    = ["cost_estimator.py"]
+  env = {
+    API_KEY = var.cost_api_key
+  }
```
TheOutdoorProgrammer, line 69: Im not sure this is going to be needed, they could just have the ENV be part of the environment thats running tofu. We'd be effectively setting the environment twice here, right? If folks need to consume variables in their runtime, they could just export those variables before running tofu. I think passing variables through the `args` parameter would be a more valid use-case.

rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
...
+provider "aws" {
+  middleware = [
+    middleware.security_scanner,    # Runs first - can fail fast on violations
+    middleware.compliance_checker,  # Runs second - sees security metadata
+    middleware.cost_estimator      # Runs last - only if security/compliance pass
+  ]
+}
+```
+
+For pre-hooks (pre-plan, pre-apply, etc.), if any middleware returns a "fail" status, execution stops immediately and subsequent middleware in the chain are not called. For post-hooks, all middleware execute regardless of individual failures, allowing comprehensive metadata collection.
```
TheOutdoorProgrammer, line 101: I think we should allow the middleware to tell tofu what to do in both situations. I cant think of any use-cases off the top of my head, but maybe if some post hook wants to stop the execution for some valid reason we should allow them to do so?

rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+### Process Lifecycle
+
+Middleware processes are managed with a simple lifecycle:
+
+1. **Startup**: When OpenTofu begins execution (init, plan, apply, etc.), all configured middleware processes are started
+2. **Initialization**: Each middleware receives an initialize request and must respond with its capabilities
+3. **Operation**: Middleware processes remain running for the entire duration of the OpenTofu command
+4. **Shutdown**: When OpenTofu completes (successfully or with errors), all middleware processes are terminated
```
TheOutdoorProgrammer, line 219: Should we consider when a middleware wants to run through all lifecycles and then wants to do some functionality during shutdown?

I know you mentioned multiple invocations above, it not being aware of previous invocations of a middleware. But Im imagining something like some security middleware wants to ensure the plan is the same between planning and applying. They could even allow the user to still apply the different plan but during shutdown they show some warning or something. Just thinking out loud.

rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Open Questions
+
+### Performance and Scalability
+- What is the performance impact of middleware on large infrastructures?
+- Should middleware calls be parallelized for independent resources?
+- How do we handle middleware that needs to see all resources before making decisions?
+
+### Security Considerations
+- Should we sanitize the input to the middleware to hide sensitive fields?
```
TheOutdoorProgrammer: IMO we should just add a field to the middleware block to let the user choose.

```hcl
provider "aws" {
   middleware = [middleware.security]
   middleware_sensitive_values = true
}
```

ImIOImI:
Is it possible to provide a middleware hook with context from OpenTofu (like the values of local variables or the output of a certain resource, etc.)? I'm especially looking at hooks during and after the planning init stage.

cam72cam:
Overall, I quite like this approach and think it would be a valuable addition for the community and the ecosystem as a whole!  At least for now, I like the restriction that it's a local process that doesn't have any networking.

Most of the questions I had were addressed or mentioned in the PR.

Additional questions:
* What happens when middleware is renamed.  We missed that during the encryption design
* Can middleware be declared in or passed to child modules / passed around as a value?  Do they have to be in the root of a project for security?

rfc/20250711-Middleware-For-Enhanced-Operations.md
```diff
@@ -0,0 +1,378 @@
...
+- Should there be restrictions on what middleware can access?
```
TheOutdoorProgrammer, line 378: If you ask me, I would say we make this as configurable as possible. That way it opens up the usage of middleware wherever possible.

apparentlymart:
This all seems pretty interesting! I'm slightly concerned that whatever we expose here will constrain how OpenTofu can evolve in future by exposing concepts that are currently hidden implementation details, but that's more of a "let's make sure to design the details carefully" than any sort of objection. It's a similar tradeoff to the JSON formats produced by `tofu show` and `tofu plan -json`, so we do at least have a starting point for what level of detail we've already committed to exposing to external integrations, and so we can limit our extra attention only to behaviours that aren't already exposed that way and that's hopefully a small set. 

I suspect that some folks will want to treat their "middleware" programs as external shared dependencies rather than including them inline in the same place as the calling configuration, so they can be easily reused across multiple callers without duplication, so we might consider changing how they are specified to include a module package source address (the kind of thing you'd write in a module call's `source` argument, though perhaps referring to a file in the package rather than a directory in this case) and then we can have `tofu init` install them as part of the existing module package fetching process. (This would not work for programs that want to hook the init process itself, of course. Are the use-cases for hooking init more important than the ability to fetch these programs automatically during init? I'm not sure. 🤷🏻‍♂️)

Separately, I wonder about including an extra tier of configuration that belongs to the execution environment where OpenTofu is running, rather than to the OpenTofu configuration that's being run. That would then allow platform teams (or TACOS-like systems acting on their behalf) to include behaviours that should happen regardless of what configuration is currently being used, outside the control of whoever wrote that configuration, which seems to be a common requirement for policy-related concerns at least, so that it isn't possible to disable the policy in the same action as introducing a violation of it (whether accidentally or maliciously). That could be controlled by one or more of our usual (unfortunately rather inconsistent) ways for external automation to influence OpenTofu's behaviour: CLI configuration, command line arguments, environment variables.

rfc/20250711-Middleware-For-Enhanced-Operations.md
cam72cam: Between multiple middlewares fighting about what an instance size should be, to simply tracing deployment failures back to the configuration, I'd that it should be read only access to the attributes.

cam72cam: It could suggest that the user change the instance in the error message.  That would still encourage them to modify the configuration that represents the "truth" that should be applied.

rfc/20250711-Middleware-For-Enhanced-Operations.md
cam72cam: Could pass a "another middleware has failed" info potentially.

rfc/20250711-Middleware-For-Enhanced-Operations.md
cam72cam: I think doing this in the middleware block itself makes more sense.

rfc/20250711-Middleware-For-Enhanced-Operations.md
Yantrio: I agree with @cam72cam here, I think we should keep it read only 100% so that we dont cause confusion, but also for security purposes.

rfc/20250711-Middleware-For-Enhanced-Operations.md
Yantrio: I think this is valid, I'll put some thought into how we can handle this nicely.

rfc/20250711-Middleware-For-Enhanced-Operations.md
Yantrio: I'm a little torn on this one. Personally I feel that the env vars should be set explicitly as this is a great way to configure the applicaton, also it helps with situations where you may have multiple middleware that do similar things. 

You may want to run multiple instances of the same middleware with different env vars.

This does raise another question though about if we should propagate the env vars and do these env vars that are in the config override the existing ones from the `tofu` execution?

rfc/20250711-Middleware-For-Enhanced-Operations.md
Yantrio: So I think we have 2 ways to handle this that "feel" clean to me.

a) We introduce a "warn" status that will produce a warning and not fail the run.
b) We introduce a new field on the middleware that sets "fail on error" to true/false

Either way could work, maybe we can do both? what do you think?

rfc/20250711-Middleware-For-Enhanced-Operations.md
Yantrio: So, I think this can be worked around using the current system by having the middleware "store" its data on disk or in a database and it being stateful. There is no rule against middleware being stateful here at all.

In situations where a plan and then an apply happen as 2 individual calls to `tofu` I feel as if we should not allow for attaching to an existing middleware instance it would just complicate things.

rfc/20250711-Middleware-For-Enhanced-Operations.md
Yantrio: That's valid 👍  I think adding it to the middleware block is preferred for me, if you wish to handle it at a provider level you can define a different set of middleware for the provider to use.

Yantrio:
>I suspect that some folks will want to treat their "middleware" programs as external shared dependencies rather than including them inline in the same place as the calling configuration, so they can be easily reused across multiple callers without duplication, so we might consider changing how they are specified to include a module package source address (the kind of thing you'd write in a module call's source argument, though perhaps referring to a file in the package rather than a directory in this case) and then we can have tofu init install them as part of the existing module package fetching process. (This would not work for programs that want to hook the init process itself, of course. Are the use-cases for hooking init more important than the ability to fetch these programs automatically during init? I'm not sure. 🤷🏻‍♂️)

**RE: installation**
This is a really interesting point for me. I've noticed with the prevalence of mcp servers running in both docker, or from package managers that have an "install and execute" mode such as `npx` that the concept of installing the software is not a concern that end users have much anymore.

To keep things simple in an initial introduction I was thinking that we should avoid handling the installation of the software that the middleware calls and assume that it is in the $PATH or will be installed by whatever command you are running. 

**RE: re-usable "packs" of middleware**
I agree that we should design this with re-usability in mind. I also believe that engineers will want to be able to distribute sets of middleware for things such as compliance across multiple tofu instances (projects/units/stacks etc), but the one part that I could not solve when thinking about this was the situation where you want to also define which providers get which middleware. I think the concept of re-usable middleware works really well for situations where we are attaching the middleware to the entire project, but for per-provider I don't think this works too well unless we find a way to define how a middleware gets attached. Do you have any ideas about how you would like this to look?

>Separately, I wonder about including an extra tier of configuration that belongs to the execution environment where OpenTofu is running, rather than to the OpenTofu configuration that's being run. 

In the grand scheme of things I've also been considering such an approach. I do have concerns that yet another type of "project" system is going to be invented but it does have some cross-cutting problem solving upsides! Personally I am a fan of cargo.toml from the rust/crab world and it's simplicity (language aside). 

Maybe we should open up another RFC to discuss the possibility of this and what problems it could solve. However I do have concerns that terragrunt recently tried to simplify this process in their project due to it being a large point of confusion for end users.

rfc/20250711-Middleware-For-Enhanced-Operations.md
cam72cam: I think the hard part to handle is chaining middleware.  I assume that the reason for this is that you don't want chained middleware not finding data it needs in the input?

Yantrio:
@cam72cam 

> What happens when middleware is renamed. We missed that during the encryption design
Right now the design says that the middleware can attach data to the state/plan file by returning some metadata. And then that metadata is stored under the name of the middleware.

Maybe we should add a "alias" field to the middleware, that way people can rename their middleware configuration block and keep the same alias for nice migrations.

> Can middleware be declared in or passed to child modules / passed around as a value? Do they have to be in the root of a project for security?

My gut feeling is that due to how providers are defined and passed to modules today (implicit or explicitly) from the root module, that we should ensure that middleware is handled in the same way. That is mainly because the target for middleware right now is only the root module (terraform block) or a provider. If we decide to allow for middleware to be attached to modules, or even individual resources/etc then we can re-consider.

Yantrio:
@ImIOImI 

>Is it possible to provide a middleware hook with context from OpenTofu (like the values of local variables or the output of a certain resource, etc.)? I'm especially looking at hooks during and after the planning init stage.

I think we should include the outputs, variables etc when calling the post-plan phase hook/middleware. 👍

rfc/20250711-Middleware-For-Enhanced-Operations.md
yottta: For security reasons, if `tofu` will be the invoker of the middleware binary, I would vote for passing only env vars specifically mentioned in the configuration.
In order to provider a mechanism to "forward" env vars `tofu` process have access to, we can provide a specific function for that (`fromenv("TF_VAR_my_var")` or something similar).
This way we can control exactly what env vars out of the `tofu` session, the middleware can use.

TBH, I am concerned about the possible plethora of malicious implementations that could leak sensitive user information, like tokens and other sensitive bits.

rfc/20250711-Middleware-For-Enhanced-Operations.md
yottta, line 47: I know that recently there was a conversation about this: could we add also a "heartbeat" operation? Could be configurable with a  1m00s default.
Some could find that quite useful.

Yantrio: What would the use-case be for writing custom middleware to happen on a "heartbeat" ?

yottta: Mostly related to process monitors. Those systems can inject a middleware configuration at the project level and have custom logic around the heartbeat.
I am mostly referring to this: #2948

Yantrio: So would you want to fail or cancel if things took too long? Would you like to just have a way to have middleware time things?

yottta: As mentioned in the ticket above, the "process monitoring" is done through logs parsing. Instead of that, can be added a heartbeat for it.

rfc/20250711-Middleware-For-Enhanced-Operations.md
TheOutdoorProgrammer: Makes sense to me, they can always just fail the run as well 👍

rfc/20250711-Middleware-For-Enhanced-Operations.md
TheOutdoorProgrammer: >TBH, I am concerned about the possible plethora of malicious implementations that could leak sensitive user information, like tokens and other sensitive bits.

IDK my thought is this is no different than a provider which can pick up the runtimes environment. Like the AWS provider for example can pick up `AWS_ACCESS_KEY_ID` etc etc.

Maybe we do both? Just like providers? You can explicitly pass things in but you can also just let the middleware pick up things from the current environment.

rfc/20250711-Middleware-For-Enhanced-Operations.md
TheOutdoorProgrammer: Same page, sounds good to me 👍

apparentlymart:
The remark in https://github.com/opentofu/opentofu/pull/3016#issuecomment-3068911487 about it no longer being so important for OpenTofu to be in control of dependency installation caused me to start thinking about what that assumption would imply if applied more broadly, but rather than derailing this discussion with that side-chatter I've started a new thing about it: https://github.com/opentofu/opentofu/pull/3027.

Yantrio:
After a lengthy discussion with @cam72cam thisafternoon I've collated some of my thoughts of an advanced set of use cases for middleware in this gist here: https://gist.github.com/Yantrio/a9bb20fcfeb15f610e11ea6b71c2112d

This gist is in no way intended to be an indication of what I think a version 1 should be but I do think that this answers some of @apparentlymart's questions around re-usable sets of middleware that could be distributed.

I will be rolling some of these changes back into the RFC i have proposed above (tomorrow)

Primarily the following points:
- Each middleware **must** define a `metadata_key` or similarly named property that will be used to store data inside the state or plan files, or to pass data to other middleware. This requirement will ensure that we have data integrity in chain-able middleware situations. And we can also ensure that provider attachments do not have duplicates (For example, you don't want 2 middleware outputting "cost" labelled metadata.

- Middleware should be passable as inputs and outputs to/from modules. This allows for redistribution of middleware collections. It introduces some complexity but I believe this is a nice middleground between only allowing root-module usage and creating a new "project file". I would like to discuss this more though as I do not wish to back us into a corner. I do not think that this should go into a v1. On the technical side of this, @cam72cam recommended that we mark the middleware so that we do not allow you to consume middleware properties in other places too easily. Middleware blocks should only be used as middleware blocks!

rfc/20250711-Middleware-For-Enhanced-Operations.md
yottta: > Maybe we do both? Just like providers? You can explicitly pass things in but you can also just let the middleware pick up things from the current environment.

Sounds good. And then this will provide control of how secure a user wants to have its configs. 👍

ImIOImI:
I was reminded of this this morning doing my daily inventory of drift. AWS is not going to fix their API so that their provider deals with lists better. This has been a known ongoing problem for years. It will not change. Therefore, I will either have to accept the toil involved at looking at my "drifted" stacks every morning to ensure that there isn't anything _really_ wrong, or write dangerous and annoying ignore changes lifecycle meta arguments. This feels like a longshot, but is there a chance that I can write middleware that does some basic checking on a plan and stops reporting things like the attached screenshot as drift?
image

cam72cam:
@ImIOImI I wonder if a better option here would be to add a "ignore_order" attribute to the lifecycle block.  You'd still have to add it to all of your configs, but it could solve a more general case potentially?

apparentlymart:
Although it's a bit of a layering violation (the language runtime would need to grow more awareness of UI layer concerns), it seems possible in principle to allow something handling the "post plan" hook to return a set of attribute paths that are "uninteresting" and thus to be ignored from diffing, and conversely also attribute paths that are _particularly interesting_ and so should always be shown even if the UI heuristics would normally cause them to be hidden.

We could potentially do similarly for the "refresh" step to allow influencing whether and how OpenTofu presents the "Objects have changed outside of OpenTofu" section of the plan output when refreshing detects a change made in the remote system.

That would then allow using some local context about what is and is not important for a particular organization's workflow to tailor the way the diffs are presented.

---

It's important that this only be able to affect how the information is displayed in the UI and not also affect the underlying data, because the provider protocol expects that OpenTofu sends providers exactly the same planned new state as was returned from the provider during the planning phase, or else this can violate assumptions the provider made in its planning logic where the remote API has constraints that span across multiple arguments. (For example, situations where the planning phase treats two arguments as mutually-exclusive because the remote API can't support them both being set at once; behavior would be unspecified if we then modified the planned object to violated that constraint during the apply phase, because providers might not recheck that in their apply time logic.)

This means that the UI hint would not prevent a change from being made. It would only skip mentioning that change in the plan diff, and then silently make the change during the apply phase anyway. That's a potential UX hazard, but perhaps we'd convince ourselves that it's okay because it's opt-in. (It being used in reusable modules is a little tricky though, because the person opting in is not necessarily the person dealing with the consequences of the unexpected change. 🤔)

We would need to think about what should happen if the only changes needed are ones under "uninteresting" attribute paths, since then OpenTofu would likely still need to report that the resource instance is going to be updated but would not have any interesting change of arguments to use to explain why.

ImIOImI:
@apparentlymart I came up with some scenarios that might fit into the suggestions you made.

>  it seems possible in principle to allow something handling the "post plan" hook to return a set of attribute paths that are "uninteresting" and thus to be ignored from diffing

We recently changed the tags on all our infra. Thousands of resources were updated and the PRs were excruciating. Being able to filter out uninteresting changes like the very expected tag changes would have been lifesaving and killed a TON of toil. Because of how many resources were changed it took a long time to review our PRs and often while reviewing someone would merge something and then we needed to update the branch and redo the plan. So the PR review started again from scratch. Having some kind of way to dismiss changes that have already been reviewed within the context of a PR would be magic as well. 

> conversely also attribute paths that are particularly interesting and so should always be shown even if the UI heuristics would normally cause them to be hidden.

Using OPA or something similar to highlight really important changes like database changes, roles changing, message busses getting deleted, etc. would be a big win as well. These changes might not be violations of policy per say, but they'd be really important to pull out of a plan. 

Furthermore, we're starting to see LLMs integrated into TACOS for interpreting plans and explaining the cause of drift. This is another very logical integration opportunity for those platforms. Imagine being able to print out a plan annotated by an LLM that's fluent in AWS logs and correlating a change in a resource with a change a specific user made in the AWS UI.

apparentlymart:
Hi all,

We're putting this RFC on hold right now just because our attention is elsewhere and so we're not actively pursuing it. We'll revisit thus at a later time.

Thanks!

PR #3017, opened 2025-07-11
Remove duplicated connection block schema

yottta:
Working on the ephemerality of the connection block, realised that the connection block schema is defined in two packages. Checked the diff between the two and seen nothing different (other than a single comment).

⚠️ Please double check on your own that the one I removed is not breaking any legacy assumptions or anything that I am not aware of 🤕 

The one kept: https://github.com/opentofu/opentofu/blob/a1ba3e24aa9c52452951ff00d6b6c1c777f23532/internal/communicator/shared/shared.go#L24
The one removed: https://github.com/opentofu/opentofu/blob/a1ba3e24aa9c52452951ff00d6b6c1c777f23532/internal/tofu/node_resource_validate.go#L166

Part of no issue, just random thing found.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/node_resource_abstract.go
+++ b/internal/tofu/node_resource_abstract.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/communicator/shared"
 	"github.com/opentofu/opentofu/internal/configs"
 	"github.com/opentofu/opentofu/internal/configs/configschema"
 	"github.com/opentofu/opentofu/internal/dag"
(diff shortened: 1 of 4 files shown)


cam72cam:
Looks like it originated in https://github.com/opentofu/opentofu/blame/fbe3219fbe6813c122a846314e5661d96d5d78c8~/terraform/eval_validate.go
And was copy-pasted in https://github.com/opentofu/opentofu/commit/987df2003a8be36e48f9e0deb7b8d8030037c42c  without any reason given as to why it was not imported in the other location.

apparentlymart (approved):
I don't really like the overly-general package name "shared", but that was already there so no problem with extending the use of it for now and considering a more contextually-relevant name later.

This "superset schema" has been around since the adoption of HCL 2 forced a schema on what had originally just been "load whatever is in the block into a map and let each communicator take whatever subset it wants", so ideally this wouldn't exist at all and each communicator would have its own schema but that's a much riskier change to make and so I agree that just making sure we only have this workaround in one place is the best compromise for now.

Issue #2970, opened 2025-06-27
s3 backend `.tflock` PutObject missing `x-amz-server-side-encryption` header


sgtoj:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
OpenTofu v1.10.1
on linux_arm64
```

### OpenTofu Configuration Files

```hcl
terraform {
  backend "s3" {
    bucket         = "<my-tfstate-bucket>"
    key            = "demo/terraform.tfstate"
    region         = "us-east-1"
    dynamodb_table = "<state-lock-table>"
    encrypt        = true
    use_lockfile   = true
  }
}
```

### Debug Output

Example of the error with redactions...
```
Error: Error acquiring the state lock

Error message: operation error S3: PutObject,
access denied: User is not authorized to perform: s3:PutObject on resource:
arn:aws:s3:::<my-tfstate-bucket>/…/terraform.tfstate.tflock with an explicit deny in a resource-based policy
```

### Expected Behavior

OpenTofu should include the same x-amz-server-side-encryption header on the `.tflock` object that it already sends with the `.tfstate` object, so the upload passes standard bucket policies that enforce encryption.

### Actual Behavior

`.tflock` upload omits the header → bucket policy denies the request → run aborts before any plan or apply can proceed.

### Steps to Reproduce

1. create/choose an S3 bucket `<my-tfstate-bucket>` and attach a strict policy:

   ```json
   {
     "Version": "2012-10-17",
     "Statement": [
       {
         "Sid": "DenyUnencryptedObjectUploads",
         "Effect": "Deny",
         "Principal": "*",
         "Action": "s3:PutObject",
         "Resource": "arn:aws:s3:::<my-tfstate-bucket>/*",
         "Condition": {
           "Null": { "s3:x-amz-server-side-encryption": "true" }
         }
       }
     ]
   }
   ```

2. create a minimal OpenTofu config:

   ```hcl
   terraform {
     backend "s3" {
       bucket         = "<my-tfstate-bucket>"
       key            = "demo/terraform.tfstate"
       region         = "us-east-1"
       dynamodb_table = "<state-lock-table>"
       encrypt        = true
       use_lockfile   = true
     }
   }
   ```

3. run `tofu init` (or `tofu plan`)

[...]

PR #3022, opened 2025-07-13
fix: Support provider references with index expressions in .tf.json

ystkfujii:
Resolves https://github.com/opentofu/opentofu/issues/2915

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ BUG FIXES:
 * S3 backend now correctly sends the `x-amz-server-side-encryption` header for the lockfile ([#2870](https://github.com/opentofu/opentofu/issues/2970))
 * Allow function calls in test variable blocks ([#2947](https://github.com/opentofu/opentofu/pull/2947))
 * The `issensitive` function now returns an unknown result when its argument is unknown, since a sensitive unknown value can potentially become non-sensitive once more information is available. ([#3008](https://github.com/opentofu/opentofu/pull/3008))
+* Provider references like "null.some_alias[each.key]" in .tf.json files are now correctly parsed ([#2915](https://github.com/opentofu/opentofu/issues/2915))
 
 ## Previous Releases
 
--- a/internal/configs/hcl2shim/util.go
+++ b/internal/configs/hcl2shim/util.go
@@ -6,8 +6,11 @@
 package hcl2shim
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
 )
 
 // exprIsNativeQuotedString determines whether the given expression looks like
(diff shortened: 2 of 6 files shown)


internal/configs/hcl2shim/util.go
```diff
@@ -84,6 +87,16 @@ func ConvertJSONExpressionToHCL(expr hcl.Expression) (hcl.Expression, hcl.Diagno
 		return nil, diags
 	}
 
+	if value.Type() != cty.String || value.IsNull() {
```
diofeher, line 90: Can you add a test case to `TestConvertJSONExpressionToHCL` to catch this check?

ystkfujii:
@diofeher 

Thanks for the review.
I've added a test. Please take a look at it when you have time.

internal/configs/hcl2shim/util_test.go
```diff
@@ -58,3 +58,46 @@ func TestConvertJSONExpressionToHCL(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertJSONExpressionToHCL_NonStringORNull(t *testing.T) {
```
diofeher: Thanks for the modifications! My last nit request would be to change this Test function name to `TestConvertJSONExpressionToHCL_Invalid`, since it's more generic (easier to extend later) and will be consistent with the rest of our codebase. Other than that, it's looking great :)

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ BUG FIXES:
 * S3 backend now correctly sends the `x-amz-server-side-encryption` header for the lockfile ([#2870](https://github.com/opentofu/opentofu/issues/2970))
 * Allow function calls in test variable blocks ([#2947](https://github.com/opentofu/opentofu/pull/2947))
 * The `issensitive` function now returns an unknown result when its argument is unknown, since a sensitive unknown value can potentially become non-sensitive once more information is available. ([#3008](https://github.com/opentofu/opentofu/pull/3008))
+* Provider references like "null.blah[each.key]" in .tf.json files are now correctly parsed ([#3022](https://github.com/opentofu/opentofu/pull/3022))
```
Gogotchuri: Two things here:
- Since the issue is more detailed than the PR, let's link the issue instead, it makes more sense
- This is a small nit, but including the word "blah" in CHANGELOG looks weird to me, let's use "some_alias" instead (I know it is what Martin used in his example). 
```suggestion
* Provider references like "null.some_alias[each.key]" in .tf.json files are now correctly parsed ([#2915](https://github.com/opentofu/opentofu/issues/2915))
```

Gogotchuri (changes requested):
Other than the small things I've mentioned and Dio's comment, everything looks great!
Thank you for your time!

CHANGELOG.md
cam72cam: Hmm, this is a divergence from all of our other entries in the changelog thus far.  I'd be open to changing this, but should probably be consistent.

internal/configs/hcl2shim/util_test.go
ystkfujii: Fixed!

CHANGELOG.md
ystkfujii: I changed it from 'blah' to 'some_alias'.
As the other entries in the CHANGELOG were mostly PRs, I left them as such.
Please leave a comment if any changes are needed.

Gogotchuri: @cam72cam We've had this discussion for this 

* S3 backend now correctly sends the `x-amz-server-side-encryption` header for the lockfile ([#2870](https://github.com/opentofu/opentofu/issues/2970))

entry with @yottta and @apparentlymart; the comment I made here was the decision we ultimately made - use whatever describes the change context the best.
It's fine if we leave it as it is. It just adds an extra click from PR to the issue.

diofeher: @Gogotchuri interesting point! I agree that linking the issue or PR who has more context would be the better, and either one of them would be accessible from each other since they're linked.

The only con of linking the issue is that the PR is on bottom of the discussion, while the in the opposite, the issue is the first thing you see on the PR, on the "Resolves" statement.

cam72cam: Ah, I had missed that.  I withdraw my objections :+1:

ystkfujii:
I’ve updated the changelog to link to the issue instead of the PR.

PR #3024, opened 2025-07-14
Smarter approach to .Equals on state objects for unordered lists

### LINKED ISSUE #3015: Drift detection "tofu plan --refresh-only --detailed-exitcode " exits with 2, even when there are no changes
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.10.2
on windows_amd64
+ provider registry.opentofu.org/hashicorp/azurerm v4.35.0
```

```hcl
provider "azurerm" {
  features {}
}

resource "azurerm_resource_group" "rg" {
  name     = "drift-detection-bug-slawek-wtw-rg"
  location = "West Europe"
  tags = {
    Environment = "Testing"
  }
}

resource "azurerm_storage_account" "sa" {
  name                            = "sadriftdetbugslawekwtw"
  resource_group_name             = azurerm_resource_group.rg.name
  location                        = azurerm_resource_group.rg.location
  account_tier                    = "Standard"
  account_replication_type        = "LRS"
  allow_nested_items_to_be_public = false

  tags = {
    Environment = "Testing"
  }
}
```

### Debug Output

https://gist.github.com/slawek-wtw/a96a6a42a60ae2d3c3558b73f499c06c
That's `INFO` log as GitHub Gist have given `500` for `TRACE`, however the bug should be easily reproduceable locally.

### Expected Behavior

The command `tofu plan -refresh-only -detailed-exitcode` should return `0` if the configuration was just deployed with `tofu apply` and there are no changes detected .

### Actual Behavior

When running the command `tofu plan -refresh-only -detailed-exitcode` it returns `2` even though it is ran right after `tofu apply` and there are no changes detected .

### Steps to Reproduce

```powershell
tofu init
tofu apply -auto-approve
tofu plan -refresh-only -detailed-exitcode
echo $LASTEXITCODE #pwsh
echo $? #bash
```

### Additional Context

_No response_

### References

The same issue have been fixed in Terraform `v1.12.0` (May 14, 2025):
https://github.com/hashicorp/terraform/releases/tag/v1.12.0

Original issue:
https://github.com/hashicorp/terraform/issues/36403

cam72cam:
Reflect.DeepEqual cares about what order of lists it is comparing.  For many of the attributes in the state resource objects, we don't care.

An alternative to this PR is to make sure the contents of the list are always sorted.  I believe that this would be more error prone and require a more thorough review of all of the places it is used within the codebase.

Resolves #3015

## Checklist

### Go checklist

 

### DIFF
--- a/internal/states/instance_object_src.go
+++ b/internal/states/instance_object_src.go
@@ -6,6 +6,9 @@
 package states
 
 import (
+	"bytes"
+	"reflect"
+
 	"github.com/zclconf/go-cty/cty"
 	ctyjson "github.com/zclconf/go-cty/cty/json"
 
@@ -69,6 +72,87 @@ type ResourceInstanceObjectSrc struct {
 	Status              ObjectStatus
 	Dependencies        []addrs.ConfigResource
 	CreateBeforeDestroy bool
+}
+
+func equalSlicesIgnoreOrder[S ~[]E, E any](a, b S, fn func(E, E) bool) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	// Not sure if this is the most efficient approach, but it works
+	for _, v := range a {
+		found := false
+		for _, o := range b {
+			if fn(v, o) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+
+	for _, v := range b {
+		found := false
+		for _, o := range a {
+			if fn(v, o) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (os *ResourceInstanceObjectSrc) Equal(other *ResourceInstanceObjectSrc) bool {
+	if os == other {
+		return true
+	}
+	if os == nil || other == nil {
+		return false
+	}
+
+	if os.SchemaVersion != other.SchemaVersion {
+		return false
+	}
+
+	if !bytes.Equal(os.AttrsJSON, other.AttrsJSON) {
+		return false
+	}
+
+	if !reflect.DeepEqual(os.AttrsFlat, other.AttrsFlat) {
+		return false
+	}
+
+	if !equalSlicesIgnoreOrder(os.AttrSensitivePaths, other.AttrSensitivePaths, cty.PathValueMarks.Equal) {
+		return false
+	}
+	if !equalSlicesIgnoreOrder(os.TransientPathValueMarks, other.TransientPathValueMarks, cty.PathValueMarks.Equal) {
+		return false
+	}
+
+	if !bytes.Equal(os.Private, other.Private) {
+		return false
+	}
+
+	if os.Status != other.Status {
+		return false
+	}
+
+	if !equalSlicesIgnoreOrder(os.Dependencies, other.Dependencies, addrs.ConfigResource.Equal) {
+		return false
+	}
+
+	if os.CreateBeforeDestroy != other.CreateBeforeDestroy {
+		return false
+	}
+
+	return true
 }
 
 // Decode unmarshals the raw representation of the object attributes. Pass the
--- a/internal/states/resource.go
+++ b/internal/states/resource.go
@@ -59,6 +59,33 @@ func (rs *Resource) EnsureInstance(key addrs.InstanceKey) *ResourceInstance {
 	return ret
 }
 
+func (rs *Resource) Equal(other *Resource) bool {
+	if rs == other {
+		// Handles both pointers being nil
+		return true
+	}
+	if rs == nil || other == nil {
+		// Handles one pointer being nil
+		return false
+	}
+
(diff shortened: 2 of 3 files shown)


internal/states/instance_object_src.go
```diff
@@ -71,6 +74,87 @@ type ResourceInstanceObjectSrc struct {
...
+		for _, o := range a {
+			if fn(v, o) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
```
Gogotchuri, line 111: Is this second loop necessary?

Each loop checks if every element in one slice has a match in the other, and given every element in `a` has a match in `b`, then by definition, every element in `b` must also have a match in `a`.

Given their lengths are equal, which we check at the top.

internal/states/instance_object_src.go
```diff
@@ -71,6 +74,87 @@ type ResourceInstanceObjectSrc struct {
...
+
+	if !bytes.Equal(os.AttrsJSON, other.AttrsJSON) {
+		return false
+	}
+
+	if !reflect.DeepEqual(os.AttrsFlat, other.AttrsFlat) {
+		return false
+	}
+
+	if !equalSlicesIgnoreOrder(os.AttrSensitivePaths, other.AttrSensitivePaths, cty.PathValueMarks.Equal) {
```
diofeher, line 138: Can we add a comment explaining why we're not considering the order for these two comparisons?

internal/states/instance_object_src.go
```diff
@@ -71,6 +74,87 @@ type ResourceInstanceObjectSrc struct {
...
+		return false
+	}
+	if !equalSlicesIgnoreOrder(os.TransientPathValueMarks, other.TransientPathValueMarks, cty.PathValueMarks.Equal) {
```
diofeher, line 143: Same here

internal/states/instance_object_src.go
cam72cam: This needs to be a comparison without taking order into account, per the original issue.

Edit: I misunderstood your question.  Commented below about duplicate elements.

diofeher: I was about to comment the same, Ilia. I'm wondering if the second loop is necessary and if could be simplified in one operation.

cam72cam: This takes into account duplicate elements within each list.

cam72cam: I'll make sure I add some clarifying comments :+1:

internal/states/instance_object_src.go
cam72cam: b8d5fabc37

internal/states/instance_object_src.go
Gogotchuri: I don't think this makes a difference. Given that their length are equal, you are checking that every element in a has a corresponding element in b and their lengths are equal. The two loops do the same thing, only instead of operational time (ab)^2, we have 2(ab)^2.

internal/states/instance_object_src.go
cam72cam: b8d5fabc37

internal/states/instance_object_src.go
cam72cam: b8d5fabc37

Gogotchuri: Ah, duplicate case, I didn't think that was possible in there. Let me do a little testing.

internal/states/instance_object_src.go
```diff
@@ -93,6 +95,8 @@ func equalSlicesIgnoreOrder[S ~[]E, E any](a, b S, fn func(E, E) bool) bool {
 		}
 	}
 
+	// Now check if all elements in b exist in a
+	// This is nessesary just in case there are duplicate entries (there should not be).
```
diofeher: typo: necessary

internal/states/instance_object_src.go
diofeher: Looking a lot better, thanks!

internal/states/instance_object_src.go
cam72cam: @Gogotchuri consider a list of:
[1,1,2,3] compared to [1, 2, 3, 4]
If I omit the second loop, this will be considered equal.

internal/states/instance_object_src.go
cam72cam: Fixed in a2c37159a5

internal/states/instance_object_src.go
Gogotchuri: Makes sense. Sorry for the confusion.
The following case fails, though - [1,2,2] compared to [2,1,1], right?
Will return true when they are actually different. Less likely to happen, but possible.

cam72cam: For our purposes here, we don't care about order or duplicates.
[1,2,2] ~== [2,1,1]

Gogotchuri: All my concerns are resolved.
Thanks for the patience.

PR #3025, opened 2025-07-14
refactor: separate entities from views/json into a new package

diofeher:
Relates to #3009

This refactoring is needed in order to remove import cycles between `jsonformat` and `views/json`.

I'm still not sure about the name of the new package, though. I've used `jsonentities` but I would be happy for hear from the other maintainers what would be an appropriate name.

## Checklist

### Go checklist

 


apparentlymart (approved):
The number of packages matching `internal/command/json*` does seem to be getting a little out of hand 😀 and so maybe we'll want to take a more comprehensive look at the organization of these at some point, but this seems like a reasonable incremental step in the meantime just to make it easier to share more code between these packages.

diofeher:
@apparentlymart Agreed! While I'm doing this task, I'll try to think more on what would be a good way to organize these packages. Thanks!

PR #3030, opened 2025-07-15
Diffing lists containing multiline strings

### LINKED ISSUE #2887: Diffing of lists of strings is not useful enough
### Community note

> [!TIP]

### OpenTofu Version

```shell
v1.11.0
```

We are used to a behaviour from other implementations that lists of multi-line strings are diffed in the plan output. In Opentofu this doesn't happen:

Image

Desired output, can be achieved by diffing list items each, one-by-one:

Image

### Attempted Solutions

Use other implementations, it works there.

### Proposal

Implement a similar solution.

Use Terraform from after the fork.

### References

_No response_

Gogotchuri:
After trying several approaches that were overkill for this case, I found an existing working code path for diffing complex slice elements, specifically objects. I was able to tap into the same code with minimal changes to other parts of the code and reuse it for multiline strings.

This PR also includes small refactorings of existing flows to make it more readable and adds comments to explain why/how those specific parts work, which took me time to understand, hopefully making it easier for future contributors working on this code. 

Resolves #2887

## Checklist

- x ] I (and other contributors to this PR) have not looked at the Terraform source code while implementing this PR.

### Go checklist

 

### DIFF
--- a/internal/command/jsonformat/jsondiff/diff.go
+++ b/internal/command/jsonformat/jsondiff/diff.go
@@ -6,10 +6,8 @@
 package jsondiff
 
 import (
-	"reflect"
-	"strings"
-
 	"github.com/zclconf/go-cty/cty"
+	"reflect"
 
 	"github.com/opentofu/opentofu/internal/command/jsonformat/collections"
 	"github.com/opentofu/opentofu/internal/command/jsonformat/computed"
@@ -120,26 +118,7 @@ func (opts JsonOpts) processArray(change structured.ChangeSlice) computed.Diff {
 
 	// This callback is used to determine if we should diff the elements in the slices instead of marking them as deleted and created.
 	shouldDiffElement := func(a, b interface{}) bool {
-		isMultilineString := false
-		tA, tB := GetType(a), GetType(b)
-		// If the values are both string, we check if one of them contains newlines to determine if we should diff them,
-		if tA == String && tB == String {
-			isMultilineString = strings.Contains(a.(string), "\n") || strings.Contains(b.(string), "\n")
-		}
-		// Only in case the strings have a common line we should diff them,
-		if isMultilineString {
-			linesA := strings.Split(a.(string), "\n")
-			linesB := strings.Split(b.(string), "\n")
-			for _, line := range linesA {
-				for _, lineB := range linesB {
-					if line == lineB {
-						return true
-					}
-				}
-			}
-		}
-		// If we haven't retuned at this point, we don't diff the strings and only care if both values are objects.
-		return tA == Object && tB == Object
+		return (GetType(a) == Object && GetType(b) == Object) || ShouldDiffMultilineStrings(a, b)
 	}
 
 	return opts.Array(collections.TransformSlice(change.Before, change.After, processIndices, shouldDiffElement))
--- a/internal/command/jsonformat/jsondiff/diff.go
+++ b/internal/command/jsonformat/jsondiff/diff.go
@@ -7,6 +7,7 @@ package jsondiff
 
 import (
 	"reflect"
+	"strings"
 
 	"github.com/zclconf/go-cty/cty"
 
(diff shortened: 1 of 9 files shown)


internal/command/jsonformat/jsondiff/diff.go
```diff
@@ -117,11 +118,31 @@ func (opts JsonOpts) processArray(change structured.ChangeSlice) computed.Diff {
...
+		// If the values are both string, we check if one of them contains newlines to determine if we should diff them,
+		if tA == String && tB == String {
+			isMultilineString = strings.Contains(a.(string), "\n") || strings.Contains(b.(string), "\n")
+		}
+		// Only in case the strings have a common line we should diff them,
+		if isMultilineString {
+			linesA := strings.Split(a.(string), "\n")
+			linesB := strings.Split(b.(string), "\n")
+			for _, line := range linesA {
+				for _, lineB := range linesB {
```
diofeher: I think we can simplify this nested loop with 

```
bStr := b.(string)
for _, line := range linesA {
   if ok := strings.Contains(bStr, line) {
     return true
   }
}
```

WDYT?

Gogotchuri: The idea is good, but prone to failure. This might test part of the line inclusion, for example:
```tf
# a
<<-EOT
line
another one
EOT

# b
<<-EOT
this is a line
nothing to do with a
EOT
```
Your version will return true for those two strings, because "line" (the first line of a) is included in string b.
When in fact those do not share a common line, hence diffing them is pointless.

diofeher: That's a good point! I commented with strings.Contains but I was thinking on the line of using a regex with optional new lines on the beginning and on the end... But I guess the performance hit here would be comparable and at the same time the code would not be more readable. Thanks for bringing the extra context to what I've missed in my comment :D

internal/command/jsonformat/jsondiff/types.go
```diff
@@ -41,3 +42,29 @@ func GetType(val interface{}) Type {
...
+// This function is used to determine if we should diff the elements in the slices instead of marking them as deleted and created.
+// Primarily used for collections.ShouldDiffElement callback.
+func ShouldDiffMultilineStrings(a, b any) bool {
+	isMultilineString := false
+	tA, tB := GetType(a), GetType(b)
+	// If the values are both string, we check if one of them contains newlines to determine if we should diff them,
+	if tA == String && tB == String {
+		isMultilineString = strings.Contains(a.(string), "\n") || strings.Contains(b.(string), "\n")
+	}
+	// Only in case the strings have a common line we should diff them,
```
cam72cam, line 57: I'm curious about this assertion.  Could you provide an example of where you think this additional check would be useful?

I also don't see a test case for this, but could have missed something.

Gogotchuri: I'm not sure which of the checks you're referring to, so I'll explain everything here.
- Type checks are necessary to avoid possible panic during casting, not likely currently, but a good safety measure for the future.
- At least one of them should contain a multiline string; otherwise, diffing them doesn't provide any additional context, other than deletion and creation, which we already do.
- We should only diff multiline strings that have at least a single common line; otherwise, we aren't providing any additional context or clarity by diffing those strings.
- Additionally, if we do not check for a common line, we might mistake an earlier line deletion for a line update and diff with the incorrect element. This is being tested in a few test cases, where deletion of the line happens before the update. For instance, `multiline_string_list_update_2` in `differ/differ_test.go`: without this check, the second element of the before array, "deleted line", would be diffed with the second element of the after array and would assume that "deleted line" was being updated with "line1\nline2+", instead of marking it as deleted.
I hope this clears things up.

cam72cam: Ah, sorry!  I should have specified!

I was curious about the common line logic.

cam72cam: That makes sense, thanks for explaining!

towolf:
Thank you very much! :bow:

PR #3032, opened 2025-07-16
Destroy nodes with dependencies on ephemeral resources

yottta:
Part of #2834

_same comment with the one in the commit_

During testing references of an ephemeral resource in a provider block
configuration, we found out that the ephemeral resources were not
planned to be opened in order to provide the relevant information to
destroy the second provider related resources.

Consider the following config:
```
provider "testprovider" {
  alias = "prov1"
}

ephemeral "testprovider_ephemeral" "res" {
  provider = testprovider.prov1
  ...
}

provider "testprovider" {
  alias = "prov2"
  cfg = ephemeral.testprovider_ephemeral.res // <- HERE
}

resource "testprovider_resource" "res2" {
  provider = testprovider.prov2
  ...
}
```

As can be seen, during destroy, in order to destroy "res2", we need to
have configured the testprovider.prov2 with the value coming from an
ephemeral resource.
In order to do so, the following has been done in this commit:
* During walkPlanDestroy, in the `PlanGraphBuilder`, we don't skip the
  `ConfigTransformer` entirely now. Instead, we want the transformer to add the
  ephemeral resource into the graph. This is needed because otherwise
  ephemeral resources cannot get into the graph since they have no state
  saved.
* In `pruneUnusedNodesTransformer`, we added a new case where we are not
  pruning the edge between a resource and a provider node to allow the
  ephemeral resource to be opened.
* After the steps above were done, we hit another issue: in
  `DestroyEdgeTransformer`, the ephemeral resources were considered as
  `creators`, ending up in creating destroy nodes dependencies on the
  ephemeral nodes. This ended up in graph cycles. Logically speaking, the
  ephemeral resource opening must have no dependencies on nodes
  destruction, only on opening, reading or other types of actions.
  Therefore, the fix was quite simple, just exclude resources with Open
  action from being labeled as `creators`.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/e2etest/testdata/ephemeral-workflow/proto5/main.tf
+++ b/internal/command/e2etest/testdata/ephemeral-workflow/proto5/main.tf
@@ -47,13 +47,13 @@ provider "simple" {
   // This is needed in two cases: during plan/apply and also during destroy.
   // This test has been updated when DestroyEdgeTransformer was updated to
   // not create dependencies between ephemeral resources and the destroy nodes.
-  // The "cfg" field is just a simple configuration attribute of the provider
+  // The "i_depend_on" field is just a simple configuration attribute of the provider
   // to allow creation of dependencies between a resources from a previously
   // initialized provider and the provider that is configured here.
-  // The "cfg" field is having no functionality behind, in the provider context,
+  // The "i_depend_on" field is having no functionality behind, in the provider context,
   // but it's just a way for the "provider" block to create depedencies
   // to other blocks.
-  cfg = local.simple_provider_cfg
+  i_depend_on = local.simple_provider_cfg
 }
 
 resource "simple_resource" "test_res_second_provider" {
--- a/internal/command/e2etest/testdata/ephemeral-workflow/proto5/main.tf
+++ b/internal/command/e2etest/testdata/ephemeral-workflow/proto5/main.tf
@@ -2,25 +2,29 @@
 // test binaries instead of reaching out to the registry.
 terraform {
   required_providers {
-    simple5 = {
+    simple = {
       source = "registry.opentofu.org/hashicorp/simple"
     }
   }
 }
 
+provider "simple" {
+  alias = "s1"
+}
+
 data "simple_resource" "test_data1" {
-  provider = simple5
+  provider = simple.s1
   value = "initial data value"
 }
 
 ephemeral "simple_resource" "test_ephemeral" {
   count = 2
-  provider = simple5
+  provider = simple.s1
   value = "${data.simple_resource.test_data1.value}-with-renew"
 }
 
 resource "simple_resource" "test_res" {
-  provider = simple5
+  provider = simple.s1
   // NOTE this is wrongly configured on purpose to force a revisit of the test once ephemeral marks are implemented.
   // Once write only arguments are also implemented, adjust the implementation of the provider to support that too
   // and use that new field instead.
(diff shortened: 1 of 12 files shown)


internal/provider-simple-v6/provider.go
```diff
@@ -101,6 +110,7 @@ func (s simple) UpgradeResourceState(_ context.Context, req providers.UpgradeRes
 }
 
 func (s simple) ConfigureProvider(context.Context, providers.ConfigureProviderRequest) (resp providers.ConfigureProviderResponse) {
+	fmt.Println("CONFIGURE SIMPLE6 PROVIDER")
```
cam72cam: debug :)

cam72cam:
This looks great, love all the additional trace logging.

cam72cam (approved):

internal/provider-simple-v6/provider.go
yottta: Removed in b5c640c1a764934ffd53045afeaaf7208202d680. Good catch.

internal/command/e2etest/testdata/ephemeral-workflow/proto5/main.tf
```diff
@@ -2,37 +2,59 @@
...
+  simple_provider_cfg = ephemeral.simple_resource.test_ephemeral[0].value
+}
+
+provider "simple" {
+  alias = "s2"
+  // NOTE: Ensure that ephemeral values can be used to configure a provider.
+  // This is needed in two cases: during plan/apply and also during destroy.
+  // This test has been updated when DestroyEdgeTransformer was updated to
+  // not create dependencies between ephemeral resources and the destroy nodes.
+  cfg = local.simple_provider_cfg
```
diofeher: I'm not sure if `cfg` is a good name for using it here, at least for me it wasn't clear enough on its usage. You mention about "ephemeral values can be used to configure a provider" but what is this configuration about? Is just to create some dependency?

Maybe I'm missing the context from the other PRs or I just should dig enough to understand it, but that was my first thought when I saw this field...

yottta: Good point! Thanks for pointing it out. Yes, this is just a random, with no functionality configuration field that is used just to see how OpenTofu is generating the nodes and edges between a resource and a provider node.
I will add some more comments around this.
Thanks

yottta: Added in 53f949d473d9ea8caa66b2486e1658fb4b19c504.
Please let me know if it makes things clearer. I greatly appreciate your feedback, because I am knees deep in this and hard to figure out what context is missing for those that are not in touch with the evolution of this implementation.

diofeher: I think it makes clearer, indeed, thanks!

Did you think about "depends_on" since it's more descriptive to what is being used? I'm not sure too of this name because we use it in other contexts, but I think something like that would be more descriptive about its usage itself

diofeher: And thanks for digging in this project, I know you're doing a lot of hard stuff there :D Big projects are complex

yottta: You mean as a name of the attribute in the `provider` block? Instead of `cfg`? If yes, TBH, I feel that it would create even more confusion since it's conflicting with the already existing functionality of `depends_on` that it's available in resource/data blocks.
But correct me if I got this wrong.

diofeher: Yea, I agree with you, I'm not sure about `depends_on` too. But I think a name meaning dependency would be better in this case

yottta: Naming is hard man! 😅 
Renamed this in f48dff18b8f27c1beffd95962b162c4a5ba13f3b. Not happy with it, but if you can suggest a better naming, we can create a new PR to do so.

PR #3049, opened 2025-07-21
Ephemeral mark and ephemeral resources refs

yottta:
Part of #2834

The main goal of this PR is to add the ephemeral mark and ensure that the values evaluated from expressions pointing to ephemeral resources are marked with the new mark.
To be able to test the ephemeral resources marks, some other blocks had to get their ephemeral mark too:
* provider
* connection
* provisioner (remote and file since local had been tackled previously)
* output
  * here only partial work has been done in this PR. More will be done in another PR, specifically for outputs 
  * moved the output validations from config to graph node
* variables
  * here only partial work has been done in this PR. More will be done in another PR, specifically for variables

The approach took here was to add a new `Ephemeral bool` in the `*configschema.Block` that is true only for the contexts mentioned above.
This way, inside evalCtx, when a block is evaluated, we can validate its schema allow to contain ephemeral values.
[Here](https://github.com/opentofu/opentofu/pull/3006/commits/8acb084af3bdb4227e77fb00c2fc72150c29a90a) is where the validation, the `Ephemeral bool` and some related changes had been added.

Because of the way we generate the ephemeral errors in the evalCtx, some existing evaluation diagnostics now are returned after calling [InConfigBody](https://github.com/opentofu/opentofu/pull/3006/commits/ec35ffd9beebb363a7c5426efc470f0ffd3e62e3).

## Checklist

### Go checklist

 


internal/configs/configschema/path.go
```diff
@@ -58,3 +58,19 @@ func (o *Object) AttributeByPath(path cty.Path) *Attribute {
 	}
 	return nil
 }
+
+// PathName is trying to get the name of the attribute from cty.Path.
+// Due to the implementation difference between cty.IndexStep and cty.GetAttrStep,
+// this method is working only with a path where the last element is a cty.GetAttrStep.
+// TODO ephemeral - is there a better way to do this?
+func PathName(path cty.Path) string {
```
cam72cam, line 68: I'm having trouble squaring the description of this function with the implementation?

It looks like it returns the first GetAttrStep in a path sequence?  Is it supposed to have a break?

yottta: Good catch! Fixed in 73f00b8193b269081d404b96d96f3ad0545d1047

internal/tofu/eval_variable.go
```diff
@@ -54,6 +54,25 @@ func prepareFinalInputVariableValue(addr addrs.AbsInputVariableInstance, raw *In
 		}
 	}
 
+	if raw.Value.HasMark(marks.Ephemeral) && !cfg.Ephemeral {
```
cam72cam: What if only a nested attribute contains the ephemeral mark?

internal/tofu/eval_variable.go
```diff
@@ -54,6 +54,25 @@ func prepareFinalInputVariableValue(addr addrs.AbsInputVariableInstance, raw *In
...
+		// For child modules variables, this logic is unnecessary since those variables
+		// do always have a SourceRange defined.
+		// We generate subj this way because of the root module variables. In many cases,
+		// the SourceRange can be missing for root module variables.
+		subj := cfg.DeclRange
+		if raw.HasSourceRange() {
+			subj = raw.SourceRange.ToHCL()
+		}
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
```
cam72cam, line 68: Can we include the variable name in either the summary or detail?

internal/tofu/evaluate.go
```diff
@@ -816,6 +835,15 @@ func (d *evaluationStateData) GetResource(ctx context.Context, addr addrs.Resour
...
 				afterMarks = combinePathValueMarks(afterMarks, schemaMarks)
 			}
+			// For ephemeral marks, we don't need to check recursively since this type of mark can be used
+			// only in other ephemeral blocks/contexts.
+			// Therefore, it cannot be used only in specific attributes of a non-ephemeral block, but only directly
+			// in blocks that are marked as ephemeral.
+			if schema.Ephemeral {
+				// When the evaluated block is ephemeral, we want to mark its value
+				// as ephemeral too, to be able to validate later where it's referenced.
+				afterMarks = combinePathValueMarks(afterMarks, []cty.PathValueMarks{{Path: make(cty.Path, 0), Marks: cty.NewValueMarks(marks.Ephemeral)}})
```
cam72cam: How does this work with nested paths?  ephemeral.foo.bar.baz?  

Should this concern be moved into schema.ValueMarks() as that already has a clear understanding of these types of rules for Senstive?

internal/tofu/eval_variable.go
yottta: Fixed in 3ffcecc6f4bf3acc7032cd65f23158f88f9c00e4

internal/tofu/eval_variable.go
yottta: Thanks. Really good point! Fixed in 3ffcecc6f4bf3acc7032cd65f23158f88f9c00e4.
This is part of the variables referencing and marking which is not complete yet, but this is reducing from the work that I will put in that particular PR later.

internal/tofu/node_output.go
```diff
@@ -387,6 +392,14 @@ If you do intend to export this data, annotate the output value as sensitive by
 					Subject: n.Config.DeclRange.Ptr(),
 				})
 			}
+			if n.Config.Ephemeral {
+				diags = diags.Append(&hcl.Diagnostic{
+					Severity: hcl.DiagError,
+					Summary:  "Invalid output configuration",
+					Detail:   "Root modules are not allowed to have outputs defined as ephemeral",
```
diofeher, line 399: Just to make sure I understood correctly: why do we accept output as ephemerals in modules but not on root modules? I assume that output values are written on the state file, so I would like to understand why would we accept on modules?

OR

We do not accept on both?

cam72cam: Output values are only kept in the serialized state file on the root module.  Non-root variables may store the values in the state objects in memory, but they are not persisted.

diofeher: Interesting, thanks for the explanation @cam72cam !

internal/tofu/evaluate.go
yottta: After several tests around this, it works well with inner objects and attributes.
But found a behaviour that was wrong (not disruptive, but wrong), where in some cases (an ephemeral block was referencing an ephemeral attribute from another resource), the evaluated block contained marks on multiple levels, where it's designed to keep the marks only at the root of the value and not on its attributes.
Changed in ce7a436ca22d5a7c9039b8612989e7c7898c6302

internal/tofu/evaluate.go
```diff
@@ -810,12 +829,31 @@ func (d *evaluationStateData) GetResource(ctx context.Context, addr addrs.Resour
 			}
 
 			afterMarks := change.AfterValMarks
+			// For ephemeral marks, we don't need to check recursively since this type of mark is meant
```
cam72cam: Duplicated comment?

yottta: Rebased in 4684cf38f4538b89d7eb6b90c108e13813b9e4f3. Thanks.

cam72cam (approved):

PR #3064, opened 2025-07-23
docs: consolidate warnings are enabled by default

diofeher:
While testing #2751 , I've noticed the docs on our website are wrong for `-consolidate-warnings`. It's enabled by default so to *don't* see consolidated warnings, you have to pass this attribute as false.

This should be backported to previous versions, too.

Wrong pages:
https://opentofu.org/docs/cli/commands/plan/#other-options
https://opentofu.org/docs/cli/commands/apply/#other-options

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/apply.go
+++ b/internal/command/apply.go
@@ -350,64 +350,64 @@ Usage: tofu [global options] apply [options] [PLAN]
 
 Options:
 
-  -auto-approve          Skip interactive approval of plan before applying.
+  -auto-approve                Skip interactive approval of plan before applying.
 
-  -backup=path           Path to backup the existing state file before
-                         modifying. Defaults to the "-state-out" path with
-                         ".backup" extension. Set to "-" to disable backup.
+  -backup=path                 Path to backup the existing state file before
+                               modifying. Defaults to the "-state-out" path with
+                               ".backup" extension. Set to "-" to disable backup.
 
-  -compact-warnings      If OpenTofu produces any warnings that are not
-                         accompanied by errors, show them in a more compact
-                         form that includes only the summary messages.
+  -compact-warnings            If OpenTofu produces any warnings that are not
+                               accompanied by errors, show them in a more compact
+                               form that includes only the summary messages.
 
   -consolidate-warnings=false  If OpenTofu produces any warnings, no consolidation
                                will be performed. All locations, for all warnings
                                will be listed. Enabled by default.
 
-  -consolidate-errors    If OpenTofu produces any errors, no consolidation
-                         will be performed. All locations, for all errors
-                         will be listed. Disabled by default
+  -consolidate-errors          If OpenTofu produces any errors, no consolidation
+                               will be performed. All locations, for all errors
+                               will be listed. Disabled by default.
 
-  -destroy               Destroy OpenTofu-managed infrastructure.
-                         The command "tofu destroy" is a convenience alias
-                         for this option.
+  -destroy                     Destroy OpenTofu-managed infrastructure.
+                               The command "tofu destroy" is a convenience alias
+                               for this option.
 
-  -lock=false            Don't hold a state lock during the operation. This is
-                         dangerous if others might concurrently run commands
-                         against the same workspace.
+  -lock=false                  Don't hold a state lock during the operation.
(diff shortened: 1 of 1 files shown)


website/docs/cli/commands/apply.mdx
```diff
@@ -57,9 +57,9 @@ The following options change how the apply command executes and reports on the a
   at least one error and thus the warning text might be useful context for
   the errors.
 
-- `-consolidate-warnings` - If OpenTofu produces any warnings, do not attempt
-  to consolidate similar messages. All locations for all warnings will be
-  listed.
+- `-consolidate-warnings` - If OpenTofu produces any warnings, attempt
+  to consolidate similar messages. It's enabled by default, so to see
+  locations for all warnings, use `-consolidate-warnings=false`.
```
apparentlymart: For the other "weird" options that don't do anything unless you write `=false` after them we've just included that as part of the option at the start of the item, like `-consolidate-warnings=false`. (For example, see `-input=false` and `-lock=false` below.

I'd suggest that we follow the same principle here so that we're always showing exactly what the user would need to type to get the effect described, and then making sure the text talks about what would happen if the user typed that.

For example:

> - `-consolidate-warnings=false` - Disable OpenTofu's default behavior of attempting to consolidate similar warning messages into a single item.

(I noticed also that our `tofu apply -help` output has a similar problem. I previously fixed this for `tofu plan -help` in 1260f2218c8afd2e2abe93fe99c0b9b11a2275c6 because I was already changing other parts of that output anyway, but I apparently didn't make the same change for `tofu apply -help` and so they are now inconsistent. 😖 )

diofeher: Oh, I didn't notice that, thanks for the great catch, @apparentlymart ! Gonna change it

diofeher: Regarding https://github.com/opentofu/opentofu/commit/1260f2218c8afd2e2abe93fe99c0b9b11a2275c6,
do we really need to change the identation of all the fields? Because when I was testing here, it seems they're auto indented on the `-help` output.

Screenshot 2025-07-23 at 13 23 33

(I didn't indent the other fields)

apparentlymart: As far as I know there is nothing transforming the help text before it gets written to the output stream.

The library we use to implement the command line handling [uses a Go template](https://github.com/mitchellh/cli/blob/e73bd003a976863797bccc7d5630e83969f20bb8/cli.go#L735-L742) to generate that output. The help text string we provide in the OpenTofu code [gets placed direcly into that `.Help` symbol](https://github.com/mitchellh/cli/blob/e73bd003a976863797bccc7d5630e83969f20bb8/cli.go#L527), so I can't find any code that would automatically adjust the widths.

(We have talked a few times about switching to a different UI library that works at a higher level of abstraction and can generate these helptext strings automatically itself, but that's a big and risky effort so we've not yet been able to prioritize doing it, and so we continue maintaining these helptext strings manually for now. :confounded: )

FWIW I don't think you necessarily need to change the help text output as part of this PR. I just mentioned it because I noticed it while I was checking to see how we described this option in some other places, but it seems fine for this PR to stay focused only on the website documentation and we can deal with the usage text separately some other time.

diofeher: Man, you're right. I just noticed that I changed the string for`apply` but I was checking the `-help` for `plan`. Sorry about that 😅

I'm fine on putting together on this PR, since for me it's about showing the right information related to `consolidate-warnings` field. Thanks again!

diofeher: Fixed on https://github.com/opentofu/opentofu/pull/3064/commits/37aab23a4a35ec846d969d5152fd6e75680dcd25

apparentlymart (approved):

website/docs/cli/commands/apply.mdx
yottta: Adding some past conversations around this:
* https://github.com/opentofu/opentofu/pull/2705#discussion_r2076061854
* https://github.com/opentofu/opentofu/pull/2705#discussion_r2073696918

PR #3066, opened 2025-07-23
RFC: conditional  `enabled` field

diofeher:
Relates to #1306 

To read the RFC: https://github.com/opentofu/opentofu/blob/rfc-direct-conditional-support/rfc/2025-07-22-conditional-enable.md

This RFC aims to consolidate all design decisions regarding an `enabled` field in single-instance resources and modules. A prototype with most of these decisions is implemented at #3042.

## Checklist

### Go checklist

 

### DIFF
--- a/rfc/2025-07-22-conditional-enable.md
+++ b/rfc/2025-07-22-conditional-enable.md
@@ -51,7 +51,8 @@ module "modcall" {
 1. Support for conditional enabling on:
     1. Resources - Add a new field on the `lifecycle` block called `enabled`;
     2. Modules - Add a new `lifecycle` block for modules but only supporting the new `enabled` field.
-2. Raise errors if a resource is used while being disabled and the dependent resources are not disabled too;
+2. By default, this field is enabled if not otherwise specified.
+3. Raise errors if a resource is used while being disabled and the dependent resources are not disabled too;
 
 ## Technical Approach
 
@@ -87,7 +88,8 @@ OpenTofu will perform the following actions:
 Plan: 0 to add, 0 to change, 0 to destroy.
 ```
 
-This behavior applies only to resources. For modules, you must write an explicit `move` block:
+This behavior only exists for resources at the moment. As part of the RFC, support will be written for implied moves within modules.
+Until then, this can be done with:
 
 ```
 moved {
@@ -165,7 +167,7 @@ The only available data access patterns for possibly disabled resources are:
 This can be extended using custom provider-defined functions. For example, if someone wanted to create a function that adds a warning instead of raising an error when trying to access the attribute,
 they could do:
 
-- `provider::custom::warning_enabled(null_resource.example.id, "default value")
+- `provider::custom::warning_enabled(null_resource.example.id, "default value")`
 
 For now, we offer the three options above. 
 `try` is the most concise option, but it should be used carefully since it can silently mask errors not caused by null value access. `can` and `!= null` are more verbose but more reliable for doing what you're trying to do.


rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,130 @@
...
+
+If a resource is disabled, it shouldn't be possible to access the attributes since the resource
+is not on the graph. We prefer to loudly tell the user they're trying to access something that is not
+available. In order to have a good user experience while they want to enable/disable stuff, we
+offer a way to silently fail while trying to access that, by using the `try` function.
+
+There's some discussion about this [topic](https://github.com/opentofu/opentofu/issues/1306#issuecomment-2398120132), and the better compromise we found is that a enabled 
+resource can be `null` when it's disabled, but no errors are going to be shown if they use the
+resource as `enabled=true` first, since we need to evaluate the `enabled` expression. In the opposite side,
+if the resource is `disabled`, and a reference is trying to access it, an error is going to be shown:
```
Gogotchuri: I have been researching this issue, and I share the same concern as @apparentlymart.
We would have no static checking capabilities if we implemented it this way. Hence, until the user sets `enabled=false`, we won't be able to check if he handled things correctly down the line.
I think having either a mandatory nullness check or `try` on the resource with an `enabled` field in the `lifecycle` block, if possible, would resolve most of this issue. I haven't tried implementing this, and I'm unsure what challenges it poses.

diofeher: Agreed! I have been researching since yesterday on how to evaluate these configs dynamically down the line and it's not straightforward. Having a mandatory nullness check is a great idea!

The main problem is that when we check the references later on the graph, we have access only to the dependencies using this struct here:

https://github.com/opentofu/opentofu/blob/72f4ed2acb555d3e1553b2ec2c9ee7831f2be872/internal/addrs/resource.go#L291-L295

They don't contain any references to the configuration, basically only to the address. We only have access to the node config when we're evaluating the node itself, but not from other nodes that are referencing it:

https://github.com/opentofu/opentofu/blob/5f67fb44674562ed4ab987074d165a5a309ca7b9/internal/tofu/node_resource_abstract.go#L519-L524

Usually we don't have the root config available on these evaluators, and if we want to have it, maybe we would be adding an unwanted dependency. At the same time, we don't have the information if they are using the `enabled` field at all, so I'm trying to find the proper place to where pass this information.

diofeher: I was talking to @apparentlymart and it seems there's no way to do that, both for evaluating the expression or just checking if the references have the `enabled` field. I'm going to revert this specific part of the RFC saying that people should do the checks themselves.

diofeher:
I'm reverting back this RFC to draft since creating the data access patterns for these resources would be trickier than I thought. I'm still experimenting in the other branch #3042. 

Until I can find a better compromise on what we can achieve, I'm gonna update it here.

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,130 @@
...
+    enabled = var.env == "production"
+  }
+}
+```
+
+## Proposed Solution
+
+1. Raise errors if a resource is used while being disabled;
+1. Support for conditional enabling on:
+  1. Resources - Add a new field on the `lifecycle` block called `enabled`;
```
tmccombs: Why not data sources?

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,130 @@
...
+
+  # null_resource.example[0] has moved to null_resource.example
+    resource "null_resource" "example" {
+        id = "8975736315378968412"
+    }
+
+Plan: 0 to add, 0 to change, 0 to destroy.
+```
+
+### Usage of `enabled` together with `for_each` and `count`.
```
tmccombs: I think it would be reasonable to have both, if the enabled flag has a higher priority than count or for_each. That is if it is disabled, it isn't created regardless of the value of count or for_each.

rfc/2025-07-22-conditional-enable.md
apparentlymart: `resource`, `data`, and `ephemeral` blocks all declare "resources" really. It's just an annoyance of history that the first kind of resource -- managed resources -- squatted on the generic block type name before anyone realized that there would be resources of other kinds. 

So this item is talking about resources of all kinds, just as with `count` and `for_each`. The same code handles all three when it comes to these meta-arguments.

(I don't intend this comment to mean that it would not still be worth changing the RFC texr to make this clearer. I suggest we directly name all of the block types that this new argument can appear in, to make it precise: `resource`, `data`, `ephemeral` and `module`, instead of just talking about the general concepts.)

rfc/2025-07-22-conditional-enable.md
apparentlymart: If we allow both then we need to decide what both of them together does, mechanically.

One answer is that combining `for_each` with `enabled = false` causes the same effect as `for_each = {}`, with the resource then appearing as an empty object.

Another answer is that `enabled = false` always just makes the resource be a null value of whatever type it would have been otherwise: alone it's a null object, with `for_each` a null map, and with `count` a null list.

The fact that it's not immediately obvious what it ought to mean gives me pause, though, since it means that everyone who encounters the combination will need to go learn what it means to know what effect it has. A `for_each` with a conditional or `for` expression in it has the advantage that if you already know what each of these three language features does in isolation you can infer how they will interact with each other without any additional information. 🤔

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,130 @@
...
+
+  lifecycle {
+    enabled = var.env == "production"
+  }
+}
+```
+
+## Proposed Solution
+
+1. Raise errors if a resource is used while being disabled;
```
yottta:
```suggestion
1. Raise errors if a resource is used while being disabled and the dependant resources are not disabled too;
```
Right?

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,130 @@
...
+
+Lifecycle block is currently not supported on `Modules`, but in our docs we're mentioning to users that it's reserved for future usage:
+
+> OpenTofu does not use the lifecycle argument. However, the lifecycle block is reserved for future versions.
+
+https://opentofu.org/docs/language/modules/syntax/#meta-arguments
+
+It would be the right timing to start to support it, but only for the `enabled` field.
+
+### Migration from `count` to `enabled`
```
yottta, line 71: I might miss some things here, but shouldn't `for_each` be handled similarly?

yottta (commented):
Neat idea! Looking forward to see it moving forward. 🙌

rfc/2025-07-22-conditional-enable.md
apparentlymart: This section is describing some pre-existing behavior where OpenTofu notices a situation where e.g. the desired state includes `aws_instance.foo` while the previous state includes `aws_instance.foo[0]`, and so it behaves as if the author had written a `moved` block like this without requiring the author to explicitly write it:

```hcl
moved {
  from = aws_instance.foo[0]
  to   = aws_instance.foo
}
```

(there's a similar rule for migrating in the opposite direction, too.)

This rule relies on the assumption that count element zero is the one that should be kept, and all of the others discarded, creating a similar effect to what would happen if you previously had `count = 3` and now you have `count = 1` -- the "zeroth" element would be kept and the others would be planned for destroy.

There is no similar automatic rule for `for_each` because there isn't any obvious rule for choosing which of the possibly-many string instance keys is the one to keep. In that case we require the author to write out an explicit `moved` block describing which instance should be kept:

```hcl
moved {
  from = aws_instance.foo["example"]
  to   = aws_instance.foo
}
```

I'd also note that this "implied `moved` block" behavior only applies to adding/removing `count` from resources. There is no similar migration for module calls because there are many more possible ways to resolve a change like that and so OpenTofu wants the author to be explicit about what they intended.

Maybe we should add some more detail like what I described above to the RFC text so that there's some additional context about what would be handled automatically and what would not be, and we can also mention that `moved` blocks are the intended solution for the more ambiguous cases that OpenTofu doesn't handle automatically.

(In case it's interesting: [the existing code that handles these "implicit moves"](https://github.com/opentofu/opentofu/blob/97ab952a396eb5f6a7cee55aba5463489b64d822/internal/refactoring/move_statement.go#L98-L153).)

yottta: Great! That makes sense!

> Maybe we should add some more detail like what I described above to the RFC text so that there's some additional context about what would be handled automatically and what would not be, and we can also mention that moved blocks are the intended solution for the more ambiguous cases that OpenTofu doesn't handle automatically.

Totally agree with this. Having a section in this RFC handling these cases could prove useful.

rfc/2025-07-22-conditional-enable.md
diofeher: Improved on https://github.com/opentofu/opentofu/pull/3066/commits/120ca027e35846a5c2fc543c8fac592373fda93e

Let me know what you think so I can resolve this conversation.

rfc/2025-07-22-conditional-enable.md
diofeher: Thanks for the question @tmccombs ! 

As @apparentlymart answered, we will be supporting all of them. I've added a note to the of the RFC: https://github.com/opentofu/opentofu/pull/3066/commits/86b2c11dedf685dffef6ee4bb9278708c185c2bf

rfc/2025-07-22-conditional-enable.md
diofeher: I prefer not to have these fields enabled together. I believe it could lead to significant confusion regarding their priority rules. If you want to "disable" a resource while using count, for example, simply set count to 0. This new feature would primarily be used for single-instance resources.

I'll leave this conversation opened though to see what others think.

rfc/2025-07-22-conditional-enable.md
diofeher: @apparentlymart Can you tell me about ambiguous cases where OpenTofu wouldn't be able to handle that besides moving from `count` and `for_each`?

diofeher: Adding a section about `for_each` https://github.com/opentofu/opentofu/pull/3066/commits/fb664c0143ea3a60b94f7682e143725d5d70162e

rfc/2025-07-22-conditional-enable.md
diofeher: You're right! Let me commit that to clarify!

diofeher: Fixed on https://github.com/opentofu/opentofu/pull/3066/commits/7f24cf12363d499bf40a6d2ca9733fd3063c4741

rfc/2025-07-22-conditional-enable.md
apparentlymart: The only case where OpenTofu does anything automatic today is switching from nothing to `count`, or the reverse of that, on resources.

Switching from `enabled = true` to `count` > 0, and the reverse of that, is effectively the same as what currently works since it causes the same instance keys to appear, so I don't think we need any extra code to make that work. 

Switching from nothing to `enabled = true` and the reverse is effectively no change, because those two situations are equivalent. No address changes are required at all.

Everything else is "ambiguous" in the way I meant that, and so is not handled automatically. Adding `enabled` doesn't change anything about the other cases.

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,187 @@
...
+│     │ null_resource.example is null
+│
+│ This value is null, so it does not have any attributes.
+```
+
+The available data access patterns for possibly disabled resources are:
+
+- `null_resource.example != null ? null_resource.example.id : "default value"`
+- `try(null_resource.example.id, "default_value")`
+- `can(null_resource.example.id) ? null_resource.example.id : "default value"`
```
Gogotchuri: To be a bit more specific. Are those the *only* available patterns for accessing resources with the enabled field? Or can those be accessed regularly, and we recommend those patterns to avoid errors in case the resource is disabled?

rfc/2025-07-22-conditional-enable.md
Gogotchuri: I'm resolving this and started a new thread for the latest set of changes.

rfc/2025-07-22-conditional-enable.md
diofeher: What I was doing was looking through the list at https://opentofu.org/docs/language/functions/ and seeing/testing which ones could be used. I couldn't find other patterns besides those three... I'll add more information to this specific point!

apparentlymart: It's true to say that there's nothing special about references to `null_resource.example` beyond the usual rules about what you can and cannot do with null values, but I do think it's worth including some examples of common patterns we're expecting so we can understand whether this design is likely to lead to something intuitive/readable.

When we finally write docs for this I think we should choose just one pattern to recommend, to encourage a consistent idiom, but we can save that decision for later.

diofeher: Good! I've added a few clarifications to it https://github.com/opentofu/opentofu/pull/3066/commits/b065ba86d82411827b166cc33a8fb5eea1f96f86

Gonna resolve it but feel free to reopen/comment if you feel we still need to put more information here.

rfc/2025-07-22-conditional-enable.md
diofeher: Hmm, ok! Do you think the current state of the RFC shows reflects what's needed here?

rfc/2025-07-22-conditional-enable.md
cam72cam: I think forbidding it for now does not mean we could not revisit this conversation again in the future and therefore does not block this RFC

diofeher: Agreed! Let me resolve this conversation so we can move this RFC forward.

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+```
+
+## Proposed Solution
+
+1. Raise errors if a resource is used while being disabled and the dependent resources are not disabled too;
+2. Support for conditional enabling on:
+  1. Resources - Add a new field on the `lifecycle` block called `enabled`;
+  2. Modules - Add a new `lifecycle` block for modules but only supporting the new `enabled` field.
```
yottta: This is not rendered correctly. Could you check this please?
Screenshot 2025-08-20 at 11 36 13

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+
+## Technical Approach
+
+### `lifecycle` support on Modules
+
+The lifecycle block is currently not supported on `modules`, but in our docs we mention to users that it's reserved for future usage:
```
yottta:
```suggestion
The lifecycle block is currently not supported on `module` block, but in our docs we mention to users that it's reserved for future usage:
```

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+    resource "null_resource" "example" {
+        id = "8975736315378968412"
+    }
+
+Plan: 0 to add, 0 to change, 0 to destroy.
+```
+
+### Migration from `for_each` to `enabled`
+
+Unlike `count`, implicit moves are not handled by OpenTofu. This means that the user should explicitly
```
yottta:
```suggestion
Unlike `count`, implicit moves are not handled by OpenTofu when migrating from `for_each`. This means that the user should explicitly
```

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+### What types of variables are supported on the conditional
+
+There are a few types of values that cannot be used in this conditional:
+
+1. Unknown values, since they cannot be used until the apply phase
+2. Sensitive values
+3. Null values
+4. Conditionals that do not evaluate to true/bool values
```
yottta, line 219:
```suggestion
4. Conditionals that do not evaluate to true/bool values
5. Ephemeral values
```

yottta (approved):
Looks good. Just some small nitpicks

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+  lifecycle {
+    enabled = var.env == "production"
+  }
+}
+```
+
+## Proposed Solution
+
+1. Raise errors if a resource is used while being disabled and the dependent resources are not disabled too;
```
Yantrio: Nitpick: it was quite strange for me to read about errors in a proposed solution before the actual proposed soltuion of adding the new property to the lifecycle block.

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Technical Approach
+
+### `lifecycle` support on Modules
+
+The lifecycle block is currently not supported on `modules`, but in our docs we mention to users that it's reserved for future usage:
+
+> OpenTofu does not use the lifecycle argument. However, the lifecycle block is reserved for future versions.
+
+https://opentofu.org/docs/language/modules/syntax/#meta-arguments
```
Yantrio: I'd recommend linking to a hard commit on github to the mdx file for this instead of opentofu.org. As that is going to change over time and we should aim to have this viewable historically

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+It would be the right timing to start supporting it, but only for the `enabled` field.
```
Yantrio:
```suggestion
This proposal could be the right time to start supporting this `lifecycle` block, but only to add the `enabled` field.
```

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Migration from existing resources
+
+### Migration from `count` to `enabled`
+
+As mentioned before, users were adding support for enabling/disabling resources by using `count = var.enabled ? 1 : 0` and this is the most common case where they want to use an `enabled` field.
```
Yantrio: Where was this mentioned before? I know you mentioned `count` in the intro but not a specific example

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+Luckily, this support is [in-place already](https://github.com/opentofu/opentofu/pull/3066#discussion_r2231518392). When using a `count` for this conditional dependency, if
```
Yantrio: >Luckily, this support is [in-place already](https://github.com/opentofu/opentofu/pull/3066#discussion_r2231518392).

I feel like it took me a long time to realise that you were talking about moving resources here! a little bit of a pre-amble to each section would be appreciated to provide context of what we're talking about aside from the title.

For example, I think we should be saying something similar to: 

```
To support moving between resources where the `count = 0` workaround was used to the new `lifecycle -> enabled` approach, opentofu will automatically support an implicit move of the resource out of the box with no changes needed
```

What we have here is very jarring to read because i had no clue what "this support" is discussing and the linked comment was quite confusing.

rfc/2025-07-22-conditional-enable.md
Yantrio: I'd just like to say that I really think we're missing out by not supporting both approaches here. And i think we should add something to the future considerations section about this.

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+```
+
+It doesn't make a lot of sense in the example above to move from `for_each` to `enabled` if more than one resource is being used. `bucket-2` in the example above would be removed. This path of migration
+should be used *only* if a single instance is being used and for_each is supporting that conditional.
+
+### Usage of `enabled` together with `for_each` and `count`.
+
+These three different arguments would behave similarly, but with different semantics.
```
Yantrio: different how?

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+It could be argued that the user wants to enable a resource with 4 instances, but in our current code,
+that conditional can be changed to `count=0`.
```
Yantrio: This makes no sense to me, can you provide some more explanation here or even some examples?

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+The proposal is that we do not support them together. Only one of them can be used, returning errors if we try to use the others together.
```
Yantrio: Please justify why in the RFC

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+Semantically, this feature would be used for single-instance resources or modules.
+
+### Errors when resource is not enabled
+
+There's some discussion about this [topic](https://github.com/opentofu/opentofu/issues/1306#issuecomment-2398120132), and the better compromise we found is that an enabled 
+resource can be `null` when it's disabled, but no errors are going to be shown statically,
+like when using `validate` or the language server, in an OpenTofu extension, like VSCode.
```
Yantrio:
```suggestion
like when using `tofu validate` or the `tofu-ls` language server
```

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+│ This value is null, so it does not have any attributes.
+```
+
+The only available data access patterns for possibly disabled resources are:
+
+- `null_resource.example != null ? null_resource.example.id : "default value"`
+- `try(null_resource.example.id, "default_value")`
+- `can(null_resource.example.id) ? null_resource.example.id : "default value"`
+
+Notice this can be extended by using custom provider functions. Let's say if someone would like to
```
Yantrio:
```suggestion
Notice this can be extended by using custom provider-defined functions. Let's say if someone would like to
```

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+- `provider::custom::warning_enabled(null_resource.example.id, "default value")
+
+For now, we will be offering the three options above. 
+`try` is the most concise option, but it should be used carefully since it can silently mask
+different errors not caused by null values access. `can` and `!= null` are more verbose but 
+are more reliable on doing what you're trying to do.
+
+Discarded options:
+
+- In a GitHub thread, someone mentioned using `?` to deal with disabled resources. Since this would be a HCL change, we discarded that option.
```
Yantrio: Can you clarify why we dont want to introduce a HCL change please?

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,195 @@
...
+
+There are a few types of values that cannot be used in this conditional:
+
+1. Unknown values, since they cannot be used until the apply phase
+2. Sensitive values
+3. Null values
+4. Conditionals that do not evaluate to true/bool values
+
+## Future considerations
+
```
Yantrio, line 223: Please add some considerations around supporting count+enabled/for-each+enabled.

rfc/2025-07-22-conditional-enable.md
apparentlymart: I think we should also note that this automatic behavior exists only for resources, and that for modules (unless we change something as part of implementing this feature) it will be necessary to write an explicit `moved` block:

```hcl
moved {
  from = module.example[0]
  to   = module.example
}
```

The current behavior of automatically generating an implied `moved` block for this situation only works for resources.

rfc/2025-07-22-conditional-enable.md
diofeher: Thanks for this catch and for the other comments! I've put all of them together at https://github.com/opentofu/opentofu/pull/3066/commits/aad14d18b2eff7d45db55ffd41ba77d11be46854

rfc/2025-07-22-conditional-enable.md
diofeher: Makes sense!

rfc/2025-07-22-conditional-enable.md
diofeher: Agreed

rfc/2025-07-22-conditional-enable.md
diofeher: Ops, this phrase is awkward! Gonna rework on it

rfc/2025-07-22-conditional-enable.md
diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R225

rfc/2025-07-22-conditional-enable.md
diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R175

rfc/2025-07-22-conditional-enable.md
diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R11-R22

rfc/2025-07-22-conditional-enable.md
diofeher: Changed the order on https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R54

rfc/2025-07-22-conditional-enable.md
diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R64

rfc/2025-07-22-conditional-enable.md
diofeher: James' comment was addressed on 

https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R75-R77

Martin's comment was addressed on https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R90-R98

rfc/2025-07-22-conditional-enable.md
diofeher: Moved close to the end of the document and reworked the text:

https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R192-R208

rfc/2025-07-22-conditional-enable.md
diofeher: Same here: https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R192-R208

rfc/2025-07-22-conditional-enable.md
diofeher: The "why" can be found here, in this new commit: https://github.com/opentofu/opentofu/pull/3066/commits/8c482a1ba87a0dd109b7b5607f586ca69665bf56#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R205-R208

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,225 @@
...
+  # null_resource.example[0] has moved to null_resource.example
+    resource "null_resource" "example" {
+        id = "8975736315378968412"
+    }
+
+Plan: 0 to add, 0 to change, 0 to destroy.
+```
+
+This behavior applies only to resources. For modules, you must write an explicit `move` block:
```
Yantrio: How do we justify to end users the difference in functionality here. Do you think we should add support to move modules in the same way as resources in this case now?

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,225 @@
...
+The only available data access patterns for possibly disabled resources are:
+
+- `null_resource.example != null ? null_resource.example.id : "default value"`
+- `try(null_resource.example.id, "default_value")`
+- `can(null_resource.example.id) ? null_resource.example.id : "default value"`
+
+This can be extended using custom provider-defined functions. For example, if someone wanted to create a function that adds a warning instead of raising an error when trying to access the attribute,
+they could do:
+
+- `provider::custom::warning_enabled(null_resource.example.id, "default value")
```
Yantrio:
```suggestion
- `provider::custom::warning_enabled(null_resource.example.id, "default value")`
```

rfc/2025-07-22-conditional-enable.md
```diff
@@ -0,0 +1,225 @@
...
+
+  lifecycle {
+    enabled = var.env == "production"
+  }
+}
+```
+
+## Proposed Solution
+
+1. Support for conditional enabling on:
```
Yantrio, line 51: I think we still need to explicitly mention in the RFC that the default for this `enabled` field is `true`

diofeher: Agreed, I missed changing that from the last round of reviews. Good catch and thanks again :D

rfc/2025-07-22-conditional-enable.md
diofeher: Let me investigate how much work would be involved on this one.

rfc/2025-07-22-conditional-enable.md
diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3066/commits/32dd71602608cb8730db2e83e5b1946aef44f3db

rfc/2025-07-22-conditional-enable.md
diofeher: It seems we would need to switch some code in this part here: https://github.com/opentofu/opentofu/blob/97ab952a396eb5f6a7cee55aba5463489b64d822/internal/refactoring/move_statement.go#L98-L153

It seems doable, I'm gonna update the RFC to reflect these changes.

diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3066/commits/3d8ecea7d3e35781cfe0af25b84353e4876b572f#diff-e73ef3cdb70f6c33c181650de9156bbc54d893a5693c864110b2193e89862ee6R91-R92

rfc/2025-07-22-conditional-enable.md
diofeher: Added a future consideration about this

cam72cam (approved):

apparentlymart (approved):

PR #3072, opened 2025-07-24
fix: resolve local mirror detection with -chdir option

### LINKED ISSUE #2531: Implied local mirror does not work with -chdir when using the "launch directory" mirror
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.9.0
on darwin_arm64
```

```hcl
terraform {
  backend "local" {}
  required_providers {
    vault = {
      source  = "hashicorp/vault"
      version = "4.6.0"
    }
  }
}
```

### Debug Output

I will post it here as I think the relevant part is already in a small bit:
```
Initializing provider plugins...
- Finding hashicorp/vault versions matching "4.6.0"...
2025-02-18T18:16:38.904+0100 [TRACE] providercache.fillMetaCache: scanning directory .terraform/providers
2025-02-18T18:16:38.904+0100 [TRACE] getproviders.SearchLocalDirectory: failed to resolve symlinks for .terraform/providers: lstat .terraform/providers: no such file or directory
2025-02-18T18:16:38.904+0100 [TRACE] providercache.fillMetaCache: error while scanning directory .terraform/providers: cannot search .terraform/providers: lstat .terraform/providers: no such file or directory
- Installing hashicorp/vault v4.6.0...
2025-02-18T18:16:38.904+0100 [TRACE] providercache.Dir.InstallPackage: installing registry.opentofu.org/hashicorp/vault v4.6.0 from terraform.d/plugins/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64
2025-02-18T18:16:38.905+0100 [TRACE] providercache.fillMetaCache: scanning directory .terraform/providers
2025-02-18T18:16:38.906+0100 [WARN]  failed to read metadata about .terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64: stat .terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64: no such file or directory
╷
│ Error: Failed to install provider
│ 
│ Error while installing hashicorp/vault v4.6.0: after installing registry.opentofu.org/hashicorp/vault it is still not detected in .terraform/providers; this is a bug in OpenTofu
╵
```

### Expected Behavior

The provider installation should have worked, e.g., by creating the correct link to the launch directory, not the -chdir directory.
You can see that it succeeded properly if you get this warning:
```
╷
│ Warning: Incomplete lock file information for providers
│ 
│ Due to your customized provider installation methods, OpenTofu was forced to calculate lock file checksums
│ locally for the following providers:
│   - hashicorp/vault
│ 
│ The current .terraform.lock.hcl file only includes checksums for darwin_arm64, so OpenTofu running on
│ another platform will fail to install these providers.
│ 
│ To calculate additional checksums for another platform, run:
│   tofu providers lock -platform=linux_amd64
│ (where linux_amd64 is the platform to generate)
╵
```

### Actual Behavior

tofu creates a link at `${LAUNCHDIR}/tofu/.terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64` which points to `${LAUNCHDIR}/tofu/terraform.d/plugins/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64` instead of `${LAUNCHDIR}/terraform.d/plugins/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64` (note the `tofu` from `-chdir` is part of the link).

### Steps to Reproduce

I prepared a few different examples attempting to solve the issue and highlighting it.

## Variant 1: mirror inside launch directory

This first example follows the documentation (https://opentofu.org/docs/cli/config/config-file/#implied-local-mirror-directories) closely:
> If a terraform.d/plugins directory exists in the current working directory then OpenTofu will also include that directory, regardless of your operating system. This behavior changes when you use the -chdir option with the init command. In that case, OpenTofu checks for the terraform.d/plugins directory in the launch directory and not in the directory you specified with -chdir.

The script reproduce.sh does the following:
- We download a provider (hashicorp/vault in this case) and store it in the implied local mirror directory in our current working (= launch) directory.
- Then we cleanup the seed, leaving us with the mirror
- Next, we create a subdirectory `tofu` which contains the same main.tf again and try to run `TF_LOG=trace tofu -chdir=tofu init`
- We observe the error above:
  > ```
  > ╷
  > │ Error: Failed to install provider
  > │ 
  > │ Error while installing hashicorp/vault v4.6.0: after installing registry.opentofu.org/hashicorp/vault it is still not detected in .terraform/providers; this is a bug in OpenTofu
  > ╵

```sh
# download a provider
cat <<-EOF > main.tf
terraform {
  backend "local" {}
  required_providers {
    vault = {
      source  = "hashicorp/vault"
      version = "4.6.0"
    }
  }
}
EOF
tofu init

# move the downloaded provider to the "directory-local mirror"
mkdir -p terraform.d/plugins/registry.opentofu.org/hashicorp
mv .terraform/providers/registry.opentofu.org/hashicorp/vault terraform.d/plugins/registry.opentofu.org/hashicorp

# cleanup seed
rm -r .terraform main.tf .terraform.lock.hcl

# Create subdir with same provider requirement
mkdir -p tofu  # -p for the other examples
cat <<-EOF > tofu/main.tf
terraform {
  backend "local" {}
  required_providers {
    vault = {
      source  = "hashicorp/vault"
      version = "4.6.0"
    }
  }
}
EOF
TF_LOG=TRACE tofu -chdir=tofu init
```

## Variant 2: mirror inside subdir

We can try to move the terraform.d directory into the subdirectory instead and try again (reproduce2.sh), in hopes that the link will now point to the correct location:

```sh
mkdir -p tofu/terraform.d/plugins/registry.opentofu.org/hashicorp
mv .terraform/providers/registry.opentofu.org/hashicorp/vault terraform.d/plugins/registry.opentofu.org/hashicorp
```

This variant does not the local mirror and install the provider from the remote, as can be seen by the trace output:
```
025-02-18T18:32:21.239+0100 [DEBUG] GET https://github.com/opentofu/terraform-provider-vault/releases/download/v4.6.0/terraform-provider-vault_4.6.0_SHA256SUMS
2025-02-18T18:32:21.239+0100 [TRACE] HTTP client GET request to https://github.com/opentofu/terraform-provider-vault/releases/download/v4.6.0/terraform-provider-vault_4.6.0_SHA256SUMS

etc.
```

## Variant 3: mirror duplicate (one to detect, one to link to)

In this variant, we create the mirror twice: once in the launch, once in the subdirectory (reproduce3.sh).

```sh
# duplicate the downloaded provider to the "directory-local mirror"
mkdir -p terraform.d/plugins/registry.opentofu.org/hashicorp
mkdir -p tofu/terraform.d/plugins/registry.opentofu.org/hashicorp
mv .terraform/providers/registry.opentofu.org/hashicorp/vault tofu/terraform.d/plugins/registry.opentofu.org/hashicorp
cp -r tofu/terraform.d/plugins/registry.opentofu.org/hashicorp terraform.d/plugins/registry.opentofu.org/hashicorp
```

I don't know why, but this variant also pulls the provider from the remote.

## Variant 4: run init once, move the mirror, run init again

In this variant, we run the init and let it fail (as in reproduce.sh), then **copy** the terraform.d to tofu/terraform.d and run it again (this time do not replace the mkdir/mv commands but append them, reproduce4.sh):

```sh
# retry with local mirror in subdirectory
cp -r terraform.d tofu/terraform.d
TF_LOG=TRACE tofu -chdir=tofu init
```

This way, the link will already be there and work correctly; and the `terraform.d` inside the launch directory will prevent tofu from replacing the link with freshly downloaded variant.
We can verify this with the following output, which is only possible because tofu did not attempt to download the SHA256SUMS file:
```
╷
│ Warning: Incomplete lock file information for providers
│ 
│ Due to your customized provider installation methods, OpenTofu was forced to calculate lock file checksums
│ locally for the following providers:
│   - hashicorp/vault
│ 
│ The current .terraform.lock.hcl file only includes checksums for darwin_arm64, so OpenTofu running on
│ another platform will fail to install these providers.
│ 
│ To calculate additional checksums for another platform, run:
│   tofu providers lock -platform=linux_amd64
│ (where linux_amd64 is the platform to generate)
╵
```

## Variant 5: create link ahead

In this case, tofu overwrites the link with the wrong one again:
```console
$ ls -l tofu/.terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64
lrwxr-xr-x@ 1 shoeffner  staff  129 Feb 18 19:18 tofu/.terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64 -> /Users/shoeffner/Projects/test/tofu-bug-local-mirror/terraform.d/plugins/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64
$ tofu -chdir=tofu init
.... Failed to install provider ...
$ ls -l tofu/.terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64
lrwxr-xr-x@ 1 shoeffner  staff  134 Feb 18 19:16 tofu/.terraform/providers/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64 -> /Users/shoeffner/Projects/test/tofu-bug-local-mirror/**tofu**/terraform.d/plugins/registry.opentofu.org/hashicorp/vault/4.6.0/darwin_arm64
```
I believe this is due to https://github.com/opentofu/opentofu/blob/a3276028ad368db23c0a4fed7a143bdb4e616715/internal/providercache/package_install.go#L263-L266

## Variant 6: sanity check, let's do it without the subdirectory

This works as expected (reproduce6.txt).
```diff
- cat <<-EOF > tofu/main.tf
+ cat <<-EOF > main.tf

- TF_LOG=TRACE tofu -chdir=tofu init
+ TF_LOG=TRACE tofu init
```

## Files

I was not allowed to upload .sh files... so here they are as txt files (bash can still interpret them just fine):

[reproduce.txt](https://github.com/user-attachments/files/18850704/reproduce.txt)
[reproduce2.txt](https://github.com/user-attachments/files/18850705/reproduce2.txt)
[reproduce3.txt](https://github.com/user-attachments/files/18850703/reproduce3.txt)
[reproduce4.txt](https://github.com/user-attachments/files/18850706/reproduce4.txt)
[reproduce5.txt](https://github.com/user-attachments/files/18851486/reproduce5.txt)
[reproduce6.txt](https://github.com/user-attachments/files/18851648/reproduce6.txt)

### Additional Context

The error is likely somewhere in [installFromLocalDir](https://github.com/opentofu/opentofu/blob/a3276028ad368db23c0a4fed7a143bdb4e616715/internal/providercache/package_install.go#L194), which performs the linking, cleanup, etc., or wherever the PackageMeta is generated (as I can imagine the `Location` there is wrong). Right now I cannot download all the build dependencies to debug it (as I am on the train – it's a miracle I managed to run the reproduce scripts...), but I could probably take a closer look later or maybe next week, if you like.

I first assumed my ~/.tofurc would override some part of this (I have a few developer settings) so I removed it – the issue persists without it.
I don't think I do anything special here.

I haven't tried other variants of local mirrors (such as XDG_DATA_DIR etc.).

Now that I have a workaround – actually, two: no chdir and copying after a failed attempt –  there is no particular hurry to fix this one, I just thought it's a good idea to put this in the backlog. For my team the local mirror itself is a workaround while we internally patch a provider until the changes are upstream or we finally get around to setting up our own registry, so it's fine.
We also considered the dev_overrides, but placing an executable form our build server into the right directory (potentially already contained in a ziptree) makes it a zero-config vs. "little config" effort, which is why the local mirror is nicer (even though we somewhat lose the automatic locking mechanism).

### References

I searched for `local mirror chdir` and nothing turned up, but of course it could be I missed something.

baa-ableton:
Resolves #2531 

## Checklist

### Go checklist

 

### DIFF
--- a/cmd/tofu/provider_source_test.go
+++ b/cmd/tofu/provider_source_test.go
@@ -19,22 +19,19 @@ import (
 
 func TestProviderSource(t *testing.T) {
 	tests := []struct {
-		name               string
-		originalWorkingDir string
-		setupFunc          func(t *testing.T) (string, string) // returns (originalDir, overrideDir)
-		expectedProvider   string
+		name             string
+		setupFunc        func(t *testing.T) (string, string) // returns (originalDir, overrideDir)
+		expectedProvider string
 	}{
 		{
-			name:               "without overrideWd - should use terraform.d/plugins in original working directory",
-			originalWorkingDir: "",
+			name: "without overrideWd - should use terraform.d/plugins in original working directory",
 			setupFunc: func(t *testing.T) (string, string) {
-				// Create terraform.d/plugins in the original working directory
-				currentDir, err := os.Getwd()
-				if err != nil {
-					t.Fatalf("Failed to get current directory: %v", err)
-				}
-				pluginsDir := filepath.Join(currentDir, "terraform.d", "plugins")
-				err = os.MkdirAll(pluginsDir, 0755)
+				// Create a temporary directory to conduct the test in
+				tempDir := t.TempDir()
+
+				// Create terraform.d/plugins in the testing directory
+				pluginsDir := filepath.Join(tempDir, "terraform.d", "plugins")
+				err := os.MkdirAll(pluginsDir, 0755)
 				if err != nil {
 					t.Fatalf("Failed to create plugins directory: %v", err)
 				}
@@ -53,21 +50,19 @@ func TestProviderSource(t *testing.T) {
 					t.Fatalf("Failed to create mock provider binary: %v", err)
 				}
 
-				return currentDir, ""
+				return tempDir, ""
 			},
 			expectedProvider: "hashicorp/test-provider",
 		},
 		{
-			name:               "with overrideWd - should still use terraform.d/plugins in original working directory",
-			originalWorkingDir: "",
+			name: "with overrideWd - should still use terraform.d/plugins in original working directory",
 			setupFunc: func(t *testing.T) (string, string) {
-				// Create terraform.d/plugins in the original working directory
-				originalDir, err := os.Getwd()
-				if err != nil {
-					t.Fatalf("Failed to get current directory: %v", err)
-				}
-				pluginsDir := filepath.Join(originalDir, "terraform.d", "plugins")
-				err = os.MkdirAll(pluginsDir, 0755)
+				// Create a temporary directory to conduct the test in
+				tempDir := t.TempDir()
+
+				// Create terraform.d/plugins in the testing directory
+				pluginsDir := filepath.Join(tempDir, "terraform.d", "plugins")
+				err := os.MkdirAll(pluginsDir, 0755)
 				if err != nil {
 					t.Fatalf("Failed to create plugins directory: %v", err)
 				}
@@ -87,9 +82,13 @@ func TestProviderSource(t *testing.T) {
 				}
 
 				// Create a temporary directory for the override working directory
-				overrideDir := t.TempDir()
+				overrideDir := filepath.Join(tempDir, "override")
+				err = os.MkdirAll(overrideDir, 0755)
+				if err != nil {
+					t.Fatalf("Failed to create override directory: %v", err)
+				}
 
-				return originalDir, overrideDir
+				return tempDir, overrideDir
 			},
 			expectedProvider: "hashicorp/test-provider",
 		},
@@ -100,19 +99,17 @@ func TestProviderSource(t *testing.T) {
 			// Setup test environment
 			originalWorkingDir, overrideWd := tt.setupFunc(t)
 
+			err := os.Chdir(originalWorkingDir)
+			if err != nil {
+				t.Fatalf("Failed to change to original working directory: %v", err)
+			}
+
 			// If we have an override directory, change to it (simulating -chdir behavior)
 			if overrideWd != "" {
 				err := os.Chdir(overrideWd)
 				if err != nil {
 					t.Fatalf("Failed to change to override directory: %v", err)
 				}
-				// Restore the original working directory after the test
-				defer func() {
-					err := os.Chdir(originalWorkingDir)
-					if err != nil {
-						t.Logf("Failed to restore original working directory: %v", err)
-					}
-				}()
 			}
 
 			// Create a mock disco service
@@ -141,9 +138,8 @@ func TestProviderSource(t *testing.T) {
 			provider := addrs.MustParseProviderSourceString(tt.expectedProvider)
 			versions, _, err := source.AvailableVersions(context.Background(), provider)
 			if err != nil {
-				// It's okay if this fails since we're not setting up actual providers
-				// We're mainly testing that the source is created correctly
-				t.Logf("AvailableVersions failed (expected for test): %v", err)
+				// If available provider versions could not be determined, something went wrong with mock provider setup
+				t.Fatalf("AvailableVersions failed (expected for test): %v", err)
 			}
 
 			t.Logf("Source created successfully for test case: %s", tt.name)


cmd/tofu/provider_source.go
```diff
@@ -147,7 +149,14 @@ func implicitProviderSource(ctx context.Context, services *disco.Disco) getprovi
...
 
-	addLocalDir("terraform.d/plugins") // our "vendor" directory
+	// When using -chdir, the terraform.d/plugins directory should be checked
+	// in the original working directory, not the directory passed as parameter
+	// to -chdir.
+	if usingChdir {
+		addLocalDir(filepath.Join(originalWorkingDir, "terraform.d/plugins"))
+	} else {
+		addLocalDir("terraform.d/plugins") // our "vendor" directory
+	}
```
Gogotchuri: I believe we can simplify this to just
```go
addLocalDir(filepath.Join(originalWorkingDir, "terraform.d/plugins"))
```
Which would make this an absolute path and work for both cases.
There would be no need to check whether we are using `-chdir`.
Also, the comments will need to be updated.

cmd/tofu/provider_source.go
```diff
@@ -92,12 +92,14 @@ func explicitProviderSource(ctx context.Context, config *cliconfig.ProviderInsta
 // one version available in a local directory are implicitly excluded from
 // direct installation, as if the user had listed them explicitly in the
 // "exclude" argument in the direct provider source in the CLI config.
-func implicitProviderSource(ctx context.Context, services *disco.Disco) getproviders.Source {
+func implicitProviderSource(ctx context.Context, services *disco.Disco, originalWorkingDir string, usingChdir bool) getproviders.Source {
```
Gogotchuri: `usingChdir` argument is not used in this function and can be removed. It doesn't change anything since we removed conditional branching.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/opentofu/svchost/disco"
+)
+
+func TestImplicitProviderSource(t *testing.T) {
+	tests := []struct {
+		name               string
+		originalWorkingDir string
+		usingChdir         bool
+		setupFunc          func(t *testing.T, originalDir string) string
+		expectedLocalDirs  []string
```
Gogotchuri: `expectedLocalDirs` are not used during the test runs.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/opentofu/opentofu/internal/command/cliconfig"
+	"github.com/opentofu/opentofu/internal/command/cliconfig/ociauthconfig"
+	"github.com/opentofu/svchost/disco"
+)
+
+func TestImplicitProviderSource(t *testing.T) {
+	tests := []struct {
+		name               string
+		originalWorkingDir string
+		usingChdir         bool
```
Gogotchuri: `usingChdir` doesn't alter the behaviour of `implicitProviderSource`; therefore, it is not needed here.

baa-ableton:
Hey @Gogotchuri 

so with this last commit: I fixed the github workflows problem that popped up becuase of the missing license headers in the test file.

I took away the usingChdir parameter which was no longer being used

I reworked the tests and used the paradigm in main.go which is to say using `os.Chdir` before calling `providerSource()` for a test case that has a declared overrideWd. 


I am not 100% on how I went about determining whether tests pass:
```
			provider := addrs.MustParseProviderSourceString(tt.expectedProvider)
			versions, _, err := source.AvailableVersions(context.Background(), provider)
```

If you think there is a better way to check whether or not a provider has been configured (at least in FS structure) properly, LMK!

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/command/cliconfig"
+	"github.com/opentofu/opentofu/internal/command/cliconfig/ociauthconfig"
+	"github.com/opentofu/svchost/disco"
+)
+
+func TestProviderSource(t *testing.T) {
+	tests := []struct {
+		name               string
+		originalWorkingDir string
```
Gogotchuri: Since you are returning the `originalDir` from the `setupFunc`, this field is not used anymore. Please clean this up.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,156 @@
...
+		setupFunc          func(t *testing.T) (string, string) // returns (originalDir, overrideDir)
+		expectedProvider   string
+	}{
+		{
+			name:               "without overrideWd - should use terraform.d/plugins in original working directory",
+			originalWorkingDir: "",
+			setupFunc: func(t *testing.T) (string, string) {
+				// Create terraform.d/plugins in the original working directory
+				currentDir, err := os.Getwd()
```
Gogotchuri: We should use the test's temporary directory here. `t.TempDir` and alter the runs based on that. After which, you will need to use `os.Chdir` in each case at the start to simulate the current behaviour.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,156 @@
...
+				return currentDir, ""
+			},
+			expectedProvider: "hashicorp/test-provider",
+		},
+		{
+			name:               "with overrideWd - should still use terraform.d/plugins in original working directory",
+			originalWorkingDir: "",
+			setupFunc: func(t *testing.T) (string, string) {
+				// Create terraform.d/plugins in the original working directory
+				originalDir, err := os.Getwd()
```
Gogotchuri: We should change this to `t.TempDir()`; See the comment in the above setup function.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,156 @@
...
+				if err != nil {
+					t.Fatalf("Failed to create provider directory: %v", err)
+				}
+
+				// Create a mock provider binary
+				providerBinary := filepath.Join(providerDir, "terraform-provider-test-provider")
+				err = os.WriteFile(providerBinary, []byte("mock provider binary"), 0755)
+				if err != nil {
+					t.Fatalf("Failed to create mock provider binary: %v", err)
+				}
```
Gogotchuri, line 82: This test binary stays in the filesystem after the test. This should be solved once we use the test temp directory.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,156 @@
...
+			originalWorkingDir, overrideWd := tt.setupFunc(t)
+
+			// If we have an override directory, change to it (simulating -chdir behavior)
+			if overrideWd != "" {
+				err := os.Chdir(overrideWd)
+				if err != nil {
+					t.Fatalf("Failed to change to override directory: %v", err)
+				}
+				// Restore the original working directory after the test
+				defer func() {
```
Gogotchuri: This defer shouldn't be needed here; each unit test run is isolated enough that restoring the working directory has no effect.

cmd/tofu/provider_source_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+			// Try to get available versions for a test provider
+			// This will help verify that the local directories are properly configured
+			provider := addrs.MustParseProviderSourceString(tt.expectedProvider)
+			versions, _, err := source.AvailableVersions(context.Background(), provider)
+			if err != nil {
+				// It's okay if this fails since we're not setting up actual providers
+				// We're mainly testing that the source is created correctly
+				t.Logf("AvailableVersions failed (expected for test): %v", err)
+			}
```
Gogotchuri: We should make this check fatal. 
You can take a look at how the `AvailableVersions` function works in `filesystem_mirror_source.go`. If the `AvailableVersions` returns an error and hence can't find the provider version, the behaviour is not correct, and we need to make this check fatal.
`AvailableVersions` scans the directory structure and determines the provider versions based on that. It doesn't validate the binary, and there is no reason for this to fail in our test cases.
 You can verify that by running this test without your fix in this PR. If you don't fail at this point, the version before your fix passes just fine.

Gogotchuri (changes requested):
Implementation looks good! Thanks for taking the time to apply my suggestions.
I have a few additional comments on the tests.

baa-ableton:
Hey @Gogotchuri 
Based on your feedback:
* All tests use t.TempDir() now, even the case without -chdir
* os.ChDir() is used for both test cases: no -chdir -> simply changes dir once to t.TempDir() and providerSource() is called from there with the same t.TempDir() passed in as originalWorkingDir. As for the case using -chdir -> os.ChDir() is called twice. Once to move to t.TempDir() and then from there into './override' where providerSource() is called with the originalWorkingDir passed in as a parameter, and not the override path.

Gogotchuri (approved):
Thanks for making the necessary changes!
This looks good to me now and should be safe to merge.

PR #3078, opened 2025-07-28
Proposed Execution Architecture RFC

cam72cam:
This RFC coalesces multiple months of discussion among the Maintainers and a variety of prototypes.  This is more high level than was originally anticipated, but an architectural vision is needed before we can nit-pick the more granular changes we have been tinkering with.

The general proposal is primarily based on prototyping from @cam72cam, followed up by discussions with @apparentlymart and @yantrio.

Much of the historical reference material is provided by @apparentlymart

This takes precedence over my previous work in https://github.com/opentofu/opentofu/pull/2921

### DIFF
--- a/rfc/20250728-execution-architecture.md
+++ b/rfc/20250728-execution-architecture.md
@@ -110,7 +110,7 @@ We are proposing a major design change to OpenTofu's internal architecture, focu
 
 Operations today re-use a lot of the same concepts and code throughout their execution.  In practice, this means that a lot of work needs to be repeated between operations.  It is also complicated that the expected result of a given codepath may change dramatically based on the operation, leading to hard to trace bugs.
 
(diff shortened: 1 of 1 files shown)


apparentlymart (approved):
Because we've already been discussing various ways to proceed with this, I must confess that I've been having trouble reading this as a high-level goal without reading lots of implementation details into it and then getting stuck thinking about them.

Putting all of those details aside though, I'm understanding this RFC as proposing, at a high level, that the likely benefits of revisiting these long-standing implementation choices are significant enough to justify the risk and opportunity cost of the work.

And as someone who has spent many years trying to improve this situation by gradual changes, with very mixed success, I feel good about this new direction. As we continue working on this we will need to watch out for common hazards like [the second-system effect](https://en.m.wikipedia.org/wiki/Second-system_effect). I think we can keep that in check by being clear with ourselves about exactly what problems we are trying to solve in each decision, and making sure those problems are all connected to either existing functionality or new functionality that we've already investigated enough to be confident that we're focusing only on essential complexity.

I'm looking forward to moving to the next level of detail here. 😀

yottta (approved):

PR #3087, opened 2025-07-31
add `for_each` attribute to the `mock_provider` block

### LINKED ISSUE #2917: Provider configurations with `for_each` are not compatible with `mock_provider` in test configurations
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.9.0
on darwin_arm64
+ provider registry.opentofu.org/hashicorp/aws v5.100.0
```

Link to public repo with reproducible example:  https://github.com/ryanpagel/opentofu_bug_report_01/tree/main/example_01

### Debug Output

will provide later

### Expected Behavior

Should be able to run `tofu test` with `mock_provider` for providers that are defined using the new `for_each` construct.

### Actual Behavior

The following error when running `tofu test` with the steps below:

```
tests/default.tftest.hcl... fail
  run "default"... fail
╷
│ Error: Provider instance not present
│ 
│ To work with aws_s3_bucket.this["us-east-1"] its original provider instance at provider["registry.opentofu.org/hashicorp/aws"].by_region["us-east-1"] is required, but it has been removed. This occurs when an element is removed from the provider configuration's for_each collection while objects created by
│ that the associated provider instance still exist in the state. Re-add the for_each element to destroy aws_s3_bucket.this["us-east-1"], after which you can remove the provider configuration again.
│ 
│ This is commonly caused by using the same for_each collection both for a resource (or its containing module) and its associated provider configuration. To successfully remove an instance of a resource it must be possible to remove the corresponding element from the resource's for_each collection while
│ retaining the corresponding element in the provider's for_each collection.
╵

Failure! 0 passed, 1 failed.
```

### Steps to Reproduce

1. clone the repo: https://github.com/ryanpagel/opentofu_bug_report_01/tree/main/example_01
2. cd example_01
3. tofu init -var-file=test.tfvars
4. tofu test -var-file=test.tfvars

### Additional Context

Creating providers using the new `for_each` makes it impossible to use the `mock_provider` for those providers within the context of `tofu test`.  We should be able to use `mock_provider` for providers that are created using `for_each`.

### References

_No response_

diofeher:
Resolves #2917

The way I decided to solve the issue above was to create a new attribute for `mock_provider`. So, when users are configuring mock providers they can do something like:

```.tftest.hcl
mock_provider "aws" {
  alias = "by_region"
  for_each = toset(var.regions)
}

variables {
  regions = [ "us-east-1", "us-west-2"]
}
```

And the instances are going to be available on the managed resources, allowing an assert like:

```
run "default" {
  assert {
    condition = aws_s3_bucket.this["us-east-1"].bucket == "rp-test-abc-us-east-1"
    error_message = "this should fail"
  }
    
  assert {
    condition = aws_s3_bucket.this["us-west-2"].bucket == "rp-test-abc-us-west-2"
    error_message = "this should fail"
  }
}

```

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/test_test.go
+++ b/internal/command/test_test.go
@@ -1542,8 +1542,8 @@ func TestTest_MockProviderValidation(t *testing.T) {
 	}
 }
 
-// TestTest_MockProviderValidation checks if tofu test runs proper validation for
-// mock_provider. Even if provider schema has required fields, tofu test should
+// TestTest_MockProviderValidationForEach checks if tofu test runs proper validation for
+// mock_provider with for_each. Even if provider schema has required fields, tofu test should
 // ignore it completely, because the provider is mocked.
 func TestTest_MockProviderValidationForEach(t *testing.T) {
 	td := t.TempDir()
--- a/internal/configs/config.go
+++ b/internal/configs/config.go
@@ -1035,13 +1035,14 @@ func (c *Config) getProviderConfigTransformForTest(evalCtx *hcl.EvalContext) tes
 	}
 }
 
-// evaluateProviderConfig evaluates code for the mock provider. In the current implementation,
-// it only evaluates the for_each expression.
+// evaluateProviderConfig evaluates code for the mock provider. for_each is the only attribute
+// that is evaluated for the mock provider, but support for other provider attributes can be added
+// here if needed.
 func (mp *MockProvider) evaluateProviderConfig(evalCtx *hcl.EvalContext) hcl.Diagnostics {
 	var diags hcl.Diagnostics
 
 	if mp.ForEach == nil {
-		// Since we're evaluating only forEach expressions, return it if it's not present.
+		// Since we're evaluating only for_each expressions, return it if it's not present.
 		return diags
 	}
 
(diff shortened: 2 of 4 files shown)


cam72cam (approved):

cam72cam:
Although there is an open question in the original issue, I think this solution is the right one.  We can always re-open and add the more "magic" functionality if it is deemed necessary.

KarstenSiemer:
Very nice, I think this solves the problem correctly

diofeher:
Gonna wait for one more review from core maintainers here to validate the design.

internal/command/test_test.go
```diff
@@ -1541,3 +1541,69 @@ func TestTest_MockProviderValidation(t *testing.T) {
 		t.Fatalf("expected status code 0 but got %d: %s", code, output.All())
 	}
 }
+
+// TestTest_MockProviderValidation checks if tofu test runs proper validation for
```
Yantrio: I think we should update this to match the test

internal/configs/config.go
```diff
@@ -1039,6 +1048,38 @@ func (c *Config) getProviderConfigTransformForTest(evalCtx *hcl.EvalContext) tes
 	}
 }
 
+// evaluateProviderConfig evaluates code for the mock provider. In the current implementation,
+// it only evaluates the for_each expression.
```
Yantrio: tiny tiny nitpick: you mention that "in the current implementation" it does something, is there ever going to be a future implementation? I don't think we need this kind of comment here.

diofeher: That's a good point! Let me rework on this comment

diofeher: Now that I'm working on it, I could remember what I was thinking: we could implement support for other features in the future, but I'll clarify this comment.

diofeher: Fixed on https://github.com/opentofu/opentofu/pull/3087/commits/fba207dedafe6c65104d68b0ea88dc80d3295654

internal/command/test_test.go
diofeher: Fixed on https://github.com/opentofu/opentofu/pull/3087/commits/fba207dedafe6c65104d68b0ea88dc80d3295654

Thanks for the catch!

Yantrio (approved):

PR #3108, opened 2025-08-06
Ephemeral variables

yottta:
Part of #2834 

An overview of the changes:
* Changed the way we pass the values of the ephemeral variables into `tofu apply` (work that relied on https://github.com/opentofu/opentofu/pull/1998).
  * when saving the plan into a file, the ephemeral variables don't need to be stored (3bb5f150714ff524eed1e538dad1efe6ec2b189e and 544169e31fc609b39dca09217ab96b374613938e) into the plan file. Instead, we need to drop those but for doing that, at the time of writing the plan file, we need to know which variables are ephemeral and which are not. Therefore, we added a new method on the `configs.Module` that returns a map with all variables and which is ephemeral and which not. This is then passed over into the plan for later use.
* Enhance the prompt for ephemeral variables to also show a hint that the input is ephemeral and additionally, when typing the value for an ephemeral value, it is not shown. (6ba5dd2f3625815ab2bf6ea4ccbd7452db8d917e)
* Add the ephemeral attribute of the variables in the structure used to generate JSON representation of the variable configuration. Used for the `tofu show` command (258737f8baed4580709ccc6d4361cfc821d0c721)
* Added a unit test for overriding ephemeral attributes of a variable configuration (d5c4fcdb8b0c403a7c3df5f1052265865ff507b2)
  * 🔴 For reviewers: should we allow this? I've seen this done for sensitive and that looks odd too. So, since that is allowed, should we allow it for ephemeral too? 🤔 . I am not sure TBH, because sensitivity configuration of a variable rarely creates errors when the variable is referenced. But if somebody is going to reference an ephemeral variable (and nothing else), that will most probably create a lot of errors, especially if the ephemerality is switched from `false` to `true`.
* Handle variable ephemeral attribute during static evaluation and ensure that whoever is using that evaluator, is not affected (4f6f295beff2b4a4e01ebea32cc6ec5daec67c9c)
* Added handling of the ephemeral values in the variables validation blocks, similar to the ones for sensitive. (4f6f295beff2b4a4e01ebea32cc6ec5daec67c9c)
* Enhance diagnostics to not leak ephemeral values if will come to that (fa86723245a32c1fe607ab4b3c9ddb4e2719e231)
* Added a bunch of TODOs for later work around marks (fae5290313e7acffc9517cfb4852e2cc41028f00)

## Checklist

### Go checklist

 

### DIFF
--- a/internal/lang/evalchecks/eval_for_each.go
+++ b/internal/lang/evalchecks/eval_for_each.go
@@ -182,7 +182,7 @@ func performValueChecks(expr hcl.Expression, hclCtx *hcl.EvalContext, allowUnkno
 			Subject:     expr.Range().Ptr(),
 			Expression:  expr,
 			EvalContext: hclCtx,
-			Extra:       DiagnosticCausedBySensitive(true),
+			Extra:       DiagnosticCausedByConfidentialValues(true),
 		})
 		resultVal = cty.NullVal(ty)
 	}
--- a/internal/tofu/eval_import.go
+++ b/internal/tofu/eval_import.go
@@ -60,12 +60,19 @@ func evaluateImportIdExpression(expr hcl.Expression, evalCtx EvalContext, keyDat
 		})
 	}
 
-	// TODO ephemeral - we might need to add ephemeral mark too
 	if importIdVal.HasMark(marks.Sensitive) {
 		return "", diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "Invalid import id argument",
 			Detail:   "The import ID cannot be sensitive.",
+			Subject:  expr.Range().Ptr(),
+		})
+	}
+	if importIdVal.HasMark(marks.Ephemeral) {
+		return "", diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid import id argument",
+			Detail:   "The import ID cannot be ephemeral.",
 			Subject:  expr.Range().Ptr(),
 		})
 	}
@@ -186,13 +193,20 @@ func parseImportIndexKeyExpr(evalCtx EvalContext, expr hcl.Expression, keyData i
 		return idx, diags
 	}
 
-	// TODO ephemeral - we might need to add ephemeral mark too
 	unmarkedVal, valMarks := val.Unmark()
 	if _, sensitive := valMarks[marks.Sensitive]; sensitive {
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "Import block 'to' address contains an invalid key",
 			Detail:   "Import block contained a resource address using an index which is sensitive. Please ensure indexes used in the resource address of an import target are not sensitive",
+			Subject:  expr.Range().Ptr(),
+		})
+	}
+	if _, ephemeral := valMarks[marks.Ephemeral]; ephemeral {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Import block 'to' address contains an invalid key",
+			Detail:   "Import block contained a resource address using an index which is ephemeral. Please ensure indexes used in the resource address of an import target are not ephemeral",
 			Subject:  expr.Range().Ptr(),
 		})
 	}
--- a/internal/tofu/eval_provider.go
+++ b/internal/tofu/eval_provider.go
@@ -76,14 +76,22 @@ func resolveProviderInstance(ctx context.Context, keyExpr hcl.Expression, keySco
 		return nil, diags
 	}
 
-	// TODO ephemeral - we might need to add ephemeral mark too
 	if keyVal.HasMark(marks.Sensitive) {
 		return nil, diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "Invalid provider instance key",
 			Detail:   "A provider instance key must not be derived from a sensitive value.",
 			Subject:  keyExpr.Range().Ptr(),
-			Extra:    evalchecks.DiagnosticCausedBySensitive(true),
+			Extra:    evalchecks.DiagnosticCausedByConfidentialValues(true),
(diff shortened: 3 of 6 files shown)


internal/command/jsonentities/diagnostic.go
```diff
@@ -394,7 +394,8 @@ func newDiagnosticExpressionValues(diag tfdiags.Diagnostic) []DiagnosticExpressi
 	values := make([]DiagnosticExpressionValue, 0, len(vars))
 	seen := make(map[string]struct{}, len(vars))
 	includeUnknown := tfdiags.DiagnosticCausedByUnknown(diag)
-	includeSensitive := tfdiags.DiagnosticCausedBySensitive(diag)
+	// this means sensitive and/or ephemeral
+	includeSensitiveValues := tfdiags.DiagnosticCausedBySensitive(diag)
```
cam72cam: I think we should re-name the function if it's now doing additional work.

internal/configs/backend.go
```diff
@@ -48,6 +48,8 @@ func decodeBackendBlock(block *hcl.Block) (*Backend, hcl.Diagnostics) {
 // for the purpose of hashing, so that an incomplete configuration can still
 // be hashed. Other errors, such as extraneous attributes, have no such special
 // case.
+// TODO ephemeral - check if ephemeral should be able to be used here or not. Seems that it shouldn't
```
cam72cam: I think forbidding it here is the safe option for now, could just extend the sensitive check below.

internal/configs/version_constraint.go
```diff
@@ -31,6 +31,8 @@ func decodeVersionConstraint(attr *hcl.Attribute) (VersionConstraint, hcl.Diagno
 	}
 	return decodeVersionConstraintValue(attr, val)
 }
+
+// TODO ephemeral - check how the ephemeral mark should be handled in this function
```
cam72cam: Should be forbidden, same as sensitive

internal/lang/evalchecks/eval_for_each.go
```diff
@@ -135,6 +135,7 @@ func performTypeAndValueChecks(expr hcl.Expression, hclCtx *hcl.EvalContext, all
 
 }
 
+// TODO ephemeral - check how ephemeral should impact this function. Check also the unit tests
```
cam72cam, line 138: Good question.  My initial thought is that it should follow the same conventions as Sensitive and only be allowed in each.value and not in each.key.

internal/tfdiags/diagnostic_extra.go
```diff
@@ -151,26 +151,26 @@ func DiagnosticCausedByUnknown(diag Diagnostic) bool {
...
 // DiagnosticExtraBecauseSensitive is an interface implemented by values in
 // the Extra field of Diagnostic when the diagnostic is potentially caused by
-// the presence of sensitive values in an expression evaluation.
+// the presence of sensitive or ephemeral values in an expression evaluation.
 //
 // Just implementing this interface is not sufficient signal, though. Callers
 // must also call the DiagnosticCausedBySensitive method in order to confirm
 // the result, or use the package-level function DiagnosticCausedBySensitive
 // as a convenient wrapper.
 type DiagnosticExtraBecauseSensitive interface {
```
cam72cam: Sensitive -> Secret?

internal/tofu/eval_provider.go
```diff
@@ -76,6 +76,7 @@ func resolveProviderInstance(ctx context.Context, keyExpr hcl.Expression, keySco
 		return nil, diags
 	}
 
+	// TODO ephemeral - we might need to add ephemeral mark too
```
cam72cam: Yep, these instance keys can't be ephemeral

internal/tofu/eval_import.go
```diff
@@ -185,6 +186,7 @@ func parseImportIndexKeyExpr(evalCtx EvalContext, expr hcl.Expression, keyData i
 		return idx, diags
 	}
 
+	// TODO ephemeral - we might need to add ephemeral mark too
```
cam72cam: Yep, these instance keys can't be ephemeral

cam72cam:
Overall, I think these changes make sense.  The one piece that gives me pause is injecting the ephemeral variables back into the plan during apply.  I can't think of a better solution at the moment though.

cam72cam:
>should we allow this? I've seen this done for sensitive and that looks odd too. So, since that is allowed, should we allow it for ephemeral too?

I think so.  Hopefully no-one ever uses it, but it follows our existing conventions

internal/tfdiags/diagnostic_extra.go
yottta: Renamed everything in 787ba01aa13cefdc4a31f7fa9d590874ce9fa0e3

internal/command/jsonentities/diagnostic.go
yottta: Renamed everything in 787ba01aa13cefdc4a31f7fa9d590874ce9fa0e3

internal/tofu/eval_import.go
yottta: Resolved in fc87b945a7f4fc5a303349e9fb43b0e8c587238b

internal/tofu/eval_provider.go
yottta: Resolved in fc87b945a7f4fc5a303349e9fb43b0e8c587238b

internal/lang/evalchecks/eval_for_each.go
yottta: Updated the TODO so that we can handle this properly later. Thanks

internal/configs/version_constraint.go
yottta: Resolved in 7e8efcacaf5b7cbcfbe33eed4b5f5593331b4519

internal/configs/backend.go
yottta: Resolved in 7e8efcacaf5b7cbcfbe33eed4b5f5593331b4519

PR #3110, opened 2025-08-06
Remove all calls to DeepCopy in resource instance write codepath

cam72cam:
This prevents a states.DeepCopy() for each state hook write, while keeping the same safety guarantees.

There are failing  tests as this is just at the prototype level for now

Resolves # 

## Checklist

### Go checklist

 


cam72cam:
@yottta with the changes in the parent PR, there is now a DeepCopy and safer state access.  Does this satisfy your concerns?

internal/states/statemgr/transient.go
```diff
@@ -68,4 +68,13 @@ type Writer interface {
 	// modified while a WriteState call is in progress. WriteState itself
 	// will never modify the given state.
 	WriteState(*states.State) error
+
+	// MutateState modifies the current transient snapshot.
+	//
+	// The caller may return an entirely distinct state object, that must not
+	// be modified after this call.  The primary use case is creating a new state
```
yottta, line 75:
```suggestion
	// be modified after this call. The primary use case is creating a new state
```

PR #3123, opened 2025-08-11
Ephemeral outputs

yottta:
Part of #2834

### Changes
* Add ephemeral attribute to the generated config for outputs. When running `tofu show config -json` we want to also indicate if an output is ephemeral. Therefore, this commit adds support for that.
* Mark the value of an module output as ephemeral when the the output is configured as ephemeral. Till now, the output would have been ephemeral only if the value generated by the expression in output.value attribute would have reference other ephemeral values. With this commit, even if the value expression is not pointing to any ephemeral values, if the output.ephemeral=true, then the value of the output during evaluation, will be marked as ephemeral. This ensures that when an output is used later, the ephemeral mark will be carried over into other expressions.
* Improve the validation of ephemeral outputs. This covers things like ephemeral output declared in root module and outputs that are getting ephemeral values without being configured as such.
* Add ephemeral outputs error tests

## Checklist

### Go checklist

 


internal/tofu/eval_conditions.go
```diff
@@ -184,6 +184,11 @@ func evalCheckRule(ctx context.Context, addr addrs.CheckRule, rule *configs.Chec
 		return checkResult{Status: checks.StatusError}, diags
 	}
 
+	// TODO ephemeral - FOR REVIEWERS: this is here for a long time the way it is.
+	//  The check's failure triggers a print of a diff between the values in the condition.
+	//  The diff is shown even if the condition expression references a sensitive or an ephemeral value.
+	//  Is this as desired? It's the first place where I see a bleeding of confidential info.
+
```
Gogotchuri: This seems like an issue, yeah, nice catch. Can we remove the expression from the diagnostic?

Gogotchuri (commented):
I left a comment on the "review TODO". That part might need to be changed.
Other than that, everything looks good to me.

internal/tofu/eval_conditions.go
yottta: We could, but probably there should be done differently to be able to still show the diff when the information involved is not confidential.
But let's see what others think about this.

cam72cam: `a print of a diff between the values in the condition` can you link to that code?

```
check "dummy" {
        data "tfcoremock_simple_resource" "foo" {
                id = sensitive("42")
        }
        assert {
                condition = data.tfcoremock_simple_resource.foo.id != "42"
                error_message = "foo"
        }
}

resource "tfcoremock_simple_resource" "foo" {
        id = "42"
}

provider "tfcoremock" {
  resource_directory = "./foo"
  data_directory = "./foo"
}

```

seems to function correctly in both human and json mode

cam72cam: The evalcontext/expr printer is pretty smart about removing sensitive values.

cam72cam: https://github.com/opentofu/opentofu/blob/main/internal/command/jsonentities/diagnostic.go#L466-L605 is where that happens on the json path

yottta: Thanks for pushing on this @cam72cam, indeed, the issue is strictly with ephemeral, not with sensitive too... Probably got confused during testing this.
I will try to provide a fix for the ephemeral values on this.

This is the [test](https://github.com/opentofu/opentofu/pull/3123/files#diff-bd91291f1bd57b53049a281c97d491dcb4b5a87d3702dc03c379b64cbf474288R204) for that.

cam72cam (approved):

internal/tofu/eval_conditions.go
yottta: As agreed with @cam72cam, we ignore for the moment the ephemeral from the diffs. Changed in 0f9c21f52544e8901e4669f7c29f3b449b54eec3.

Those particular diffs have been introduced in https://github.com/opentofu/opentofu/pull/3009.
This is not an urgent matter right now, so we can defer it for later.

Should we create a ticket to tackle this later? Or are we ok with only the TODO left in there?

cc: @diofeher Maybe you have something to say against this particular part since you are the author of the functionality this change it touches.

diofeher: Thanks for pinging me into this discussion and sorry for the late answer, @yottta !

I think what you've done makes sense. Later, when we revisit that part of the code, I think we can use the [Change](https://github.com/opentofu/opentofu/blob/main/internal/command/jsonplan/plan.go#L124) struct to hold something similar as it's doing for sensitive values, so we could show something like:

```
3 -> (ephemeral value omitted)
```

You can see on how we do something similar on this [function](https://github.com/opentofu/opentofu/blob/main/internal/command/jsonplan/plan.go#L598).

Gogotchuri (approved):
Nice work!

internal/tofu/eval_conditions.go
yottta: Thanks a lot for checking this and thanks for the suggestion. 

I wanted to do something similar, but the `Change` struct's general purpose is for handling plan and state information about changes. The sensitive information was already there and is useful in other parts too, but the ephemeral information is not meant to be included in that struct in order to move it around to decide on this.
We might be able to do something more clever based on the marks of the after/before values and have a custom type func in the `jsonentities.Diagnostic` instead of the current `Difference` one, and in those funcs we could generate the diff output directly. But we can take this tehnical discussion in a separate issue that will be handled later and have a more transparent decision made there.

Created a new issue to track this: #3151

PR #3131, opened 2025-08-12
Validations should only be run during non-destroy operations

### LINKED ISSUE #3126: OpenTofu 1.9.0+ incorrectly handles data source dependencies during destroy operations
### Community note

> [!TIP]

### OpenTofu Version

```shell
tested with multiple opentofu versions:
- 1.8.11
- 1.9.3
- 1.10.5
```

sample repo to reproduce error:
https://github.com/stefano-franco/opentofu-destroy-bug

### Debug Output

https://gist.github.com/stefano-franco/959daaafe9ab1c6a596416427c3ca112

### Expected Behavior

`tofu destroy` shouldn't cause an error.

### Actual Behavior

OpenTofu 1.9.0+ incorrectly handles data source dependencies during destroy operations when the data source uses for_each with computed values and the output is passed to a module with validation rules.

- The data source value is known during the plan phase
- Module validation should pass since the value is available
- During destroy, OpenTofu incorrectly treats the data source as "unknown"
- This causes a "for_each" validation error even though the resource already exists

### Steps to Reproduce

1. Checkout sample repo: https://github.com/stefano-franco/opentofu-destroy-bug
2. Install OpenTofu 1.9.0+
3. Setup AWS credentials and run `tofu init`
6. `tofu apply` → ✅ Success
7. `tofu destroy` → ❌ Error

### Additional Context

This issue was introduced with OpenTofu 1.9.0 and will only appear on destroy.
Older versions of OpenTofu and any version of Terraform don't have this issue.

### References

_No response_

cam72cam:
This fixes an issue I originally introduced in https://github.com/opentofu/opentofu/pull/2216.  The original PR and this PR try to modify the graph transformation process as little as possible due to it's complexity.

Unfortunately, I introduced a path in #2216 for apply that does not apply to destroy operations.  In both apply and destroy, we aggressively prune the graph to minimize what actually needs to be evaluated in order to make the necessary changes.  This work is done in the pruneUnusedNodesTransformer and relies upon some very specific interface implementation and dense logic.

For #2216, I needed to ensure that variable blocks that have validations are not pruned out from the graph between plan -> apply.  I introduced an interface implementation that allows the node to say "Hey!  I still need to exist!  Don't prune me!".  That worked reasonably well for the standard apply scenario, and I added a test to confirm.

However, I did not fully understand the implications for destroy.  During a destroy action, the variables should be pruned from the graph and the check not run, as the data it's targeting won't exist (the reason for the check/validation).  When it is not pruned, it keeps "configuration" nodes within the apply graph which do not function correctly during destroy operations.

This is not an ideal solution, but it has a very small likelihood of breaking existing workflows.  As evidenced by the complexity described above, this is a very tricky thing to modify correctly, and the smaller the change we can make, the better.  Longer term, I hope that something that comes out of #3078 will allow us to simplify this whole process.

I still need to add a bespoke test case to context_apply2_test.go and update the changelog

Resolves #3126 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/transform_destroy_edge.go
+++ b/internal/tofu/transform_destroy_edge.go
@@ -287,11 +287,11 @@ type pruneUnusedNodesTransformer struct {
 	// destroy. Planing normally involves all nodes, but during a destroy plan
 	// we may need to prune things which are in the configuration but do not
 	// exist in state to evaluate.
-	skip bool
+	Op walkOperation
 }
 
 func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) error {
-	if t.skip {
+	if t.Op != walkPlanDestroy && t.Op != walkDestroy && t.Op != walkApply {
 		return nil
 	}
 
@@ -314,7 +314,7 @@ func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) err
 				case graphNodeTemporaryValue:
 					// root module outputs indicate they are not temporary by
 					// returning false here.
-					if !n.temporaryValue() {
+					if !n.temporaryValue(t.Op) {
 						log.Printf("[TRACE] pruneUnusedNodes: temporary value vertex %q kept because it's not a temporary value vertex", dag.VertexName(n))
 						return
 					}
@@ -342,7 +342,7 @@ func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) err
 							// root module, and so it's not actually important
 							// to expand it and so this lets us do a bit more
 							// pruning than we'd be able to do otherwise.
-							if tmp, ok := v.(graphNodeTemporaryValue); ok && !tmp.temporaryValue() {
+							if tmp, ok := v.(graphNodeTemporaryValue); ok && !tmp.temporaryValue(t.Op) {
 								log.Printf("[TRACE] pruneUnusedNodes: expanding vertex %q kept because another expanding vertex %q with non-temporary value is one of its dependencies", dag.VertexName(n), dag.VertexName(v))
 								continue
 							}
--- a/internal/tofu/transform_targets.go
+++ b/internal/tofu/transform_targets.go
@@ -204,6 +204,7 @@ func (t *TargetingTransformer) getTargetedOutputNodes(targetedNodes dag.Set, gra
 
 		// root module outputs indicate that while they are an output type,
 		// they not temporary and will return false here.
+		// We use walkInvalid here as we only care about the op as a workaround for nodeVariableReference, which does not apply here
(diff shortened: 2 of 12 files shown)


yottta (commented):
Could you please provide some comments (in code or in the PR description) regarding the thinking process about this? I would be curious what takes drove these changes.

cam72cam:
@diofeher you are right, let me move this back to draft as I get that (and the checks) sorted :+1:

diofeher:
@cam72cam Now that I've seen your updated description at the PR, it makes sense! 

cc @yottta

internal/tofu/transform_destroy_edge.go
```diff
@@ -287,11 +287,11 @@ type pruneUnusedNodesTransformer struct {
 	// destroy. Planing normally involves all nodes, but during a destroy plan
```
diofeher, line 287: I think we should update the comment above `Op walkOperation` to match the new field. It seems to be explaining why skip is being used, so we would want to explain why `Op walkOperation` is there instead.

internal/tofu/transform_targets.go
```diff
@@ -204,7 +204,7 @@ func (t *TargetingTransformer) getTargetedOutputNodes(targetedNodes dag.Set, gra
 
 		// root module outputs indicate that while they are an output type,
 		// they not temporary and will return false here.
-		if tv.temporaryValue() {
+		if tv.temporaryValue(walkInvalid) {
```
diofeher, line 208: From what I understood, it shouldn't matter what's being passed to this `temporaryValue` here since this value is only being used at `nodeVariableReference`, and the type of nodes here are only the ones implementing `GraphNodeTargetable`.

With that being said, I'm wondering if we should comment explicitly here saying why `walkInvalid` is being used because it can look confusing to someone reading the code. But I would say this is really a nitpick 😅

Gogotchuri (changes requested):
This looks good to me!

internal/tofu/transform_destroy_edge.go
cam72cam: Updated in efffb7498c7913848184e7ca8f9999028386a46a

internal/tofu/transform_targets.go
cam72cam: Documented in c2d5dbd55f9c1dffeaab8eb81815fba8b73f51dd

Yantrio (approved):

Yantrio:
This looks good to me, I think it may be nice to have a test for destroy transformers too though @cam72cam

cam72cam:
@Yantrio given the complexity of interactions between components in the tofu package, the vast majority of functionality is tested via the context tests.

PR #3140, opened 2025-08-14
Simplify the nightly build process

Yantrio:


yottta (approved):
So no signing of the release? 🤔

Yantrio:
I don't see the need to sign the nightly releases. it also reduces the build time drastically to skip the signing

PR #3170, opened 2025-08-22
Add `tofu.applying` (and `terraform.applying`)

cam72cam:
## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/static_scope.go
+++ b/internal/configs/static_scope.go
@@ -212,6 +212,9 @@ func (s staticScopeData) GetTerraformAttr(_ context.Context, addr addrs.Terrafor
 	// TODO this is copied and trimmed down from tofu/evaluate.go GetTerraformAttr.  Ideally this should be refactored to a common location.
 	var diags tfdiags.Diagnostics
 	switch addr.Name {
+	case "applying":
+		return cty.False.Mark(marks.Ephemeral), nil
+
 	case "workspace":
 		workspaceName := s.eval.call.workspace
 		return cty.StringVal(workspaceName), diags
--- a/website/docs/language/expressions/references.mdx
+++ b/website/docs/language/expressions/references.mdx
@@ -130,8 +130,13 @@ The following values are available:
   advanced cases where OpenTofu is run from a directory other than the root module
   directory. We recommend using `path.root` or `path.module` over `path.cwd` where
   possible.
-- `terraform.workspace` is the name of the currently selected
+- `terraform.workspace` / `tofu.workspace` is the name of the currently selected
   [workspace](../../language/state/workspaces.mdx).
+- `terraform.applying` / `tofu.applying` is true when the apply phase is running
+  and false in any other phase (plan, validate). This keyword is related to the
+  apply phase and not to the apply command, meaning that when running `tofu apply`,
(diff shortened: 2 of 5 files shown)


cam72cam:
image

internal/configs/static_scope.go
```diff
@@ -212,6 +212,9 @@ func (s staticScopeData) GetTerraformAttr(_ context.Context, addr addrs.Terrafor
 	// TODO this is copied and trimmed down from tofu/evaluate.go GetTerraformAttr.  Ideally this should be refactored to a common location.
 	var diags tfdiags.Diagnostics
 	switch addr.Name {
+	case "applying":
```
yottta, line 215: Do we want to add it in the static evaluator? 🤔 What would be the use case? Isn't the static evaluator used only in the initialisation? If so, we should not allow this here. Maybe I miss something

cam72cam: I could see someone trying to use this unintentionally in a static scope.  We need to support the attribute, but the question is what the value should be.  I think either what I have here or `cty.UnknownVal(cty.Bool)`

yottta: I would keep the false one. I don't think that we should play with unknown values in this context.

apparentlymart: I'd suggest making it return an error instead of any actual value, because anything we decide during early eval will be relevant during both the plan and apply phases and so it doesn't really make sense to use this symbol as part of those expressions.

With that said, I suppose it's pretty debatable what it even means for there to be an ephemeral value during early eval... almost everything we decide with early eval gets saved to disk somewhere anyway so maybe attenpts to use this would fail downstream somewhere anyway. Still, I expect we can return a clearer error message about this one if we do it here since we can be explicit in the message that it's invalid because the result will be used during both the plan and apply phases and must stay consistent between them.

website/docs/language/expressions/references.mdx
```diff
@@ -130,8 +130,13 @@ The following values are available:
...
   directory. We recommend using `path.root` or `path.module` over `path.cwd` where
   possible.
-- `terraform.workspace` is the name of the currently selected
+- `terraform.workspace` / `tofu.workspace` is the name of the currently selected
   [workspace](../../language/state/workspaces.mdx).
+- `terraform.applying` / `tofu.applying` is true when the apply phase is running
+  and false in any other phase (plan, validate). This keyword is related to the
+  apply phase and not to the apply command, meaning that when running `tofu apply`,
+  it will first be set to false during the plan plase and then set to true during
+  the apply phase. This is an ephemeral value that should be handled accordingly.
```
apparentlymart, line 139: I've not been paying close attention to how we've been documenting the other parts of this, but it could be nice to have some central doc section somewhere that defines what "ephemeral value" means and what it means to "handle it accordingly" so that we can make the words here, and presumably in various other places, link to that definition.

(This can wait for a later PR that finalizes all of the docs for this whole area of functionality though, if we don't yet have everything in place.)

internal/configs/static_scope.go
cam72cam: Addressed in 75693b81a1

PR #3171, opened 2025-08-22
Ephemeral write only attributes

yottta:
Part of #2834

Overview of the changes:
* The diff renderer gets some changes because for the write-only attributes that have no values to generate the rendering action based on, we need to propagate the action of the parent block to be used for the write-only renderers. This leads to **always** rendering write-only attributes when the parent block is not null.
* In order to ensure the robustness of the write-only attributes work through the system, on **every call that can return the resource state**, we added a validation to ensure that the provider is not misbehaving and all the attributes configured as write-only are nil.
* Because the write-only attributes are always null, in state and in the responses from the provider, if the plan calls return such an attribute in the `RequiresReplace` list we cannot rely on the logic existing till now where the replacement of the resource it's performed only if the value changed. For write-only, we instead check if the attribute is configured as such so we record it as a valid candidate for triggering the replacement. Later, we check that the paths recorded are indeed write-only.
  * In order to test this, I had to do some hacky things with the testing provider because the [PlanModifiers](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#requiresreplace) do not work for the write-only, even though it's supposed to. Took a note to open a PR and an issue on [terraform-provider-framework](https://github.com/hashicorp/terraform-plugin-framework). The issue is [here](https://github.com/hashicorp/terraform-plugin-framework/blob/51dc6877ded6018bdd55c18406b4c80ecab2f99e/resource/schema/mapplanmodifier/requires_replace_if.go#L63). Will never run the modifier because the values for write-only are null in state and plan. 
* In [2838](https://github.com/opentofu/opentofu/pull/2838/files#diff-9c36ed2e45b438049e9222a2e4b0ed42cae723cf105a59fc61f6b80b184a2cc4R316) we updated the plan validation logic to allow a the write-only attributes to be null in the plan even if the configuration is not. That logic had to be slightly move up to cover any write-only configured nested attribute.
## Checklist

### Go checklist

 


internal/tofu/write_only.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/opentofu/opentofu/internal/configs"
+	"github.com/opentofu/opentofu/internal/configs/configschema"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// validateWriteOnlyAttributes checks that the write-only attributes are not returned back with an actual value.
+func validateWriteOnlyAttributes(schema *configschema.Block, v cty.Value, resCfg *configs.Resource) (diags tfdiags.Diagnostics) {
```
cam72cam: I have mixed feelings about this.  It's a valid solution, but leaks the "provider should be implemented correctly" concerns outside the providers.Interface.

What do you think about implementing this logic within the objects that fulfill providers.Interface instead?

internal/command/jsonformat/computed/renderers/write_only.go
```diff
@@ -0,0 +1,26 @@
...
+
+package renderers
+
+import (
+	"fmt"
+
+	"github.com/opentofu/opentofu/internal/command/jsonformat/computed"
+)
+
+var _ computed.DiffRenderer = (*writeOnlyRenderer)(nil)
```
cam72cam: We already assert this in the constructor below, why duplicate this check?

cam72cam:
This is quite comprehensive, well done!

internal/tofu/write_only.go
yottta: Thanks! I struggled with finding a good place for this and I don't like it either the way it is now. Let me try and fix this.

internal/command/jsonformat/computed/renderers/write_only.go
yottta: Fixed in b8b2fa5c4a4e4994eac28a4a543200095363325b. Was there strictly because I followed the way sensitive renderer was implemented. But I agree that it's redundant.

internal/tofu/write_only.go
yottta: Fixed in 906d5e6e704de6328995140ba854cb849086fc49.
The issue with this approach is that at the Interface level we have no way to reference back the resource configuration as was done before.
But that shouldn't be a problem since the issue that it's validated here is not per resource instance but it's checking the required functionality of the whole provider instance. We should be fine.

cam72cam: Thanks!

internal/command/jsonformat/computed/renderers/testing.go
```diff
@@ -121,6 +135,12 @@ func validateKeys[C, V any](t *testing.T, actual map[string]C, expected map[stri
 		if diff := cmp.Diff(actualAttributes, expectedAttributes); len(diff) > 0 {
 			t.Errorf("actual and expected attributes did not match: %s", diff)
 		}
+	} else {
+		gotKeys := slices.Sorted(maps.Keys(actual))
+		wantKeys := slices.Sorted(maps.Keys(expected))
+		if diff := cmp.Diff(wantKeys, gotKeys); len(diff) > 0 {
+			t.Errorf("keys not match: %s", diff)
+		}
 	}
```
Gogotchuri, line 144: Could you explain why this else is needed here? Wouldn't it be the same if `validateKeys` was just this else block?

yottta: There is none, indeed.
TBH, I wanted to write it as
```go
	if len(actual) != len(expected) {
       // ...
       return
    }
    gotKeys := slices.Sorted(maps.Keys(actual))
    wantKeys := slices.Sorted(maps.Keys(expected))
    if diff := cmp.Diff(wantKeys, gotKeys); len(diff) > 0 {
       t.Errorf("keys not match: %s", diff)
    }
```
but just forgot...

I can push another fix for this if you would consider this useful.

Gogotchuri: It can get even simpler, like this:
```go
func validateKeys[C, V any](t *testing.T, actual map[string]C, expected map[string]V) {
        gotKeys := slices.Sorted(maps.Keys(actual))
        wantKeys := slices.Sorted(maps.Keys(expected))
        if diff := cmp.Diff(wantKeys, gotKeys); len(diff) > 0 {
	        t.Errorf("keys not match: %s", diff)
        }
}
 ```

Gogotchuri: This covers everything in both conditions. Anyway, putting much effort into this is not worth it. Thanks!

yottta: After checking this again, yeah, makes total sense to leave it only as you said. Thanks a lot.
Pushed [here](https://github.com/opentofu/opentofu/pull/3176).

PR #3176, opened 2025-08-25
Remove the else branch since does not make sense anymore

yottta:
As discussed in [#3171](https://github.com/opentofu/opentofu/pull/3171#discussion_r2297813019)

### DIFF
--- a/internal/command/jsonformat/computed/renderers/testing.go
+++ b/internal/command/jsonformat/computed/renderers/testing.go
@@ -8,7 +8,6 @@ package renderers
 import (
 	"maps"
 	"slices"
-	"sort"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


cam72cam (approved):

diofeher:
nit: on the PR title you mentioned about removing the `else` branch, but on the discussion, what's removed is the block on the if. So we can clarify better on the PR title @yottta

yottta:
> nit: on the PR title you mentioned about removing the `else` branch, but on the discussion, what's removed is the block on the if. So we can clarify better on the PR title @yottta

Done. I have a 7th sense today. I merge the PRs right before somebody else is submitting its review 😅

diofeher:
@yottta haha don't worry! When I clicked on submit, I saw you just merged, but I wanted to leave as reference

diofeher:
Oh, it should be good since we're not merging on main, only on your branch (just noticed that)

PR #3192, opened 2025-08-27
Rework the way ephemeral variables are used when given on `tofu apply` command

yottta:
We added the initial parsing in #3108 (commit 3bb5f150714ff524eed1e538dad1efe6ec2b189e). We took a note to try a different approach on handling this.

Attempts to resolve [this review](https://github.com/opentofu/opentofu/pull/3108#issuecomment-3161067954).

This commit explores a different path: saving the variable names into the plan file and use those later to determine what variables are required to have a value passed in during the apply command.
This way, we have a more transparent way of what variables are meant to be configured during the apply and what variables should not.

**Update for the rest of the commits:** 
Additionally, In order to maintain a similar logical structure with the already existing PlanOpts, we introduced a new structure called ApplyOpts. This new structure is meant to handle variable values for the apply phase in the same way PlanOpts handle those for the plan phase. `ApplyOpts#SetVariables` will be only used for the variables that **are found** in the plan but **as ephemeral**. With this, we moved the [initial implementation](https://github.com/opentofu/opentofu/pull/3108/commits/3bb5f150714ff524eed1e538dad1efe6ec2b189e) from the backend into a more appropriate place, `tofu.Context`.
The initial approach, of injecting the `tofu apply` `-var`s into the existing plan, was definitely hacky.
This new implementation wants to provider a similar framework of handling of `-var`s between `tofu plan` and `tofu apply`.

We also moved all the validations from the backend regarding variables added in #1998 to the `tofu.Context` and merged the existing implementation with all these merges and validations.

Part of #2834

## Checklist

### Go checklist

 


Gogotchuri (approved):

yottta:
Dropped existing reviews since this has been reworked. Please check the description and the second commit message for more details.

internal/plans/planfile/tfplan.go
```diff
@@ -240,11 +241,18 @@ func readTfplan(r io.Reader) (*plans.Plan, error) {
 	}
 
 	for name, rawVal := range rawPlan.Variables {
+		// For the variables written in the plan with a null value, we only record that as ephemeral
+		// but we don't load its null value in the plan object since that would add some complexity later.
+		if valueIsEmpty(rawVal) {
```
cam72cam: Is this necessary now that they are no longer set below?

yottta: If you refer to the fact that we check for empty value, yes. Otherwise there is no way to mark that variable as an ephemeral one or not.
Maybe I am misunderstanding this. Please correct me if I am wrong.

internal/tofu/context_apply.go
```diff
@@ -22,6 +23,18 @@ import (
 	"github.com/opentofu/opentofu/internal/tracing"
 )
 
+type ApplyOpts struct {
+	// SetVariables are the raw values for root module variables as provided
+	// by the user who is requesting the run, prior to any normalization or
+	// substitution of defaults.
```
Yantrio, line 34: Can you please add a comment that this is based on the PlanOpts.

This is mentioned in our contributing guidelines here: https://github.com/opentofu/opentofu/blob/main/contributing/DEVELOPING.md#a-note-on-copyright

> When you copy/paste code from within the OpenTofu code, always make it explicit where you copied from. This helps us resolve issues later on.

internal/tofu/context_apply2_test.go
```diff
@@ -5710,3 +5711,188 @@ ephemeral "test_ephemeral_resource" "a" {
...
+				for vName, ok := range plan.EphemeralVariables {
+					if !ok {
+						continue
+					}
+					delete(plan.VariableValues, vName)
+				}
+			}
+
+			// check apply
+			newState, diags := ctx.Apply(context.Background(), plan, m, tt.applyOpts)
```
Yantrio:
```suggestion
			newState, diags := ctx.Apply(t.Context(), plan, m, tt.applyOpts)
```

internal/tofu/context_apply2_test.go
```diff
@@ -5710,3 +5711,188 @@ ephemeral "test_ephemeral_resource" "a" {
...
+
+	for name, tt := range cases {
+		t.Run(name, func(t *testing.T) {
+			h := &testHook{}
+			ctx := testContext2(t, &ContextOpts{
+				Hooks: []Hook{h},
+			})
+
+			// check plan
+			plan, diags := ctx.Plan(context.Background(), m, states.NewState(), &PlanOpts{
```
Yantrio:
```suggestion
			plan, diags := ctx.Plan(t.Context(), m, states.NewState(), &PlanOpts{
```

internal/tofu/context_apply.go
```diff
@@ -242,7 +225,122 @@ func (c *Context) ApplyGraphForUI(plan *plans.Plan, config *configs.Config) (*Gr
...
+// during the apply too, not only during the plan command.
+// Therefore, when a plan is created via `tofu plan -out <planfile>` and then applied with `tofu apply <planfile>`
+// we need to be able to specify -var/-var-file/etc to allow configuring the variables that are not kept in the
+// plan (encryption configuration, ephemeral variables).
+//
+// # mergePlanAndApplyVariables
+// This gets the plan and the *ApplyOpts and builds the InputValues. The values saved in the plan have
+// priority *when defined*, but the variables found in the plan with a null value are considered to be
+// ephemeral and values for those are searched in the ApplyOpts.
+func (c *Context) mergePlanAndApplyVariables(config *configs.Config, plan *plans.Plan, opts *ApplyOpts) (InputValues, tfdiags.Diagnostics) {
```
Yantrio, line 255: I think it would be a good idea here to add unit tests for this method.

internal/tofu/context_apply.go
yottta: Fixed in 6dd7dd6060aa0ae88edf6839edf4973811147ec9

internal/tofu/context_apply2_test.go
yottta: Fixed in 6dd7dd6060aa0ae88edf6839edf4973811147ec9

internal/tofu/context_apply2_test.go
yottta: Fixed in 6dd7dd6060aa0ae88edf6839edf4973811147ec9

internal/tofu/context_apply.go
yottta: Fixed in 5ea694808390f136967b1d3c32262d6e9dc6ca3b.
Tried to cover all of the cases and also added comments into the actual method to make it understandable later.
Please let me know if you see other cases that I don't.

Yantrio (approved):

internal/plans/planfile/tfplan.go
```diff
@@ -470,13 +478,16 @@ func changeFromTfplan(rawChange *planproto.Change) (*plans.ChangeSrc, error) {
...
-	if len(rawV.Msgpack) == 0 { // len(0) because that's the default value for a "bytes" in protobuf
+	if valueIsEmpty(rawV) {
 		return nil, fmt.Errorf("dynamic value does not have msgpack serialization")
 	}
-
 	return plans.DynamicValue(rawV.Msgpack), nil
 }
 
+func valueIsEmpty(rawV *planproto.DynamicValue) bool {
+	return rawV == nil || len(rawV.Msgpack) == 0 // len(0) because that's the default value for a "bytes" in protobuf
```
cam72cam: So if I understand correctly, we are determining plan.EphemeralVariables from the null/emptyness of a Dynamic value.

I'm wondering if we want to store plan.EphemeralValues directly in the proto file instead of re-deriving it later on.  I could see this causing some weird bugs down the line?

yottta: Since the concept is new, as long as those are wrote and read correctly from the plan, there should be no issue.
The only thing that I am not sure about is how the proto changes will impact downgrades, in case are needed 🤔

cam72cam: Plan files are locked to specific versions and thus can not be upgraded/downgraded.

IMO it's safer both now and when new features are introduced to explicilty store this information in the proto representation.

yottta: Fixed in 730f22ec7be404b3c8e42588f5529608c498b9af.
I love this! I don't know why I am holding myself back from updating the state/plan proto files... bad habits die hard.

cam72cam (approved):

PR #3214, opened 2025-08-29
Use parallel errgroup to increase file copy speed

cam72cam:
Reduces the impact of the issue reported in #1086 

More specifically: When using monorepo/centralized modules with lots of files, this starts multiple IO copies in parallel.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/copy/copy_dir.go
+++ b/internal/copy/copy_dir.go
@@ -6,11 +6,14 @@
 package copy
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"golang.org/x/sync/errgroup"
 )
 
 // CopyDir recursively copies all of the files within the directory given in
@@ -44,6 +47,8 @@ func CopyDir(dst, src string) error {
 		return fmt.Errorf("failed to evaluate symlinks for source %q: %w", src, err)
 	}
 
+	var errg errgroup.Group
+
 	walkFn := func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return fmt.Errorf("error walking the path %q: %w", path, err)
@@ -65,13 +70,6 @@ func CopyDir(dst, src string) error {
 		// destination with the path without the src on it.
 		dstPath := filepath.Join(dst, path[len(src):])
 
-		// we don't want to try and copy the same file over itself.
-		if eq, err := SameFile(path, dstPath); err != nil {
-			return fmt.Errorf("failed to check if files are the same: %w", err)
-		} else if eq {
-			return nil
-		}
-
 		// If we have a directory, make that subdirectory, then continue
 		// the walk.
 		if info.IsDir() {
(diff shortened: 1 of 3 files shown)


joe-a-t:
I tested this and it improved time from 52 seconds on Tofu 1.10.5 to 21 seconds. It's still not anywhere near as good as if we just did symlinks instead of copying files which would reduce that time down to 2.5 seconds in my test, but this is still a significant improvement that is worth doing while we continue to push for unifying the behavior of remote sources to use symlinks instead of copying the files.

Gogotchuri (approved):
Looks good!

internal/copy/copy_dir.go
```diff
@@ -87,22 +85,35 @@ func CopyDir(dst, src string) error {
...
-		// If the current path is a symlink, recreate the symlink relative to
-		// the dst directory
-		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
-			target, err := os.Readlink(path)
-			if err != nil {
-				return fmt.Errorf("failed to read symlink %q: %w", path, err)
+		errg.Go(func() error {
+
+			// we don't want to try and copy the same file over itself.
+			if eq, err := SameFile(path, dstPath); err != nil {
```
Yantrio, line 91: Could you elaborate on why you moved this into the errgroup instead of having it before the `mkdir`?

cam72cam: I identified SameFile as a bit expensive in the TOFU_CPU_PROFILEing I did.  It makes a bunch of syscalls which are better in the go routine.

The mkdir I left out as I didn't want them contending to make dirs and sub dirs.  It was also a very small amount of the overall runtime.

Yantrio (approved):

internal/copy/copy_dir.go
```diff
@@ -87,22 +85,35 @@ func CopyDir(dst, src string) error {
...
 			}
-			return nil
-		}
-		return copyFile(dstPath, path, info.Mode())
+			return copyFile(dstPath, path, info.Mode())
+		})
+		return nil
 	}
-	return filepath.Walk(src, walkFn)
+	err = filepath.Walk(src, walkFn)
```
yhakbar, line 114: Curious as to why you're using `filepath.Walk` over `filepath.WalkDir` here. Would you consider adjusting that in a future PR (or is it better to use `filepath.Walk` in this scenario)?

cam72cam: Of the testing I've done, the os.Lstat calls are under 3% of the total runtime (according to the profile).

I'd be open to making the change in a subsequent PR, I wanted to keep this one focused to just the parallelism

cam72cam: image

cam72cam: @yhakbar if you can find some scenarios where TOFU_CPU_PROFILE shows that the overhead in lstat is significant, please open an issue with a reproduction case!

yhakbar: I doubt it would result in significant overhead. Just called it out as it's the recommended alternative in the [standard library docs](https://pkg.go.dev/path/filepath#Walk), and you might prefer to standardize on the established best practice unless you need the stat calls.

cam72cam: We do end up using the stat info in most branches of this codepath

PR #3284, opened 2025-09-17
fix `internal/backend/local` tests by closing file handle properly

diofeher:
Relates to #1201 

There are two parts to this PR:

1. The state file is not being closed after `PersistState`, so when DeleteWorkspace is called at `TestLocal_backup`, it doesn't work because the file remains open on Windows. To fix this, the only way to close the file handle is to call `Unlock`. That's why I added a call to `Lock` before persisting and then unlocking it afterward.

https://github.com/opentofu/opentofu/blob/197135b4af4dde49068c55000073a2dbbd0699f1/internal/states/statemgr/filesystem.go#L436

2. runtime.GC is still needed to close the open file handles to avoid `TempDir RemoveAll cleanup: unlinkat` errors.

This PR fixes:

```
--- FAIL: TestLocal_backend (1.63s)
    backend_test.go:32: delete workspace foo err: unlinkat terraform.tfstate.d\foo\terraform.tfstate: The process cannot access the file because it is being used by another process.
    testing.go:1369: TempDir RemoveAll cleanup: unlinkat C:\Users\RUNNER~1\AppData\Local\Temp\TestLocal_backend96626496\001\terraform.tfstate.d\bar\terraform.tfstate: The process cannot access the file because it is being used by another process.
FAIL
FAIL	github.com/opentofu/opentofu/internal/backend/local	3.[42](https://github.com/opentofu/opentofu/actions/runs/17794392461/job/50578690287#step:5:43)1s
```

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/testing.go
+++ b/internal/backend/testing.go
@@ -190,12 +190,24 @@ func TestBackendStates(t *testing.T, b Backend) {
 		fooState := states.NewState()
 		barState := states.NewState()
 
+		// Lock before writing
+		lockInfo := statemgr.NewLockInfo()
+		lockInfo.Operation = "init"
+		lockId, err := foo.Lock(t.Context(), lockInfo)
+		if err != nil {
+			t.Fatalf("error locking foo: %s", err)
+		}
 		// write a known state to foo
 		if err := foo.WriteState(fooState); err != nil {
 			t.Fatal("error writing foo state:", err)
 		}
 		if err := foo.PersistState(t.Context(), nil); err != nil {
 			t.Fatal("error persisting foo state:", err)
+		}
+
+		// Unlock the state after writing
+		if err := foo.Unlock(t.Context(), lockId); err != nil {
+			t.Fatalf("error unlocking foo: %s", err)
 		}
 
 		// We'll make "bar" different by adding a fake resource state to it.
--- a/internal/backend/local/backend_test.go
+++ b/internal/backend/local/backend_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


internal/backend/testing.go
```diff
@@ -219,13 +219,25 @@ func TestBackendStates(t *testing.T, b Backend) {
...
+
 		// write a distinct known state to bar
 		if err := bar.WriteState(barState); err != nil {
 			t.Fatalf("bad: %s", err)
 		}
 		if err := bar.PersistState(t.Context(), nil); err != nil {
 			t.Fatalf("bad: %s", err)
 		}
+		// Unlock the state after writing
+		if err := foo.Unlock(t.Context(), lockId); err != nil {
```
yottta: This feels a little bit weird to have it here. Please correct me if I am wrong.
Wouldn't be better to have the locking done right after creating `foo` state manager? And then the unlocking right after all the tests (aka before deleting workspaces)?

And another confusing part, this `foo` lock/unlock pair wraps operations on `bar`. Maybe I am confused for nothing, but doesn't look right to me 🤔 
Could you help understanding why is it needed this way?

Thanks

diofeher: Hey, @yottta !

You're right, this works, but it's confusing. When I was writing the code, I thought I was wrapping around foo, but I probably placed it in the wrong spot due to poor cleanup.

It works just because any call after writing to the foo state would be enough, because there's a method for closing the `state` as stated on the PR's description. On Windows, if we don't close it (calling unlock), we can't call delete due to mandatory locking.

Let me know if that's clear for you. I created another commit to make this code more readable: https://github.com/opentofu/opentofu/pull/3284/commits/601fb6771790e3ae068d1bf01d2ed9eee0ed2999

yottta (approved):

PR #3294, opened 2025-09-22
fix: Prevent nil panic in marshalProviderConfigs when inSingleModuleMode

Yantrio:
When calling `tofu show -json -module=DIR` (introduced in #3003 and (called another module) it would cause a panic due to a nil reference.

This is because in single module mode we do not populate the children for the module. So it's impossible to recurse down

This commit aims to fix that by exiting a little bit earlier if we are in single module mode

---

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonconfig/config.go
+++ b/internal/command/jsonconfig/config.go
@@ -290,6 +290,16 @@ func marshalProviderConfigs(
 		m[key] = p
 	}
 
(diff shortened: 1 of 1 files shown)


yottta (approved):

cam72cam (approved):

diofeher:
I just saw this PR, but shouldn't it be a good idea to add an automated test to cover this change? @Yantrio

PR #3324, opened 2025-10-02
fix: typo on the function's comment

diofeher:
This function's comment is referring to another name instead of the name being used on the function.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/addrs/module_instance.go
+++ b/internal/addrs/module_instance.go
@@ -500,7 +500,7 @@ func (m ModuleInstance) HasSameModule(other ModuleInstance) bool {
 	return true
 }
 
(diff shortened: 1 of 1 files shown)


apparentlymart (approved):
If I recall correctly, I think this happened because the first version of this function was `func (m ModuleInstance) HasSameModule(other ModuleInstance)` -- comparing with another `ModuleInstance` rather than with a `Module` directly -- but then we realized that the callers we were trying to improve had been converting their `Module` into a `ModuleInstance` just to compare, so it was better to make this compare with `Module` directly instead.

As is often the case, updated everything except the redundant copy of the name at the start of the doc comment. 🙃

diofeher:
Thanks for the context @apparentlymart!

PR #3327, opened 2025-10-03
wip: implicit moves for modules

### LINKED ISSUE #3326: Add implicit move for modules like what's existing in resources

diofeher:
Resolves #3326 

This is a draft for implementing the implicit move statement for modules previously using `count` and then moving to use `enabled`, or for implicit use of `enabled`.

This is not working yet, but I'm planning to make it work first and then will refactor it.

## Checklist

### Go checklist

 


internal/tofu/context_plan2_test.go
```diff
@@ -3360,52 +3361,144 @@ output "output" {
...
-	// an entry in the resource address's _module path_ that we'll find
-	// missing, rather than the resource's own instance key, and so
-	// our analyses need to handle that situation by indicating that all
-	// of the resources under the missing module instance have zero
-	// instances, regardless of which resource in that module we might
-	// be asking about, and do so without tripping over any missing
-	// registrations in the instance expander that might lead to panics
-	// if we aren't careful.
-	//
-	// (For some history here, see https://github.com/hashicorp/terraform/issues/30110 )
```
diofeher, line 3380: This test can be removed because the module will be moved implicitly, avoiding orphaning the resources.

internal/refactoring/move_statement.go
```diff
@@ -86,11 +86,20 @@ func findMoveStatements(cfg *configs.Config, into []MoveStatement) []MoveStateme
 // We should think very hard before adding any _new_ implication rules for
 // moved statements.
 func ImpliedMoveStatements(rootCfg *configs.Config, prevRunState *states.State, explicitStmts []MoveStatement) []MoveStatement {
-	return impliedMoveStatements(rootCfg, prevRunState, explicitStmts, nil)
+	return impliedMoveStatements(rootCfg, prevRunState, explicitStmts)
```
cam72cam: We can merge the public/private function here as you have untangled the additional parameter.

internal/refactoring/move_statement.go
```diff
@@ -100,63 +109,123 @@ func impliedMoveStatements(cfg *configs.Config, prevRunState *states.State, expl
...
+func impliedMoveStatementsForModules(
+	prevRunState *states.State,
+	parentCfg *configs.Config,
+	modCallName string,
+	modCallCfg *configs.ModuleCall,
+	explicitStmts []MoveStatement) []MoveStatement {
+	var into []MoveStatement
+	var fromKey, toKey addrs.InstanceKey
+
+	parentState := prevRunState.Module(addrs.RootModuleInstance)
```
cam72cam: I'm curious how this works for module calls a few levels deep within the configuration?  I would expect to see a addrs.Module address used here?  Perhaps parentCfg.Path?

I don't see a test for this scenario, could ensure there is one?

cam72cam (changes requested):
Overall looks good!  One concern about how the code functions within multiple nested layers.

internal/refactoring/move_statement.go
diofeher: Good point! Gonna add in the next commit!

internal/refactoring/move_statement.go
diofeher: I had that on my to-do list, but overlooked it when cleaning up the PR :(

Thanks for the catch! I'm going to add the modification and a test covering this case.

diofeher: Addressed by https://github.com/opentofu/opentofu/pull/3327/commits/2302c891f9c82b32ab4771f6c6dec924b373cc3d

internal/refactoring/move_statement.go
diofeher: Addressed by https://github.com/opentofu/opentofu/pull/3327/commits/2302c891f9c82b32ab4771f6c6dec924b373cc3d

internal/tofu/context_plan2_test.go
```diff
@@ -3360,52 +3361,169 @@ output "output" {
...
+				"child/grandchild/main.tf": `resource "test_object" "a" {}`,
+			}),
+			addr:     mustResourceInstanceAddr("module.child.module.grandchild[0].test_object.a"),
+			prevAddr: mustResourceInstanceAddr("module.child.module.grandchild.test_object.a"),
+			prevState: states.BuildState(func(s *states.SyncState) {
+				s.SetResourceInstanceCurrent(mustResourceInstanceAddr("module.child.module.grandchild.test_object.a"), &states.ResourceInstanceObjectSrc{
+					AttrsJSON: []byte(`{}`),
+					Status:    states.ObjectReady,
+				}, mustProviderConfig(`provider["registry.opentofu.org/hashicorp/test"]`), addrs.NoKey)
+				s.SetResourceInstanceCurrent(mustResourceInstanceAddr("module.child.test_object.a"), &states.ResourceInstanceObjectSrc{
```
yottta: Please correct me if I am wrong, but why is here a resource in the previous state?

Debugging through the implementation in `impliedMoveStatementsForModules` for this particular case, the statement for the `grandchild` module call seems to be generated only because in the previous state, the `child` module is recorded strictly because of this resource.

Removing this resource from the previous state, fails the test case because the `grandchild` module does not get the expected statement.

Without the resource in the `child` module:

Image

With this resource in the `child` module:
Image

What this means, is that the call to `prevRunState.ModuleInstances(parentCfg.Path)` does not work as intended. It does not return the `child` module to check the calls inside of it to be able to generate a moved statement for the `grandchild` module call.

diofeher: Thanks for taking the time to debug @yottta! I'm actually debugging the exact implementation to see if there's another approach to these nested modules! I'll let you know as soon as I find something.

diofeher: I'm actually seeing three possible approaches here:

1. Trying to get the parent node by brute-forcing. We can't know if the instance key used here was an `addrs.NoKey` or an `addrs.IntKey(0)`, for example.
2. Saving the module information on the `prevRunState`, so we have module information there even if there are no resources/outputs related to it.
3. We do the support only for parent modules that have resources and state that as a limitation of the feature.

I'm in favor of the second approach, so that's what I'm investigating right now.

EDIT - James mentioned that having the empty module in the state could be a problem because it could break a lot of existing tooling.

diofeher: @yottta Found another approach with Christian's help: https://github.com/opentofu/opentofu/pull/3327/commits/0268a8b72d63438cee079ed2b7a833bd7f42c4a4

cc @cam72cam 

ps: https://github.com/opentofu/opentofu/commit/0268a8b72d63438cee079ed2b7a833bd7f42c4a4#diff-d6a0bc0942ee242a1e6e68853890287a93c8c3dc16bf18db8122265e21d45248R188-R189
Adding a resource is a different approach here; it's just to put the child module on the state, but not a parent with a resource, as I was doing before.

internal/refactoring/move_statement.go
```diff
@@ -100,63 +105,141 @@ func impliedMoveStatements(cfg *configs.Config, prevRunState *states.State, expl
...
+		absCallAddr := addrs.AbsModuleCall{
+			Module: callerAddr,
+			Call:   callAddr.Call,
+		}
+
+		// Only one instance of the module can be dealt, because moving from single
+		// to repeated can't support multiple instances. The other instances are going
+		// to be removed or created, depending on the direction of the change.
+		fromKey := callAddr.Key
+		if fromKey != addrs.NoKey && fromKey != addrs.IntKey(0) {
```
diofeher, line 161: This code was added to avoid this error:

```
│ Error: Ambiguous move statements
│
│   on child/main.tf line 1:
│    1: module "grandchild" {
│
│ A statement at child/main.tf:1,1 declared that module.child.module.grandchild[0] moved to module.child.module.grandchild,
│ but this statement instead declares that module.child.module.grandchild[1] moved there.
│
│ Each module instance can have moved from only one source instance.
```

internal/tofu/context_plan2_test.go
yottta: Now looks better! Thanks for checking this.
Could we remove the second resource from the test though? That just adds confusion.

diofeher: Addressed on https://github.com/opentofu/opentofu/pull/3327/commits/42b77d0ac33fbfc36ca414790a10dbf9ae0d0dd3

Let me know if that's what you meant! Thanks again for the detailed review :)

yottta: Exactly what I meant! Thanks. Now it's more clearer.

internal/tofu/context_plan2_test.go
```diff
@@ -3360,52 +3361,165 @@ output "output" {
...
-			Status:    states.ObjectReady,
-		}, mustProviderConfig(`provider["registry.opentofu.org/hashicorp/test"]`), addrs.NoKey)
-	})
+func TestContext2Plan_moduleImplicitMove(t *testing.T) {
+	// Modules are being moved implicitly to use the `enabled` field when nothing
+	// is declared on the block. Alternatively, they are implicitly being moved from
+	// using `enabled` as true or without declaring `enabled` to use count.
+	var tests = map[string]struct {
+		name      string
+		addr      addrs.AbsResourceInstance
```
yottta:
```suggestion
		expectedAddr      addrs.AbsResourceInstance
```

yottta (approved):
Looks ok now FMPOV. Well done! 🙌

PR #3366, opened 2025-10-12
create automated tests for the internal/flock package

### LINKED ISSUE #3093: test: create automated tests for the `internal/flock` package
During the work on https://github.com/opentofu/opentofu/pull/3090, we noticed we don't have any automated tests covering that work.

We should create tests for both for windows and unix functionality.

baa-ableton:
Resolves #3093 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/flock/filesystem_lock_windows.go
+++ b/internal/flock/filesystem_lock_windows.go
@@ -92,7 +92,6 @@ func LockBlocking(ctx context.Context, f *os.File) error {
 
 func Unlock(*os.File) error {
 	// the lock is released when Close() is called
-	// TODO: Consider together with maintainer whether this should return f.Close()
 	return nil
 }
 


internal/flock/filesystem_lock_windows.go
```diff
@@ -96,7 +96,7 @@ func Unlock(*os.File) error {
 	return nil
 }
 
-func lockFileEx(h syscall.Handle, flags, reserved, locklow, lockhigh uint32, ol *syscall.Overlapped) (err error) {
+func lockFileEx(h syscall.Handle, flags, reserved, locklow, lockhigh uint32, ol *syscall.Overlapped) err error {
```
diofeher: There's a syntax error in the declaration of `lockFileEx`.

baa-ableton: my bad, just pushed the fix

internal/flock/filesystem_lock_test.go
```diff
@@ -0,0 +1,357 @@
...
+	// Test unlocking
+	err = Unlock(f)
+	if err != nil {
+		t.Fatalf("Failed to unlock: %v", err)
+	}
+}
+
+func TestLock_Contention(t *testing.T) {
+	// Tests back-to-back Lock calls on a single file
+
```
diofeher, line 46: This test is breaking on Windows:

```
--- FAIL: TestLock_Contention (0.00s)
    filesystem_lock_test.go:90: Second lock failed as expected - true contention detected: The process cannot access the file because another process has locked a portion of the file.
    filesystem_lock_test.go:102: Second lock should succeed after first unlock: The process cannot access the file because another process has locked a portion of the file.
```

internal/flock/filesystem_lock_test.go
```diff
@@ -0,0 +1,357 @@
...
+	err = Unlock(f)
+	if err != nil {
+		t.Fatalf("Failed to unlock: %v", err)
+	}
+}
+
+func TestLockBlocking_Cancellation(t *testing.T) {
+	// Tests cancellation of LockBlocking on a single file while a Lock is already in place
+	// Doesn't really test anything in POSIX systems since the same process can hold multiple locks
+	// on the same file
```
diofeher, line 163: This test is breaking on Windows:

```
    filesystem_lock_test.go:200: System supports real contention between same-process handles
    filesystem_lock_test.go:221: Lock was properly cancelled: context deadline exceeded (took 202.3348ms)
    filesystem_lock_test.go:225: Cancellation took too long: 202.3348ms
```

internal/flock/filesystem_lock_test.go
```diff
@@ -0,0 +1,357 @@
...
+	} else {
+		t.Fatalf("Expected timeout/cancellation error, got: %v", err)
+	}
+}
+
+func TestLockBlocking_EventualSuccess(t *testing.T) {
+	// Tests eventual success of LockBlocking on a single file while a Lock is 
+	// already in place which is then released
+	// Doesn't really test anything in POSIX systems since the same process can hold multiple locks
+	// on the same file
```
diofeher, line 238: This test is panicking on Windows:

```
panic: test timed out after 10m0s
	running tests:
		TestLockBlocking_EventualSuccess (10m0s)

goroutine 16 [running]:
testing.(*M).startAlarm.func1()
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:2682 +0x345
created by time.goFunc
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/time/sleep.go:215 +0x2d

goroutine 1 [chan receive]:
testing.(*T).Run(0xc000003500, {0x7ff618a4756f?, 0x7ffac65a3fb0?}, 0x7ff618a4e568)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:2005 +0x469
testing.runTests.func1(0xc000003500)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:2477 +0x37
testing.tRunner(0xc000003500, 0xc000049c70)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:1934 +0xc3
testing.runTests(0xc0000080f0, {0x7ff618b8b700, 0x6, 0x6}, {0x7?, 0xc00005c8c0?, 0x7ff618b90720?})
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:2475 +0x4b4
testing.(*M).Run(0xc000060140)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:2337 +0x63a
main.main()
	_testmain.go:55 +0x9b

goroutine 13 [sync.WaitGroup.Wait]:
sync.runtime_SemacquireWaitGroup(0xc000036800?, 0xa0?)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/runtime/sema.go:114 +0x2e
sync.(*WaitGroup).Wait(0xc00000a460)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/sync/waitgroup.go:206 +0x85
github.com/opentofu/opentofu/internal/flock.TestLockBlocking_EventualSuccess(0xc0001421c0)
	D:/a/opentofu/opentofu/internal/flock/filesystem_lock_test.go:279 +0x2c5
testing.tRunner(0xc0001421c0, 0x7ff618a4e568)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:1934 +0xc3
created by testing.(*T).Run in goroutine 1
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/testing/testing.go:1997 +0x44b

goroutine 14 [chan receive]:
github.com/opentofu/opentofu/internal/flock.LockBlocking({0x7ff618a7eed8, 0x7ff618bd79e0}, 0xc00004c148)
	D:/a/opentofu/opentofu/internal/flock/filesystem_lock_windows.go:91 +0xba
github.com/opentofu/opentofu/internal/flock.TestLockBlocking_EventualSuccess.func1()
	D:/a/opentofu/opentofu/internal/flock/filesystem_lock_test.go:268 +0x65
created by github.com/opentofu/opentofu/internal/flock.TestLockBlocking_EventualSuccess in goroutine 13
	D:/a/opentofu/opentofu/internal/flock/filesystem_lock_test.go:265 +0x2ab

goroutine 15 [sleep]:
time.Sleep(0x5f5e100)
	C:/hostedtoolcache/windows/go/1.25.2/x64/src/runtime/time.go:363 +0x167
github.com/opentofu/opentofu/internal/flock.LockBlocking.func1()
	D:/a/opentofu/opentofu/internal/flock/filesystem_lock_windows.go:82 +0xbe
created by github.com/opentofu/opentofu/internal/flock.LockBlocking in goroutine 14
	D:/a/opentofu/opentofu/internal/flock/filesystem_lock_windows.go:62 +0xa5
```

baa-ableton:
@diofeher 

I noticed the windows part of this code is different than what I had anticipated:
```
func Unlock(*os.File) error {
	// the lock is released when Close() is called
	return nil
}
```

Unlock doesn't do anything which is why all the tests are failing for windows I believe 🪟 . I want to get your take on how to proceed. I can change the tests so that the file descriptor is closed instead of calling Unlock - in the case of Windows systems. 

I could also attempt at implementing Unlock for windows which would surpass the original scope of the issue which is why I'd like to get your take. Please LMK

diofeher:
Sorry for taking so long to get back to you @baa-ableton! I should have time later this week, and I'll look this PR in detail.

diofeher:
@baa-ableton Sorry for taking so long to get back to you. Just had another look at your PR. 

It would be ideal if you could run these tests using the same code on both platforms. I know there are some caveats on Windows systems due to Mandatory Locking. Can you implement the tests without changing the Windows implementation?

Sorry again for taking so long to get back to you and thanks for the contribution!

baa-ableton:
@diofeher 
hey no worries. I just tested on my Windows machine and the tests are failing due to locks not being released. I'll look into it tomorrow

internal/flock/filesystem_lock_windows.go
```diff
@@ -92,6 +92,7 @@ func LockBlocking(ctx context.Context, f *os.File) error {
 
 func Unlock(*os.File) error {
 	// the lock is released when Close() is called
+	// TODO: Consider together with maintainer whether this should return f.Close()
```
baa-ableton: @diofeher FYI

baa-ableton:
Hey @diofeher I made sure the tests are passing on my windows machine. It had to do with the Unlock in Windows needing a os.File.close().

I also left a note to discuss with you and potentially other maintainers about this

internal/flock/filesystem_lock_windows.go
diofeher: That's a good point! I'm going to try to gather some historical context on why it wasn't implemented that way, because I don't think the best approach would be to do that in the test file (implementing instead of testing). Still, there may be a reason not to close the file on the Unlock. I'll try to get back to you ASAP

diofeher: I re-checked this, and `Close` shouldn't be called in this function. You can remove this TODO.

internal/flock/filesystem_lock_test.go
```diff
@@ -0,0 +1,363 @@
...
+	//   - CIFS locking can have different semantics than local filesystems
+	err = Lock(f2)
+	if err != nil {
+		// Case 1: True contention detected (expected on some systems or configurations)
+		t.Logf("Second lock failed as expected - true contention detected: %v", err)
+		
+		// Verify that unlocking the first handle allows the second to succeed
+		// We are likely in a Windows OS now and our Unlock implementation for Windows is a no-op
+		// As commented there, we need to use Close
+		err = f1.Close()
```
diofeher, line 96: Question, I just noticed this `f1.Close()` is called above with a `defer`. Do you want to duplicate this call here? What do you think we should do with the first call at line 55?

internal/flock/filesystem_lock_test.go
```diff
@@ -0,0 +1,363 @@
...
+	go func() {
+		defer wg.Done()
+		ctx := context.Background()
+		lockErr = LockBlocking(ctx, f2)
+	}()
+
+	// Release first lock after short delay
+	time.Sleep(50 * time.Millisecond)
+	if runtime.GOOS == "windows" {
+		err = f1.Close()
```
diofeher, line 276: Question, I just noticed this `f1.Close()` is called above with a `defer`. Would you like to duplicate this call here? What should we do with the first call at line 248?

internal/flock/filesystem_lock_test.go
diofeher: Fixed, thanks!

internal/flock/filesystem_lock_test.go
diofeher: Fixed, thanks!

internal/flock/filesystem_lock_test.go
diofeher: Fixed, thanks!

internal/flock/filesystem_lock_test.go
diofeher: By the way, I think these are the last remaining issues; we should be ready to merge after fixing those two!

baa-ableton: @diofeher That's a great point. We probably shouldn't duplicate the call, delete the defer at line 55 and then manually close the file at the end of the POSIX flow, at the bottom of the `else` clause. I don't think we need to implement something similar for `f2` however. What do you think, LMK.

internal/flock/filesystem_lock_test.go
baa-ableton: hmm I see how there is 2 instances of the same problem, maybe we can generalize a solution. Please consider what I propose in this comment for the issue you mentioned earlier in the file/conversation:

For all first files opened `f1`, we can replace the current defers with the following:
```go
defer func () {
// Tests have logic built-in to call Close manually if OS is windows
    if runtime.GOOS != "windows" {
        f1.Close()
    }
}()
```

So for the test `TestLock_Contention` I could simply implement this defer for `f1` and not implement the f1.Close() at the bottom of the `else` clause - like I mentioned above.

For the test that is relevant to this comment thread, `TestLockBlocking_EventualSuccess`, we just replace the defer for `f1` with this block and as far as I can see, everything will still work out of the box.

internal/flock/filesystem_lock_test.go
diofeher: What you proposed makes sense. We can't have a single `close` call because of different behavior between Windows/Unix, so let's do it like that!

Since `f2` does not impact the behavior of the test, just having a single `defer f2.close()` should be good enough.

internal/flock/filesystem_lock_test.go
baa-ableton: @diofeher FYI^ please lmk about this as well

diofeher: That's fair. Actually, the solution of placing the `close` in the else is bad because if errors occur, the file won't be closed. In the end, I think the best solution would be to keep what we currently have rather than creating the block! lol

diofeher: I'll quickly test something on Windows and then get back to you.

internal/flock/filesystem_lock_test.go
diofeher: If the file is already closed, we're not checking for the errors, so it should be good (when defer is called). Let's leave it as is.

internal/flock/filesystem_lock_test.go
diofeher: If the file is already closed, we're not checking for the errors, so it should be good (when defer is called). Let's leave it as is.

diofeher:
Thanks for your contribution, @baa-ableton!

internal/flock/filesystem_lock_windows.go
diofeher: You can remove this line and we should be good to go :)

baa-ableton: @diofeher Done

diofeher (approved):

PR #3385, opened 2025-10-16
go.mod: go get cloud.google.com/go/storage@v1.57.0

yottta:
This library is a dependency of the `gcs` backend.
While updating this, `cloud.google.com/go/kms` got updated to (not to the latest version) and that is used by the state encryption.

After the update, the tests for the encryption key using GCP KMS works ok. With and without SA impersonation.
For the `gcs` backend, I had to update the error handling of the bucket setup since the NotFound error when getting a bucket now is wrapped starting with cloud.google.com/go/storage@1.51.0 ([googleapis/google-cloud-go#11519](https://github.com/googleapis/google-cloud-go/pull/11519)).
In the same note, wanted to change the error handling in [`Workspace()` method](https://github.com/opentofu/opentofu/blob/9eeca90c1985a66c249b0ddbacdd4acb9c7b6642/internal/backend/remote-state/gcs/backend_state.go#L41) too, but that's out of scope for this PR. The [official docs of storage@1.57.0](https://pkg.go.dev/cloud.google.com/go/storage#BucketIterator.Next) use `iterator.Done` error to compare against, and their snippet is still following the same approach (not using `errors.Is()`).

We should be safe nonetheless because the integration tests for the `gcs` backend cover listing of the workspaces in several ways. In the future, if the `iterator.Done` will get wrapped, those tests will definitely fail.

As a safety measure, I tested this with one of my tofu configurations, with a user that impersonated KMS key usage through a SA and also impersonated the same SA for reading/writing to the state dedicated GCS bucket.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/gcs/client.go
+++ b/internal/backend/remote-state/gcs/client.go
@@ -34,7 +34,7 @@ type remoteClient struct {
 func (c *remoteClient) Get(ctx context.Context) (payload *remote.Payload, err error) {
 	stateFileReader, err := c.stateFile().NewReader(ctx)
 	if err != nil {
-		if err == storage.ErrObjectNotExist {
+		if errors.Is(err, storage.ErrObjectNotExist) {
 			return nil, nil
 		} else {
 			return nil, fmt.Errorf("Failed to open state file at %v: %w", c.stateFileURL(), err)


diofeher (changes requested):
> I had to update the error handling of the bucket setup since the NotFound error when getting a bucket now is wrapped starting with cloud.google.com/go/storage@1.51.0 (https://github.com/googleapis/google-cloud-go/pull/11519).

Good catch!

> In the same note, wanted to change the error handling in [Workspace() method](https://github.com/opentofu/opentofu/blob/9eeca90c1985a66c249b0ddbacdd4acb9c7b6642/internal/backend/remote-state/gcs/backend_state.go#L41) too, but that's out of scope for this PR. The [official docs of storage@1.57.0](https://pkg.go.dev/cloud.google.com/go/storage#BucketIterator.Next) use iterator.Done error to compare against, and their snippet is still following the same approach (not using errors.Is()).

From what I understood reading the linked issue, you don't need to change the approach on iterator.Done, but only on `ErrObjectNotExist` and `ErrBucketNotExist`.

Besides that, I found one possible place that should be fixed as well: 

https://github.com/opentofu/opentofu/blob/d42dcc3b0a222a3b7ca76a848506b777b0d2b0d8/internal/backend/remote-state/gcs/client.go#L37

Can you double-check if that's right?

yottta:
> > I had to update the error handling of the bucket setup since the NotFound error when getting a bucket now is wrapped starting with cloud.google.com/go/storage@1.51.0 ([googleapis/google-cloud-go#11519](https://github.com/googleapis/google-cloud-go/pull/11519)).
> 
> Good catch!
> 
> > In the same note, wanted to change the error handling in [Workspace() method](https://github.com/opentofu/opentofu/blob/9eeca90c1985a66c249b0ddbacdd4acb9c7b6642/internal/backend/remote-state/gcs/backend_state.go#L41) too, but that's out of scope for this PR. The [official docs of storage@1.57.0](https://pkg.go.dev/cloud.google.com/go/storage#BucketIterator.Next) use iterator.Done error to compare against, and their snippet is still following the same approach (not using errors.Is()).
> 
> From what I understood reading the linked issue, you don't need to change the approach on iterator.Done, but only on `ErrObjectNotExist` and `ErrBucketNotExist`.
> 
> Besides that, I found one possible place that should be fixed as well:
> 
> https://github.com/opentofu/opentofu/blob/d42dcc3b0a222a3b7ca76a848506b777b0d2b0d8/internal/backend/remote-state/gcs/client.go#L37
> 
> Can you double-check if that's right?

Great catch! Somehow I managed to miss exactly that 🤦 
Fixed in a4a503d5e55c0a922ac25e7115aa9b4543bac559

diofeher (approved):
🚀

PR #3403, opened 2025-10-21
deps: upgrade Azure-related libraries

diofeher:
Related to #3386

- [x] azcore: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/CHANGELOG.md#1191-2025-09-11
- [x] azidentity: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/CHANGELOG.md#1130-2025-10-07
- [x] armstorage: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/storage/armstorage/CHANGELOG.md#200-2025-09-25
- [x] azblob: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/CHANGELOG.md#163-2025-10-16

I couldn't find significant differences in what we're using from these dependencies.
I've fixed a regression in the acceptance tests that only supported `az cli`, so now it can also support environment variables. 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/azure/README.md
+++ b/internal/backend/remote-state/azure/README.md
@@ -26,6 +26,7 @@ export ARM_TENANT_ID='xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
 ```
 
 We recommend using the Azure CLI (`az`) to authenticate to Azure for the infrastructure bootstrapping steps, which create a resource group, storage account, and blob storage container in your configured subscription.
+It's also possible to run these tests by setting TF_AZURE_TEST_CLIENT_ID and TF_AZURE_TEST_CLIENT_SECRET if the Azure CLI cannot be used.
 
 With all of these configured, you'll be able to run the following tests:
 - TestAccBackendAccessKeyBasic
@@ -41,7 +42,7 @@ To run the secrets test, you will need these variables:
 
 ```bash
 export TF_AZURE_TEST_CLIENT_ID=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
-export TF_AZURE_TEST_SECRET=some~secret~string
+export TF_AZURE_TEST_CLIENT_SECRET=some~secret~string
 ```
 
 You can get those by going into the meta-test directory and following the instructions there, or manually obtaining a secret client. Instructions for manually creating a Client Secret for your existing app registration can be found [here](https://learn.microsoft.com/en-us/azure/industry/training-services/microsoft-community-training/public-preview-version/frequently-asked-questions/generate-new-clientsecret-link-to-key-vault#check-and-update-client-secret-expiration-date) in the first part of this guide. You do not need to put the secret in a vault or update the application configuration.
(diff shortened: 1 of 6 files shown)


lbordowitz:
I want to bring your attention to [these two env vars](https://github.com/opentofu/opentofu/pull/3403/files#diff-e3f0603e4252c5616a78e81f21bb3e2d4d9c760e1e6d192ee7f18a64eca58c98R312-R313) in one of the tests. I intentionally used different names than the ones used for the actual tests. This solves the problem of environment isolation in tests.

When you run `backend.TestBackendConfig`, it calls the `ConfigureFunc` defined in the backend schema. That means every default value function is called, including those which obtain environment variables. By setting `ARM_CLIENT_ID` and `ARM_CLIENT_SECRET`, you may accidentally trigger the test to use client credentials instead of the intended auth acceptance method.

This is not hypothetical. In the previous implementation, there was an OIDC backend acceptance test which never once went through the OIDC auth flow, and did not even require an OAuth token to run.

diofeher:
@lbordowitz Good point!

> I want to bring your attention to [these two env vars](https://github.com/opentofu/opentofu/pull/3403/files#diff-e3f0603e4252c5616a78e81f21bb3e2d4d9c760e1e6d192ee7f18a64eca58c98R312-R313) in one of the tests. I intentionally used different names than the ones used for the actual tests. This solves the problem of environment isolation in tests.

I was able to test `TestAccBackendServicePrincipalClientSecret` but didn't test `TestAccBackendServicePrincipalClientCertificate` because I don't have proper certificate access, and I can't create one :(

> When you run backend.TestBackendConfig, it calls the ConfigureFunc defined in the backend schema. That means every default value function is called, including those which obtain environment variables. By setting ARM_CLIENT_ID and ARM_CLIENT_SECRET, you may accidentally trigger the test to use client credentials instead of the intended auth acceptance method.

That's a good point. We need to better separate this code from the issue of not using the wrong authentication method in these acceptance tests. Let me think about what a better approach would be here.

EDIT: Continuing our conversation from Slack, it makes sense to use two different environment variable names for these client credentials and properly document them. Gonna work on it, thanks for the review!

cam72cam (approved):

diofeher:
I'm gonna wait for @lbordowitz's approval before merging this one.

internal/backend/remote-state/azure/README.md
```diff
@@ -26,6 +26,7 @@ export ARM_TENANT_ID='xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
 ```
 
 We recommend using the Azure CLI (`az`) to authenticate to Azure for the infrastructure bootstrapping steps, which create a resource group, storage account, and blob storage container in your configured subscription.
+It's also possible to run these tests by setting TF_AZURE_TEST_CLIENT_ID and TF_AZURE_TEST_CLIENT_SECRET if the Azure CLI cannot be used.
```
lbordowitz, line 29: Thank you for adding this.

PR #3421, opened 2025-10-23
Ensure we use goreleaser-pro for release build workflow

Yantrio:
We currently use goreleaser pro to build and deploy opentofu. However, when we introduced the new `nightlies` feature, we forgot to update our `release.yml` github workflow file to use the go-releaser pro version. This resulted in goreleaser saying that the `nightly` value in the yaml file is unknown.


yottta:
For posterity: this change was needed because the release workflow was using the non-pro version, which when encounters a ["nightly" section](https://github.com/opentofu/opentofu/blob/eff8f3faba6e770175a85ffd60735b9fb0f5490d/.goreleaser.yaml#L353) in the configuration file, fails with undefined configuration, since that's a feature only available in the pro version.

PR #3435, opened 2025-10-27
cliconfig: Create testable shim for fs actions

lbordowitz:
This is the first part of addressing the problem of nonstandard CLI configuration directories in OpenTofu. See #2655.

Before any functionality is changed, the current behavior of how CLI configuration files are loaded should be thoroughly tested. We do have `cliconfig` tests already, but those heavily depend on environment variables. Moreover, they do not consider the precedence with which different configurations might be loaded. Finally, they don't actually check the local filesystem; this is probably because a test looking at the real file system of the developer isn't feasible.

Therefore, I've created a mockable shim for the filesystem and added regression tests looking at real locations. My original pull request implementing this, #3337, had some architecturally dubious design decisions. This present change implements a more surgical change, never venturing outside of the `cliconfig` package.

Maybe "shim" isn't the right word; is there existing terminology for this testing practice? I also welcome requests for changes to make this code more idiomatic.

A future pull request will add functionality differences between Mac and Unix. So, while the test files are completely identical in this PR, the next change will cause them to diverge.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/cliconfig/cliconfig.go
+++ b/internal/command/cliconfig/cliconfig.go
@@ -496,16 +496,17 @@ func (cl *ConfigLoader) cliConfigFile() (string, tfdiags.Diagnostics) {
 		}
 	}
 
-	log.Printf("[DEBUG] Attempting to stat CLI config file: %s", configFilePath)
-	_, err := cl.Stat(configFilePath)
+	log.Printf("[DEBUG] Attempting to open CLI config file: %s", configFilePath)
+	f, err := cl.Open(configFilePath)
 	if err == nil {
+		f.Close()
 		return configFilePath, diags
 	}
 
 	if mustExist || !errors.Is(err, fs.ErrNotExist) {
 		diags = append(diags, tfdiags.Sourceless(
 			tfdiags.Warning,
-			"Unable to stat CLI configuration file",
+			"Unable to open CLI configuration file",
 			fmt.Sprintf("The CLI configuration file at %q does not exist.", configFilePath),
 		))
 	}
--- a/internal/command/cliconfig/config_loader.go
+++ b/internal/command/cliconfig/config_loader.go
@@ -1,17 +1,33 @@
 // Copyright (c) The OpenTofu Authors
 // SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
 
 package cliconfig
 
-import "os"
+import (
+	"io/fs"
+	"os"
+)
 
+// ConfigFileSystem is an abstraction layer for file system operations
+// involved in loading the CLI configuration.
+//
+// The logic of choosing CLI configuration files is highly dependent on
+// the location of those files in the filesystem. Since it's not feasible
+// to test that in a real filesystem, we separate the filesystem operations
+// from the logic of how the files are chosen. The real CLI config loader
+// uses standard os package functions, and the test config loader uses
+// fstest-based functions.
 type ConfigFileSystem interface {
-	// ReadFile can read files
-	ReadFile(name string) ([]byte, error)
-	// Stat can stat files
-	Stat(name string) (os.FileInfo, error)
-	// ReadDir can read files in a directory
+	// Open opens the named file for reading.
+	Open(name string) (fs.File, error)
+	// ReadDir reads the named directory, returning all its directory entries sorted by filename.
 	ReadDir(name string) ([]os.DirEntry, error)
+	// ReadFile reads the named file and returns the contents.
+	ReadFile(name string) ([]byte, error)
+	// Stat returns a [FileInfo] describing the named file. If there is an error, it will be of type [*PathError].
+	Stat(name string) (os.FileInfo, error)
 }
 
 type standardFileSystem struct{}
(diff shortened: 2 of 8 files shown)


internal/command/cliconfig/config_loader.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+
+package cliconfig
+
+import "os"
+
+type ConfigFileSystem interface {
```
yottta, line 22: For the posterity, I would suggest to add a comment on this type explaining the motivation and what's meant to achieve.

internal/command/cliconfig/config_locations_macos_test.go
```diff
@@ -0,0 +1,294 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"testing/fstest"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func (tfs *testFileSystem) trim(name string) string {
```
yottta, line 22: `os.PathSeparator` is already built based on the architecture:

https://pkg.go.dev/cmd/go#hdr-Build_constraints
Image

Was there a different reason of having this in here instead of `config_locations_helper_test.go`?

internal/command/cliconfig/config_locations_macos_test.go
```diff
@@ -0,0 +1,294 @@
...
+						"modules.v0": "https://config0.example.com/",
+					},
+				},
+			},
+		},
+		{
+			locationTestParameters: locationTestParameters{
+				name:  ".tofurc and .terraformrc",
+				files: []string{filepath.Join(home, ".terraformrc"), filepath.Join(home, ".tofurc")},
+			},
```
Gogotchuri: This is a small thing, but on first reading this test case, I got confused.
The name is ".tofurc and .terraformrc" and I assumed without reading the files property that this was the order of files for the test, and I thought the "expected" part was a mistake, because it was taking the index 1 instead of index 0 (Which, due to the name, in my mind, was .tofurc).
Suggestion:

```suggestion
				name:  ".terraformrc and .tofurc",
				files: []string{filepath.Join(home, ".terraformrc"), filepath.Join(home, ".tofurc")},
			},
```

internal/command/cliconfig/config_locations_macos_test.go
```diff
@@ -0,0 +1,294 @@
...
+	"testing/fstest"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func (tfs *testFileSystem) trim(name string) string {
+	return strings.TrimLeft(name, string(os.PathSeparator))
+}
+
+var common_env_vars = []string{
```
yottta:
```suggestion
var commonEnvVars = []string{
```

Gogotchuri (approved):
This is definitely cleaner and a good start.
I have a tiny comment, but overall this looks awesome!

internal/command/cliconfig/config_loader.go
Gogotchuri: Agreed. 
When reading this as a part of this PR with the whole context in mind, it definitely doesn't need any additional explanation.
But when you stumble upon it without additional context, details of why will definitely help.

internal/command/cliconfig/cliconfig.go
```diff
@@ -476,17 +496,16 @@ func cliConfigFile() (string, tfdiags.Diagnostics) {
 		}
 	}
 
-	log.Printf("[DEBUG] Attempting to open CLI config file: %s", configFilePath)
-	f, err := os.Open(configFilePath)
+	log.Printf("[DEBUG] Attempting to stat CLI config file: %s", configFilePath)
+	_, err := cl.Stat(configFilePath)
```
cam72cam: This changes the sematics slightly.  You need to check the stat result to ensure it's a file and not a directory.  I'm also not 100% sure, but I suspect that Open() also does a more comprehensive permission check?

internal/command/cliconfig/config_loader.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) The OpenTofu Authors
```
cam72cam, line 1: Our general guidance is to put the full header here, even if it's a brand new file.  This ensures that if anyone moves code into this file down the line or refactors it, the header is all-encompassing.

It definitely feels odd to do as you wrote this whole thing independently of the rest of the codebase, but that's what we've been advised to do at this point.

internal/command/cliconfig/config_unix.go
```diff
@@ -84,7 +84,7 @@ func homeDir() (string, error) {
 	return user.HomeDir, nil
 }
 
-func pathExists(path string) bool {
-	_, err := os.Stat(path)
+func pathExists(cfs ConfigFileSystem, path string) bool {
```
cam72cam: Why is this not a function on (cl *ConfigLoader)?

internal/command/cliconfig/config_locations_macos_test.go
lbordowitz: Windows is different; it has to trim out the drive letter.

yottta: Oh, good point. Now I just realised that I checked only the unix and macos specific files.
And the next question would be "why do we have separate files for macos and unix?" but I feel that the answer is already given in the description of the PR: "while the test files are completely identical in this PR, the next change will cause them to diverge".

lbordowitz: Yup. After this is merged, I'm going to quickly follow up with another PR that changes functionality. The goal of this PR is to set up regression tests to make sure current behavior continues to be supported regardless of changes to the underlying logic that I make in the next change.

PR #3445, opened 2025-10-29
Unify core functions address handling

yottta:
This tackles a recently added TODO to remove some of the duplication on handling core functions addresses.
This commit removes the const `lang.CoreNamespace` and replaces its usage with the actual `addrs.Function` struct that is meant to be used to parse and work with function addresses.

Before doing the changes, checked that there are enough tests and added others where the coverage was insufficient to guarantee no regression.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/addrs/provider_function.go
+++ b/internal/addrs/provider_function.go
@@ -62,9 +62,13 @@ func (f Function) IsNamespace(namespace string) bool {
 	return len(f.Namespaces) > 0 && f.Namespaces[0] == namespace
 }
 
-// CoreNamespaced returns the function with the [FunctionNamespaceCore] namespace
-// if has no namespace already defined.
-func (f Function) CoreNamespaced() Function {
+// FullyQualified returns a new [Function] where the [Function.Namespaces] is guaranteed to be filled.
+// For the functions that already have a namespace defined (e.g.: provider::test::func, core::tolist, etc), this
+// method will return the object that was called on.
+// For the functions that have no namespace defined (tolist, tomap, ephemeralasnull, sensitive, etc), this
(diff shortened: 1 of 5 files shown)


internal/addrs/provider_function.go
```diff
@@ -62,6 +62,18 @@ func (f Function) IsNamespace(namespace string) bool {
 	return len(f.Namespaces) > 0 && f.Namespaces[0] == namespace
 }
 
+// CoreNamespaced returns the function with the [FunctionNamespaceCore] namespace
+// if has no namespace already defined.
+func (f Function) CoreNamespaced() Function {
```
cam72cam: I like the idea behind this function, but the name makes reading the code a bit confusing.

Based on the name, I think "put this function in the core namespace", which is not what it does.

func.WithNormalizedNamespace() maybe?  I'm terrible at naming things so take this all with a grain of salt.

cam72cam:
Other than naming concerns, I like it!

internal/addrs/provider_function.go
yottta: Renamed it to `FullyQualified` on ced344f79e53043e044b8e6e12bebdc85d80de01. Please let me know if that describes better the intent of it.

cam72cam: LGTM!

cam72cam (approved):

apparentlymart (approved):
This seems like a nice compromise!

I do wonder if in future we could change `addrs.ParseFunction` to "normalize" with the core prefix immediately, just like we do with some of the other address types that have multiple equivalent ways of writing the same thing, but that's a riskier change so I'm happy to take this for now and wait to see if we want to change it more at some later time.

PR #3471, opened 2025-11-05
[RFC]Proposal to remove `Meta` struct, rework flags handling and pave the way for a possible new CLI layer

yottta:
[Rendered version](https://github.com/opentofu/opentofu/blob/cli-meta-refactor-rfc/rfc/20251105-refactor-tofu-cli.md)

To enable OpenTofu to iterate faster and to be able to add more functionalities on the CLI layer, this RFC attempts to initiate the refactoring process of the whole startup layer:
* Extract and reorder the content from the [`realMain`](https://github.com/opentofu/opentofu/blob/6a8e73110e8cdf95a4fc7d8a7d68d39b9bb700ae/cmd/tofu/main.go#L72) function to allow execution of specific bits to be deferred to the place where those are actually needed.
* Extract the definition, parsing and validation of the flags in components that can isolate the logic driven by logically grouped flags.
* In the end, break the "god" [`Meta`](https://github.com/opentofu/opentofu/blob/6a8e73110e8cdf95a4fc7d8a7d68d39b9bb700ae/internal/command/meta.go#L50) structure and have each command depend strictly on the components it needs.

As also mentioned in the content of the RFC, there is no single way of approaching such a refactor, so this RFC attempts to collect and emphasise the requirements and boundaries of such an action.

The proposed steps for refactoring are purposely vague since the complexity of the whole startup layer is quite high. Hence, there cannot be a clear path forward on how this can be performed.

## Checklist

### Go checklist

 

### DIFF
--- a/rfc/20251105-refactor-tofu-cli.md
+++ b/rfc/20251105-refactor-tofu-cli.md
@@ -174,8 +174,16 @@ Then, there are some bits that depend mostly (if not solely) on flags and defaul
   * this is mainly dependent on the 2 components that we described above: UI and Workdir. Once those will be extracted,
     this can be extracted too.
 
-So starting with the least dependent bits will allow walking down the chain and extract everything in separate
+Therefore, starting with the least dependent bits will allow walking down the chain and extract everything in separate
 components that can be instantiated inside the `Run` method of the commands.
+
+> [!NOTE]
+> About the `ui/view`. After did some work proposed by this RFC, became clearer that we can continue without
+> the migration of all commands to the Views concept from the old UI, but in doing so will add more shim
+> code to make things work properly.
+> 
+> Therefore, as part of this RFC, we can carry on the migration of all the commands to the Views abstraction,
+> which will make any subsequent work way easier.
 
 During this work, all logically grouped flags should be moved to these structs and allow the struct logic to
 record these on a particular FlagSet or parse the values for those directly.


yottta:
Thanks a lot for your review @apparentlymart.

> The general idea of [package arguments](https://pkg.go.dev/github.com/opentofu/opentofu/internal/command/arguments) is to centralize all of the functions that translate from the raw args []string representation into a more semantic representation, and therefore we can test that translation (which tends to include important validation rules that the rest of the system relies on) separately from the logic that relies on it.
>
> I don't think I'm telling you anything you didn't already figure out from reading the code. But hopefully we can continue the idea so that package arguments is the only place which has functions that take args []string, and then the logic implemented elsewhere exclusively uses the prevalidated results from those functions so it can just assume the arguments are always valid.

Agree here, and that's why I added a note in the RFC about this. I would like to use what's already in there and unify that for all the commands. 
Will update that part to make things clearer.

---

> I think one of the more troubling parts of package command's current design is [Meta.Backend](https://github.com/opentofu/opentofu/blob/85de3d40fae67efb8cfe9020bba738433b49c409/internal/command/meta_backend.go#L95), which has grown to include a monstrous amount of functionality just because it happens to be a central function that is called by almost every command.
>
> While it would certainly be easier to leave that as-is while improving other aspects first, I expect that'll be hard to do in practice given how much cross-cutting behavior is hidden behind that method. We might need to make a plan for how to break up that method into smaller parts before we can make much more progress on everything else, but I hope I'm wrong about that!

That's a good point! I will try to take a look at that and to add a possible path forward. Therefore, even though you approved this, I will drop your approval and will draft the PR in the meantime.

---

> We should think carefully about the tradeoff of doing less work in realMain. While I do agree that there are some cases today where it initializes subsystems that a specific command doesn't need, that does have a benefit of giving the operator fast feedback if there's something "globally wrong" with their environment, such as having an invalid CLI configuration file.
>
> If we start initializing things more selectively then it might be harder for folks to notice they've set something up incorrectly, because they'd need to run a larger set of commands in order to fully exercise all of that initialization logic.
>
> But I don't mean to say that we definitely shouldn't do it. Only that I think we should think about the UX for incorrectly-configured environments at the same time as we think about the rest of this, to make sure we're not making it significantly harder for orchestration/automation developers to debug certain kinds of problems.

I totally agree! That's why I mentioned in the introduction of the RFC this part:
> The changes proposed in this RFC must not change any existing material functionality.

Though, to make it clear, I will add your point about the general UX as a sub-point of the one quoted above.
I just want to note that even if I agree strongly with that, there are still some commands and some functionality that shouldn't care about that, as `-v/--version`, `metadata functions`, `fmt` and maybe others that I might miss right now. But the idea is that any command that is doing no action related to the configuration (or/and state) should not initialise those unless needed.

---

> There is a broad design idea in the existing system of having all cross-cutting concerns initialized in a central place and then passed to the other components that need them in a lightweight version of dependency inversion style. This particularly applies to discovering settings from outside of the system, like environment variables, the CLI configuration files, etc.
>
> I hope that we can keep following that general principle as we change things, though of course there are lots of ways to achieve that and having a single huge realMain function at the front of the program probably isn't the best way. 😀
>
> This point isn't meant to say that what you wrote here is inconsistent with that idea, just that I think this is one of the main tensions we'll be facing as we decide how to refactor this logic: can we shrink the amount of code we're running in package main while still ensuring there's some easy-to-read, linear code somewhere that maintainers can use to find out what all of the possible external inputs are, without having to study the entire codebase to learn about them all.

True indeed. Regarding this, [one particular recent piece of work](https://github.com/opentofu/opentofu/issues/3186) kept popping into my mind while worked on this RFC. The way _I see_ this part, is that there are some behaviours configurable from flags and also environment variables, while others have only one source. 
This inconsistency still bothers me, especially because I've seen its benefits in some of my personal projects (and what I found in other known projects, like [Renovate](https://docs.renovatebot.com/self-hosted-configuration/) for example), where the same configuration can be given via cli _or_ environment variable.

The issue in OpenTofu is that there is no such thing till now, no clear cut where some configurations do come from, and at some point I would love to have this expanded. The usability in a CI system improves dramatically from this, but that's another conversation that we could have later.

Bottom line, is that I would like _not_ to add that constraint (or requirement) in this RFC right not. Instead I think that it's important to be sure that the way environment variables work today will keep working even after the changes. One reason for that is that once we will migrate to a more capable CLI library, that will allow us to play around in different ways with how we can prioritise different sources of configuration, enabling a more granular and clear definition of the sources where a configuration can come from.

yottta:
> I think one of the more troubling parts of package command's current design is [Meta.Backend](https://github.com/opentofu/opentofu/blob/85de3d40fae67efb8cfe9020bba738433b49c409/internal/command/meta_backend.go#L95), which has grown to include a monstrous amount of functionality just because it happens to be a central function that is called by almost every command.
>
> While it would certainly be easier to leave that as-is while improving other aspects first, I expect that'll be hard to do in practice given how much cross-cutting behavior is hidden behind that method. We might need to make a plan for how to break up that method into smaller parts before we can make much more progress on everything else, but I hope I'm wrong about that!

Coming back to this...

> We might need to make a plan for how to break up that method into smaller parts before we can make much more progress on everything else

Initially I thought the same, but after playing around with some refactoring around the backend, it kinda confirmed my initial assumption: I don't think that would be wise in this RFC to rework the backend implementation since it covers a lot of edge cases, legacy cases and I am not aware of the reasoning behind those.
I would like to just "extract" the whole logic in its own component and have it isolated to make it easier to reason later about it.

> I expect that'll be hard to do in practice given how much cross-cutting behavior is hidden behind that method.

Yes, indeed, there will be some small refactoring in such a way to inject the dependencies in the component, and that might lead to a good amount of dependencies, but at least will make it clear what it depends on and what each command needs to prepare in order to execute the new backend component.

Updated in cfafa30f5312795f53d7f7ce4289873f9a85d9ec

cam72cam (approved):
I think this is a reasonable proposal and look forward to reviewing the individual PRs

rfc/20251105-refactor-tofu-cli.md
```diff
@@ -0,0 +1,248 @@
...
+>
+> This is suggested as such because we might be actually able to improve the startup performance if the refactor
+> will allow initialisation only of the components needed by each command.
+
+#### Meta backend
+The most complex and sensitive part of the `Meta` structure is the [backend implementation](https://github.com/opentofu/opentofu/blob/85de3d40fae67efb8cfe9020bba738433b49c409/internal/command/meta_backend.go#L95).
+
+The implementation visible in [`meta_backend.go`](https://github.com/opentofu/opentofu/blob/d24b85fc685be224da93de546b5566429ca7b62f/internal/command/meta_backend.go)
+and in [`meta_backend_migrate.go`](https://github.com/opentofu/opentofu/blob/d24b85fc685be224da93de546b5566429ca7b62f/internal/command/meta_backend_migrate.go)
+tackles a lot of edge cases and legacy concerns that the reasons behing it is not that clear.
```
lbordowitz: "tackles a lot of edge cases and legacy conerns; the reasons _behind_ that implementation are not clear."

yottta: Fixed in e99cd7c9ca770819c6aa59b98404346c2bf9917c

yottta:
> There are a couple of awkward phrasings here and there, but I approve of the purpose and overall approach of the RFC.

Thanks a lot for the review! I would appreciate if you could suggest better phrasing of the ones that sound weird to you.

PR #3473, opened 2025-11-05
RFC: databricks backend

zcking:
This submits a RFC for adding the Databricks backend with locking to OpenTofu. This RFC allows more users to leverage OpenTofu for IaC with reduced friction for data practitioners, and provides unified governance on the state file.

## Checklist

### DIFF
--- a/rfc/20251021-databricks-backend.md
+++ b/rfc/20251021-databricks-backend.md
@@ -1,0 +1,211 @@
+# Databricks Backend with Locking for OpenTofu
+
+This RFC proposes adding Databricks as a new state backend for OpenTofu, enabling teams using Databricks with limited access to Cloud Service Providers to manage their infrastructure state. The backend utilizes Unity Catalog volumes for state storage, offering flexible locking options: [Databricks Lakebase](https://docs.databricks.com/aws/en/oltp/instances/about) (a managed PostgreSQL instance) or [Delta Lake](https://docs.delta.io/) tables as a fallback.
+
+### Motivation
+
+- **Dissuasion to implement IaC** due to a lack of cloud storage backends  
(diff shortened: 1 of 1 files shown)


yottta:
Hi @zcking,

Thanks a lot for your submission.
The functionality proposed by this RFC fits quite well in the Backends as Plugins issue #382. I recommend checking that out and let us know if your use case would be covered by that.

The maintainers will discuss this soon and we will provide a more clearer feedback once we have an agreement upon this.

zcking:
@yottta thank you for pointing me in this direction. 

Yes, #382 would fit our use case very well if/when it is approved.

R7L208:
@yottta - is #3476 the implementation of #382 that you mentioned? My team would benefit from this backend as we have no access to AWS but want to use opentofu for IaC of some of our Databricks resources

yottta:
Hi @R7L208, 
Yes, you identified that correctly. 

Please note though, that it is only an experiment (therefore the note from that PR: `One possible solution to #382`) that is dependent on other work yet to be done on other projects.
We don't have yet an ETA for it but we would love to deliver this in a future 1.12. _This is not a guarantee, because of the dependencies we have._

R7L208:
Thanks, @yottta ! Is there anywhere I can watch the dependency work? Just wondering what the best way to keep up to date on progress would be. 

>This is not a guarantee, because of the dependencies we have.

Yeh, completely understand this but would like to follow along if possible

yottta:
> Thanks, @yottta ! Is there anywhere I can watch the dependency work? Just wondering what the best way to keep up to date on progress would be.
> 
> > This is not a guarantee, because of the dependencies we have.
> 
> Yeh, completely understand this but would like to follow along if possible

The work that you can follow is in [this comment](https://github.com/opentofu/opentofu/issues/382#issuecomment-3541873563) (in #382).

PR #3475, opened 2025-11-06
fix: use variables at `enabled` in nested modules 

diofeher:
Resolves https://cloud-native.slack.com/archives/C05PXGAB05R/p1762418361124039?thread_ts=1761238300.596699&cid=C05PXGAB05R

From Roger Siegenthaler:

```
Finally got around to testing the 1.11-rc1 release and am encountering an issue with the new enabled meta-argument on modules.
I have a variable that controls whether a client_secret is needed. This controls the creation of a blue/green timer (in a module) aswell as a few other resources. The enabled argument works for the resources (including correctly identifying the implicit “moved” blocks), however for the module I’m receiving an error:
│ Error: Reference to undeclared input variable
│ 
│   on .terraform/modules/.../service-account.tf line 27, in module "bg_rotation":
│    8:     enabled = var.client_secret.enabled
│ 
│ An input variable with the name "client_secret" has not been declared. This variable can be declared with a variable "client_secret" {} block.
This happens if I reference the module without setting the client_secret input or if I set it to {enabled = false} or = true.
Code snippet:
variable "client_secret" {
  description = "A config for the client secret."
  type = object({
    enabled       = bool
    duration_in_M = optional(number, 24)
    overlap_in_M  = optional(number, 3)
  })
  default = {
    enabled = false
  }
  validation {
    # ...
  }
}

module "bg_rotation" {
  source = "..."

  duration_in_M = var.client_secret.duration_in_M
  overlap_in_M  = var.client_secret.overlap_in_M

  lifecycle {
    enabled = var.client_secret.enabled
  }
}

resource "azuread_application_password" "blue" {
  display_name   = "..."
  application_id = azuread_application_registration.app_registration.id

  start_date = module.bg_rotation.blue_start_date
  end_date   = module.bg_rotation.blue_end_date

  rotate_when_changed = {
    rotation = module.bg_rotation.blue_rotation_date
  }

  lifecycle {
    enabled = var.client_secret.enabled
  }
}
```

For some reason I had the context evaluation from the parent on the module because using the already existing `evalCtx` didn't work properly. But now, after re-testing, it seems this works fine even with variables defined in the parent and this is what other fields use as well (like `for_each` and `count`). I've added a regression test too.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_plan2_test.go
+++ b/internal/tofu/context_plan2_test.go
@@ -3522,25 +3522,51 @@ func TestContext2Plan_moduleImplicitMove(t *testing.T) {
 		"from nested enabled-module multiple-resource to enabled-module single-resource": {
 			config: testModuleInline(t, map[string]string{
 				"main.tf": `
+					module "parent" {
+						source        = "./parent"
+						child_enabled = true
+					}`,
+				"parent/main.tf": `
+					variable "child_enabled" {
+						type = bool
+					}
+
 					module "child" {
-						source = "./child"
+						source   = "../child"
+						lifecycle {
+							enabled = var.child_enabled
+						}
+						grandchild1_enabled = true
 					}`,
 				"child/main.tf": `
-					variable "enabled" { default = true }
-					module "grandchild" {
-						source = "./grandchild"
+					variable "grandchild1_enabled" {
+						type = bool
+					}
+					variable "grandchild2_enabled" {
+						type    = bool
+						default = true
+					}
+					module "grandchild1" {
+						source = "../grandchild"
 						lifecycle {
-							enabled = var.enabled
+							enabled = var.grandchild1_enabled
 						}
-					}`,
-				"child/grandchild/main.tf": `resource "test_object" "a" {
+					}
+					module "grandchild2" {
+						source = "../grandchild"
+						lifecycle {
+							enabled = var.grandchild2_enabled
+						}
+					}
+					`,
+				"grandchild/main.tf": `resource "test_object" "a" {
 					count = 1
 				}`,
 			}),
-			expectedAddr: mustResourceInstanceAddr("module.child.module.grandchild[0].test_object.a"),
-			prevAddr:     mustResourceInstanceAddr("module.child.module.grandchild.test_object.a"),
+			expectedAddr: mustResourceInstanceAddr("module.parent.module.child.module.grandchild1.test_object.a[0]"),
+			prevAddr:     mustResourceInstanceAddr("module.parent.module.child.module.grandchild1.test_object.a"),
 			prevState: states.BuildState(func(s *states.SyncState) {
-				s.SetResourceInstanceCurrent(mustResourceInstanceAddr("module.child.module.grandchild.test_object.a"), &states.ResourceInstanceObjectSrc{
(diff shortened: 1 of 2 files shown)


siegenthalerroger:
Thanks for the quick fix! Are dev builds made for merge requests? I could test this MR directly if wanted.

diofeher:
@siegenthalerroger If you have that project locally, I think the fastest way to test would be downloading this branch, using `make build`,  then calling the binary on your project: `cd ~/your-project/ && ~/opentofu/tofu plan`.

apparentlymart (approved):
This _is_ quite confusing, but after re-reading the code here I think I can at least explain why the new code is correct. (I cannot explain why it didn't work when you originally tried it.)

One important thing here is that this graph node type implements the special `GraphNodeReferenceOutside` interface:

https://github.com/opentofu/opentofu/blob/f82265dc640609103a7d07397ee99006686df519/internal/tofu/node_module_expand.go#L117-L120

This activates some tricky additional behavior in the code which detects references during graph construction:

https://github.com/opentofu/opentofu/blob/a961f737b7938f51a52202fdc6de5a2917b3c250/internal/tofu/transform_reference.go#L503-L525

https://github.com/opentofu/opentofu/blob/a961f737b7938f51a52202fdc6de5a2917b3c250/internal/tofu/transform_reference.go#L527-L549

Notice that in the second snippet we use the second result from `ReferenceOutside` as the module that the node _makes references into_ -- that is, the module whose evaluation scope any expressions in the module call are resolved in -- and `nodeExpandModule` returns `n.Addr.Parent()` in that case.

The loop you modified in this PR then matches that by iterating over the expanded instances of the _parent_ module:

https://github.com/opentofu/opentofu/blob/a961f737b7938f51a52202fdc6de5a2917b3c250/internal/tofu/node_module_expand.go#L127-L131

...so by the time we're calling `evaluateEnabledExpression` we're already evaluating in the parent module.

(I'm sorry if you already learned all of this while debugging! I'm mainly just writing it down here hoping it would help other reviewers understand the context without repeating the debugging work I did to understand how this was working.)

---

I think part of why this is confusing is that the variable `evalCtx` is overwritten in each iteration of the loop with a differently-scoped evaluation context. Perhaps we should rename these variables to match the convention from [Naming convention for internal variables representing "contexts"](https://github.com/opentofu/opentofu/blob/main/rfc/20250108-naming-convention-for-context-vars.md#proposed-solution):

> These plain names are appropriate to use whenever there is only one variable of a type in scope and there is no further meaning to communicate beyond what type it has. In some situations multiple contexts of the same type are in scope, which warrants additional prefixes to clarify the relationships between them. For example, functions that deal with the transition from static objects from the configuration to dynamic instances declared by count/for_each/etc will tend to have `globalEvalCtx` and `moduleEvalCtx`, both of type `tofu.EvalContext` but with the former being unscoped and the latter scoped to a particular module instance path. The general rule then is to add the modifier word to the front of the name but to retain the above naming conventions for the type-specifying suffix.

The situation in this function actually exactly matches the example from this text! So, if we were to follow that example, then the argument should be called `globalEvalCtx` and the variable inside the loop should be called `moduleEvalCtx`. `globalEvalCtx` would be used only to get the `InstanceExpander`, and so all of the other references would be to `moduleEvalCtx`.

siegenthalerroger:
@diofeher just tested this with a local build and the plan showed 0 changes as expected (with the correct implicit moves).

PR #3476, opened 2025-11-06
Pluggable Backend Prototype using tfplugin6

cam72cam:
This is a prototype I built a while ago that explores what using the new provider PSS concept could look like in OpenTofu.

It started by investigating https://github.com/hashicorp/terraform-plugin-go/pull/563, which did not support locking at the time of this initial work.  I ended up using a local fork of that repository with Locking implemented.  That has since been added to the tf-plugin-go PR and is no longer required.

Given that when I did this work, the PR above did contain the full proto specification.  I instead pulled a non-stable proto file from https://github.com/hashicorp/terraform/blob/main/docs/plugin-protocol/tfplugin6.proto. Our understanding is that since https://github.com/hashicorp/terraform/blob/main/docs/plugin-protocol/.copywrite.hcl specifically marks these files as MPL-2.0, this source is reasonable to use for prototyping in OpenTofu.  We still prefer to pull the proto files from terraform-plugin-go when possible.

Example Provider:
https://github.com/cam72cam/terraform-provider-pss

Example Usage:
```hcl
terraform {
    required_providers {
        pss = { source = "terraform.local/local/pss" }
    }
    state_store "pss_cam" {
        // Provider configuration to use
        provider "pss" {
        }
        // Values to pass for state store config
        ident = "foo"
    }
}

resource "tfcoremock_simple_resource" "foo" {
}
```

The actual structure of the configuration is a functional prototype and not fully thought out ergonomically at this point.

One possible solution to #382 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/module.go
+++ b/internal/configs/module.go
@@ -329,7 +329,7 @@ func (m *Module) appendFile(file *File) hcl.Diagnostics {
 	}
 
 	for _, c := range file.StateStoreConfigs {
-		c.Provider = m.ImpliedProviderForUnqualifiedType(c.ProviderConfigAddr().LocalName)
+		c.Provider = m.ImpliedProviderForUnqualifiedType(c.LocalProvider.LocalName)
 
 		if m.StateStoreConfig != nil {
 			diags = append(diags, &hcl.Diagnostic{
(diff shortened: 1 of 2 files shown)


kvokka:
@cam72cam may i ask if there are any updates on this one? thank you

PR #3481, opened 2025-11-07
[Hackathon][POC] Cobra integration

yottta:
This is a POC created during the hackathon to see how [cobra](https://github.com/spf13/cobra) can be used to replace our current cli library.

This is a flaky POC of one of the steps described in the RFC #3471

---

Sadly, there is not much to show showcase here since the outcome of this work is to ensure that with cobra we can have a similar behaviour with the one in the main branch.

The changes here do not cover everything, but most of the things.
Issues still to be solved:
* `-version`/`--version`/`-v` flag parsing
* `-h`/`-help` flag parsing

---

**EDIT**
This POC also highlights another thing that I hoped that will not be necessary: the amount of customisation needed to achieve a similar behavior with the existing one.
As you can see, in order to be able to render the help (usage) text correctly (especially regarding flags format), I added a custom build function that is doing a similar work with the one in [tofu/help.go](https://github.com/opentofu/opentofu/blob/cc8e86c99842f3ee943419a333bc996095834b0a/cmd/tofu/help.go#L19), with the clear benefit that what I added in this PR is way more dynamic than what we have.

That said, the dynamic nature of that is not _a given_ by the cobra library, but it's another in-house made implementation to overcome the problem that we have. Therefore, to overcome our pain points with the management of the cli flags, we could reuse (in its entirety or partially) the implementation added in here to achieve a similar, more dynamic, result in our current implementation.

One of the value that I personally see, is that by using cobra (or more specifically [`pflag`](https://github.com/spf13/pflag)) together with its baked in [`CopyToGoFlagSet`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.CopyToGoFlagSet), we can manage our flags and their documentation more easily. But that it's not a unique behaviour of the `pflag` library itself, but can be achieved by using the go stdlib `flag` package too. 
In the end this is more of a personal preference on what "looks better" when comparing these 2, especially because we still need to use the go `flag` package for parsing to hold the one dash format instead of the [GNU format](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html).

In the end, it feels that the "biggest" advantage cobra would bring to the table is the commands hierarchy.  But that can be achieved with most of the other existing CLI libraries (like [urfave](https://cli.urfave.org/) or like the one that we already use).

Personally, I consider this POC a success. We drawn better the issues we have, we highlighted better what pain points are there and we created an [RFC](https://github.com/opentofu/opentofu/pull/3471) that, hopefully, will guide us to have soon a clear path on improving this part of the system and to remove some of the pain points most of us have with it.


diofeher:
Really cool stuff, @yottta! Congrats

yottta:
Closing this since this was just an experiment. Not needed anymore.

Issue #3489, opened 2025-11-11
tofu apply crash with ephemeral resource


RolandColored:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
OpenTofu v1.11.0-rc1
on linux_amd64
```

### OpenTofu Configuration Files

```hcl
ephemeral "random_password" "openid_client_secret" {
  length           = 20
  override_special = "!#%&*()-_=+[]{}<>:?"
}

resource "aws_secretsmanager_secret_version" "keycloak" {
  secret_id                = aws_secretsmanager_secret.keycloak.id
  secret_string_wo         = ephemeral.random_password.openid_client_secret.result
  secret_string_wo_version = 1
}

resource "keycloak_openid_client" "service_development" {
  realm_id  = var.keycloak_realm
  client_id = "service-${var.service_name}"
  name      = "service-${var.service_name}"

  client_secret_wo         = ephemeral.random_password.openid_client_secret.result
  client_secret_wo_version = 1

  access_type              = "CONFIDENTIAL"
  service_accounts_enabled = true # enables Client Credentials flow
}
```

### Debug Output

```
!!!!!!!!!!!!!!!!!!!!!!!!!!! OPENTOFU CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
  
  OpenTofu crashed! This is always indicative of a bug within OpenTofu.
  Please report the crash with OpenTofu[1] so that we can fix this.
  
  When reporting bugs, please include your OpenTofu version, the stack trace
  shown below, and any additional information which may help replicate the issue.
  
  [1]: https://github.com/opentofu/opentofu/issues
  
  !!!!!!!!!!!!!!!!!!!!!!!!!!! OPENTOFU CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
  
  no expansion has been registered for module.auth["i18n"].ephemeral.random_password.openid_client_secret
  goroutine 2111 [running]:
  runtime/debug.Stack()
  	runtime/debug/stack.go:26 +0x5e
  runtime/debug.PrintStack()
  	runtime/debug/stack.go:18 +0x13
[...]

PR #3492, opened 2025-11-12
Do not prune the ephemeral expansion node - temp solution

yottta:
This is a temporary solution for #3489. Created this PR mostly for the integration test. The solution is not final.

The issue is that the expansion node for the ephemeral resource gets pruned and only the instance node is left in the graph.
Because of that, when the instance node is executed, it reaches for the repetition data, and because the expansion node has been pruned, the repetition data was not added into the evaluation context.

This temporary solution always keeps the ephemeral expansion node in the graph, but it's not the right approach because it might not be always the case to keep the node (even though the side effects should be minimal), but the most important part, is that the instance node does not have dependency on the expansion node. Because of that, the instance node could be executed before the expansion one ending up in the same situation as the one encountered in the reported issue.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/transform_destroy_edge.go
+++ b/internal/tofu/transform_destroy_edge.go
@@ -311,13 +311,6 @@ func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) err
 			func() {
 				n := nodes[i]
 				switch n := n.(type) {
-				case *nodeExpandApplyableResource:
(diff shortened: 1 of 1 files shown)


yottta:
Closing [in favor](https://github.com/opentofu/opentofu/pull/3495#pullrequestreview-3453270800) of #3495

PR #3495, opened 2025-11-12
Ensure Ephemeral values are handled by the diff transformer

Yantrio:
This is a slightly different approach to what @yottta has done in #3494 / #3492

When investigating #3489 with @yottta we discovered that the ephemeral opening was being planned but it was being pruned from the graph, resulting in the nodes not being registered with the expander.

As a quick workaround for this we attempted a workaround fix with 2288aa16a8c033aed818348ff2e3ca7911ea2e8a. Whilst this "worked", we felt that it was a bit of a band-aid over an underlying issue: The node has no edges, so its being pruned.

This approach instead attempts to solve this issue by altering the Diff Tranformer to handle the opening of the ephemeral resource and ensure that it's correctly attached to the correct resource instances that use it.

This code works on a one primary assumption:
**- Ephemeral resources are ALWAYS opened.** 

---

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_apply2_test.go
+++ b/internal/tofu/context_apply2_test.go
@@ -5759,6 +5759,7 @@ ephemeral "test_ephemeral_resource" "a" {
 		Result: cty.ObjectVal(map[string]cty.Value{
 			"id":     cty.StringVal("id val"),
 			"secret": cty.StringVal("val"),
+			"input":  cty.NullVal(cty.String),
 		}),
 	}
 
@@ -5795,7 +5796,7 @@ ephemeral "test_ephemeral_resource" "a" {
 		Instances: map[addrs.InstanceKey]*states.ResourceInstance{
 			addrs.NoKey: {
 				Current: &states.ResourceInstanceObjectSrc{
-					AttrsJSON:          []byte(`{"id":"id val","secret":"val"}`),
+					AttrsJSON:          []byte(`{"id":"id val","input":null,"secret":"val"}`),
 					Status:             states.ObjectReady,
 					AttrSensitivePaths: []cty.PathValueMarks{},
 					Dependencies:       []addrs.ConfigResource{},
--- a/internal/tofu/context_plan2_test.go
+++ b/internal/tofu/context_plan2_test.go
@@ -8815,6 +8815,7 @@ ephemeral "test_ephemeral_resource" "a" {
 		Result: cty.ObjectVal(map[string]cty.Value{
 			"id":     cty.StringVal("id val"),
 			"secret": cty.StringVal("val"),
+			"input":  cty.NullVal(cty.String),
 		}),
 	}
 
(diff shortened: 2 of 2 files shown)


internal/tofu/transform_diff.go
```diff
@@ -162,12 +164,41 @@ func (t *DiffTransformer) Transform(_ context.Context, g *Graph) error {
...
+			// to the resource node, we create it here.
+			// If we do not have this logic here, ephemeral resource instances that are not used by any other resource
+			// or output would be pruned away during the unused nodes pruning step, and thus not opened.
+			abstract := NewNodeAbstractResourceInstance(addr)
+			var node dag.Vertex = abstract
+			if t.Concrete != nil {
+				node = t.Concrete(abstract)
+			}
+
+			g.Add(node)
```
yottta, line 179: I am not sure that it worth doing it, but maybe we can extract somehow this logic since is duplicated with the one from the `if update` block.

yottta:
These changes fix one thing that was missed during ephemerals implementation: expansion of modules.

In other words, when the address of an ephemeral resource generated during planning was containing a module instance address (ie: `module.call["i18n"]`), it wasn't matched correctly with the config generated address that contains no instance address yet (ie: `module.call`).
Because of this, [the logic in `DiffTransformer`](https://github.com/opentofu/opentofu/blob/541fc48a7bf5e7239ecf78979ce2dd9cd59b1173/internal/tofu/transform_diff.go#L207-L208) that is linking an apply expansion node to its execution node was not able to match correctly the information from the plan with the one in the configuration, leading to the panic while opening the ephemeral resource: it was having no expansion information in the eval context.

The expansion of the ephemeral resource itself does not affect this behavior because in the `DiffTransformer` we call `addr.ContainingResource()` which returns the unkeyed absolute address of the [resource](https://github.com/opentofu/opentofu/blob/541fc48a7bf5e7239ecf78979ce2dd9cd59b1173/internal/addrs/resource.go#L307). So it was always matching the config generated address with the plan time generated address.

---

This whole idea relies on the simple fact that we store a stub of the ephemeral resources in the plan to be able to rebuild the apply graph. We also revisited previously the RFC to add [this information](https://github.com/opentofu/opentofu/pull/2996/files#diff-1b793b731c0b3e0929632ddfa87429e5d0d2ddbe95bd3f9a09dea465ac1d406dR812).

Gogotchuri (approved):
Looks great! Nice work!

internal/tofu/transform_diff.go
Yantrio: So, whilst I'm a big fan of keeping code DRY, I really would prefer to only de-duplicate code if it's repeated 3 or more times. In this case it's a short amount of code and only duplicated once.

Yantrio:
Please note that I have left the `I have written all code in this PR myself OR I have marked all code I have not written myself (including modified code, e.g. copied from other places and then modified) with a comment indicating where it came from.` item unchecked solely because part of this PR is authored by @yottta.

I can confirm that the part that I have written has been written by myself. 

@yottta can you also confirm yours for the tests?

yottta:
> Please note that I have left the `I have written all code in this PR myself OR I have marked all code I have not written myself (including modified code, e.g. copied from other places and then modified) with a comment indicating where it came from.` item unchecked solely because part of this PR is authored by @yottta.
> 
> I can confirm that the part that I have written has been written by myself.
> 
> @yottta can you also confirm yours for the tests?

I can confirm that the part that I have written has been written by myself.

PR #3517, opened 2025-11-20
Remove nil check that was returning nil schema for actual existing resources

yottta:
The nil check removed was buggy in the sense that for a provider with no managed resource schemas, it would have returned nil schema for existing data or ephemeral resources.
The only reason why this didn't create issues was that in the proto layer, when the schema is retrieved from the provider, the managed resource map was always initialised. ([proto5](https://github.com/opentofu/opentofu/blob/82fdad27fcea6654fed8f1084d94223a7e6043c5/internal/plugin/grpc_provider.go#L129) and [proto6](https://github.com/opentofu/opentofu/blob/82fdad27fcea6654fed8f1084d94223a7e6043c5/internal/plugin6/grpc_provider.go#L129))

Initially I added the unit test visible in this PR and then removed the check. As expected, the tests were failing in the expected fashion:
image

Resolves _no issue existing for this as it was discovered during ephemeral implementation_

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/schemas.go
+++ b/internal/tofu/schemas.go
@@ -51,9 +51,6 @@ func (ss *Schemas) ProviderConfig(provider addrs.Provider) *configschema.Block {
 // redundant.
 func (ss *Schemas) ResourceTypeConfig(provider addrs.Provider, resourceMode addrs.ResourceMode, resourceType string) (block *configschema.Block, schemaVersion uint64) {
 	ps := ss.ProviderSchema(provider)
-	if ps.ResourceTypes == nil {
-		return nil, 0
-	}
 	return ps.SchemaForResourceType(resourceMode, resourceType)
 }
 
--- a/internal/tofu/schemas_test.go
+++ b/internal/tofu/schemas_test.go
@@ -7,12 +7,269 @@ package tofu
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


Gogotchuri (changes requested):
Won't this now panic if `ManagedResourceMode` was used and `ps.ResourceTypes` was nil?
Can we add those checks in `SchemaForResourceType` instead?

----

Do we currently not have any tests for that use case?

yottta:
> Won't this now panic if `ManagedResourceMode` was used and `ps.ResourceTypes` was nil? Can we add those checks in `SchemaForResourceType` instead?
> 
> Do we currently not have any tests for that use case?

Not needed since accessing a nil map works as expected and returns the zero value of the map value type: https://go.dev/play/p/VokRUEYvMdG

In our context, accessing `ss.ResourceTypes` (or `ss.DataSources` or `ss.EphemeralResources`) will return a zero value for [`providers.Schema`](https://github.com/opentofu/opentofu/blob/82fdad27fcea6654fed8f1084d94223a7e6043c5/internal/providers/provider.go#L197) that is a struct value of that type having a nil `*configschema.Block` inside, which is returned back.

And yes, there are test cases for that, pretty much all the test cases with `exist: false` will check that.

Do you think that would be helpful to add some comments somewhere to clear these things?

PR #3544, opened 2025-12-05
registry: Remove legacy "regsrc" package, use regaddr instead

apparentlymart:
This registry client code is _very, very_ old and in particular predates the creation of [the separate "registry-address" library](https://pkg.go.dev/github.com/opentofu/registry-address/v2) and OpenTofu's own [`package addrs`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/addrs), so it had its own type for representing module addresses.

It also had some historical confusion about the difference between the concepts of module vs. module package, which meant the code was pretty unclear about whether it was dealing in module packages or module source addresses: in practice we were always leaving the "Submodule" field empty because the registry client isn't the component responsible for dealing with modules in package subdirectories, but this made the code confusing to read and maintain.

This change therefore removes the legacy `package regsrc` altogether and adopts [`regaddr.ModulePackage`](https://pkg.go.dev/github.com/opentofu/registry-address/v2#ModulePackage) as the canonical representation of registry module package addresses. This therefore makes the module registry client agree with the module installer (in `package initwd`) about what vocabulary types we're using, and so we no longer need shim code for converting between the two representations.

To make it abundantly clear that the registry client deals only in package addresses, the client methods and arguments are renamed to reflect that, and the callers updated to match.

While in the area anyway I also adjusted a few others things that were not aligned with our modern conventions, but tried to keep it limited to relatively localized changes for ease of review. There's probably room for more improvement here in subsequent commits.

---

The motivation for working on this now is that my work for https://github.com/opentofu/opentofu/issues/3313 was extending this package in ways that were spreading the technical debt into more places. I want to improve this first as a separate step so that the planned new functionality will be easier to review and maintain.

This should not change the external behavior of OpenTofu in any _important_ way, so I'm not adding a changelog entry.

Note though that it _does_ slightly change some of the error message text related to registry requests, but only in a way that makes the registry client more consistent with later-adopted conventions elsewhere: it uses fully-qualified package addresses, always including the registry hostname, when reporting problems that resulted from communication with the registry. This convention is because it's easier to debug network-related problems if you know what hostname OpenTofu was actually trying to communicate with.


internal/registry/client.go
```diff
@@ -139,59 +137,39 @@ func (c *Client) ModuleVersions(ctx context.Context, module *regsrc.Module) (*re
...
-}
-
-// ModuleLocation find the download location for a specific version module.
+// ModulePackageLocation find the download location for a specific module package version.
+//
 // This returns a string, because the final location may contain special go-getter syntax.
-func (c *Client) ModuleLocation(ctx context.Context, module *regsrc.Module, version string) (string, error) {
+func (c *Client) ModulePackageLocation(ctx context.Context, packageAddr regaddr.ModulePackage, version string) (string, error) {
 	ctx, span := tracing.Tracer().Start(ctx, "Find Module Location", tracing.SpanAttributes(
-		traceattrs.OpenTofuModuleCallName(module.RawName),
```
apparentlymart, line 165: I removed the `OpenTofuModuleCallName` attribute here because the way it was being populated here is not consistent with how we defined that attribute in the `traceattrs` package.

The "module call name" is the name given to the `module` block, like "foo" in `module "foo"`, but the registry client doesn't have access to that information and so it's actually the _module installer's_ responsibility to report that attribute, which it's already doing in a parent span here:

https://github.com/opentofu/opentofu/blob/b1d4708e0324b58cef5f72a8563997e8e29ce543/internal/initwd/module_install.go#L240-L246

The registry client _does_ know what package it was being asked to resolve though, and a module package address is _close enough_ to a "module source address" for tracing purposes: it uses the same syntax except that the subdir portion is always empty.

PR #3547, opened 2025-12-08
Consolidate provider resource mocking and overrides

cam72cam:
As part of a larger effort, I'm trying to consolidate provider and provisioner complexity into a single plugin manager.  In that process, I ran into some of the interesting compromises made to support mocked and overridden test resources.

In an attempt to simplify the plugin manager, I tried in this PR to move all of the resource mocking and override complexity to the single getProvider() function on NodeResourceAbstractInstance.  This makes it a *lot* easier to reason about and allows for the removal of some intertwined code between different areas of the tofu package.

This is not strictly necessary for the plugin manager I'm working on, but it allows that whole concern to be much simpler if the mocking and overrides is handled as a separate concern.  If in review this is not seen as a reasonable refactor, I have a few other paths I could potentially take.

Part of #3414 
## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/provider_for_test_framework.go
+++ b/internal/tofu/provider_for_test_framework.go
@@ -19,8 +19,10 @@ import (
 
 var _ providers.Interface = &providerForTest{}
 
-// providerForTest is a wrapper around a real provider to allow a specific resources to be overridden
-// for testing framework.
+// providerForTest is a wrapper around a real provider to allow a specific resource to be overridden
+// for the testing framework.  It is used by [NodeResourceAbstractInstance.getProvider] to fulfil the mocks
+// and overrides for that one specific resource instance, any other usage is a bug in OpenTofu and should
+// be corrected.
 type providerForTest struct {
 	// providers.Interface is not embedded to make it safer to extend
 	// the interface without silently breaking providerForTest functionality.
--- a/internal/tofu/transform_provider.go
+++ b/internal/tofu/transform_provider.go
@@ -53,10 +53,7 @@ type GraphNodeProvider interface {
 	GraphNodeModulePath
 	ProviderAddr() addrs.AbsProviderConfig
 	Name() string
-}
-
-type GraphNodeProviderTestExtension interface {
-	GraphNodeProvider
+	// For test framework
 	MocksAndOverrides() (IsMocked bool, MockResources []*configs.MockResource, OverrideResources []*configs.OverrideResource)
 }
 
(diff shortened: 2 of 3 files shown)


internal/tofu/provider_for_test_framework.go
```diff
@@ -20,43 +19,26 @@ import (
 
 var _ providers.Interface = &providerForTest{}
 
-// providerForTest is a wrapper around a real provider to allow certain resources to be overridden
-// (by address) or mocked (by provider and resource type) for testing framework.
+// providerForTest is a wrapper around a real provider to allow a specific resources to be overridden
+// for testing framework.
```
apparentlymart: It took me some time to understand what the new meaning of this type is from reading the rest of the changes here, and a future reader of this won't necessarily have the benefit of this diff to refer to, so here's an attempt to put more context here for a future maintainer (and also, as a side-effect, to test whether I understood correctly):


```suggestion
// providerForTest is a special implementation of [providers.Interface] which implements
// only a specific managed or data resource type, implementing it using a fixed override
// value instead of real provider behavior.
//
// This is used by [NodeResourceAbstractInstance.getProvider] when it finds that its
// associated resource instance has had an "override" applied to it using features of the
// module testing language. It creates an object of this type that must only be used
// for operations relating to _that specific resource instance_: using that object for
// any other resource instance will cause unspecified and confusing results.
```

I also wonder if we should reinforce the idea that an object of this type is only for one specific resource instance by including an `addrs.AbsResourceInstance` field inside it. That field would admittedly be mostly unused, but it would at least help a skimming reader quickly notice that (unlike most provider instance types) this particular one has a resource instance address and therefore something weird must be going on.

Once it's there though, we could _actually_ use it in a couple different ways to help detect if we accidentally use this improperly: In `ReadDataSource`, `UpgradeResourceState`, `ReadResource`, `PlanResourceChange`, and `ApplyResourceChange` we could fail immediately if the `Mode` of the resource doesn't make sense for whatever operation was requested, and we could also check if the `req.TypeName` field matches the `Type` of the resource. This wouldn't catch a problem of using the result with a different instance of the same resource type, but it's still a little more robust against mistakes than not checking anything at all. 🤔

internal/tofu/transform_provider.go
```diff
@@ -55,6 +55,11 @@ type GraphNodeProvider interface {
 	Name() string
 }
 
+type GraphNodeProviderTestExtension interface {
+	GraphNodeProvider
+	MocksAndOverrides() (IsMocked bool, MockResources []*configs.MockResource, OverrideResources []*configs.OverrideResource)
+}
+
```
apparentlymart: I don't think we have _that_ many concrete implementations of `GraphNodeProvider` in today's system. Could we simplify this by adding the `MocksAndOverrides` method directly to `GraphNodeProvider` instead, and then have any implementations that don't rely on this method just immediately return `false, nil, nil`?

(I ask this just because it's been my experience that using the "extension interface" pattern with the node types has made maintenance harder because it forces us to carefully study what each node type implements to understand the behavior. It's sometimes the best available option anyway, but I wonder if that's true here: the question of whether the interface is implemented seems redundant with the question of whether the method's first return value is true.)

internal/tofu/provider_for_test_framework.go
cam72cam: I've updated the text a little in d4852f1cc5, but am not sure about if it's worth the complexity of adding in the additional resource checks.  In practice, it's only initialized in one function that has tight control of it's usage.

internal/tofu/transform_provider.go
cam72cam: Addressed in d4852f1cc5

Gogotchuri (approved):

PR #3552, opened 2025-12-08
fix: remove leftover binary

mikelolasagasti:
The binary was introduced in commit 7f6a6b9cc61fbdafe102031456777be66fc0ad9e

It's part of 1.10.8 release tarball.


yottta:
Thank you for the submission.
You are right indeed and somehow we missed it.

diofeher (approved):
Thanks for the catch @mikelolasagasti

PR #3568, opened 2025-12-10
Add changelog entries for #3564 #3566 #3567

yottta:
Changelog entries for:
* https://github.com/opentofu/opentofu/pull/3564
* https://github.com/opentofu/opentofu/pull/3566
* https://github.com/opentofu/opentofu/pull/3567

## Checklist

### Go checklist

 


cam72cam:
@yottta these only need to go on the v1.11 branch AFAIK

yottta:
> @yottta these only need to go on the v1.11 branch AFAIK

If I understand this correctly, you say that the bug fixes changelogs should go strictly to the targeted version branch? 🤔 
We should have this clearly stated somewhere, because it's confusing.

Personally, I consider that we should add these in all the newer versions changelogs to be transparent that it was fixed in those too 🤷

apparentlymart:
@yottta our typical approach has been to mention a bug was fixed only if that bug was present in one of the stable releases in the same release series.

Since there haven't yet been any stable releases in the v1.12.x series, from the perspective of someone only running stable releases these bugs were introduced in v1.11.0 and fixed in v1.11.1, so by the time v1.12.0 is released these fixes will be uninteresting/irrelevant. We would only include this in the main branch changelog if our plan was to leave these bugs unfixed throughout all of the v1.11.x releases and so the first release with the fix would be v1.12.0.

There's an implicit assumption in this policy that the intended audience of the changelog is end-users who are upgrading between our stable releases, and that the changelog is _not_ intended for people who are running unreleased code from our development branches -- they should rely on the commit history directly, in that case. Is that assumption what you are disagreeing with, or do you have a different concern?

yottta:
Thanks @apparentlymart for the arguments. That sounds more than reasonable.
Closing this then.

Issue #3565, opened 2025-12-10
Lifecycle error when resource in module


joserey-veeva:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
tofu version
OpenTofu v1.11.0
on darwin_arm64
+ provider registry.opentofu.org/hashicorp/aws v6.25.0
```

### OpenTofu Configuration Files

Main file:
```hcl
module "ecs-cluster" {
  source                          = "../infra-iac-module-aws-containers-ecs-ec2-cluster"
  application                   = var.application
  environment                 = var.environment
  product                         = var.product
  create_asg_and_lt       = false
```

Module file:
```hcl
[...]
resource "aws_sns_topic" "drainer_topic" {
  name  = "${var.application}-${var.environment}-${var.product}-drainer-topic"
  lifecycle {
    enabled = var.create_asg_and_lt
  }
}
```

### Debug Output

Error: Failed to write plan file
 
The plan file could not be written: failed to write plan: resource module.ecs-cluster.aws_sns_topic.drainer_topic has unsupported action reason ResourceInstanceDeleteBecauseEnabledFalse.

### Expected Behavior

Plan will succeed

### Actual Behavior

Does not allow to continue planning

### Steps to Reproduce

**Do not use lifecycle** in the module resource 
1. `tofu init`
2. `tofu plan`
3. `tofu apply`

Add lifecycle to the module resource
1. `tofu init`
2. `tofu plan`

### Additional Context

If the variable `var.create_asg_and_lt` is `true` it does **not** produce an error
The error **only** occurs with a resource that was **already** in the state file

diofeher:
Thanks for the report @joserey-veeva! I'm investigating right now, and I'll get back to you as soon as possible.

cam72cam:
Backported to v1.11 and will be released in v1.11.1

PR #3569, opened 2025-12-10
Retaining resources during destruction - action and reason marshalling extension

Gogotchuri:
The issue https://github.com/opentofu/opentofu/issues/3565 and @diofeher (Thanks Dio!) - made me realize that I haven't fully handled tfplan marshalling during the #3409.
This PR extends the planfile.proto to include the new action and action reason.
It also fixes the issue where orphan resources have their action reason overwritten incorrectly.
I have also changed the naming of the action reason, which wasn't quite matching the others, `ReasonForgotBecauseOfLifecycleDestroy` -> `ReasonForgotBecauseLifecycleDestroy`

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/node_resource_plan_orphan.go
+++ b/internal/tofu/node_resource_plan_orphan.go
@@ -233,10 +233,11 @@ func (n *NodePlannableResourceInstanceOrphan) managedResourceExecute(ctx context
 		return diags
 	}
 
+	// In case we haven't already set a reason for the action,
+	// we might be able to offer an approximate reason for why we are
+	// planning a certain action for this object. (This is best-effort; we might
+	// sometimes not have a reason.)
 	if change.ActionReason == plans.ResourceInstanceChangeNoReason {
-		// We might be able to offer an approximate reason for why we are
-		// planning to delete this object. (This is best-effort; we might
-		// sometimes not have a reason.)
 		change.ActionReason = n.deleteActionReason(evalCtx)
 	}
 


diofeher (approved):
🚀

internal/tofu/node_resource_plan_orphan.go
```diff
@@ -233,10 +233,12 @@ func (n *NodePlannableResourceInstanceOrphan) managedResourceExecute(ctx context
 		return diags
 	}
 
-	// We might be able to offer an approximate reason for why we are
-	// planning to delete this object. (This is best-effort; we might
-	// sometimes not have a reason.)
-	change.ActionReason = n.deleteActionReason(evalCtx)
+	if change.ActionReason == plans.ResourceInstanceChangeNoReason {
```
yottta, line 240: Why this change? I might miss some context, so better to double check. 🤔

Gogotchuri: If you take a look few lines at the top, we are writing the new action reasons before this statement.
That part was being overridden with this line. Here we've been trying to decern if we need to add some action reason to add additional context for the user. And in case we already have an action reason, we no longer need to try and come up with one.

Before this change, that function couldn't find any additional action reason and it was overwriting the existing (already set) reason with "no reason".

yottta: Thanks, that clears things out.
Maybe a comment would help in the future, but it's just a suggestion.

PR #3580, opened 2025-12-11
rfc: Improvements for Contribution Documents

lbordowitz:
[Read the RFC](https://github.com/opentofu/opentofu/blob/contribution_rfc/rfc/20251211-contribution-docs.md)

Back in late October 2025, @cam72cam and I went through each of the (non-forked) repositories in the `opentofu` GitHub organization and made some notes for improvements. I cobbled together a set of issues, but the general feedback was that they all seemed fairly unrelated. It was requested that an RFC be made, for the sake of understanding the greater context and having a venue to provide feedback on the overall direction of this effort.

All but one of these issues are related to community outreach. I've included a profile for the platonic "casual contributor" based on some admittedly perfunctory research. I believe that casual contributors are the first step towards more long-term contributors, and that improving our efforts to court this type of developer will increase the size of the funnel, so to speak.

This also includes the issue about maintaining forks, which is unrelated but still part of the repository upkeep tracker I made back in October. I don't know whether it needs to be included here, but I thought it worth a mention, if only to cross my "t"s and dot my "i"s.

### DIFF
--- a/rfc/20251211-contribution-docs.md
+++ b/rfc/20251211-contribution-docs.md
@@ -40,7 +40,7 @@ A casual contributor should, at a glance, be able to understand all expectations
 
 Note that none of those guidelines are specific to any development. We may have some development guidelines which are in a supervised repository's `CONTRIBUTING.md`. However, if those were moved to their own document, then `CONTRIBUTING.md` could be standardized and mostly identical across all supervised repositories.
 
-This also means that [any fixes](https://github.com/opentofu/org/issues/26) done to `CONTRIBUTING.md` and `MAINTAINERS.md` could propagate across all relevant repositories.
+This also means that [any fixes](https://github.com/opentofu/org/issues/26) done to some central `CONTRIBUTING.md` and `MAINTAINERS.md` documents in `org` could be referred to across all relevant repositories.
 
 We may also want to list different ways to contribute that [may not include code](https://opensource.guide/how-to-contribute/#what-it-means-to-contribute), or code contributions ancillary to the main repository (like CI/CD).
 


rfc/20251211-contribution-docs.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Contributing guidelines should be standardized across supervised repositories
+
+https://github.com/opentofu/org/issues/25
+
+A casual contributor should, at a glance, be able to understand all expectations and limitations for their contributions. This includes, but is not limited to, our stance on AI and the scope of prohibition on external code.
+
+Note that none of those guidelines are specific to any development. We may have some development guidelines which are in a supervised repository's `CONTRIBUTING.md`. However, if those were moved to their own document, then `CONTRIBUTING.md` could be standardized and mostly identical across all supervised repositories.
+
+This also means that [any fixes](https://github.com/opentofu/org/issues/26) done to `CONTRIBUTING.md` and `MAINTAINERS.md` could propagate across all relevant repositories.
```
yottta: Reading in https://github.com/opentofu/org/issues/25 too, I want to ask if syncing these documents manually is the right approach.
I am not sure what the end goal is for the content of these documents, but for the contributing one, I would expect to have repo-specific information too.
Hence, I would suggest to have one globally applicable, general contribution guide in our [org repo](https://github.com/opentofu/org) and have that referenced in all the supervised repositories as the main document. Anything else that might be mentioned in the supervised repository contribution guide should be made clear that are amendments to the main document.
This way we will not go through the, possible, pain of trying and not override repository specifics.

rfc/20251211-contribution-docs.md
```diff
@@ -0,0 +1,75 @@
...
+
+#### Appendix: Forked repositories
+
+The only issue from the tracker not yet mentioned in this RFC regards [forked repositories](https://github.com/opentofu/org/issues/27), which is barely related to community outreach. It is included here solely for the sake of completeness.
+
+OpenTofu fork many repositories, but we don't have a present long-term strategy to keep these forks up-to-date. Is this a chore we should do? Are there automated tools we could use that might ease the process? What are the criteria for choosing which repositories to fork and which to contribute upstream?
```
yottta, line 75: At a first glance, I would suggest to put in place monitoring for these and not rebase automatically. We could create tickets automatically, the way we do for the [go vulnerabilities](https://github.com/opentofu/opentofu/blob/main/.github/scripts/govulncheck-submit-issues.sh#L64).
This way we can take action based on the received notifications (or created issues) and have these decided in our weekly triage.

Obviously, this is just a suggestion. It would be great to have more ideas around this shared and to try to agree upon at least one.

rfc/20251211-contribution-docs.md
diofeher: I agree with @yottta on this one!

rfc/20251211-contribution-docs.md
lbordowitz: It's a good suggestion. We should track this further [here](https://github.com/opentofu/org/issues/27).

rfc/20251211-contribution-docs.md
lbordowitz: Updating verbiage to weaker language: "referring to," which could either be linking to the documents in `org` repo, or copying them.

diofeher (approved):

PR #3589, opened 2025-12-15
Remove global schema cache and clean up tofu schema/contextPlugins

cam72cam:
Given how we use tofu.NewContext(), we still do the same amount of plugin "work". The advantage here is that we don't have to manage a fully global schema cache and how it interacts with multiple layers of abstraction.

This also fixes a few bugs around providers with invalid schemas bypassing validation checks due to aforementioned mixed levels of cache complexity.

Potential precursor to https://github.com/opentofu/opentofu/issues/3414, but also related to extended providers and other upcoming work

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_plugins.go
+++ b/internal/tofu/context_plugins.go
@@ -97,7 +97,10 @@ func (cp *contextPlugins) ProviderSchema(ctx context.Context, addr addrs.Provide
 		})
 		cp.providerSchemas[addr] = entry
 	}
-	// We don't defer unlock as the majority of the work of this function happens in calling the sync.OnceValues function defined above and called below
+	// This lock is only for access to the map. We don't need to hold the lock when calling
+	// "entry" because [sync.OnceValues] handles synchronization itself.
+	// We don't defer unlock as the majority of the work of this function happens in calling "entry"
+	// and we want to release as soon as possible for multiple concurrent callers of different providers
 	cp.providerSchemasLock.Unlock()
 
 	return entry()
@@ -164,7 +167,10 @@ func (cp *contextPlugins) ProvisionerSchema(addr string) (*configschema.Block, e
 		})
 		cp.provisionerSchemas[addr] = entry
 	}
-	// We don't defer unlock as the majority of the work of this function happens in calling the sync.OnceValues function defined above and called below
+	// This lock is only for access to the map. We don't need to hold the lock when calling
+	// "entry" because [sync.OnceValues] handles synchronization itself.
+	// We don't defer unlock as the majority of the work of this function happens in calling "entry"
+	// and we want to release as soon as possible for multiple concurrent callers of different provisioners
 	cp.provisionerSchemasLock.Unlock()
 
 	return entry()


internal/plugin/grpc_provider.go
```diff
@@ -94,53 +97,44 @@ type GRPCProvider struct {
...
-	if !p.Addr.IsZero() {
-		// Even if the schema is cached, GetProviderSchemaOptional could be false. This would indicate that once instantiated,
-		// this provider requires the get schema call to be made at least once, as it handles part of the provider's setup.
-		// At this point, we don't know if this is the first call to a provider instance or not, so we don't use the result in that case.
-		if schemaCached, ok := providers.SchemaCache.Get(p.Addr); ok && schemaCached.ServerCapabilities.GetProviderSchemaOptional {
-			logger.Trace("GRPCProvider: GetProviderSchema: serving from global schema cache", "address", p.Addr)
-			return schemaCached
-		}
+
+	// For testing only
```
Yantrio: i think this should be removed ( same in plugin6)

internal/providers/schemas.go
```diff
@@ -39,3 +42,66 @@ func (ss ProviderSchema) SchemaForResourceType(mode addrs.ResourceMode, typeName
...
+		mu.Lock()
+		defer mu.Unlock()
+
+		if schema.Provider.Block != nil {
+			return schema
+		}
+		schema = getSchema()
+		return schema
+	}
+}
```
Yantrio, line 103: nitpick: I think we could swap some of the manual mutex stuff out with sync.Once to make our lives a little easier.

internal/plugin/grpc_provider.go
```diff
@@ -94,53 +97,44 @@ type GRPCProvider struct {
...
 	ctx context.Context
 
-	mu sync.Mutex
-	// schema stores the schema for this provider. This is used to properly
-	// serialize the requests for schemas.
-	schema providers.GetProviderSchemaResponse
+	// SchemaCache stores the schema for this provider. This is used to properly
+	// serialize the requests for schemas.  This is shared between instances
+	// of the provider.
+	SchemaCache      func(func() providers.GetProviderSchemaResponse) providers.GetProviderSchemaResponse
```
Yantrio: I think the func(func T) T pattern took a while for me to put together in my head what it was doing and I think its great to do a memoizing/caching approach this way.

However I actually think it would be a little bit nicer to provider a small struct that wrapped it so its a tiny bit clearer, especially at execution time. I would expect to call a method called something like `providerCache.GetOrFetch()` maybe? (Naming is hard!)

internal/providers/schemas.go
```diff
@@ -39,3 +42,66 @@ func (ss ProviderSchema) SchemaForResourceType(mode addrs.ResourceMode, typeName
...
+
+func (resp ProviderSchema) Validate(addr addrs.Provider) error {
+	if resp.Diagnostics.HasErrors() {
+		return fmt.Errorf("failed to retrieve schema from provider %q: %w", addr, resp.Diagnostics.Err())
+	}
+
+	if resp.Provider.Version < 0 {
+		// We're not using the version numbers here yet, but we'll check
+		// for validity anyway in case we start using them in future.
+		return fmt.Errorf("provider %s has invalid negative schema version for its configuration blocks,which is a bug in the provider ", addr)
```
Yantrio:
```suggestion
		return fmt.Errorf("provider %s has invalid negative schema version for its configuration blocks,which is a bug in the provider", addr)
```

internal/plugin/grpc_provider.go
cam72cam: Ah, I already have an aliased type for this that I need to drop in

internal/plugin/grpc_provider.go
cam72cam: If we don't do this, I need to make a *lot* of changes to the grpc_provider_test.go in both plugin and plugin6.

internal/plugin/grpc_provider.go
cam72cam: Pushed a change to replace this with the alias

internal/providers/schemas.go
cam72cam: Fixed in 1fa9f7f4f0

internal/command/meta_providers.go
```diff
@@ -406,18 +408,18 @@ func providerFactory(meta *providercache.CachedProvider) providers.Factory {
 
 // initializeProviderInstance uses the plugin dispensed by the RPC client, and initializes a plugin instance
 // per the protocol version
-func initializeProviderInstance(plugin interface{}, protoVer int, pluginClient *plugin.Client, pluginAddr addrs.Provider) (providers.Interface, error) {
+func initializeProviderInstance(plugin interface{}, protoVer int, pluginClient *plugin.Client, schemaCache func(func() providers.ProviderSchema) providers.ProviderSchema) (providers.Interface, error) {
```
Gogotchuri: Should we use `SchemaCache` type here too?

internal/plugin/grpc_provider.go
Gogotchuri: Yeah, leaving an additional conditional branching here, just for tests, makes me a bit uneasy too.

internal/command/meta_providers.go
cam72cam: Good catch! fixed in 1e3dc277b2

internal/plugin/grpc_provider.go
cam72cam: I'll take the better approach of adding a constructor in the tests.  It balloons the changes a bit, but it's the right way to fix it.

cam72cam: Fixed in 0f956f1178

internal/plugin/grpc_provider.go
```diff
@@ -27,6 +25,16 @@ import (
...
 var logger = logging.HCLogger()
 
+// Some providers may generate quite large schemas, and the internal default
+// grpc response size limit is 4MB. 64MB should cover most any use case, and
+// if we get providers nearing that we may want to consider a finer-grained
+// API to fetch individual resource schemas.
+// Note: this option is marked as EXPERIMENTAL in the grpc API. We keep
+// this for compatibility, but recent providers all set the max message
+// size much higher on the server side, which is the supported method for
+// determining payload size.
```
apparentlymart: I think I already said this on a different version of this code somewhere else 🙃, but in case not:

"This option" in this comment is referring to `grpc.MaxRecvMsgSizeCallOption`, which was mentioned close to where this comment originally appeared but is no longer related to this factored-out constant alone. I'd suggest moving this second part of the comment back nearby the mention of that symbol so it'll hopefully be clearer to future maintainers what this is talking about.

internal/tofu/context_plugins.go
```diff
@@ -68,72 +80,27 @@ func (cp *contextPlugins) NewProvisionerInstance(typ string) (provisioners.Inter
...
+			}
+			defer provider.Close(ctx)
+
+			schema := provider.GetProviderSchema(ctx)
+			return schema, schema.Validate(addr)
+		})
+		cp.providerSchemas[addr] = entry
 	}
+	// We don't defer unlock as the majority of the work of this function happens in calling the sync.OnceValues function defined above and called below
+	cp.providerSchemasLock.Unlock()
```
apparentlymart: I wasn't sure what this comment was communicating at first, but I think the main idea is that this mutex is only guarding access to the `cp.providerSchemas` map itself and isn't needed for calling this function because `sync.OnceValues` does its own synchronization. Is that right?

If so, maybe something like this:

```
// This lock is only for access to the map. We don't need to hold the lock when calling
// "entry" because [sync.OnceValues] handles synchronization itself.
```

🤷‍♂️

apparentlymart (approved):
This looks good to me. I'm glad to finally be rid of this hidden global state that bleeds between all instances of `tofu.Context`!

I hope one day we can push more of this down into `package tofu` (or its successors) so that the CLI layer doesn't need to be so intimately familiar with how the plugin GRPC clients work, but that'd be a far more invasive change and so it makes sense to me to start with this compromise.

internal/tofu/context_plugins.go
```diff
@@ -68,72 +80,27 @@ func (cp *contextPlugins) NewProvisionerInstance(typ string) (provisioners.Inter
...
+			provider, err := cp.NewProviderInstance(addr)
+			if err != nil {
+				return providers.ProviderSchema{}, fmt.Errorf("failed to instantiate provider %q to obtain schema: %w", addr, err)
+			}
+			defer provider.Close(ctx)
+
+			schema := provider.GetProviderSchema(ctx)
+			return schema, schema.Validate(addr)
+		})
+		cp.providerSchemas[addr] = entry
```
yottta, line 98: To make things clearer regarding the lock, I would extract this logic into another function and use defer in there, or just use an anonymous func to do so:
```go
func() {
	cp.providerSchemasLock.Lock()
 	defer cp.providerSchemasLock.Unlock()
    entry, ok := cp.providerSchemas[addr]
    if !ok {
        entry := ...
        cp.providerSchemas[addr] = entry
    }
}()
```

This is trying to also tackle the comment Martin left on this one. Obviously, the suggested code is kinda ugly so I would suggest to create a fully separate unexported method 🤷

yottta (approved):
Other than one suggestion, this looks really nice!
Love the pattern with the `sync.Once` you used in here to cache the schema. Lovely! ❤️

internal/plugin/grpc_provider.go
cam72cam: Fixed in c6f99e38cd

internal/tofu/context_plugins.go
cam72cam: Fixed in 14b8541c1e

internal/tofu/context_plugins.go
cam72cam: I've tinkered around with this a bit, and IMO it's a lot less readable.

yottta: Ok. As said, was more of a suggestion. I don't have strong opinions on it.

Issue #3590, opened 2025-12-15
Tofu plan incompatibility issues from version v1.10.7 to v1.11.1


CidHighwindVII:
### Community note

Hello everyone.

In version v1.10.7 I was able to do a "Tofu Plan" with the following terraform.

module "eks" {
  source          = "terraform-aws-modules/eks/aws"
  version         = "20.31.1"

...
}

module "eks_blueprints_addons" {
  source  = "aws-ia/eks-blueprints-addons/aws"
  version = "1.21.1"
...
}

data "aws_eks_cluster_auth" "cluster_auth" {
  depends_on = [module.eks.cluster_id]
  name       = module.eks.cluster_name
}

data "aws_eks_cluster" "this" {
  depends_on = [module.eks.cluster_id]
  name     = module.eks.cluster_name
}

provider "helm" {
  kubernetes {
    host                   = data.aws_eks_cluster.this.endpoint
    cluster_ca_certificate = base64decode(data.aws_eks_cluster.this.certificate_authority[0].data)
    exec {
      api_version = "client.authentication.k8s.io/v1beta1"
      command     = "aws"
      args        = ["eks", "get-token", "--cluster-name", module.eks.cluster_name, "--region", var.region]
      env = {
        AWS_ACCESS_KEY_ID     = var.REAL_AWS_ACCESS_KEY_ID
        AWS_SECRET_ACCESS_KEY = var.REAL_AWS_SECRET_ACCESS_KEY
      }
    }
  }
}

module "traefik_addon" {
  source  = "aws-ia/eks-blueprints-addon/aws"
  version = "~> 1.0"
...
  depends_on = [ module.eks_blueprints_addons ]
}

As of version v1.11.1 it is no longer possible as i'm presented with the following error:

Error: Provider configuration is incomplete

  on .terraform/modules/ch_1111250000010000001_p.ch_eks.eks_blueprints_addons/helm.tf line 5, in resource "helm_release" "this":
   5: resource "helm_release" "this" {

The provider was unable to work with this resource because the associated
provider configuration makes use of values from other resources that will
not be known until after apply.

To work around this, use the planning option
-exclude="module.ch_1111250000010000001_p.module.ch_eks.module.eks_blueprints_addons.helm_release.this[\"smb-csi-driver\"]"
to first apply without this object, and then apply normally to converge.

Has anything changed from v1.10.7 to version v1.11.1 that could break this compatibility?

Thank you in advance.

### OpenTofu Version

```shell
Working
[...]

PR #3592, opened 2025-12-15
providers: Remove explicit handling of "deferred" signal from providers

apparentlymart:
This removes most of the code previously added in 491969d29def, because we since learned that the `hashicorp/helm` provider signals deferral when any unknown values are present in provider configuration even though in practice it can sometimes successfully plan changes in spite of those unknown values.

That therefore made the `hashicorp/helm` provider behavior worse under this change than it was before, returning an error when no error was actually warranted.

The ephemeral resources implementation landed later and was also interacting with this change, and so this isn't a line-for-line revert of the original change but still removes everything that was added in support of handling provider deferral signals so that we'll be able to start fresh with this later if we find a better way to handle it.

---

If merged, this will be backported to the `v1.11` branch for inclusion in a patch release. The subsequent backport PR will contain the changelog entry, so that it will end up in the v.11 changelog.

This is for https://github.com/opentofu/opentofu/issues/3590 but that should not be closed until the backport is merged.

---


apparentlymart:
I have left this as a draft right now because I want to confirm (in https://github.com/opentofu/opentofu/issues/3590#issuecomment-3657892232) whether there's a provider with a deferral-capable ephemeral resource type that I ought to test this against.

If the answer turns out to be no then I intend to mark this ready for review as-is, and so if anyone wants to review it in the meantime despite it being a draft then that could help us get this in faster. :crossed_fingers:

diofeher (approved):

yottta:
The only provider that I managed to test this with, was a custom provider written specifically to develop the whole ephemeral feature.
After searching in multiple providers, I found none using the deferred functionality.
Most probably, that's also because of the fact that the deferred is [marked as experimental](https://github.com/hashicorp/terraform-plugin-framework/blob/5c81090d2fdbc1d5d55f77b28e3f5b5977902045/ephemeral/open.go#L21-L22) for the moment so it's not recommended to be used.

---

In terms of the changes from this PR, there are 2 things that I would like to note:
* The deferred functionality for ephemerals is not a deal breaker because by looking at the [exposed reasons by the terraform-provider-framework](https://github.com/hashicorp/terraform-plugin-framework/blob/main/ephemeral/deferred.go#L6-L21), the most critical one, [where a resource configuration is unknown](https://github.com/opentofu/opentofu/blob/ddc704c371b4e0620fb9e11f2620d0ddeec59ee4/internal/tofu/node_resource_abstract_instance.go#L3088-L3092), is already handled on the OpenTofu's part (from a raw configuration POV, aka unknown values), so during planning, the deferral will be done based on that.
* Second, I wouldn't remove the whole deferred functionality from the provider layer, and instead remove it from the affected bits. Since each provider method is capable of handling the capabilities of the client, I see no reason on why we should remove that from the open ephemeral calls. I would suggest to remove that strictly from the other calls that create the issue submitted above. 
  * [Looking into the terraform-provider-framework](https://github.com/hashicorp/terraform-plugin-framework/blob/main/internal/fromproto5/openephemeralresource.go#L42), it is just forwarding the capability to the provider implementation of the OpenEphemeral, so it's only up to the provider what to do with it.

yottta:
I submitted #3593 that proposes a more cautious approach.

---

To add more context, while working on the ephemeral resources, I remember that I postponed my work on deferral signals from providers on the open action, because I've seen the work was already in progress for #2643.

So the difference between this PR and #3593 was meant to be added nonetheless.

---

@apparentlymart, considering that I couldn't find any provider that implements deferral for the open call, feel free to merge any of the PRs (and close the other). Even if you merge this PR, we could handle any future situations on the ephemeral when the time will come.

PR #3593, opened 2025-12-16
Do not handle yet the deferred signal from providers for actions other than open ephemeral resource

yottta:
Details in #3592

Resolves # 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/plugin6/grpc_provider_test.go
+++ b/internal/plugin6/grpc_provider_test.go
@@ -14,7 +14,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/zclconf/go-cty/cty"
--- a/internal/providers/deferral.go
+++ b/internal/providers/deferral.go
@@ -5,10 +5,6 @@
 
 package providers
 
-import (
-	"github.com/opentofu/opentofu/internal/tfdiags"
-)
-
 // DeferralReason is an enumeration of the different reasons a provider might
 // return when reporting why it is unable to perform a requested action with
 // the currently-available context.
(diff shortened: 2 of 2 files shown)


internal/plugin/grpc_provider.go
```diff
@@ -881,7 +853,10 @@ func (p *GRPCProvider) OpenEphemeralResource(ctx context.Context, r providers.Op
 		Config: &proto.DynamicValue{
 			Msgpack: config,
 		},
-		ClientCapabilities: clientCapabilities,
+		ClientCapabilities: &proto.ClientCapabilities{
+			DeferralAllowed:            true,
+			WriteOnlyAttributesAllowed: clientCapabilities.WriteOnlyAttributesAllowed,
+		},
```
apparentlymart, line 859: I think this (and the equivalent code in `plugin6` is the main potential concern here:

As far as we know (based on closed-box testing and reviewing the framework code) Terraform always sends the same capability information to all requests, whereas with this change OpenTofu would report different capabilities for `OpenEphemeral` than it does for any other type of request. We also know that the framework allows a provider to return deferral in the "configure" request and then remembers that answer to return on other requests, which suggests that the framework is _expecting_ that the client will report consistent capabilities across all requests. But we also know from reading the framework code that _as it's currently written_ this is unlikely to cause a problem.

So by doing this we are taking on a risk that future changes to either the framework or providers using the framework will rely on Terraform's behavior and behave incorrectly with OpenTofu's different behavior.

From re-reviewing the `hashicorp/kubernetes` providers I noticed that the latest versions now have some ephemeral resource types, which means that for anyone using those OpenTofu would set `DeferralAllowed: false` when configuring the provider but `DeferralAllowed: true` when opening the ephemeral resource instance. It seems like that _does_ work today, but it means that the deferral signal doesn't work at all because `hashicorp/kubernetes` only checks it during its configure function in practice.

All of that is to say: while the risk of doing this seems relatively low with the providers that currently exist, there seems to be no _benefit_ to it because we don't know of any provider currently existing that checks for this capability during `OpenEphemeral`. So I think my preference is to take the more complete removal I did in https://github.com/opentofu/opentofu/pull/3592 and then wait to see if there is ever any provider that checks for this capability during `OpenEphemeral`, instead of checking it only during `ConfigureProvider`. Then we can make sure that the provider's new use of the feature makes sense before we try to use it.

apparentlymart (commented):
After reading your patch and thinking more about the consequences of it I'm afraid I reached a different opinion than I had when we discussed this verbally earlier.

I left a comment inline trying to explain why, but the short version is: we don't get any benefit from supporting deferral for `OpenEphemeral` until there's at least one provider actually using it, and we already know that providers generate inaccurate deferral signals from `ConfigureProvider`, so I think the risk that a provider will eventually implement deferral _incorrectly_ for `OpenEphemeral` outweighs the (nonexistent, IMO) benefit of supporting it.

However, I'm not going to take any action on either PR today because I want to first ask if you see a benefit that I didn't notice. Also, I admit that I tend to be more risk-averse than others on the team when it comes to imagining potential future changes in provider implementations, so I'm happy to go with the maintainer team's consensus on this.

internal/plugin/grpc_provider.go
yottta: Thanks a lot for the detailed answer.
That sounds reasonably and as I said in another [comment in #3592](https://github.com/opentofu/opentofu/pull/3592#issuecomment-3660061742), I don't mind going either way.

So I will close this one and you can merge the other PR without any additional friction.

yottta:
Closed in favor of https://github.com/opentofu/opentofu/pull/3592

PR #3604, opened 2025-12-17
main: Log about reliance on GODEBUG settings

apparentlymart:
The Go runtime provides a number of configuration knobs that subtly change its behavior, and we cannot control which of these is available over time as we change which version of Go we're building with.

It's therefore possible that an OpenTofu user might be intentionally or unintentionally relying on one of these settings for OpenTofu to work on their system, in which case they would be broken if they upgraded to a newer version of OpenTofu which uses a different Go version that no longer supports that setting.

These log lines are intended to help us more quickly notice that possibility if someone opens a bug report describing an unexpected behavior change after upgrading to a new OpenTofu minor release series. We can ask the reporter to share `TF_LOG=debug` output from both the previous and new releases to compare, and then these log lines should appear in the older version's output so we can search the Go codebase and issue tracker for each of the mentioned names to learn if the handling of that setting has changed between Go versions.

---

This is a further addition related to the general concern of how we can manage this constantly-evolving set of tweakable knobs. They represent a (thankfully small) liability for OpenTofu's evolution because they are a compatibility surface that we cannot directly control, and we cannot directly measure how many people are relying on them because OpenTofu is run on end-user systems rather than on infrastructure we control.

This particular change is for a retroactive strategy of more quickly identifying what might've caused a compatibility break that has already happened, rather than for preventing them from occurring in the first place.

It remains an open question whether and how we could _proactively_ measure reliance on these, for issues such as https://github.com/opentofu/opentofu/issues/3415 and https://github.com/opentofu/opentofu/issues/2642 where we'd like to stop relying on specific settings but can't estimate the impact of doing so. But at least this should help us notice if something slips through that we didn't intentionally change.

### DIFF
--- a/cmd/tofu/version.go
+++ b/cmd/tofu/version.go
@@ -6,9 +6,62 @@
 package main
 
 import (
+	"log"
+	"runtime/metrics"
+	"strings"
+
+	"github.com/opentofu/opentofu/internal/logging"
 	"github.com/opentofu/opentofu/version"
 )
 
 var Version = version.Version
 
 var VersionPrerelease = version.Prerelease
+
+// logGodebugUsage produces extra DEBUG log lines if the Go runtime's metrics
+// suggest that code that was run so far relied on any non-default "GODEBUG"
+// settings, which could be helpful in reproducing a bug report if the
+// behavior differs based on such a setting.
+//
+// For this to be useful it must be run just before we're about to exit, after
+// we've already performed all of the requested work.
+func logGodebugUsage() {
+	// These constants reflect the documented conventions from the
+	// runtime/metrics package, so we can filter for only the metrics that
+	// are relevant to this function.
+	const godebugMetricPrefix = "/godebug/non-default-behavior/"
+	const godebugMetricSuffix = ":events"
+
+	if !logging.IsDebugOrHigher() {
+		// No point in doing any of this work if the log lines are going to
+		// be filtered out anyway.
+		return
+	}
+
+	metricDescs := metrics.All()
+	for _, metric := range metricDescs {
+		if !metric.Cumulative || metric.Kind != metrics.KindUint64 {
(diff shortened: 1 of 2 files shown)


cmd/tofu/version.go
```diff
@@ -6,9 +6,62 @@
...
+		}
+		if !strings.HasPrefix(metric.Name, godebugMetricPrefix) {
+			continue // irrelevant metric
+		}
+		if !strings.HasSuffix(metric.Name, godebugMetricSuffix) {
+			continue // irrelevant metric
+		}
+		samples[0].Name = metric.Name
+		metrics.Read(samples[:])
+		if count := samples[0].Value.Uint64(); count != 0 {
```
diofeher, line 62: Out of curiosity: the samples code got me confused by being on the outside of the loop, because I thought it was going to be populated by different values every iteration. But from what I'm reading here, it seems we just need a slice that's going to be populated by the `metric.Name` on the [0] index. 

I'm wondering if we should put the `samples` and that explanation of the array close to these assignments, and `metrics.Read` to increase the readability of this function.

apparentlymart: I put the variable declaration outside of the loop to be clear that we're intentionally reusing the same memory for each iteration, rather than allocating a separate backing array for each iteration.

With that said: I was originally writing this with efficiency in mind because I hadn't yet remembered I could add the `if !logging.IsDebugOrHigher()` check above to completely avoid running this code in most cases. Now that this only runs when logging is enabled, it's not particularly important to optimize this for speed anymore, so I guess we could just move the symbol into the loop.

(And I expect in practice the Go compiler's optimizer would realize it can reuse some stack space between iterations here anyway, so the "efficiency" was more about helping human readers maintain the invariant rather than _actually_ optimizing anything.)

diofeher (approved):

Issue #3615, opened 2025-12-19
Crash when indexing a non-existing resource in `import` + `for_each`


Geod24:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
OpenTofu v1.11.1
on linux_amd64
+ provider registry.opentofu.org/hashicorp/azuread v3.1.0
+ provider registry.opentofu.org/hashicorp/external v2.3.5
```

### OpenTofu Configuration Files

```hcl
terraform {
  backend "consul" { ... }
  required_providers {
    azuread = {
      source  = "hashicorp/azuread"
      version = "~> 3.1.0"
    }
  }
}

provider "azuread" {
  tenant_id = "..."
}

import {
  for_each = tomap({foo = "bar" })
  to = azuread_application.apps[each.value]
  id = "/applications/${each.key}"
}
```

### Debug Output

None actually ?

### Expected Behavior

Throwing an error as `apps` doesn't exist.

### Actual Behavior

```
% tofu plan
Acquiring state lock. This may take a few moments...

!!!!!!!!!!!!!!!!!!!!!!!!!!! OPENTOFU CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

OpenTofu crashed! This is always indicative of a bug within OpenTofu.
Please report the crash with OpenTofu[1] so that we can fix this.

When reporting bugs, please include your OpenTofu version, the stack trace
shown below, and any additional information which may help replicate the issue.

[1]: https://github.com/opentofu/opentofu/issues

!!!!!!!!!!!!!!!!!!!!!!!!!!! OPENTOFU CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

runtime error: invalid memory address or nil pointer dereference
goroutine 56 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:26 +0x5e
runtime/debug.PrintStack()
        runtime/debug/stack.go:18 +0x13
github.com/opentofu/opentofu/internal/logging.panicHandler({0x3cf6d80, 0x6f1c570}, {0xc00017d800, 0x416, 0x800})
        github.com/opentofu/opentofu/internal/logging/panic.go:96 +0xd6
github.com/opentofu/opentofu/internal/backend/local.(*Local).Operation.PanicHandlerWithTraceFn.func2()
[...]

PR #3616, opened 2025-12-19
Do not make use of the GenerateConfigPath during validate

yottta:
For #3615

The issue is that the transformer responsible with converting configuration to nodes contains logic for the config generation, and when is the case, if `import.to` points to a missing configuration, it needs to add a node and consider the configuration valid.
But since the validate phase does not have the information if the config generation is triggered or not, it tried initially to do a "best effort" and use the logic behind the "configGeneration". With this new change, there is a special path for the generation of the node in case of walkValidate.

ℹ️ needs backporting to v1.11
## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_validate_test.go
+++ b/internal/tofu/context_validate_test.go
@@ -2751,3 +2751,60 @@ func TestContext2Validate_importIntoModuleResource(t *testing.T) {
 		})
 	}
 }
+
+func TestContext2Validate_importIntoUnexistingResourceBlock(t *testing.T) {
+	// This checks that validate walk adds an import node into the graph even if the targeted
+	// configuration block does not exist.
+	// This is useful for the situations where the config generation flag is turned on.
+	// In those cases, the execution should run as intended without the configuration block,
+	// one of the purpose being to generate the missing block.
+	// Related to: https://github.com/opentofu/opentofu/issues/3615
+	p := simpleMockProvider()
+	hook := new(MockHook)
+	ctx := testContext2(t, &ContextOpts{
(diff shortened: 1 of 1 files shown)


internal/tofu/graph_builder_plan.go
```diff
@@ -149,6 +149,7 @@ func (b *PlanGraphBuilder) Steps() []GraphTransformer {
 
 			// We only want to generate config during a plan operation.
 			generateConfigPathForImportTargets: b.GenerateConfigPath,
+			operation:                          b.Operation,
```
apparentlymart: I don't really like propagating the raw "operation" around like this, because it makes it easier for there to be hidden differences between the different walk types that we don't expect under maintenance.

Could we make this be a boolean field named something like `forceAddImportTargets`, and define it here as `forceAddImportTargets: b.Operation == walkValidate`? That way the graph transformer is just being directly told what to do, and we don't need to look so deep to find all of the potential differences by operation.

If you think this would make the code more complex or harder to read then I don't feel super strongly about it. This is just a preference I have based on previous experience being surprised by hidden uses of operation type in deep parts of the system.

apparentlymart (approved):
This solution makes sense to me. The fake config path when validating is not something I would've expected when debugging/maintaining, so I'm glad this allows us to remove it.

internal/tofu/graph_builder_plan.go
yottta: Fixed in c54efbb2884af61c8e03f271bd29f3fb17956885

PR #3617, opened 2025-12-19
Replace EvalContext.Provider calls with direct links

cam72cam:
This is a precursor to introducing a plugin manager (draft PR coming soon).  The end goal is to consolidate as much of the provider and provisioner management as possible.

Specifically,  this PR looks to reduce the responsibilities of the EvalContext "god" object that tracks and manages all providers.  I had originally planned to remove CloseProvider as well, but that has some interesting interactions with stopping providers during a graph walk interruption.

Potential precursor to https://github.com/opentofu/opentofu/issues/3414, but also related to extended providers and other upcoming work

I've now opened a very rough draft for how this work could be leveraged to centralize the provider and provisioner management in the tofu package (and other future packages).  https://github.com/opentofu/opentofu/pull/3652.  We still need to determine if it is worth going in that direction, but it shows that the changes here allow for something of that nature.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/node_resource_validate_test.go
+++ b/internal/tofu/node_resource_validate_test.go
@@ -190,7 +190,7 @@ func TestNodeValidatableResource_ValidateResource_managedResource(t *testing.T) 
 		NodeAbstractResource: &NodeAbstractResource{
 			Addr:             mustConfigResourceAddr("test_foo.bar"),
 			Config:           rc,
-			ResolvedProvider: mustResolvedProviderInRoot("aws", p),
+			ResolvedProvider: mustResolvedProviderInRoot("test", p),
 		},
 	}
 
@@ -256,7 +256,7 @@ func TestNodeValidatableResource_ValidateResource_managedResourceCount(t *testin
 				NodeAbstractResource: &NodeAbstractResource{
 					Addr:             mustConfigResourceAddr("test_foo.bar"),
 					Config:           rc,
-					ResolvedProvider: mustResolvedProviderInRoot("aws", p),
+					ResolvedProvider: mustResolvedProviderInRoot("test", p),
 				},
 			}
 
@@ -290,7 +290,7 @@ func TestNodeValidatableResource_ValidateResource_ephemeralResource(t *testing.T
 		NodeAbstractResource: &NodeAbstractResource{
 			Addr:             mustConfigResourceAddr("ephemeral.test_foo.bar"),
 			Config:           rc,
-			ResolvedProvider: mustResolvedProviderInRoot("aws", p),
+			ResolvedProvider: mustResolvedProviderInRoot("test", p),
 		},
 	}
 
@@ -379,7 +379,7 @@ func TestNodeValidatableResource_ValidateResource_dataSource(t *testing.T) {
 		NodeAbstractResource: &NodeAbstractResource{
 			Addr:             mustConfigResourceAddr("test_foo.bar"),
 			Config:           rc,
-			ResolvedProvider: mustResolvedProviderInRoot("aws", p),
+			ResolvedProvider: mustResolvedProviderInRoot("test", p),
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


internal/tofu/node_provider_abstract.go
```diff
@@ -33,7 +33,6 @@ type NodeAbstractProvider struct {
 var (
 	_ GraphNodeModulePath                 = (*NodeAbstractProvider)(nil)
 	_ GraphNodeReferencer                 = (*NodeAbstractProvider)(nil)
-	_ GraphNodeProvider                   = (*NodeAbstractProvider)(nil)
```
apparentlymart, line 36: Am I understanding correctly that `NodeAbstractProvider` is now acting as a "partial" implementation of `GraphNodeProvider`, which other node types are expected to embed and then complete by adding an `Interface` method?

If so, I'd suggest we document that a little more explicitly on the methods that _are_ from that interface, since this "partial implementation by embedding" idea is not a very common Go pattern and so I wouldn't expect future maintainers to understand it immediately without having this diff as helpful context. 😀 

(It's annoying that there are multiple methods and therefore no clear single place to put that documentation. I don't have a better idea than choosing one of the methods to be the "main" one that has the documentation, and then making the docs for all of the others include a link to that main one, so that the main language is only in one place but folks can still find it when they're looking at another method.)

internal/tofu/transform_provider.go
```diff
@@ -53,6 +54,7 @@ type GraphNodeProvider interface {
 	GraphNodeModulePath
 	ProviderAddr() addrs.AbsProviderConfig
 	Name() string
+	Instance(addrs.InstanceKey) providers.Configured
```
apparentlymart: Based on how we implemented provider `for_each` in the first round, I'd guess that `addrs.InstanceKey` here can still potentially be either the provider's instance key directly _or_ an instance key from one of the ancestor modules, depending on how the provider configuration was declared. Is that true?

If so, we should probably add a doc comment here making that very clear because I expect an unfamiliar reader would guess that this `addrs.InstanceKey` is only the one for the provider instance itself.

(If I've misunderstood and this is actually always just the provider instance key then this wouldn't apply.)

cam72cam: This is just the provider instance key, but a comment to that effect would be useful

apparentlymart: Cool! That's some good evidence that I ought to read this more closely to better understand how things are flowing together here.

internal/tofu/node_provider_abstract.go
cam72cam: Added comments in a1829fc8bd

internal/tofu/transform_provider.go
cam72cam: Added comments in a1829fc8bd

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3212,9 +3212,13 @@ func resourceInstancePrevRunAddr(evalCtx EvalContext, currentAddr addrs.AbsResou
 }
 
 func (n *NodeAbstractResourceInstance) getProvider(ctx context.Context, evalCtx EvalContext) (providers.Interface, providers.ProviderSchema, error) {
-	underlyingProvider, schema, err := getProvider(ctx, evalCtx, n.ResolvedProvider.ProviderConfig, n.ResolvedProviderKey)
+	underlyingProvider := n.ResolvedProvider.Instance(n.ResolvedProviderKey)
```
Yantrio: nitpick: Should we add some nil checking just incase for `n.ResolvedProvider`?

Yantrio: (Same in import and validate)

internal/tofu/node_provider.go
```diff
@@ -88,8 +110,13 @@ func (n *NodeApplyableProvider) initInstances(ctx context.Context, evalCtx EvalC
...
 	}
 
+	if n.instances == nil {
+		n.instances = map[addrs.InstanceKey]providers.Configured{}
+	}
+
 	for _, key := range initKeys {
-		_, err := evalCtx.InitProvider(ctx, n.Addr, key)
+		instance, err := evalCtx.InitProvider(ctx, n.Addr, key)
+		n.instances[key] = instance
```
Yantrio: This will store the instance if the err is non-nil. I think we shouldn't do that.

internal/tofu/node_resource_abstract_instance.go
cam72cam: eeac2209c5

internal/tofu/node_provider.go
cam72cam: afad60b67f

PR #3625, opened 2025-12-26
cleanup: remove stale FIXME in tf provider

AndrewCharlesHay:
Resolves # 

## Checklist

### Go checklist

 

- [X] I have locally started the website as [described here](https://github.com/opentofu/opentofu/blob/main/website/README.md) and checked my changes.

Removes a stale FIXME comment in internal/builtin/providers/tf/provider.go. The suggested validation logic is already present in dataSourceRemoteStateValidate.

Discovered while investigating #3501.

### Connection to Terraform
This codebase inherited this stale FIXME from the upstream Terraform project. An identical cleanup has been submitted to Terraform in **hashicorp/terraform#38031**.

**Timeline**:
1. **July 2018**: The FIXME was introduced in hashicorp/terraform#18446 when the provider was created.
2. **May 2019**: The validation logic was implemented in hashicorp/terraform#21205, but the FIXME comment was accidentally left behind.
3. **Dec 2025**: This PR removes the stale comment.

### DIFF
--- a/internal/builtin/providers/tf/provider.go
+++ b/internal/builtin/providers/tf/provider.go
@@ -61,9 +61,7 @@ func (p *Provider) ValidateProviderConfig(_ context.Context, req providers.Valid
 
 // ValidateDataResourceConfig is used to validate the data source configuration values.
 func (p *Provider) ValidateDataResourceConfig(_ context.Context, req providers.ValidateDataResourceConfigRequest) providers.ValidateDataResourceConfigResponse {
-	// FIXME: move the backend configuration validate call that's currently
-	// inside the read method  into here so that we can catch provider configuration
-	// errors in tofu validate as well as during tofu plan.
+
 	var res providers.ValidateDataResourceConfigResponse
 
 	// This should not happen


Yantrio:
Hi @AndrewCharlesHay. Would you please mind re-adding back the checklist of items from the pull request template please? We ask that everybody fills that in for all pull requests. 

You can find a copy of this here: https://raw.githubusercontent.com/opentofu/opentofu/refs/heads/main/.github/pull_request_template.md

The most important part is that we would like to ensure that you are the author of these changes, especially since you mentioned the changes in the terraform repo. This is a formality that we would prefer to stick to here please, I'm sure you understand. Thanks!

AndrewCharlesHay:
@Yantrio done!

cam72cam (changes requested):
~Looking at the implementation of dataSourceRemoteStateValidate and dataSourceRemoteStateRead, this FIXME is still relevant.~

~Validate just checks that the backend block exists and is formatted somewhat correctly, but does not actually initialize the backend and call b.Configure.~

AndrewCharlesHay:
Do I merge or does someone else do it?

apparentlymart:
We're not allowed to look at pull requests from Terraform as a general project policy, but using `git blame` on our own codebase I found 083af21d302268863a5de0d2c5faa5150b0d6be0 that seems likely to be the OpenTofu commit equivalent to the Terraform fix mentioned in the initial writeup, since that fix happened before the fork.

Although I authored that commit it was so long ago that I don't recall what motivated that work at the time. But I would agree that the change itself and the way I described it in the commit message both suggest that I was intending to improve the behavior in the same way as that FIXME commit indicated, and so removing that comment now does seem valid.

PR #3630, opened 2025-12-30
lint: reintroduce QF1008 static check

lbordowitz:
This static check linter was removed pending completion of `--new-from-rev` tag work. That work was completed in #2931 around June 2025, but these static check exclusions slipped through the cracks. This PR aims to incrementally rectify this. 

The static check to [Omit embedded fields from selector expression ](https://staticcheck.dev/docs/checks/#QF1008) is re-enabled by this pull request, and all violations are fixed in this PR as well.

## Checklist

### Go checklist

 


yottta (approved):
Approving this since was agreed previously to go forward with this but removing all those references, especially for `Meta` related fields, seems to make things more confusing 🤷 
Let's wait for other reviews on this.

cam72cam:
I agree with @yottta, I like the explicit nature of calling out embedded structs in golang.

I'm not opposed to this, but I would rather leave this style choice up to what is chosen during implementation.

diofeher:
I agree with others, and we gain some time to read and understand the context by using the embedded structure explicitly. 

One thing to note: references work fine in IDEs using `gopls` when the embedded field is removed.

yottta:
@lbordowitz Do we want to revive this one? Or should we close it?

diofeher:
@yottta +1 on closing this one. We agreed that our preference is to explicitly call out these embedded structs for better readability.

lbordowitz:
I'm going to close it for now. We can revisit this if we ever care to define and enforce a formal style guide.

PR #3634, opened 2025-12-30
lint: reintroduce ST1005 static check

lbordowitz:
This draft pull request for removing a linter static check exclusion is for demonstration purposes only.

For the sake of user friendliness, we capitalize some errors and add punctuation at the end as well. This chafes against [ST1005](https://staticcheck.dev/docs/checks/#ST1005), which is concerned with the composability of error messages. I will recommend a couple different paths:

- Maintain the status quo: keep this linter disabled.
- Adjust all error messages to conform to Golang best practices, and modify error messages as complete sentences when they reach a system/user interface.

I have my preference, but I'll keep it close to the vest for the sake of healthy conversation.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,7 @@ linters:
     staticcheck:
       # For now, we will disable some static checks to match golang-ci-lint@v1 functionality.
       # These should be addressed once the --new-from-rev work is taken care of.
-      checks: ["all", "-QF1008", "-ST1003", "-ST1005", "-ST1012", "-ST1016"]
+      checks: ["all", "-QF1008", "-ST1003", "-ST1012", "-ST1016"]
   exclusions:
     generated: lax
     presets:


yottta:
AFAICT, there are different paths for an error to be handled:
1. Just wrapping by using `fmt.Errorf("%w")`
2. By appending the error directly to the diagnostics by using `tfdiags.Diagnostics.Append(err)`

IMO, the first part shouldn't be that concerning because if an error bubbles up outside of a diagnostic context, we use `Ui.Error()` to print it.
To analyse and postprocess a chain of wrapped errors could end up to be a quite unreliable way of doing this.

I would focus more on the diagnostic side of things, where we can create a custom error type and use that to carry information for raw error rendering and for diagnostic rendering. For example, we might work something in [here](https://github.com/opentofu/opentofu/blob/4958febbb07bb03e55f4529e83550429da2c3f71/internal/tfdiags/diagnostics.go#L77-L89).
For now I would focus only on the errors that are reported by the linter.

Maybe there is already something similar for diagnostics, so I would be curious to hear from the other maintainers that has more knowledge on this side

lbordowitz:
I'm not interested in introducing this lint checker.

PR #3652, opened 2026-01-08
Introduce plugin library and managers

cam72cam:
As we continue to evolve the existing tofu codebase with extensions to the plugin model (as well as building our new execution engine), certain areas can heavily benefit from being broken out into their own concerns.

This PR contains two linked concepts, a plugin library and a plugin manager.

The plugin library represents all plugins installed in the current tofu execution.  It also serves as a common place to cache the schemas between different managers.  This is the first commit in this work as the code change is small, but the test change is mechanically large.

The remaining commits are the more interesting.  We introduce the concept of plugin managers.  They are responsible for starting, managing, and stopping/closing plugins.  These also interact with the libraries schema cache to prevent duplicate work where possible

This is built on top of the work in https://github.com/opentofu/opentofu/pull/3617 and continues the path started there.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/plugins/provider.go
+++ b/internal/plugins/provider.go
@@ -6,10 +6,18 @@
 package plugins
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"log"
+	"sync"
+	"sync/atomic"
 
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/providers"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+	"github.com/opentofu/opentofu/version"
+	"github.com/zclconf/go-cty/cty"
 )
 
 type ProviderFactories map[addrs.Provider]providers.Factory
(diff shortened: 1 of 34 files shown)


cam72cam:
I'm still working on adding a few comments here and there, but overall this is ready for review.  I highly suggest reviewing it as two distinct commits.

internal/plugins/provider.go
```diff
@@ -27,3 +33,180 @@ func (p ProviderFactories) NewInstance(addr addrs.Provider) (providers.Interface
...
+	entry, ok := p.providerSchemas[addr]
+	if !ok {
+		entry = sync.OnceValue(func() providerSchemaResult {
+			log.Printf("[TRACE] plugins.providerManager Initializing provider %q to read its schema", addr)
+
+			var diags tfdiags.Diagnostics
+
+			if p.closed {
+				return providerSchemaResult{diags: tfdiags.Diagnostics{}.Append(fmt.Errorf("bug: unable to start provider %s, manager is closed", addr))}
+			}
```
Yantrio: What happens if the providermanager is closed before we call this, and then we want to call it again. Are we consistently going to return the diagnostics instead of retrying fetching the schema?

Yantrio: (Same for provisioners)

internal/plugins/provider.go
```diff
@@ -27,3 +33,180 @@ func (p ProviderFactories) NewInstance(addr addrs.Provider) (providers.Interface
...
+	if preparedCfg != cty.NilVal && !preparedCfg.IsNull() && !preparedCfg.RawEquals(unmarkedConfigVal) {
+		log.Printf("[WARN] ValidateProviderConfig from %q changed the config value, but that value is unused", addr)
+	}
+
+	configResp := provider.ConfigureProvider(ctx, providers.ConfigureProviderRequest{
+		// We aren't actually Terraform, so we'll just pretend to be a
+		// Terraform version that has roughly the same functionality that
+		// OpenTofu currently has, since providers are permitted to use this to
+		// adapt their behavior for older versions of Terraform.
+		TerraformVersion: "1.13.0",
```
Yantrio: Do you think we should move this to a more centralized location with a big comment about how this is intended to be consumed? This is not the place I would expect to find this magic string

internal/plugins/provisioner.go
```diff
@@ -26,3 +32,135 @@ func (p ProvisionerFactories) NewInstance(typ string) (provisioners.Interface, e
...
+	var diags tfdiags.Diagnostics
+	for name, prov := range p.instances {
+		err := prov.Stop()
+		if err != nil {
+			diags = diags.Append(fmt.Errorf("provisioner.Stop %s: %w", name, err))
+		}
+	}
+	clear(p.instances)
+	return diags.Err()
+}
```
Yantrio: These methods should be more inline with what is happening in the provider manager.

I also think we should not be clearing the instances on `Stop`

internal/tofu/eval_context.go
```diff
@@ -41,20 +40,8 @@ type EvalContext interface {
...
-	// configuration does not match the Path() of the EvalContext.
-	InitProvider(ctx context.Context, addr addrs.AbsProviderConfig, key addrs.InstanceKey) (providers.Interface, error)
-
-	// ProviderSchema retrieves the schema for a particular provider, which
-	// must have already been initialized with InitProvider.
-	//
-	// This method expects an _absolute_ provider configuration address, since
-	// resources in one module are able to use providers from other modules.
-	ProviderSchema(context.Context, addrs.AbsProviderConfig) (providers.ProviderSchema, error)
+	Providers() plugins.ProviderManager
```
Yantrio, line 45: please add comments for these new fields

internal/tofu/graph_walk_context.go
```diff
@@ -59,11 +57,7 @@ type ContextGraphWalker struct {
 	variableValuesLock sync.Mutex
 	variableValues     map[string]map[string]cty.Value
 
-	providerLock  sync.Mutex
-	providerCache map[string]map[addrs.InstanceKey]providers.Interface
-
-	provisionerLock  sync.Mutex
-	provisionerCache map[string]provisioners.Interface
+	providerLock sync.Mutex
```
Yantrio: please rename this

internal/plugins/provider.go
cam72cam: c8fa44236c

internal/plugins/provisioner.go
cam72cam: c8fa44236c

internal/plugins/provider.go
cam72cam: 1dc5be9836

internal/tofu/graph_walk_context.go
cam72cam: 2812ff6086

internal/tofu/eval_context.go
cam72cam: edd46248a8

internal/plugins/provisioner.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return provisioner.ValidateProvisionerConfig(provisioners.ValidateProvisionerConfigRequest{
+		Config: config,
+	}).Diagnostics
+}
+
+func (p *provisionerManager) ProvisionResource(ctx context.Context, typ string, config cty.Value, connection cty.Value, output provisioners.UIOutput) tfdiags.Diagnostics {
+	provisioner, err := p.provisioner(typ)
+	if err != nil {
+		return tfdiags.Diagnostics{}.Append(fmt.Errorf("failed to instantiate provisioner %q to validate config: %w", typ, err))
```
Yantrio: i think this is a copy and paste from above, we're not validating config here.

internal/plugins/provider.go
```diff
@@ -0,0 +1,239 @@
...
+	return p.providerFactories.HasProvider(addr)
+}
+
+func (p *providerManager) GetProviderSchema(ctx context.Context, addr addrs.Provider) (providers.ProviderSchema, tfdiags.Diagnostics) {
+	if p.isShutdown {
+		// It's technically possible, but highly unlikely that a manager could be shutdown while fetching the schema
+		// In that scenario, we will start and then stop the corresponding provider internally to this function and not
+		// interfere with the set of known instances.
+		return providers.ProviderSchema{}, tfdiags.Diagnostics{}.Append(fmt.Errorf("bug: unable to start provider %s, manager is shutdown", addr))
+	}
```
Yantrio: i think this should be inside the providerSchemasLock because the `shutdown` method writes it under that lock.

cam72cam: I'm attempting to avoid locking this whole function so we can run GetProviderSchema from multiple routines.

As my comment states, the worst case scenario here is that we have one additional provider start/stop cycle and won't leave any dangling procs.

internal/plugins/provisioner.go
cam72cam: 182a9dc3f9

internal/plugins/provider.go
apparentlymart: I agree that functionally it doesn't really matter if we inadvertently start and then immediately stop the provider here, but this seems like it's still a data race if `Shutdown` and `GetProviderSchema` are both called concurrently from different goroutines.

Perhaps we should use [`atomic.Bool`](https://pkg.go.dev/sync/atomic#Bool) for the `isShutdown` field as a compromise. Given that bool is sub-word-sized and we have plenty of other synchronization going on around these accesses I expect that the current implementation of Go would do just fine without it when we're not running under the race detector, but using atomic operations would make it explicit that we're accepting and tolerating concurrent access to this field, in a way that the race detector (and, if relevant some day, potentially other implementations of Go) can also understand.

(I would also remove any interaction with `instanceLock` in the `Shutdown` method, so that mutex is exclusively for access to the `instances` slice, and not to any other field.)

cam72cam: Good point @apparentlymart, I've made that change in 9228cf6872.

Does that seem like a reasonable compromise to you @Yantrio ?

Yantrio: Sounds great to me. I like this approach 👍

Yantrio (approved):

internal/plugins/provisioner.go
diofeher, line 1: I know this PR has already been merged, but it would be great if we could add some unit tests to this package.

Or are you planning to do it later, in case the design of it changes?

 cc @cam72cam

Issue #3595, opened 2025-12-16
RFC tracker: CLI layer rework and integration of a new CLI library (cobra)


yottta:
This issue is an [RFC Tracker](https://github.com/opentofu/opentofu/blob/main/rfc/20241023-rfc-tracking-issues.md) of three RFCs:
* Rework the startup logic in OpenTofu to clean legacy, hard to maintain code ([PR](https://github.com/opentofu/opentofu/pull/3471))
* Replace [mitchellh/cli](https://github.com/mitchellh/cli) with [spf13/cobra](https://github.com/spf13/cobra) ([PR](https://github.com/opentofu/opentofu/pull/3541))
* Instead of switching to [spf13/cobra](https://github.com/spf13/cobra), switch to [urfave/cli](https://github.com/urfave/cli) ([PR](https://github.com/opentofu/opentofu/pull/4445))

---

Identified issues that will be handled by this:
  * https://github.com/opentofu/opentofu/issues/748
  * https://github.com/opentofu/opentofu/issues/3500
  * https://github.com/opentofu/opentofu/issues/2239
  * https://github.com/opentofu/opentofu/issues/3050 (partially)
  * https://github.com/opentofu/opentofu/issues/3044
  * https://github.com/opentofu/opentofu/issues/3741
  * https://github.com/opentofu/opentofu/issues/3976
  * https://github.com/opentofu/opentofu/issues/420
  * https://github.com/opentofu/opentofu/issues/4020
  * https://github.com/opentofu/opentofu/issues/4343

---

There are no known blockers to this, other than the work suggested in the RFC #3541 cannot be done till the work in the RFC #3471 is not done.

---

The list below contains the work items that are part of the work on these RFCs. The list will be updated as we go:
- [x] #3713
- [x] #3725 
- [x] #3765 
- [x] #4055
- [x] #4072
- [x] #4073
- [x] #4112

---

The following list is meant to record the future possible work that we would like to do after all this is done:
* For the `tofu providers` added in [#3811](https://github.com/opentofu/opentofu/pull/3811#pullrequestreview-3890070527), we would want to add a JSON view. 
  * We could ignore this for the moment and wait for a community request to see how people would like to use the output of such a view on this command. That way, we will have more arguments for the format that we would go with.

paul-civitas:
One of the things I'm hoping can be solved here is:

Some things can only be set via env variable, such as TF_WORKSPACE and TF_LOG
[...]

PR #3659, opened 2026-01-12
add documentation for tab-completion with aliases

CouscousPie:
Resolves # 

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ ENHANCEMENTS:
 - The `yamldecode` function now supports the "merge" tag, most commonly written as `<<` where a map key would be expected, with sequences of mappings rather than just individual mappings. ([#3607](https://github.com/opentofu/opentofu/pull/3607))
 - New CLI argument `-json-into=<outfile>` has been added to support emitting both human readable and machine readable logs ([#3606](https://github.com/opentofu/opentofu/pull/3606))
 - The `s3` backend now automatically discovers and uses AWS credentials issued using [the `aws login` command](https://docs.aws.amazon.com/cli/latest/reference/login/) in AWS CLI. ([#3767](https://github.com/opentofu/opentofu/pull/3767))
+- Add a clarifying section on how to use Tab-completion with aliases to documentation as tofu don't allow the usual methods. ([#3659](https://github.com/opentofu/opentofu/pull/3659))
 
 BUG FIXES:
 
--- a/website/docs/cli/commands/index.mdx
+++ b/website/docs/cli/commands/index.mdx
@@ -116,3 +116,8 @@ manually in the shell profile, run the following command:
 ```bash
 tofu -uninstall-autocomplete
 ```
+To enable tab-completion for a shell alias such as `tf`, add the following to
+your shell profile after running the install command:
+```bash
+complete -C tofu tf
+```


website/docs/cli/commands/index.mdx
```diff
@@ -116,3 +116,8 @@ manually in the shell profile, run the following command:
 ```bash
 tofu -uninstall-autocomplete
 ```
+To enable tab-completion for a shell alias such as `tf`, add the following to
```
yottta, line 119: Image

Image

If you refer to something like this, this works out of the box.
Once `tofu -install-autocomplete` is executed and a new shell session is started, any alias pointing to the actual `tofu` binary will have the autocomplete scripts working.

Maybe I misunderstood it, so please correct me if I am wrong here.

CouscousPie: For me it does not seem to work like this on Debian 13.2 with Bash 5.2.37. Furthermore, it appears to me a little unusual to use a sub-shell in an alias instead of simply specifying something like `alias tf="tofu"`. If this works for you, that is great. Many ways lead to the same place, but let's not discuss which one to take, but how to help people find their way. Currently the documentation is not clear about how to get to Tab-completion with an alias. Hence, why I opened this pull-request.

To me the whole completion installation is a little bit weird, to begin with. Is there a specific reason, why it doesn't adhere to conventional `source <(tofu completion bash)`? Just putting something in the users dotfiles, without asking them or at least tell them about it looks in my eyes to be kind of the opposite of what "open" means to me, where the user is in full control. Not to start with how that may be prone to errors. I personally think the whole completion should receive an overhaul, but adding a clarification how to use the current state with aliases is the smallest possible step to improve UX.

yottta: I tested it with the debian:13.2 docker container and yes, indeed, does not work as it does in my environment.

> Is there a specific reason, why it doesn't adhere to conventional source <(tofu completion bash)?

Legacy reasons that few of us know (if any).

> I personally think the whole completion should receive an overhaul, but adding a clarification how to use the current state with aliases is the smallest possible step to improve UX.

We plan on working on that. In case you want to follow along, feel free to check https://github.com/opentofu/opentofu/issues/3595

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ ENHANCEMENTS:
 - Improve performance around provider checking and schema management. ([#2730](https://github.com/opentofu/opentofu/pull/2730))
 - `tofu init` now fetches providers and their metadata in parallel. Depending on provider size and network properties, this can reduce provider installation and checking time. ([#2729](https://github.com/opentofu/opentofu/pull/2729))
 - The `yamldecode` function now supports the "merge" tag, most commonly written as `<<` where a map key would be expected, with sequences of mappings rather than just individual mappings. ([#3607](https://github.com/opentofu/opentofu/pull/3607))
+- Add a clarifying section on how to use Tab-completion with aliases to documentation as tofu don't allow the usual methods. ([#3659](https://github.com/opentofu/opentofu/pull/3659))
```
lbordowitz, line 23: Please remove this changelog entry

website/docs/cli/commands/index.mdx
Yantrio, line 119: Would you mind putting a different alias in here instead of `tf` please?

I do not think its a good idea to encourage people to confuse OpenTofu and HashiCorp Terraform which are both in this space.

I'm not sure on the best alternative, but with `tf` being a very popular way to address HashiCorp's Terraform I think it's best if we avoid it. Maybe just `t` is fine?

Yantrio:
Would you also be able to sign the DCO please? you can find more information in the checks below, or at this URL: https://github.com/opentofu/opentofu/pull/3659/checks?check_run_id=63091184758

cam72cam:
@CouscousPie Are you still interested in working on this PR? We can not continue the review process until the DCO is signed.

PR #3662, opened 2026-01-12
New `tofu graph` implementations for the new runtime approach

apparentlymart:
I originally started this naively thinking it'd be a relatively little thing I could use to reload context back into my brain after the holiday period.

Unfortunately it quickly got out of hand and in retrospect I think this is too much new complexity to include at this early stage of the work on https://github.com/opentofu/opentofu/issues/3414.

I'm expecting that we will want something like this _eventually_ though, and so I'm opening this as a draft for now just to make it easier to find in future in case it's useful as a starting point for some later work. I don't intend to ever merge this in its current form though, and expect to close this at some point before that overall project is finished.

### DIFF
--- a/internal/lang/eval/config_debug_graph.go
+++ b/internal/lang/eval/config_debug_graph.go
@@ -7,11 +7,12 @@ package eval
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"strings"
 
 	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/dag"
+	"github.com/opentofu/opentofu/internal/dag/graphviz"
 	"github.com/opentofu/opentofu/internal/lang/eval/internal/configgraph"
 	"github.com/opentofu/opentofu/internal/lang/eval/internal/evalglue"
 	"github.com/opentofu/opentofu/internal/lang/grapheval"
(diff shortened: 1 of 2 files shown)


apparentlymart:
Just to make it easier for a future reader to understand what this code actually does, here are some images generated from the expected output of the two main test cases in this PR.

---

This is a "resource instance graph" rendering, from the expected output of `TestConfigInstanceWriteGraphvizGraphForDebugging`:

graphviz-rendered version of the expected output of the test TestConfigInstanceWriteGraphvizGraphForDebugging, described more below

This shows how the new evaluator is able to determine which resource instances are desired as part of constructing the graph, whereas today's `tofu graph` is based on the traditional runtime's planning graph builder which only finds relationships between entire resources. In particular, the dynamic analysis strategy detects that `foo.beep[0]` only depends on `foo.baz[0]`, whereas today's runtime assumes that all instances of `foo.beep` depend on all instances of `foo.baz`.

The node `foo.deferred[*]` is also illustrating that this can handle resources whose `count` is not yet known, using a dashed outline to indicate that this graph node is just a placeholder for zero or more objects we can't predict yet.

---

This is an "execution graph" rendering, from the expected output of `TestGraphWriteGraphvizRepr`:

graphviz-rendered version of the expected output of the test TestGraphWriteGraphvizRepr, described more below

In this case we model the data flow into specific operation arguments rather than into entire operations, which gives a little more insight into what each of the dependencies represents. For example, the edges related to `r[0]`, representing the `OpenProvider` result, connect with the "provider client" arguments on the other operations.

(Because operation graph arguments are currently positional rather than named it relies on the reader knowing the signature of each of the operation types. The graph renderer doesn't actually know what each of the arguments represents.)

PR #3672, opened 2026-01-14
Make it clear for the external key providers when it needs to generate decryption key

### LINKED ISSUE #3666: External key provider documentation/examples are incorrect about `null` inputs
Hi,

I'm just looking to implement a simple external key provider that retrieves a static encryption key from 1Password by utilizing the 1Password CLI.

The documentation at https://opentofu.org/docs/language/state/encryption/#external-experimental currently states that the input would be `null` if only encryption is needed and no decryption.
In practice however, what I'm seeing is instead of `null`, the actual input is `{"external_data":null}`, which also breaks at least the POSIX Shell and Python examples from behaving the way they're intended to be, which are expecting a root-level `null`.
Once an encrypted state exists it will instead pass an empty dict for `external_data`: `{"external_data":{}}`.

The POSIX Shell example also has an additional issue where it is using `echo -n`, which isn't POSIX-compatible, as POSIX echo does not have flags: https://www.shellcheck.net/wiki/SC3037.

Reproduction steps:
1. Start in an empty folder with the tofu code and `key-provider.sh` below
2. `tofu init` - this will pass `{"external_data":null}` to the key provider
3. `tofu plan` - this will still pass `{"external_data":null}` to the key provider
4. `tofu apply` - this will still pass `{"external_data":null}` to the key provider
5. `tofu apply` - now that we have an encrypted state this will pass `{"external_data":{}}` instead

```tofu
terraform {
  encryption {
    key_provider "external" "ext" {
      command = [ "./key-provider.sh" ]
    }

    key_provider "pbkdf2" "main" {
      chain = key_provider.external.ext
    }

    method "aes_gcm" "main" {
      keys = key_provider.pbkdf2.main
    }

    state {
      method   = method.aes_gcm.main
      enforced = true
    }

    plan {
      method   = method.aes_gcm.main
      enforced = true
    }
  }
}

output "test" {
  value = 1
}
```

```sh
#!/bin/sh

set -e

# Output the header as a single line:
echo '{"magic":"OpenTofu-External-Key-Provider","version":1}'

# Read the input metadata.
INPUT=$(echo -n $(cat))

echo "$(date +%F-%T) - $INPUT" >> key-provider.log

if [ "${INPUT}" = "null" ]; then
  # We don't have metadata and shouldn't output a decryption key.
  cat << EOF
{
  "keys":{
    "encryption_key":"AQIDBAUGBwgJCgsMDQ4PEA=="
  },
  "meta":{
    "external_data":{}
  }
}
EOF
else

  # We have metadata and should output a decryption key. In our simplified case
  # it is the same as the encryption key.
  cat << EOF
{
  "keys":{
    "encryption_key":"AQIDBAUGBwgJCgsMDQ4PEA==",
    "decryption_key":"AQIDBAUGBwgJCgsMDQ4PEA=="
  },
  "meta":{
    "external_data":{}
  }
}
EOF
fi
```

OpenTofu version: v1.11.2

yottta:
Before this change, the payload that OpenTofu was sending to the external key-provider was always containing data (eg: `{"external_data": {}}`), which is in disagreement with the initial feature design.
With these changes, when the key provider is configured for only encryption, it sends a plain `null` payload and when it requires decryption key, it sends a non null payload containing at least the key for external data (eg: `{"external_data": {}}`).

This PR also fixes some inconsistencies and best practices in the scripts used given as an example for an external key provider.

Resolves #3666

Resolves # 

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ BUG FIXES:
 - Fixed `tofu test` with `mock_provider` failing during cleanup when `lifecycle { ignore_changes }` references a block. ([#3644](https://github.com/opentofu/opentofu/issues/3644))
 - In JSON syntax, the state encryption method configuration now allows specifying keys using both normal expression syntax and using template interpolation syntax. Previously only the template interpolation syntax was allowed, which was inconsistent with other parts of the encryption configuration. ([#3654](https://github.com/opentofu/opentofu/issues/3654))
 - No longer generate spurious error messages about incorrectly-detected provider reference problems when modules fail to load during the construction of a configuration tree. ([#3681](https://github.com/opentofu/opentofu/pull/3681))
+- OpenTofu consistently sends "null" to external `key_provider` programs when only encryption key is requested ([#3672](https://github.com/opentofu/opentofu/pull/3672))
 
 ## Previous Releases
 


cam72cam (approved):

Yantrio (commented):
This looks good to me, I would approve but i think this is worthy of a changelog entry incase someone relied on this behaviour

yottta:
> This looks good to me, I would approve but i think this is worthy of a changelog entry incase someone relied on this behaviour

Good point! Done in 3f2ae9207675b888dfb8ebfc23fa7fd7444e4b75

PR #3674, opened 2026-01-15
Tenuous initial wiring of the new-style apply engine

apparentlymart:
Continuing along the path toward the "walking skeleton" milestone of https://github.com/opentofu/opentofu/issues/3414, this is some initial work to stub out the apply engine and wire it up to everything else just enough to see it fail due to the execution graph model not being quite right yet. :grinning: 

Because we're still in an exploration phase of this project I'm moving forward in larger steps than we typically would, and so this PR actually deals with a few different concerns at once:

- While arranging to actually save real execution graphs from the planning phase I found some gaps in the planning engine implementation that I dealt with in the various commits affecting `engine/planning`.
- The saved plan workflow between `tofu plan -out=tfplan` and `tofu apply tfplan` now minimally works under the new runtime engine, with the execution graph traveling as a sidecar on the `plans.Plans` data structure for now.
- The `engine/apply` package has some minimal initial work to stub out how its different parts might fit together and wire it up to the evaluator just enough to let an execution graph run, though any non-trivial graph is likely to immediately fail because the opcode implementations in `package execgraph` are not yet complete.

    Notably though I have seen the `opOpenProvider` operation succeed, illustrating that the overall compilation and execution behavior is working even though the operation-specific logic is incomplete.

Because of the kinda-scattered nature of the work here, it might be easier to [review it on a commit-by-commit basis](https://github.com/opentofu/opentofu/pull/3674/commits).

I decided to submit this now, at a rather-arbitrary stage of development, because my next goal will be to rework the `execgraph` model to be a better match for the work we actually need to do during the apply phase, and it'll be easier to review _that_ diff without all of the stubbing noise from this PR.

Some of those gaps are highlighted in this diff as `FIXME`/`TODO` comments, but more broadly that means deciding on a more complete set of opcodes for the execution graph, updating the planning engine to generate graphs using those new opcodes, and then making the opcode implementations actually complete enough for successful basic execution.

As before, all of the code here is guarded behind both a compile-time and runtime experimental feature opt-in, so nothing here is reachable in stable OpenTofu releases, and even in builds that have experiments enabled this code won't run unless opted in with the `TOFU_X_EXPERIMENTAL_RUNTIME` environment variable.

### DIFF
--- a/internal/engine/applying/exec_context.go
+++ b/internal/engine/applying/exec_context.go
@@ -34,6 +34,7 @@ type executionContext struct {
 	// state, and so we could simplify this to be just an immutable map of
 	// pre-decoded state objects without any mutex.
 	priorState *syncState
+	oracle     *eval.ApplyOracle
 	plugins    plugins.Plugins
 
 	// graph refers back to the execution graph that this object was
@@ -45,7 +46,7 @@ type executionContext struct {
 	graph *execgraph.CompiledGraph
 }
 
-func compileExecutionGraph(ctx context.Context, plan *plans.Plan, plugins plugins.Plugins) (*execgraph.CompiledGraph, *executionContext, tfdiags.Diagnostics) {
+func compileExecutionGraph(ctx context.Context, plan *plans.Plan, oracle *eval.ApplyOracle, plugins plugins.Plugins) (*execgraph.CompiledGraph, *executionContext, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 
 	execGraph, err := execgraph.UnmarshalGraph(plan.ExecutionGraph)
@@ -89,12 +90,21 @@ func compileExecutionGraph(ctx context.Context, plan *plans.Plan, plugins plugin
 	}
 	execCtx.priorState = priorState
 
+	execCtx.oracle = oracle
+	execCtx.plugins = plugins
+
 	return compiledGraph, nil, diags
 }
 
 // DesiredResourceInstance implements [execgraph.ExecContext].
 func (c *executionContext) DesiredResourceInstance(ctx context.Context, addr addrs.AbsResourceInstance) *eval.DesiredResourceInstance {
-	panic("unimplemented")
+	// FIXME: We're currently just ignoring the diagnostics completely here
+	// because this API was modelled after the apply phase where we proactively
+	// walk the entire configuration collecting diagnostics for everything,
+	// but we don't do that during the apply phase so we should instead model
+	// the "desired resource instance" request as a fallible operation here.
+	ret, _ := c.oracle.DesiredResourceInstance(ctx, addr)
+	return ret
 }
 
 // NewProviderClient implements [execgraph.ExecContext].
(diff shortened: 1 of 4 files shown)


internal/engine/applying/exec_context.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/opentofu/opentofu/internal/plans"
+	"github.com/opentofu/opentofu/internal/providers"
+	"github.com/opentofu/opentofu/internal/states"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+type executionContext struct {
+	// priorState is the state that was current at the end of the planning
+	// phase, which the actions in the execution graph are intended to start
+	// from. This must remain unmodified throughout graph execution.
```
yottta, line 29: > This must remain unmodified throughout graph execution

I just want to comment that I love the idea! Please do! In this case, immutability FTW! ❤️

internal/engine/applying/exec_context.go
```diff
@@ -0,0 +1,142 @@
...
+	// pre-decoded state objects without any mutex.
+	priorState *syncState
+	oracle     *eval.ApplyOracle
+	plugins    plugins.Plugins
+
+	// graph refers back to the execution graph that this object was
+	// instantiated for.
+	//
+	// The compiled graph also has a pointer to the execution context, so this
+	// field must be populatd only after the graph has already been successfully
```
yottta:
```suggestion
	// field must be populated only after the graph has already been successfully
```

internal/engine/applying/exec_context.go
```diff
@@ -0,0 +1,142 @@
...
+	// There is a cyclic dependency between the execution context and the
+	// compiled execution graph, so we compile the graph against an unpopulated
+	// execution context first and then populate it afterwards. This is okay
+	// because the execution graph compiler guarantees that it won't call
+	// any methods on the execution context until the graph is actually
+	// executed.
+	execCtx := &executionContext{}
+	compiledGraph, moreDiags := execGraph.Compile(execCtx)
+	diags = diags.Append(moreDiags)
+	if moreDiags.HasErrors() {
```
yottta, line 78:
```suggestion
	if diags.HasErrors() {
```

internal/engine/applying/exec_context.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, nil, diags
+	}
+	execCtx.graph = compiledGraph
+
+	// We pre-populate a bunch of other information here too so that we can
+	// have a single place to catch errors involving inconsistencies in the
+	// input, missing provider dependencies, etc.
+	priorState, moreDiags := syncStateFromPriorState(ctx, plan.PriorState, plugins)
+	diags = diags.Append(moreDiags)
+	if moreDiags.HasErrors() {
```
yottta, line 88:
```suggestion
	if diags.HasErrors() {
```

internal/engine/applying/exec_context.go
apparentlymart: This particular thing is intentional: it's asking "have we added any _new_ errors?", since anything that was already in `diags` before the append was already checked for errors previously.

In this particular case it's not an important difference because none of these codepaths can currently generate warnings, but this is the typical pattern we follow for diagnostics elsewhere so that when there _are_ warnings we don't keep re-checking whether they are errors every time.

PR #3680, opened 2026-01-15
Provider validation for `enabled` at module calls

diofeher:
This validation on provider config was missed while I was working on the `enabled` field for module calls.

This needs to be backported to v1.11.

The context for this fix is the following:

@cam72cam noticed we missed this validation while working on other things. If this validation doesn't exist, if a module has a provider config inside it, and if the module uses expansion on it (like count, for_each, or enabled), and if the provider is gone, the resource won't be deleted if the module is deleted conditionally.

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 
-- Change the validation of modules using provider config to not allow `enabled` configuration. ([#3680](https://github.com/opentofu/opentofu/pull/3680))
+- Modules containing local provider configurations now also reject the `enabled` argument, matching existing behavior for `count`, `for_each`, and `depends_on`. ([#3680](https://github.com/opentofu/opentofu/pull/3680))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 
+- Change the validation of modules using provider config to not allow `enabled` configuration. ([#3680](https://github.com/opentofu/opentofu/pull/3680))
```
Yantrio: This isn't super clear to me if i just read the changelog and the github issue, im not sure on some good wording but i think we should maybe re-write this with the changelog-only reader in mind.

diofeher: Do you think that wording works better? https://github.com/opentofu/opentofu/pull/3680/commits/9e5aebff8b64a4ee58b88b0514e83f830989f4fd

apparentlymart (approved):

PR #3681, opened 2026-01-15
configs: Skip provider ref validation when modules fail to load

apparentlymart:
We previously already did some partial skipping of provider reference validation when there are modules in the tree that failed to load, but we were still running the main logic that checks whether the providers passed between module boundaries are consistent.

Because that validation involves checking consistency between different modules it will tend to generate spurious errors when run against a partially-loaded module tree. Therefore we must also skip the other validation steps whenever at least one module failed to load, so that the reported diagnostics will only describe the problem loading the module and not produce additional misleading reports describing downstream problems caused by a module's content being incomplete due to those errors.

This one is kinda awkward to write a test for because it's a very specific interaction that seems unlikely to arise again, but just to have _some_ representation of this historical problem I wrote a simple test that intentionally makes one module fail to load and then tests that the result only contains the diagnostic related to that intentional failure. Before this fix that test would fail because of the spurious additional errors produced by the functions we're now skipping.

Without this fix, the test fixture added here generates spurious additional error diagnostics complaining about non-existent problems with the provider references between modules:

```
╷
│ Error: Provider type mismatch
│ 
│   on uninit-module-and-provider-refs.tf line 25, in module "child":
│   25:     example.a = example.a
│ 
│ The local name "example.a" in the root module represents provider
│ "hashicorp/example", but "example.source" in module.child
│ represents "example/example".
│ 
│ Each provider has its own distinct configuration schema and provider
│ types, so this module's "example.a" can be assigned only a
│ configuration for example/example, which is not required by
│ module.child.
╵
```

(This bug was originally reported in https://github.com/hashicorp/terraform/issues/38067 .)


yottta (approved):
As you said in the description, ran the new test after reverted the config_build.go changes and indeed confirms the fix.
Thanks for this.

PR #3689, opened 2026-01-16
go.mod: Upgrade to Go 1.25.6

apparentlymart:
This minor revision includes a fix of a problem where a crafted zip file can cause denial of service by taking a very long time to open the first file in the archive, which can potentially be triggered by a maliciously-crafted provider or module distribution archive fetched during `tofu init`, as discussed in https://github.com/golang/go/issues/77102.

I'm intending to open a second PR making a similar change in the v1.11 branch. Because there haven't yet been any v1.12 stable releases, I plan to add the changelog entry for this into the v1.11 changelog in that other PR.


yottta (commented):
Shouldn't the [.go-version](https://github.com/opentofu/opentofu/blob/2e09323694cb5eb9ef11ff3588a679e2b8497ead/.go-version#L1) be updated too? 🤔 
I know that we barely use that file, but for the sake of consistency, I would update that too.

PR #3706, opened 2026-01-27
Read state refactor

### LINKED ISSUE #3643: Legacy state reader incorrectly triggered when TF_DATA_DIR contains terraform.tfstate
## Description

When `TF_DATA_DIR` points to a directory containing a `terraform.tfstate` file, OpenTofu reads it through the legacy code path in `internal/legacy/tofu/state.go`, which rejects valid version 4 state files with an incorrect error message.

## Steps to Reproduce

```bash
# Setup: any tofu project with providers configured
mkdir -p /tmp/test-state
cd /path/to/tofu/project

# Set TF_DATA_DIR and run tofu
export TF_DATA_DIR="/tmp/test-state"
tofu init
tofu apply -auto-approve  # Creates terraform.tfstate in TF_DATA_DIR

# Later, any state operation fails:
tofu state list
```

## Expected Behavior

State operations work normally since the state file is valid version 4 format created by the same OpenTofu version.

## Actual Behavior

```
Error: Failed to load state: OpenTofu 1.11.2 does not support state version 4, please update.
```

The state file clearly shows it was created by OpenTofu 1.11.2:
```json
{"version":4,"terraform_version":"1.11.2","serial":1,...}
```

## Root Cause Analysis

The error originates from `internal/legacy/tofu/state.go` which has:

```go
const StateVersion = 3  // Not updated for version 4

func ReadStateV3(jsonBytes []byte) (*State, error) {
    // ...
    if state.Version > StateVersion {  // 4 > 3 = true
        return nil, fmt.Errorf("OpenTofu %s does not support state version %d, please update.",
            tfversion.SemVer.String(), state.Version)
    }
    // ...
}
```

When `TF_DATA_DIR` contains a `terraform.tfstate` file, the backend initialization code appears to read it through this legacy path instead of the modern state reader in `internal/states/statefile/`.

## Workaround

Store the state file in a different directory from `TF_DATA_DIR`:

```bash
export TF_DATA_DIR="/tmp/test-state/data"  # For modules/providers only
mkdir -p "$TF_DATA_DIR"
tofu init
tofu apply -state=/tmp/test-state/terraform.tfstate  # State file outside TF_DATA_DIR
```

## Environment

- **OpenTofu Version:** 1.11.2
- **OS:** Linux (Debian 12/13, also reproduced on Proxmox VE 8.x)
- **Provider:** bpg/proxmox v0.91.0 (though likely provider-independent)

## Additional Context

This was discovered during E2E testing with isolated state directories per environment. The pattern of using `TF_DATA_DIR` for state isolation (setting it to a directory containing both the terraform data and state file) triggers this bug consistently.

The legacy state code in `internal/legacy/tofu/state.go` appears to be intended for very old state formats, but it's being triggered incorrectly for modern version 4 states when they happen to be located in `TF_DATA_DIR`.

KrishnaSindhur:
Resolves #3643 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/clistate/cli_state_test.go
+++ b/internal/command/clistate/cli_state_test.go
@@ -157,9 +157,6 @@ func TestWriteState_SetsVersion(t *testing.T) {
 
 func TestNewState(t *testing.T) {
 	state := NewState()
-	if state == nil {
-		t.Fatal("NewState returned nil")
-	}
 	if state.Version != StateVersion {
 		t.Errorf("expected version %d, got %d", StateVersion, state.Version)
 	}
@@ -171,10 +168,6 @@ func TestCLIState_DeepCopy(t *testing.T) {
 	}
 
 	copied := original.DeepCopy()
-	if copied == nil {
-		t.Fatal("DeepCopy returned nil")
-	}
-
 	if copied.Version != original.Version {
 		t.Errorf("version mismatch: %d != %d", copied.Version, original.Version)
 	}


internal/command/clistate/cli_state.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/opentofu/opentofu/internal/plans"
+	tfversion "github.com/opentofu/opentofu/version"
+	"github.com/zclconf/go-cty/cty"
+	ctyjson "github.com/zclconf/go-cty/cty/json"
+)
+
+// StateVersion is the current supported version for CLI state files.
+const StateVersion = 3
+
+type CLIState struct {
```
yottta, line 28: `TFVersion`, `Serial` and `Lineage` could be removed too IINW.

yottta (commented):
Generally, it looks good.
Could you please check these places too:
* https://github.com/KrishnaSindhur/opentofu/blob/c4f49afc10060ceaaba3ac7367557afcf75f061d/internal/command/command_test.go#L453
* https://github.com/opentofu/opentofu/blob/5f8ce97dcd6def43b2c593faa82593617eedac19/internal/command/command_test.go#L776
* https://github.com/opentofu/opentofu/blob/5f8ce97dcd6def43b2c593faa82593617eedac19/internal/command/command_test.go#L497
* https://github.com/opentofu/opentofu/blob/5f8ce97dcd6def43b2c593faa82593617eedac19/internal/command/command_test.go#L716
* https://github.com/opentofu/opentofu/blob/5f8ce97dcd6def43b2c593faa82593617eedac19/internal/command/workspace_command_test.go#L384

internal/command/clistate/cli_state.go
KrishnaSindhur: Ok I will check removing and running this once

KrishnaSindhur: It might run but legacy `State` also have this!!

yottta: Yes, it does, but since this configuration is meant to store only the backend configuration, we would like to keep mostly (if not only) that field.
There is a mention of this in the issue discussion (https://github.com/opentofu/opentofu/issues/3643#issuecomment-3739547627):
> ...just the Backend field and its contents.

KrishnaSindhur: `TFVersion` is not consumed anywhere meaningful mostly in test. Other usages are in LocalState and init(). Should I try to refactor this and remove that dependency as well?

yottta: Your changes from 69cf4b3da0ef09aa36e6d6d692fef019f65a8bf8 look good.

Revisiting again all the changes and the guidance provided in the issue, I can safely say that `Remote` can be removed too.
There is a validation [there](https://github.com/opentofu/opentofu/blob/1e0a920a3418eded87bf55990e2762fb41e6ed78/internal/command/meta_backend.go#L636) on that field but it's the only usage that I see, except tests.
Also, if you want to get a better understanding, try creating a new configuration, apply it and check the contents of `.terraform/terraform.tfstate` (or to be more precise, check `<TF_DATA_DIR>/terraform.tfstate` content).

KrishnaSindhur: yes will check from here `.terraform/terraform.tfstate` will give more confidence

internal/command/meta_backend.go
```diff
@@ -633,17 +632,6 @@ func (m *Meta) backendFromConfig(ctx context.Context, opts *BackendOpts, enc enc
 		}
 	}()
 
-	if !s.Remote.Empty() {
```
yottta, line 636: @opentofu/maintainers Just throwing it out there: by removing this bit here, wouldn't be advisable to update our migration guides to suggest this intermediary step that we remove the error for?

Or maybe we can add back a dummy `Remote` object into the new `CLIState` and during unmarshaling of the CLI state, to return an error from there if the object is non-nil?

Asking to be sure that I am not missing something that could hinder the UX once we have this merged.

internal/command/clistate/cli_state.go
apparentlymart: I think this was clear already but just to make sure, in case this historical context is useful:

The "remote" field was where much older versions of Terraform (v0.8 and earlier) stored remote state settings. It was replaced by "backend" in Terraform v0.9, but some support for "remote" was kept for a while to allow automatic migration from the old remote state style to the new backend style.

Terraform v0.12 finally removed that support aside from the error message that @yottta found, because that was the release that first introduced the format we currently use for the main state snapshots, and so at that point this legacy format was _only_ for representing the "backend" information that we're refactoring in this PR.

Therefore I agree that it's okay to remove all of the handling of "remote". It has never been supported in any version of OpenTofu, and even in Terraform it hasn't been supported since a long-obsolete older version.

internal/command/meta_backend.go
apparentlymart: I just left a longer comment on a different part of the review before I noticed this :man_facepalming:, but just to repeat the relevant part:

This check is here primarily so that folks who were upgrading directly from Terraform v0.8 to v0.12 would get direct feedback that they need to upgrade to and use Terraform v0.11, first to get their `.terraform/terraform.tfstate` file updated to the new structure.

Therefore I think we can safely remove any handling of it from OpenTofu. It would only be relevant if someone was trying to upgrade directly from Terraform v0.8 to OpenTofu, but there are already plenty of reasons why that migration wouldn't work.

lbordowitz (commented):
This looks great!

Were you considering removing the `State` struct and related tests from legacy?

KrishnaSindhur:
> Were you considering removing the `State` struct and related tests from legacy?


Yes sure I can try that, will take a look!!

internal/command/clistate/read_file_test.go
```diff
@@ -0,0 +1,336 @@
...
+	if err == nil {
+		t.Fatal("expected error for missing version, got nil")
+	}
+	if !strings.Contains(err.Error(), "does not support state version 0") {
+		t.Errorf("unexpected error message: %v", err)
+	}
+}
+
+func TestReadState_UnsupportedVersion(t *testing.T) {
+	reader := strings.NewReader(`{"version": 999, "serial": 1, "lineage": "test"}`)
```
yottta: Can you convert this test to a table test to check also the versions that used previously? This way we make it clear that this is decoupled from the actual [state versioning](https://github.com/opentofu/opentofu/blob/2585a63df090607cdd1e38a5a05dbf9eff51466e/internal/states/statefile/read.go#L132-L139).

internal/command/clistate/cli_state_test.go
```diff
@@ -0,0 +1,336 @@
...
+	if err := WriteState(original, buf); err != nil {
+		t.Fatalf("WriteState failed: %v", err)
+	}
+
+	result, err := ReadState(buf)
+	if err != nil {
+		t.Fatalf("ReadState failed: %v", err)
+	}
+
+	if result.Backend == nil || result.Backend.Type != "s3" {
```
yottta: I see that the `Version` check has been removed. This is more of a nitpick since this is already tested in above tests.

KrishnaSindhur:
> > Were you considering removing the `State` struct and related tests from legacy?
> 
> Yes sure I can try that, will take a look!!

@lbordowitz If its okay, I can create a new PR removing some legacy state struct that still has dependencies in a few files.

internal/command/clistate/cli_state.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	ver := &jsonVersionOnly{}
+	if err := json.Unmarshal(jsonBytes, ver); err != nil {
+		return nil, fmt.Errorf("decoding CLI state file version failed: %w", err)
+	}
+
+	if ver.Version != StateVersion {
+		return nil, fmt.Errorf(
+			"OpenTofu %s does not support state version %d, please update.",
```
yottta:
```suggestion
			"OpenTofu %s does not support CLI state version %d, please update.",
```
To keep this consistent with the rest of errors above.

internal/command/clistate/cli_state.go
```diff
@@ -0,0 +1,181 @@
...
+	mu sync.Mutex
+}
+
+func NewState() *CLIState {
+	s := &CLIState{}
+	s.init()
+	return s
+}
+
+func (s *CLIState) Init() {
```
yottta: This method looks unused which makes the mutex unneeded too.
Please double check, and if it's the case, please remove.

And since `init()` (unexported method) is used only in `NewState()` you might get rid of it too and set the version in the `CLIState` struct initialisation.

internal/command/meta_backend.go
```diff
@@ -633,17 +625,6 @@ func (m *Meta) backendFromConfig(ctx context.Context, opts *BackendOpts, enc enc
...
-		// Legacy remote state is no longer supported. User must first
-		// migrate with Terraform 0.11 or earlier.
-		diags = diags.Append(tfdiags.Sourceless(
-			tfdiags.Error,
-			"Legacy remote state not supported",
-			"This working directory is configured for legacy remote state, which is no longer supported from Terraform v0.12 onwards, and thus not supported by OpenTofu, either. To migrate this environment, first run \"terraform init\" under a Terraform 0.11 release, and then upgrade to OpenTofu.",
-		))
-		return nil, diags
-	}
-
```
diofeher, line 646: I believe it's important to include a CHANGELOG for this specific update since it alters the current behavior of no longer supporting legacy remote state. (I know it's a really old version, but it would be good to document it.)

internal/command/clistate/cli_state_test.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		t.Fatalf("ReadState failed: %v", err)
+	}
+
+	if result.Version != StateVersion {
+		t.Errorf("version mismatch: got %d, want %d", result.Version, StateVersion)
+	}
+
+	if result.Backend == nil || result.Backend.Type != "s3" {
+		if result.Backend == nil {
```
diofeher: Can we simplify this check?

Something like 

```
  if result.Backend == nil {
      t.Error("expected backend to be non-nil")
  }
  if result.Backend.Type != "s3" {
      t.Errorf("backend type: got %q, want %q", result.Backend.Type, "s3")
  }
```

internal/command/meta_backend.go
yottta: I don't know if you've seen [this conversation](https://github.com/opentofu/opentofu/pull/3706#discussion_r2797097743).
If you didn't, can you check it? I am not sure that after that conversation a changelog is necessary 🤷 . But I might be wrong.

diofeher: omg, I missed this one. Thanks @yottta  

What Martin said makes sense.

PR #3710, opened 2026-01-28
engine/ephemeral: Wire together basic ephemeral functionality

cam72cam:
This is based on the work started in https://github.com/opentofu/opentofu/pull/3698 and is working toward the "walking skeleton" phase of https://github.com/opentofu/opentofu/issues/3414

There is a *lot* missing on error handing, handing provider result values, and testing.  It does however show how we could wire together the open/close logic for both providers and ephemeral.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/internal/exec/ephemeral.go
+++ b/internal/engine/internal/exec/ephemeral.go
@@ -1,3 +1,8 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
 package exec
 
 import (
--- a/internal/engine/planning/execgraph_ephemeral.go
+++ b/internal/engine/planning/execgraph_ephemeral.go
@@ -1,3 +1,8 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
 package planning
 
 import (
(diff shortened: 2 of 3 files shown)


apparentlymart:
While working on https://github.com/opentofu/opentofu/pull/3728 I realized that the execgraph opcodes I previously designed for ephemeral resource instances don't really fit the intended design constraint that objects should flow directly between nodes in the execution graph, rather than keeping track of things in sidecar data structures: the current argument passed to `EphemeralClose` is not something we can directly use to close the ephemeral resource instance.

I'm considering adjusting that to split `EphemeralOpen` into two opcodes, where `EphemeralOpen` returns a new type of object that has these methods:

```go
func (ei *ActiveEphemeralResourceInstance) State() *exec.ResourceInstanceObject
func (ei *ActiveEphemeralResourceInstance) Close() tfdiags.Diagnostics
```

That would then leave us with the following three opcodes:

- `EphemeralOpen(*DesiredResourceInstance) -> *ActiveEphemeralResourceInstance` to open it
- `EphemeralState(*ActiveEphemeralResourceInstance) -> *ResourceInstanceObject` to call `State` to get what we need to produce the result for ongoing evaluation
- `EphemeralClose(*ActiveEphemeralResourceInstance, await(...))` to call `Close` once we're done working with this ephemeral resource instance.

I'm looking at this from an execgraph lens here but I think this hypothetical `ActiveEphemeralResourceInstance` type could also be how the planning engine keeps track of stuff it needs to eventually close, and can then encapsulate the refreshing behavior to share it between the two phases.

Maybe we could give something like this a try as part of adapting this after https://github.com/opentofu/opentofu/pull/3728 is merged.

apparentlymart:
In some subsequent work I spotted an unhandled complexity of this which I didn't catch during initial review, which I've recorded as a comment here:

https://github.com/opentofu/opentofu/blob/c08c63225d62ec27d6e43cfe3d4360c02fcf0795/internal/engine/planning/plan_ephemeral.go#L69-L98

PR #3713, opened 2026-01-29
`-chdir` unification with the workdir logic. Removal of workdir proxy methods from Meta

yottta:
Part of #3595 

This PR unify the usage of `-chdir` in the [`workdir` package](https://github.com/opentofu/opentofu/tree/workdir-logic-in-one-place/internal/command/workdir).

Because of the amount of changes done in the unit tests, I recommend reviewing this commit by commit, to be able to properly visualise the important changes.

ℹ️ The changes in this PR follow the general approach suggested in [the RFC](https://github.com/opentofu/opentofu/pull/3471), reducing the footprint of the `Meta` structure. Future PRs regarding the aforementioned RFC will do similar work with the final intent that once `Meta` contains only "components", to move those out from the `Meta` struct directly into the command objects.

## Checklist

### Go checklist

 

### DIFF
--- a/cmd/tofu/main.go
+++ b/cmd/tofu/main.go
@@ -223,7 +223,9 @@ func realMain() int {
 	// Initialize the backends.
 	backendInit.Init(services)
 
-	binName, args := binNameAndArgs()
+	// Get the command line args.
+	binName := filepath.Base(os.Args[0])
+	args := os.Args[1:]
 
 	// Create the workdir and apply the -chdir options.
 	// The logic inside [NewWorkdir] handles the TF_DATA_DIR env var too.
@@ -498,8 +500,4 @@ func mkConfigDir(configDir string) error {
 	}
 
 	return err
-}
-
-func binNameAndArgs() (string, []string) {
-	return filepath.Base(os.Args[0]), os.Args[1:]
 }


internal/command/meta.go
```diff
@@ -354,24 +354,9 @@ func (m *Meta) Colorize() *colorstring.Colorize {
...
 }
 
-// fixupMissingWorkingDir is a compensation for various existing tests which
-// directly construct incomplete "Meta" objects. Specifically, it deals with
-// a test that omits a WorkingDir value by constructing one just-in-time.
-//
-// We shouldn't ever rely on this in any real codepath, because it doesn't
-// take into account the various ways users can override our default
-// directory selection behaviors.
-func (m *Meta) fixupMissingWorkingDir() {
```
lbordowitz, line 364: ❤️

cmd/tofu/main.go
```diff
@@ -229,32 +223,26 @@ func realMain() int {
 	// Initialize the backends.
 	backendInit.Init(services)
 
-	// Get the command line args.
-	binName := filepath.Base(os.Args[0])
-	args := os.Args[1:]
+	binName, args := binNameAndArgs()
```
cam72cam: Why remove the comment and refactor out this one line function?

yottta: I consider it cleaner and wanted to get the handling of `os.Args` in a separate function, as in a future stage we will have a way different way to handle these. 
I can revert if you consider this useless.

internal/command/workdir/dir_test.go
```diff
@@ -0,0 +1,116 @@
...
+				chdirModule := path.Join(tempDir, "root_module")
+				if err := os.Mkdir(chdirModule, 0777); err != nil {
+					t.Fatalf("failed to create %q: %s", chdirModule, err)
+				}
+				return []string{fmt.Sprintf("-chdir=%s", chdirModule), "-anotherflag=test"}
+			},
+			wantNewArgs:          []string{"-anotherflag=test"},
+			wantDataDir:          filepath.Clean("/just/a/random/path/since/it/is/not/checked"),
+			wantWorkingDirSuffix: "root_module",
+		},
```
Yantrio, line 77: This test is actually passing chdir, is this intended or should we rename this test?

internal/command/workdir/dir_test.go
```diff
@@ -0,0 +1,116 @@
...
+	cases := map[string]struct {
+		setup     func(t *testing.T, tempDir string) []string
+		tfDataDir string
+
+		wantNewArgs          []string
+		wantDataDir          string
+		wantMainDir          string
+		wantOriginalDir      string
+		wantWorkingDirSuffix string
+	}{
```
Yantrio, line 35: it might be nice to have a test for ONLY `TF_DATA_DIR` and no `-chdir`

internal/command/workdir/chdir.go
```diff
@@ -0,0 +1,89 @@
...
+			break
+		}
+		if arg == argName || arg == argPrefix {
+			return "", args, fmt.Errorf("must include an equals sign followed by a directory path, like -chdir=example")
+		}
+		if strings.HasPrefix(arg, argPrefix) {
+			argPos = i
+			argValue = arg[len(argPrefix):]
+		}
+	}
```
Yantrio, line 69: Am I correct in thinking that if somehow someone has passed multiple `-chdir` args that this code is intended to only grab the latest one in the list?

internal/command/workdir/chdir.go
```diff
@@ -0,0 +1,89 @@
...
+	// When we fall out here, we'll have populated argValue with a non-empty
+	// string if the -chdir=... option was present and valid, or left it
+	// empty if it wasn't present.
+	if argValue == "" {
+		return "", args, nil
+	}
+
+	// If we did find the option then we'll need to produce a new args that
+	// doesn't include it anymore.
+	if argPos == 0 {
```
Yantrio, line 80: Maybe zero could be a valid value for this dependant on how the args slice is constructed. What do you think about having `-1` as the default value for argPos (or use a *int) so that we can easily determine the not-found case?

internal/command/workdir/chdir.go
yottta: Yes, you are right. When we will do the migration to a proper flags parsing lib, we will get rid of this ancient bit.
But for the moment I intend to keep the existing logic untouched, even if I move it somewhere else. This was in main.go

internal/command/workdir/chdir.go
yottta: I am not sure that I understand the suggestion 🤔 
The "not found" case is the if statement above this one (ie: `if argValue == "" {return "", args, nil}`).
This is an optimisation to avoid allocation new slices but using the underlying one of the `args`

internal/command/workdir/dir_test.go
yottta: You are right. Good catch! Fixed in f52e99558b47b349f813628ed4b43c253f958efa

internal/command/workdir/dir_test.go
yottta: By fixing the wrong setup for a correctly named test, this got fixed as well in f52e99558b47b349f813628ed4b43c253f958efa

_The intent was there, but copy-paste won the battle_ 😅

Yantrio (approved):

cmd/tofu/main.go
yottta: Reverted in 21fa4e1992ccbba043361c2e360862999695ec61.
I consider that is more important to keep the git history than to add a low value change.

PR #3718, opened 2026-02-03
Move new engine integration point into tofu.Context

cam72cam:
This is working toward the "walking skeleton" phase of https://github.com/opentofu/opentofu/issues/3414

This changes the integration of the new engine from a local backend concern into a tofu concern.  In other words, instead of calling into the graph logic from within the tofu.Context, we redirect
 the call to the new engine.

This has a few benefits:
* Less duplication of the external calling logic as we experiment with the new engine
* Ability to run existing tofu.TestContext* tests against the new engine

It also has a few downsides:
* This adds a more substantial integration point between the new and old engine
* Most of the tests are not yet useful
* Some of the tests are not applicable to the new engine

Therefore, it is agreed that if this shim gets in the way of meaningful work on the engine, it will be ripped out and replaced. Probably with a more fleshed out connection in the backend.

In the mean time, if the interface to the new engine changes substantially, Christian will make an effort to maintain a compatibility shim. This will only be attempted if it does not add too much of a burden on the development process and will be abandoned in favor of ripping it out, as mentioned previously.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_temp_runtime.go
+++ b/internal/tofu/context_temp_runtime.go
@@ -74,7 +74,6 @@ func experimentalRuntimeEnabled() bool {
 
 func (c *Context) newEngineShim(ctx context.Context, config *configs.Config, inputValuesRaw InputValues) (*eval.ConfigInstance, plugins.Plugins, func(), tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
-	log.Println("[WARN] Using plan implementation from the experimental language runtime")
 
 	rawInput := map[string]cty.Value{}
 	for key, value := range inputValuesRaw {


internal/tofu/context_temp_runtime.go
```diff
@@ -0,0 +1,247 @@
...
+		return false
+	}
+
+	optIn := os.Getenv("TOFU_X_EXPERIMENTAL_RUNTIME")
+	return optIn != ""
+}
+
+func (c *Context) newEngineShim(ctx context.Context, config *configs.Config, inputValuesRaw InputValues) (*eval.ConfigInstance, plugins.Plugins, func(), tfdiags.Diagnostics) {
+	var diags tfdiags.Diagnostics
+	log.Println("[WARN] Using plan implementation from the experimental language runtime")
```
apparentlymart:
```suggestion
```

(This is just left over from factoring this code out of the `newEnginePlan` function.)

internal/tofu/context_temp_runtime.go
```diff
@@ -0,0 +1,247 @@
...
+	rawInput := map[string]cty.Value{}
+	for key, value := range inputValuesRaw {
+		if !value.Value.IsNull() {
+			rawInput[key] = value.Value
+		}
+	}
+
+	inputValues := exprs.ConstantValuer(cty.ObjectVal(rawInput))
+
+	tempLoader, _ := configload.NewLoader(&configload.Config{})
```
apparentlymart, line 87: I expect we'll need to do something a little more elaborate here once we want to support non-local modules, but this is fine for now.

internal/tofu/context_temp_runtime.go
cam72cam: Removed in 65eb7d7969

internal/tofu/context_temp_runtime.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// The new config-loading system wants to work in terms of module source
+	// addresses rather than raw local filenames, so we'll ask the
+	// addrs package to parse the path we were given. We need to adjust
+	// a little though, because this function was designed for parsing
+	// the "source" argument in a module block, not a plain filepath.
+	// We should add a function in package addrs that's actually intended for
+	// turning arbitrary filesystem paths in to addrs.LocalSource in the long
+	// run, but this will do for now.
+	configDir := config.Module.SourceDir
```
apparentlymart, line 118: @cam72cam, after we talked about something tangentially related today I belatedly remembered that the old version of this code in the local backend had a comment that I think we've lost in this refactor:

```go
	// FIXME: The configuration during apply is supposed to come from the
	// saved plan file rather than the real filesystem, but our codepaths for
	// that are all tangled up with the old-style config loader and so we
	// can't use them here. For now we just load the config directly from the
	// working directory just like the plan phase does, but we should eventually
	// implement this properly.
```

In the old version we had a separate copy of the config-loading code in the apply function which this comment was then attached to. Now that we have all of the functions sharing this same `newEngineShim` method I'm not sure where is best to attach this comment, but I'd like to find _somewhere_ to put it so that we don't forget to fix this as part of a later change to restructure how we're doing config loading here.

apparentlymart:
One thing I noticed while testing today is that the way the planning engine is integrated here means that the old runtime still ends up running the old-style apply graph builder after it's finished executing the new-style planning engine, as part of checking whether the generated plan is actually valid to apply.

It's not a big deal, but it does make the `TF_LOG=trace` output a little confusing because there's a bunch of old-style graph builder logs after the new-style planning engine logs. Maybe it'd be worth skipping that followup graph build altogether when the new engine is enabled just to make it a little easier to read the logs when debugging. :man_shrugging:

PR #3725, opened 2026-02-03
Move all implementation of `flag.Value` in one package

yottta:
Part of #3595

This is the first step needed to move the rest of the commands to the `View` implementation that [was started a long time ago](https://github.com/opentofu/opentofu/commit/1ae3d303834ea8bac484fd868c1dd4ce7fba48a0).
We need to do this, because the way the old `UI` is still used in some of the commands, makes the whole Meta structure destructuring quite a mess.

This first step moves all the implementation of our custom flags to one package, to avoid cyclic dependencies.

This `flags` package could be deleted in the future, once we move to a different style of declaring and parsing flags.
For the moment, this will do.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/flags/raw_flags.go
+++ b/internal/command/flags/raw_flags.go
@@ -14,17 +14,17 @@ import (
 // names and values to a slice. This is used to collect a sequence of flags
 // with possibly different names, preserving the overall order.
 type RawFlags struct {
-	FlagName string
-	Items    *[]RawFlag
+	Name  string
+	Items *[]RawFlag
 }
 
 var _ flag.Value = RawFlags{}
 
-func NewRawFlags(flagName string) RawFlags {
+func NewRawFlags(name string) RawFlags {
 	var items []RawFlag
 	return RawFlags{
-		FlagName: flagName,
-		Items:    &items,
+		Name:  name,
+		Items: &items,
 	}
 }
 
@@ -44,8 +44,8 @@ func (f RawFlags) AllItems() []RawFlag {
 
 func (f RawFlags) Alias(flagName string) RawFlags {
 	return RawFlags{
-		FlagName: flagName,
-		Items:    f.Items,
+		Name:  flagName,
+		Items: f.Items,
 	}
 }
 
@@ -55,7 +55,7 @@ func (f RawFlags) String() string {
 
 func (f RawFlags) Set(str string) error {
 	*f.Items = append(*f.Items, RawFlag{
-		Name:  f.FlagName,
+		Name:  f.Name,
 		Value: str,
 	})
 	return nil


internal/command/flags/flags.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"flag"
+	"fmt"
+)
+
+// FlagNameValueSlice is a flag.Value implementation that appends raw flag
+// names and values to a slice. This is used to collect a sequence of flags
+// with possibly different names, preserving the overall order.
+//
+// FIXME: this is a copy of rawFlags from command/meta_config.go, with the
```
cam72cam: Is this comment out of date now?

yottta: Thanks a lot! Fixed in 0979f4d86c5a2618c186f3d39ac1f79b9ee0f46d

yottta: Also, in https://github.com/opentofu/opentofu/pull/3725/commits/4d453b905ae1111b06e7f7e73476c326a1120d1f, removed one of the flags implementation since FlagNameValue/FlagNameValueSlice was exactly the same with RawFlag/RawFlags.
If you think that we should keep the FlagNameValue based naming, please let me know and I can keep that instead.

internal/command/flags/flags.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// FlagIsSet returns whether a flag is explicitly set in a set of flags
+func FlagIsSet(flags *flag.FlagSet, name string) bool {
+	isSet := false
+	flags.Visit(func(f *flag.Flag) {
+		if f.Name == name {
+			isSet = true
+		}
+	})
```
Yantrio, line 19: Nitpick: It may be nice to have some sort of early return from this, maybe having the Visit return a bool about if it should continue or not could be good. Just incase someone puts in a LOT of flags.

internal/command/flags/raw_flags.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"flag"
+	"fmt"
+)
+
+// RawFlags is a flag.Value implementation that appends raw flag
+// names and values to a slice. This is used to collect a sequence of flags
+// with possibly different names, preserving the overall order.
+type RawFlags struct {
+	FlagName string
```
Yantrio: Nitpick: `RawFlag` (the singular type in this file) has `Name` but this has `FlagName`. What do you think about just `Name` here,

internal/command/flags/flags.go
yottta: I agree that would be nice, but once the `flags.Visit` is called, there is no way to signal back that it needs to stop.

internal/command/flags/raw_flags.go
yottta: Fixed in c4b8ae07651228be0252d890c2a427a81a1eaad0

PR #3749, opened 2026-02-10
Refactor `init` command to use View instead of Ui

yottta:
_don't panic, most of the changes are in tests_ 😅

ℹ️ Before starting reviewing, please read the note from #3765 about the things that should be checked carrefully.

Part of #3765 

This is migrating the `tofu init` command from Ui to View.
We want to do this for all the commands in order to get rid of the Ui and its dependencies.

This is a quite large PR so I recommend you to review it commit by commit. Most of the commits have extensive details in the commit message so please check that out for a more clear context.
Let me try and make a quick map:
* Preparation for the actual changes
  * Move the initialisation of the base `cli.Ui` from main.go to the `commands` package: a420546cc254a7e70952db54e517ef8a5499ce44
  * Add the flags and validations into `arguments/init.go`. This have the same logic as it was having in `command/init.go`: 331c4e5d8651aa2bfa28fa76771b45cd9a375518
  * In order to preserve 1 to 1 output from Views compared with the Ui, added an extension on the View to be able to configure how the diagnostics are printed: 71f8654979e97baaab87223fbea64d8af01486a0
  * Purple was not included in the View colors, but only in the Ui. This fixes it: 613856e75ebc3eaaafae327f24bcd870cea88304
* The init command View implementation and a View implementation for module installation hooks: e36cf8c449453b0695cd2c7a2d58fa6f251c0773
* Migrate `command/init.go` to the new `arguments/init.go` and `views/init.go`: 02a6660b073cd3e7929538637044806290cf11c5
  * Read carefully the commit message. There is a lot of context
* Update an e2e test since now, the json output is not just converted from human form: bce48e106f25aa8f4c9bd331e435b2db3521a2e5
* Migrate all init command tests to the view functionality: 4e4f40d2ba7c5780c8090de18e726b8f83af6157
  * From the changes listed in the commit message, the first 3 cover most of the line changes. I suggest to look mainly at the bottom 3 points. Those are more interesting (and maybe debatable)
* Migrate tests for other commands, tests that are dependant on the init command: dfbea985b67fa420a521c1b8c95e46e128723c78
* And some fixes after checks have been executed, like [copyright headers](https://github.com/opentofu/opentofu/pull/3749/changes/f7eb1c93ba3853e4edc726c646c7f391601c5331) and [a mismatched output stream for one of the messages](https://github.com/opentofu/opentofu/pull/3749/changes/ba8b40ca168cc9702989b1fd2345532692b7aae0)

## Checklist

[...]

PR #3752, opened 2026-02-10
Refactor `get` command to use View instead of Ui

yottta:

Part of #3765  

This PR migrates the `tofu get` command to the View abstraction.

* WrappedUi has been removed since this was the last command using it. The Meta logic around this WrappedUi was removed too.
* The `get` command implementation was not using tracing, so [I added that too](https://github.com/opentofu/opentofu/pull/3752/changes#diff-9e7bbc3a2b8ee7dca3d4d1f6eaa97866ee911c56a67c2e4fb190049ce1e30295R26-R28).
*  `internal/command/get.go#getModules` uses the hook implementation [added in #3749](https://github.com/opentofu/opentofu/pull/3749/changes/e36cf8c449453b0695cd2c7a2d58fa6f251c0773#diff-6814d1d1b6dcbbbca52a2b1ca4ecabecc3e571a1f8feb5ca8d25ee7f94ab3bbaR12).

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/get.go
+++ b/internal/command/get.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"strings"
 
+	"github.com/mitchellh/cli"
 	"github.com/opentofu/opentofu/internal/command/arguments"
 	"github.com/opentofu/opentofu/internal/command/flags"
 	"github.com/opentofu/opentofu/internal/command/views"
@@ -54,8 +55,7 @@ func (c *GetCommand) Run(rawArgs []string) int {
 
 	if diags.HasErrors() {
 		view.Diagnostics(diags)
-		view.HelpPrompt()
-		return 1
+		return cli.RunResultHelp
 	}
 	c.GatherVariables(args.Vars)
 
--- a/internal/command/get_test.go
+++ b/internal/command/get_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/mitchellh/cli"
 	"github.com/opentofu/opentofu/internal/getmodules"
 )
 
@@ -58,7 +59,7 @@ func TestGet_multipleArgs(t *testing.T) {
 	}
 	code := c.Run(args)
 	getOutput := getDone(t)
-	if code != 1 {
+	if code != cli.RunResultHelp {
 		t.Fatalf("bad: \n%s", getOutput.Stdout())
 	}
 }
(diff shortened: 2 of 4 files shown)


internal/command/get.go
```diff
@@ -21,76 +22,77 @@ type GetCommand struct {
...
 
-	path, err := modulePath(cmdFlags.Args())
-	if err != nil {
-		c.Ui.Error(err.Error())
-		return 1
-	}
-
-	path = c.Meta.WorkingDir.NormalizePath(path)
+	// This gets the current directory as full path.
+	path := c.WorkingDir.NormalizePath(c.WorkingDir.RootModuleDir())
```
lbordowitz, line 67: This works because `WorkingDir` in `Meta` is set as `os.GetWd` during `initCommands`

yottta: Exactly! Since we create the working directory in [`main.go`](https://github.com/opentofu/opentofu/blob/a59b4843a097b9b060e2649b536b4a7caac28ff1/cmd/tofu/main.go#L218), together with the logic that applies `-chdir`, I see no reason on keep calling `os.Getwd()` and handle the errors, which is done by `modulePath`.
Though, it's necessary to say that `modulePath` also validates that `args` is empty, validation that has been moved now to the [`arguments.ParseGet()`](https://github.com/opentofu/opentofu/pull/3752/changes#diff-4d67b77359faac323ea04c991de133c62be2a088c0ddca802dd9778332821e11R49).

internal/command/get.go
```diff
@@ -21,76 +22,77 @@ type GetCommand struct {
...
+	view := views.NewGet(args.ViewOptions, c.View)
+	// ... and initialise the Meta.Ui to wrap Meta.View into a new implementation
+	// that is able to print by using View abstraction and use the Meta.Ui
+	// to ask for the user input.
+	c.Meta.configureUiFromView(args.ViewOptions)
+
+	if diags.HasErrors() {
+		view.Diagnostics(diags)
+		view.HelpPrompt()
 		return 1
```
cam72cam: Could this return `cli.RunResultHelp` and skip adding HelpPrompt?

yottta: As such? 5545a379545e038b3dfd75ce35125a50d3d73c00
I like it more TBH

cam72cam: Yep!

PR #3757, opened 2026-02-11
Refactor `login` command to use View instead of Ui

yottta:

Part of #3765 

This PR migrates the `tofu login` command to the View abstraction.

* The `login` command implementation was not using tracing, so [I added that too](https://github.com/opentofu/opentofu/blob/finish-view-refactor-login/internal/command/login.go#L57-L59).
* A weird twist of this PR is that the command does not use `-var`/`-var-file` but the flags are defined. Therefore, I added the flags, but there is no processing of those (ie: `GatherVariables`)

## Checklist

### Go checklist

 


yottta:
Tested this manually too and for the cases that I was able to test, everything works just fine.

Issue #3613, opened 2025-12-18
Upgrade to Go 1.26


apparentlymart:
By the time we publish the v1.12.0 release, Go 1.26 should be the most recent stable Go major release. Our usual policy is to begin each OpenTofu minor release series with the newest available release so that we can have the longest possible security support period, so we should expect to complete this Go upgrade before we enter the v1.12.0 prerelease period.

---

Based on [the draft release notes](https://tip.golang.org/doc/go1.26) and some initial testing with go1.26rc1, we have the following potential areas of concern with this upgrade:

- [x] Changes to the test system's vet checks for `fmt.Stringer` implementation: https://github.com/opentofu/opentofu/pull/3612
- [ ] There are some significant changes to the garbage collector and memory allocation behavior in this release.

    This is hopefully a benefit rather than a problem, but we may wish to use our various performance testing facilities to confirm that these changes have not significantly regressed OpenTofu's specific usage patterns.
- [ ] The hybrid [SecP256r1MLKEM768](https://tip.golang.org/pkg/crypto/tls#SecP256r1MLKEM768) and [SecP384r1MLKEM1024](https://tip.golang.org/pkg/crypto/tls#SecP384r1MLKEM1024) post-quantum key exchanges are now enabled by default.

    We've previously needed to disable other post-quantum crypto support for compatibility with remote state implementations (https://github.com/opentofu/opentofu/issues/2642) so we should proactively test against a similar set of endpoints to see if we need to preemptively disable these too.
- [ ] `url.Parse` now rejects URLs where the hostname component includes a colon, unless it's in square brackets indicating an IPv6 address.

    This seems unlikely to cause problems in practice since a request to such a URL would presumably have failed DNS resolution anyway. Systems that use URLs as unique identifiers rather than as endpoints to send requests to could potentially be impacted, but OpenTofu doesn't do that.
- [ ] Go 1.26 is the last release that will run on macOS 12 Monterey. Go 1.27 will require macOS 13 Ventura or later.

    Therefore we should preannounce that OpenTofu v1.12 will be the last release that supports macOS 12 Monterey, also.

We should revisit the release notes as the go1.26 release process continues, because new details may be added there over time.

apparentlymart:
I reviewed all of our _direct_ uses of `url.Parse` and confirmed that they are all doing URL parsing with the goal of sending requests to the parsed URL, rather than just using it as an identifier.

[...]

PR #3760, opened 2026-02-11
go.mod: Upgrade to Go 1.26

apparentlymart:
This is just a routine upgrade, with the intention that the forthcoming OpenTofu v1.12 series will be based on this Go release series and so will be under security support until February 2027 when this Go release will cease to be supported.

As with the v1.11 series, we cannot predict exactly which day of February Go 1.28 will be released on and so we'll be conservative and promise support until the first day of that month, but in practice we're likely to continue adopting relevant Go 1.26 minor releases for additional weeks of February until the Go team stops publishing them.

This Go upgrade does not immediately cause any new user-facing behavior changes that we'd want to mention in our changelog, aside from the announcement that this will be the last release to support macOS 12 Monterey which we added previously in https://github.com/opentofu/opentofu/pull/3618.

Closes https://github.com/opentofu/opentofu/issues/3613

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+The v1.12.x release series is supported until **February 1 2027**.
+
 ## 1.12.0 (Unreleased)
 
 UPGRADE NOTES:
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/opentofu/opentofu
 
-go 1.25.6
+go 1.26
 
 // At the time of adding this configuration, the new Go feature introduced here https://github.com/golang/go/issues/67061,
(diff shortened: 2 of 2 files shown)


apparentlymart:
The OS support change for macOS is already in our changelog from https://github.com/opentofu/opentofu/pull/3618.

The other ones are about platforms that are already unsupported for OpenTofu, so I don't think we need to say anything about those. 

Do you see another one that you suggest that we mention?

yottta:
> The OS support change for macOS is already in our changelog from #3618.
> 
> The other ones are about platforms that are already unsupported for OpenTofu, so I don't think we need to say anything about those.
> 
> Do you see another one that you suggest that we mention?

No, that's perfect! I didn't see it. Thanks

PR #3761, opened 2026-02-12
Refactor `logout` command to use View instead of Ui

yottta:

Part of #3765 

This PR migrates the `tofu logout` command to the View abstraction.

* The `logout` command implementation was not using tracing, so [I added that too](https://github.com/opentofu/opentofu/blob/finish-view-refactor-logout/internal/command/logout.go#L30-L32).

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/logout.go
+++ b/internal/command/logout.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/mitchellh/cli"
 	"github.com/opentofu/opentofu/internal/command/arguments"
 	"github.com/opentofu/opentofu/internal/command/views"
 	"github.com/opentofu/opentofu/internal/tracing"
@@ -56,8 +57,7 @@ func (c *LogoutCommand) Run(rawArgs []string) int {
 	c.Meta.configureUiFromView(args.ViewOptions)
 	if diags.HasErrors() {
 		view.Diagnostics(diags)
-		view.HelpPrompt(c.credentialsFileForHelp())
-		return 1
+		return cli.RunResultHelp
 	}
 
 	// FIXME: the -input flag value is needed to initialize the backend and the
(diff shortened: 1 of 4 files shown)


internal/command/logout_test.go
```diff
@@ -22,72 +21,87 @@ import (
...
-		if err != nil {
-			t.Fatalf("unexpected error storing credentials: %s", err)
-		}
-
-		status := c.Run(tc.args)
-		if status != 0 {
-			t.Fatalf("unexpected error code %d\nstderr:\n%s", status, ui.ErrorWriter.String())
-		}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
```
lbordowitz, line 70: Besides having nicer test output, this lets you set up the `testView`, and hence `logoutDone` on a per-test basis, something the previous mock UI did not need. This is good! It could lead to parallelization, because now there isn't this nasty global-ish thing all the tests depend on.

internal/command/views/logout_test.go
```diff
@@ -0,0 +1,230 @@
...
+			},
+			wantStdout: "Success! OpenTofu has removed the stored API token for app.example.com.\n\n",
+		},
+		"helpPrompt": {
+			viewCall: func(logout Logout) {
+				logout.HelpPrompt("/home/user/.terraform.d/credentials.tfrc.json")
+			},
+			wantJson:   []map[string]any{{}},
+			wantStdout: "",
+			wantStderr: "\nUsage: tofu [global options] logout [hostname]\n\n  Removes locally-stored credentials for specified hostname.\n\n  Note: the API token is only removed from local storage, not destroyed on the\n  remote server, so it will remain valid until manually revoked.\n      /home/user/.terraform.d/credentials.tfrc.json\n\n",
```
lbordowitz: nit: I've seen a lot of strings like this in the message code. I'd prefer multiline strings, using the backtick character.

yottta: Yeah, I didn't follow a strict rule on this, but generally I liked it this way because whenever somebody else will come on these tests in the future, having the new line escape character is more visible in its intent than a backtick block.
Maybe I am just biased here, so please let me know if you disagree with the argument.

lbordowitz: I can see how it can help with the leading and trailing newlines.... gotcha. Like I said, this is only a nitpick, nothing that should block.

internal/command/logout.go
```diff
@@ -23,27 +26,46 @@ type LogoutCommand struct {
...
+	// diagnostics according to the desired view
+	view := views.NewLogout(args.ViewOptions, c.View)
+	// ... and initialise the Meta.Ui to wrap Meta.View into a new implementation
+	// that is able to print by using View abstraction and use the Meta.Ui
+	// to ask for the user input.
+	c.Meta.configureUiFromView(args.ViewOptions)
+	if diags.HasErrors() {
+		view.Diagnostics(diags)
+		view.HelpPrompt(c.credentialsFileForHelp())
 		return 1
```
cam72cam: return cli.RunResultHelp ?

yottta: Like so? f613a6cc2a1939e8681307fa08c64019385e60c8

PR #3764, opened 2026-02-12
Refactor `console` command to use View instead of Ui

yottta:

Part of #3765  

This PR migrates the `tofu console` command to the View abstraction.

* ⚠️ The first two commits (5790e8cd1a7dbe61cf4c64d5461ffd7d7a540a94 && e388013bb8b4eb3bd71413b5f60dbb1127202b1e) could be part of a different PR, but I didn't want to disconnect the changes from the context that required those. If you still think that would be advisable to move those to a different PR, please leave it in the comments.
* There is no need for the `console` command to use traces, so skipped adding that for this command
* I had to move `DefaultStateFilename` from the `command` package to the `arguments` one. Feel free to suggest a better place to put it.
* Because there are some places where the information is printed fully colored (red or yellow) for errors and warnings, I added in the base `View` struct two methods to have that easily accessible in specific commands view implementations. 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/views/view.go
+++ b/internal/command/views/view.go
@@ -269,5 +269,5 @@ func (v *View) colorMessage(message string, color string) string {
 		return message
 	}
 
-	return v.colorize.Color(fmt.Sprintf("%s%s[reset]", color, message))
+	return v.colorize.Color(fmt.Sprintf("%s%s", color, message))
 }
--- a/internal/command/views/view_test.go
+++ b/internal/command/views/view_test.go
@@ -448,7 +448,6 @@ func TestView_errorln(t *testing.T) {
 		validate func(*testing.T, string)
 	}{
 		"simple error message": {
-
 			message: "This is an error",
 			noColor: true,
 			validate: func(t *testing.T, output string) {
@@ -461,7 +460,6 @@ func TestView_errorln(t *testing.T) {
 			},
 		},
 		"error with color disabled": {
-
 			message: "Colored error",
 			noColor: true,
 			validate: func(t *testing.T, output string) {
@@ -471,12 +469,12 @@ func TestView_errorln(t *testing.T) {
 			},
 		},
 		"error with color enabled": {
-
 			message: "Colored error",
 			noColor: false,
 			validate: func(t *testing.T, output string) {
-				if !strings.Contains(output, "Colored error") {
-					t.Errorf("expected error message in output, got %q", output)
+				expected := "\x1b[31mColored error\x1b[0m\n"
+				if diff := cmp.Diff(expected, output); diff != "" {
+					t.Errorf("got unexpected message (-want, +got):\n%s", diff)
 				}
 			},
 		},
@@ -504,7 +502,6 @@ func TestView_warnln(t *testing.T) {
 		validate func(*testing.T, string)
 	}{
 		"simple warning message": {
-
 			message: "This is a warning",
 			noColor: true,
 			validate: func(t *testing.T, output string) {
@@ -517,7 +514,6 @@ func TestView_warnln(t *testing.T) {
 			},
 		},
 		"warning with color disabled": {
-
 			message: "Colored warning",
 			noColor: true,
 			validate: func(t *testing.T, output string) {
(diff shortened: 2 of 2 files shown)


internal/command/views/view.go
```diff
@@ -242,3 +247,27 @@ func (v *View) outputHorizRule() {
...
+}
+
+// colorMessage is a utility method to easily color the whole given message in one color.
+func (v *View) colorMessage(message string, color string) string {
+	if color == "" {
+		return message
+	}
+
+	return v.colorize.Color(fmt.Sprintf("%s%s[reset]", color, message))
+}
```
apparentlymart, line 273: This is an interesting situation because the idea of printing entire sentences or paragraphs in non-default colors is something we'd intentionally been moving away from, and the migration to using views had previously been the opportunity to do that due to moving away from the UI package.

(This is because colored text is more likely to be hard to read than default-colored text in an arbitrary terminal, but it's safer to assume that someone's configured their default foreground color to have good contrast with their default background color. That's why, for example, the diagnostic rendering only colors the parts of the output that are always the same -- the "Error:" or "Warning:" prefix, and the line drawing characters -- but leaves most of the text in the default color.)

But I can also see how it's desirable to preserve the previous behavior while we're refactoring, and then make the behavior changes separately later. So I think this is okay as long as we use these helpers consistently for all situations where we're directly swapping the old UI package and then we can use "Find references" in our editors to find all of the remaining uses to fix them later.

yottta: I didn't know about this past decision, but I agree that the lack of consistency in these coloured messages is annoying. 
In a new tool, this might create confusion for the users, but seems that `tofu` users got used with these inconsistencies 🤷 

And yes, this is my intent right now. I want to keep the behaviour (even if it's inconsistent) as much as possible similar with the one before the migration.
Later, we can create another piece of work to unify the way different types of messages are printed.

internal/command/views/view_test.go
```diff
@@ -0,0 +1,555 @@
...
+
+			stderr := output.Stderr()
+			if !strings.Contains(stderr, fmt.Sprintf("tofu %s -help", cmd)) {
+				t.Errorf("expected help prompt to contain 'tofu apply -help', got %q", stderr)
+			}
+		})
+	}
+}
+
+func TestView_errorln(t *testing.T) {
```
lbordowitz, line 444: Maybe this is a nit, but do either this test or `warnln` contain checks that the expected color is shown?

yottta: d4db21d85dcc0644d81428e19f9e76958f7fdedb
Updated two of the existing tests to check for the escape codes too.
Adding the test, highlighted a duplicated reset escape code.
`colorstring.Colorize` is always initialised with `Reset = true`, meaning that it always adds the reset escape code to the colorised message.

yottta:
> Looks good, with the caveat of the state locking. Is that going to be removed prior to the 1.12 release?

Just to be sure that I got what you are asking for here, is [this](https://github.com/opentofu/opentofu/pull/3764/changes#diff-5c5b5ae504fabf75611915b88827dd51226a04b3c4111fc5c53995af2e1bb308R62-R64) the thing?

If that it is, then presumably, yes, I intend to remove it before v1.12.
But that is dependent on how much work I will be able to do and get merged.
To remove those, there are several big steps that we need to go through, and the biggest one is to extract the whole backend handling from the Meta structure (pretty much everything in the [meta_backend.go](https://github.com/opentofu/opentofu/blob/cfb0e9c410e184e1938071be45e56a763a18ab67/internal/command/meta_backend.go) and [meta_backend_migrate.go](https://github.com/opentofu/opentofu/blob/cfb0e9c410e184e1938071be45e56a763a18ab67/internal/command/meta_backend_migrate.go)) in a separate component, with its own arguments and view.

I have a point for that in #3765 (`Add View implementation for the backend migration flow`) and from there are several other steps that I have to do before extracting the whole backend logic in its own component.

Hopefully, will be able to do all of those before v1.12, but the main goal of the way I decided to do it, is to migrate slowly, in multiple iterations and to avoid breaking or changing the behavior in significant ways.

That's why v1.12 will be delivered with a lot of `TODO meta-refactor` comments all over the place.

--- 

If this was not the thing you asked for, please let me know and I will happily answer to that too. 😄

internal/command/views/view.go
yottta: Took a note and I will act on it after all commands are migrated.

PR #3766, opened 2026-02-12
[OTel]  Add resource type trace attribute to all resource instance spans

Yantrio:
This adds the resource type (ie, "aws_instance.foo" has a type of "aws_instance") to the traces that apply to resources. This helps end users create queries based on the types of the resources.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/node_resource_abstract_instance.go
+++ b/internal/tofu/node_resource_abstract_instance.go
@@ -59,7 +59,7 @@ const traceAttrResourceInstanceAddr = "opentofu.resource_instance.address"
 //
 // The value of this should be populated from the Type field of the
 // [addrs.Resource] embedded in the relevant [addrs.AbsResourceInstance].
-const traceAttrResourceType = "opentofu.resource_instance.type"
+const traceAttrResourceType = "opentofu.resource.type"
 
 // traceAttrPlanRefresh is a standardized trace span attribute name that we use
 // for a boolean attribute describing whether the refresh step is enabled


internal/tofu/node_resource_abstract_instance.go
```diff
@@ -53,6 +53,14 @@ const traceNameApplyResourceInstance = "Apply resource instance changes"
 // a value of type [addrs.AbsResourceInstance].
 const traceAttrResourceInstanceAddr = "opentofu.resource_instance.address"
 
+// traceAttrResourceType is a standardized trace span attribute name that we
+// use for recording the type of the resource that a particular span is
+// concerned with, such as "aws_instance" or "google_compute_disk".
+//
+// The value of this should be populated from the Type field of the
+// [addrs.Resource] embedded in the relevant [addrs.AbsResourceInstance].
+const traceAttrResourceType = "opentofu.resource_instance.type"
```
apparentlymart: I guess you were going for consistency with the prefix of `opentofu.resource_instance.address` above, and I think that is a reasonable thing to aim for.

I would probably have chosen just `opentofu.resource.type` here because "type" is property of the resource rather than of the instances of a resource (all instances of a resource must always have the same type), but I also don't think this matters a great deal so I'm just mentioning it to see what you think, rather than trying to convince you to change it.

(Perhaps one specific argument in favor of changing it is that we might use this same attribute on other trace spans that are talking about something we're doing for resources _as a whole_, such as calculating their "dynamic expansion", and in that case using resource instance _would_ be more clearly "wrong". But for what you used it for in this commit that argument doesn't apply.)

Yantrio: You're right that I was aiming for consistency here, but after sleeping on what you have mentioned, I think you're right that it should be `opentofu.resource.type` and not specifically for instances. I will move this across now :)

PR #3767, opened 2026-02-12
Update AWS SDK dependencies and recommend using the new `aws login` command

### LINKED ISSUE #3688: S3 backend: support credentials created using the new `aws login` command
### Community note

> [!TIP]

### OpenTofu Version

```shell
1.11.3
```

```hcl
terraform {
  backend "s3" {
    bucket = "a-bucket-for-state"
    key    = "my/project.tfstate"
    region = "eu-west-1"
    profile = "default" # Optional, tried to point it directly
    shared_config_files = [ "~/.aws/config" ] # Optional, tried to point it directly
  }
}
```

### Debug Output

https://gist.github.com/ppabis/f6adfbfba6c327e7096b1a138a7860cc

### Expected Behavior

Credentials in `~/.aws/config` should be respected. The config is the following

```
[default]
region = eu-west-1
login_session = arn:aws:iam::123456789012:user/admin
```

### Actual Behavior

Credentials generated by `aws login` are ignored.

### Steps to Reproduce

1. Use `aws login` to authenticate to CLI.
2. Create an S3 backend config.
3. Try `tofu init`

### Additional Context

This seems to also not be resolved in HashiCorp Terraform.

Seems to have been already since AWS SDK v1.40 and the `main` branch has `1.41`, however latest release `1.11.3` still stays on `v1.39.2`: https://github.com/opentofu/opentofu/blob/f53d9e3a6f6e04a4f5f22b7c32c1ddc1272ba547/go.mod#L33

```
# Release (2025-11-19.2)

## General Highlights
* **Dependency Update**: Updated to the latest SDK module versions

## Module Highlights
* `github.com/aws/aws-sdk-go-v2`: v1.40.0
  * **Feature**: Add support for AWS Login credentials (package credentials/logincreds) to the default credential chain.
```

### References

- https://github.com/opentofu/opentofu/commit/ea203931f79aab473e10b0b1f215098b16270428
- https://github.com/hashicorp/terraform-provider-aws/issues/45316
- https://github.com/hashicorp/terraform/issues/37976

apparentlymart:
Although we would've done these upgrades in preparation for our v1.12 release anyway, in this case the upgrade also allows us to close https://github.com/opentofu/opentofu/issues/3688 by making the S3 backend support using credentials issued by the `aws login` command that was recently added to AWS CLI.

In response to that, this also includes an overhaul of the credentials-related parts of the S3 backend documentation which make opinionated recommendations about how to configure credentials to this backend, and also a stronger recommendation _against_ configuring credentials directly inline using the backend config arguments.

I also took this opportunity to introduce links to relevant parts of the AWS CLI documentation, since those docs can go into far more detail than we can and that additional information will probably be useful for anyone who needs to debug why their setup isn't working.


apparentlymart:
Note for anyone who might try to test this:

- The `aws logout` command removes the cached OAuth refresh token issued by `aws login`, but it _does not_ remove the `login_session` setting from the AWS configuration file and so subsequent attempts to use AWS-integrated tools will fail with a recommendation to use `aws login` again, rather than using other lower-priority ambient credentials sources as normal.

    If you want to return to another strategy for authentication after you complete testing, you'll need to manually remove that setting from your AWS CLI configuration file.
- `aws login` is not currently compatible with AWS IAM Identity Center logins, which includes the logins that many members of the OpenTofu maintainer team typically use for testing.

    Currently, trying to use such a login causes the OAuth flow to be broken by a generic "400 Bad Request" error response. However, there is a different CLI-based login strategy documented for AWS IAM Identity Center too, which I linked from the docs here.

I already ran the `s3` backend acceptance tests with these upgraded dependencies and found no regressions. We cannot reasonably test the `aws login` behavior directly in our acceptance tests because it all lives in the upstream libraries and relies on ambient configuration files created through an interactive OAuth approval flow in a web browser, and so I tested it manually as described in https://github.com/opentofu/opentofu/issues/3688#issuecomment-3888235981 and we have a volunteer from that issue willing to test this on their own AWS account once it's been included in a nightly build.

yottta (approved):

website/docs/language/settings/backends/s3.mdx
```diff
@@ -158,10 +158,42 @@ data.terraform_remote_state.network:
...
+
 ### Credentials and Shared Configuration
 
-:::danger Warning
-We recommend using environment variables to supply credentials and other sensitive data. If you use `-backend-config` or hardcode these values directly in your configuration, OpenTofu will include these values in both the `.terraform` subdirectory and in plan files. Refer to [Credentials and Sensitive Data](../../../language/settings/backends/configuration.mdx#credentials-and-sensitive-data) for details.
+:::warning Warning
+Whenever possible we recommend keeping all settings related to authentication in the standard AWS CLI configuration files used by AWS CLI and other AWS-integrated tools, instead of using the `s3` backend arguments directly.
+
+A `backend "s3"` block should ideally contain only the `region`, `bucket`, and `key` arguments, focusing on specifying just where state snapshots will be stored rather and not specifying how to authenticate when making API requests.
```
Yantrio:
```suggestion
A `backend "s3"` block should ideally contain only the `region`, `bucket`, and `key` arguments, focusing on specifying just where state snapshots will be stored rather than specifying how to authenticate when making API requests.
```

Yantrio (approved):
One tiny tiny nitpick. otherwise this looks good to me.

apparentlymart:
The similar behavior in Terraform got some feedback here about a gap in its interaction with another `s3` backend argument: https://github.com/hashicorp/terraform/issues/38517 .

I've not yet checked whether OpenTofu's implementation has a similar quirk, but since we got this new behavior purely by upgrading to a new version of [`hashicorp/aws-sdk-go-base`](https://github.com/hashicorp/aws-sdk-go-base) (without modifying the S3 backend code at all) we may have also inherited the same bug.

Issue #3765, opened 2026-02-12
Migrate all commands to the `arguments` and `views` packages


yottta:
This issue is part of a bigger scope #3595, as mentioned in the [initial RFC](https://github.com/opentofu/opentofu/blob/b108d3fb5350c1de55356749e1cbbc6c5ee3cd50/rfc/20251105-refactor-tofu-cli.md?plain=1#L180-L186).

Below, can be found the list of work items planned at the moment for this issue:

- [x] `tofu init`: #3749
- [ ] `tofu get`: #3752
- [ ] `tofu login`: #3757
- [ ] `tofu logout`: #3761
- [ ] `tofu console`: #3764
- [ ] `tofu workspace`
- [ ] `tofu workspace show`
- [ ] `tofu import`
- [ ] Remove `Meta.extendedFlagSet`, remove all `GatherVariables` methods and consolidate the logic in the `arguments package`
- [ ] `tofu graph`
- [ ] `tofu fmt`
- [ ] `tofu version`
- [ ] `tofu metadata functions`
- [ ] `tofu providers`
- [ ] `tofu providers lock`
- [ ] `tofu providers mirror`
- [ ] `tofu providers schema`
- [ ] `tofu state list`
- [ ] `tofu state mv`
- [ ] `tofu state pull`
- [ ] `tofu state push`
- [ ] `tofu state replace-providers`
- [ ] `tofu state rm`
- [ ] `tofu state show`
- [ ] `tofu taint`
- [ ] `tofu untaint`
- [ ] `tofu force-unlock`
- [ ] `tofu workspace delete`
- [ ] `tofu workspace list`
- [ ] `tofu workspace new`
- [ ] `tofu workspace select`
- [ ] Meta cleanup:
  - Remove `Meta.process` and cleanup anything related to that
  - Remove `Meta.defaultFlagSet` and anything related to that
  - Remove `Meta.ignoreRemoteVersionFlagSet`
  - Remove `Meta.varFlagSet`
  - Remove `Meta.uiHook` and the views implementation of it
- [ ] Add View implementation for the backend migration flow
- [ ] Check where `Meta.showDiagnostics` is still used and add an interface with only the `Diagnostics` methods and use that to pass the command's view
- [ ] Remove everything related to the Ui from the Meta struct

PR #3770, opened 2026-02-16
Refactor all `workspace` commands to use View instead of Ui

yottta:

Part of #3765 

This PR migrates **all** `tofu workspace` commands (`list`, `show`, `new`, `delete`, `select`) to the View abstraction.

* First commit from this PR adds the View implementation used by all workspace commands since a lot of outputs are common among these.
  * The lack of consistency on how errors and information is printed by these `workspace` commands, led to the following decisions:
    * everything that was _mostly_ printing an `error` type directly, was replaced with a diagnostic. I didn't want to create a generic method in the View for printing `error` types since that can be overused, breaking the purpose of the view abstraction.
    * there were some messages that were used in one of the commands as they were, and in others with `strings.TrimSpace`. To not create custom implementation strictly for new line reasoning, some commands might have their output a little bit different, but consistent with the sibling ones.
* Then each commit contains the arguments and the migration of one of the commands
  * There are some commands that are not registering the flags for -json/-json-into since there is no reason to have such functionality for those commands (`tofu workspace` && `tofu show`). If you disagree, please let me know.
  * For those that allow `-json`, when that is provided by the user, we don't return the error code that prints the help text, since that's human formatted and has no reason to be printed for json format.
* In the last 2 commits the tests are migrated too.

## Checklist

### Go checklist

 


internal/command/arguments/workspace_select.go
```diff
@@ -0,0 +1,55 @@
...
+
+type WorkspaceSelect struct {
+	WorkspaceName string
+
+	CreateIfMissing bool
+
+	ViewOptions ViewOptions
+
+	Vars *Vars
+}
```
apparentlymart, line 25: Do the blank lines between these field declarations mean something?

I ask just because if I were adding a new field in later maintenance I'm not sure how I would decide whether to group it with one of the existing fields or to add another empty line separating it. 🤔

internal/command/views/workspace.go
```diff
@@ -0,0 +1,365 @@
...
+func (v *WorkspaceJSON) WorkspaceInvalidName(name string) {
+	v.view.Error(fmt.Sprintf("The workspace name %q is not allowed. The name must contain only URL safe characters, and no path separators", name))
+}
+
+func (v *WorkspaceJSON) ListWorkspaces(workspaces []string, current string) {
+	for _, workspace := range workspaces {
+		isCurrent := current == workspace
+		v.view.log.Info("workspace listing", "workspace", workspace, "is_current", isCurrent)
+	}
+}
```
apparentlymart, line 319: For an external program consuming this data, I expect I would prefer to get all of the workspace names together as a single JSON array, inside a single JSON "log line".

Could we instead just have a single `v.view.log.Info` call with the whole `workspaces` slice as a single argument, instead of a separate call for each element of that slice, and then report the current workspace name as a string instead of a bool? 🤔 

I also wonder how an external program consuming this should distinguish the log line containing the workspace list from all of the other log lines. We don't seem to have any machine-readable identifier to distinguish these ones. [The "Machine-readable UI" docs](https://opentofu.org/docs/internals/machine-readable-ui/) mention that `type` is a property name for that:

> * `type`: defines which kind of message this is and determines how to interpret other keys which may be present

...so maybe all this needs is a `"type": "workspace_list"`. (I've inferred from examples on that doc page that lowercase with underscores between words is the convention for how to name these "types", e.g. `"type": "resource_drift"` appears in one of them.)

(I mention these concerns now just because once we've shipped this we'll be constrained by compatibility promises, so we should assume that whatever we ship here can basically never change again. Since we're only returning human-oriented text for most of the others right now we don't have to worry so much about those, but this one seems like it's intended to be machine-readable already.)

internal/command/arguments/workspace_select.go
yottta: Doesn't. Let me add some comments and group them.
It was just pure ignorance 😞

yottta:
Drafted the PR since I want to go through the changes again since I have found some places that are concerning when it comes to the state locking flags handling. Will provide more information after I check this part again.

internal/command/arguments/workspace_select.go
yottta: Changed in 04053ff0da5d0f424172bb231124649d7628624d

internal/command/workspace_delete.go
```diff
@@ -103,32 +113,40 @@ func (c *WorkspaceDeleteCommand) Run(args []string) int {
...
+			tfdiags.Error,
+			"Failed to fetch the state",
+			fmt.Sprintf("Fetching state failed: %s", err),
+		)})
 		return 1
 	}
 
 	var stateLocker clistate.Locker
-	if stateLock {
-		stateLocker = clistate.NewLocker(c.stateLockTimeout, views.NewStateLocker(arguments.ViewOptions{ViewType: arguments.ViewHuman}, c.View))
```
yottta, line 129: During applying some of the suggestion from the previous reviews, identified that `c.stateLockTimeout` was used here.
That threw me on the path of rechecking this part of the logic and fixed with the actual timeout given by the user.
This was a bug before too. If you look at the beginning of this method, you can see that the `-lock-timeout` flag is mapped to a different variable and not on the `Meta.stateLockTimeout`, meaning that it was an issue before these changes too.

Not having the flags mapped to the `Meta` related attributes, can lead to behavior different from the one requested by the user. This method calls also `Meta.Backend()` which internally might use `Meta.stateLock` and `Meta.stateLockTimeout`, but not setting those fields after flags parsing, ignores the values given by the user.

Therefore, in f9944bc1a562502bd0f65852556a47bc3733672c, I initialised the Meta fields with the user provided values and used the timeout in the explicit locking in this method.

internal/command/workspace_new.go
```diff
@@ -10,83 +10,95 @@ import (
...
 
-func (c *WorkspaceNewCommand) Run(args []string) int {
+func (c *WorkspaceNewCommand) Run(rawArgs []string) int {
 	ctx := c.CommandContext()
-	args = c.Meta.process(args)
-	envCommandShowWarning(c.Ui, c.LegacyName)
-
-	var stateLock bool
-	var stateLockTimeout time.Duration
-	var statePath string
```
yottta, line 38: The same issue as described [above](https://github.com/opentofu/opentofu/pull/3770#discussion_r2845021260) is also in the `workspace new` command.
Fixed too in f9944bc1a562502bd0f65852556a47bc3733672c

internal/command/views/workspace.go
yottta: > Could we instead just have a single v.view.log.Info call with the whole workspaces slice as a single argument, instead of a separate call for each element of that slice, and then report the current workspace name as a string instead of a bool? 🤔

Initially I implemented exactly as you said, but I rebased so there is no history of that.
But seems that we have different takes on what "easy" would mean in the context of processing json logs. The reason of changing that was to be able to do something like this:
```shell
> cat json-out | jq -r 'select(.workspace != null) | "\(.workspace):\(.is_current)"'
default:false
test:true
```
I agree that we can add a "type" field to these entries too.

But as said, before when I had it with one line, it was kinda hard to parse it since there is an array of strings and you need to look into another property of the log (or even worse, another log line) to check for the currently selected workspace.
IMO, this is easier to create scripts on, but obviously, is debatable. 
WDYT?

apparentlymart: I guess I can't really make a tradeoff here because that `jq` example seems pretty contrived.

What would that look like in a _real_ script? and in that real script, does it actually matter that it takes a second run of `jq` to extract the current workspace name?

The sort of user of this I have in mind is someone writing a custom UI for OpenTofu in a general-purpose programming language, and so I expect them to have a fully-general JSON parser and be able to use arrays.

yottta: I was thinking that the user of it might be somebody that wants to handle this in a script, but I might be subjective indeed.
I agree that long term, there might be tools that will handle the output to provide different functionalities, so I agree that your argument is way stronger.
Will change this tomorrow.

yottta: Changed in 4bf0bcc8b8de51e1165aae98691d7e4198616fbe.
I am not that happy with the `@message` content. If you have a good idea, please let me know.

apparentlymart (approved):
This all seems reasonable to me! Thanks for interating on the JSON representation of the workspace list. 

I agree that the `@message` for the workspace list is awkward, but I also think it's acceptable because the message property was included in these JSON logs mainly so that a program that is consuming _only_ the JSON logs could have something to display for items it doesn't understand, but OpenTofu no longer really needs that because we added `-json-into` so that applications can use the JSON only for the specific items they care about while still showing the human output for everything else. 

(Arguably we don't really need that `@message` property anymore at all for that reason, but I don't propose we go to the extreme of removing it right now. We can perhaps use this slightly-useless message on this single low-priority command as a way to test the assumption that the message is unimportant, and then improve it in a later release if we learn we were wrong since the exact text in the message property is not covered by compatibility promises the way that the machine-readable properties are.)

PR #3772, opened 2026-02-17
Refactor `import` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu import` command to the View abstraction and to the `arguments` package.
* The `import` command implementation was not using tracing, so [I added that too](https://github.com/opentofu/opentofu/blob/finish-view-refactor-import/internal/command/import.go#L37-L38).
* Cleaned up [`Meta.uiHook`](https://github.com/opentofu/opentofu/pull/3772/changes/857c90f0fb248788a5af87276c00b7134f80e8ed) since the last "user" of it was the `import` command implementation.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/import_test.go
+++ b/internal/command/arguments/import_test.go
@@ -19,42 +19,43 @@ import (
 func TestParseImport_basicValidation(t *testing.T) {
 	wd := workdir.NewDir(".")
 	testCases := map[string]struct {
-		args []string
-		want *Import
+		args        []string
+		want        *Import
+		wantErrText string
 	}{
 		"defaults": {
-			[]string{"addr", "id"},
-			importArgsWithDefaults(func(imp *Import) {
+			args: []string{"addr", "id"},
+			want: importArgsWithDefaults(func(imp *Import) {
 				imp.ResourceAddress = "addr"
 				imp.ResourceID = "id"
 			}),
 		},
 		"parallelism flag": {
-			[]string{"-parallelism=5", "addr", "id"},
-			importArgsWithDefaults(func(imp *Import) {
+			args: []string{"-parallelism=5", "addr", "id"},
+			want: importArgsWithDefaults(func(imp *Import) {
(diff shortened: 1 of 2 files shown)


internal/command/import.go
```diff
@@ -287,24 +288,65 @@ func (c *ImportCommand) Run(args []string) int {
...
+	view.Success()
+	view.Diagnostics(diags)
 	if diags.HasErrors() {
 		return 1
 	}
 
 	return 0
 }
 
+// configureBackendFlags is a temporary shim until we move the flags for state management to a better placce
```
diofeher:
```suggestion
// configureBackendFlags is a temporary shim until we move the flags for state management to a better place
```

internal/command/arguments/import_test.go
```diff
@@ -0,0 +1,243 @@
...
+			err := diags.Err()
+			if err == nil {
+				t.Fatal("expected error but got nil")
+			}
+
+			if got := err.Error(); !strings.Contains(got, tc.wantError) {
+				t.Errorf("wrong error message\n got: %s\nwant: %s", got, tc.wantError)
+			}
+
+			if len(diags) > 0 {
```
diofeher: Since we already test the size of diags here: https://github.com/opentofu/opentofu/pull/3772/changes#diff-c4e9844d685544e0f24f44f061774abecdaaefb82512adb26b1403514407527bR148

I think we can drop this `if len(diags) > 0`

yottta: Thanks. Indeed was redundant. I reworked a little bit the tests to include these in the above test function: 450ecb4fc67c941f87dae67a4ea0b628e5ea7068

internal/command/import.go
yottta: Fixed in 450ecb4fc67c941f87dae67a4ea0b628e5ea7068

PR #3779, opened 2026-02-18
Store BeforeIdentity and AfterIdentity and display it in the plan json output

Yantrio:
This is part of the resource identity work in #3671 and is a PR ontop of that branch.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/internal/execgraph/operations_test.go
+++ b/internal/engine/internal/execgraph/operations_test.go
@@ -325,6 +325,7 @@ func (m *managedResourceInstanceMockProvider) UpgradeResourceState(context.Conte
 	panic("unimplemented")
 }
 
+
 // UpgradeResourceIdentity implements providers.Configured.
 func (m *managedResourceInstanceMockProvider) UpgradeResourceIdentity(context.Context, providers.UpgradeResourceIdentityRequest) providers.UpgradeResourceIdentityResponse {
 	panic("unimplemented")


internal/plans/internal/planproto/planfile.proto
```diff
@@ -184,10 +184,15 @@ message Change {
     // the change, as an HCL string.
     string generated_config = 6;
 
-    // PlannedIdentity is the identity value returned by the provider during
+    // BeforeIdentity is the identity value from the known state of the resource instance
+    // before the plan is executed.
+    // Only relevant for managed resources, not outputs.
+    DynamicValue before_identity = 7;
```
diofeher, line 190: Is it ok to reuse this field 7 here? Is there any potential issues with older states using this `planned_identify` as 7?

yottta (approved):

internal/plans/internal/planproto/planfile.proto
apparentlymart: That is a good general rule about protobuf schema evolution, but for `planfile.proto` in particular it's not so important because a plan file created by one version of OpenTofu cannot be used by any other version of OpenTofu -- the version number is saved as part of the plan file and we check it before decoding anything else.

We probably do still need to be careful about reusing a field index when the old and new fields have types with different [wire formats](https://protobuf.dev/programming-guides/encoding/) because an incompatible selection there might cause the unmarshal step to fail with a parse error before we get a chance to check the version number, but if `planned_identity` was also a `DynamicValue` then that shouldn't be a problem.

With all of that said, there are plenty of integers so we could just choose a different one anyway to avoid the question! :grinning:

diofeher: Ah, that's a good point. If we've checked this field before and the versions were incompatible before reaching this part, it should be good enough to stay as is. ^^

diofeher (approved):

internal/command/jsonplan/values.go
```diff
@@ -252,8 +252,8 @@ func marshalPlanResources(changeMap map[string]*plans.ResourceInstanceChangeSrc,
 		}
 		resource.SensitiveValues = v
 
-		if changeV.PlannedIdentity != cty.NilVal && !changeV.PlannedIdentity.IsNull() {
-			identityJSON, err := ctyjson.Marshal(changeV.PlannedIdentity, changeV.PlannedIdentity.Type())
+		if changeV.AfterIdentity != cty.NilVal && !changeV.AfterIdentity.IsNull() {
```
yottta, line 255: ~~Why the change here? Probably I am missing some context.~~
Based on the changes in planfile.proto, this is actually the new "plannedIdentity", right?

Yantrio: Correct.

PR #3784, opened 2026-02-19
Fix panic in local-exec and file provisioners when required attributes are null

### LINKED ISSUE #3783: Crash when local-exec provisioner has a null command
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.5
on linux_amd64
+ provider registry.opentofu.org/hashicorp/null v3.2.4
```

```hcl
terraform {
  required_providers {
    null = {
      source = "hashicorp/null"
    }
  }
}

resource "null_resource" "test" {
  provisioner "local-exec" {
    command    = null
  }
}
```

### Debug Output

For now, let me just share the normal error outputs (without trace):
https://gist.github.com/meeque/fee5a1b55734394da66d2c283921ec73

### Expected Behavior

OT should not crash.

Instead, OT should print an error message, saying that a `null` `command` is not allowed. Maybe similar to what happens when passing an empty string as `command`.

Or, OT should simply skip a `local-exec` `provisioner` with a `null` `command`.
(The latter would be my personal preference, because it would allow me running provisioners conditionally.)

### Actual Behavior

Looks like OT crashes. 

It prints a crash banner and a stack-trace, then exits with status `11`.

### Steps to Reproduce

1. In an empty directory, create a file (e.g. `main.tf`) with the above OT configuration.
2. Run `tofu init`.
3. Run `tofu apply -auto-approve`.

The above configuration is self-contained and it crashes reliably.
Some other variations also crash, e.g. omitting the `command` altogether and using an empty `provisioner` block.

### Additional Context

Nothing special about my environment, I think. It's an Ubuntu 25.10 box.

### References

_No response_

Yantrio:
Previously, passing null to a required provisioner attribute (e.g. command = null in local-exec, destination = null in file) caused a crash due to calling AsString() on a null cty.Value.

This commit attempts to fix that by adding validation in both the validate and apply phases.

Resolves #3783

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ BUG FIXES:
 - The `azurerm` backend's MSI authentication method will now respect the provided client ID ([#3586](https://github.com/opentofu/opentofu/issues/3586))
 - Add `universe_domain` option in the `gcs` backend to support sovereign GCP services ([#3758](https://github.com/opentofu/opentofu/issues/3758))
 - Fixed `tofu init` crashing when a module `version` uses a variable and the module is referenced from a test file. ([#3686](https://github.com/opentofu/opentofu/issues/3686))
+- Fixed `local-exec` and `file` provisioners crashing when a required attribute (`command` or `destination`) is set to `null`. ([#3783](https://github.com/opentofu/opentofu/issues/3783))
 
 ## Previous Releases
 


apparentlymart (approved):
TLDR: this looks good to me!

---

The way provisioner arguments and their associated `connection` block arguments are handled today is more-or-less just a direct translation of how they worked in Terraform v0.11 and earlier to use the `cty` API instead of using `map[string]any`, and such.

For example, here's the much older code equivalent to the code you modified at the start of the local provisioner's `ProvisionResource` method:

https://github.com/opentofu/opentofu/blob/57db3cfe5164404c0e353fda7596364911868417/builtin/provisioners/local-exec/resource_provisioner.go#L47-L51

The original code here was relying on the legacy Terraform plugin SDK's tendency to quietly swap out `nil` for the zero value of the schema type[^1], making this `.(string)` assertion always succeed but sometimes return `""`. 

[^1]: By total coincidence, that behavior is fresh in my mind because [it also seems to be the root cause of a bug in the `hashicorp/aws` provider](https://github.com/hashicorp/terraform-provider-aws/issues/38755#issuecomment-3924220090) that I was poking at after hours yesterday. 😀

All of that is to say: this approach to fixing it makes sense to me because it's effectively making the new-style code better mimic how the much older version of this code would've behaved under the Terraform plugin SDK. Whereas the plugin SDK did its best to avoid ever returning `nil`, `cty` expects the caller to check for and handle its null values and unknown values explicitly, and that's what all of your newly-added checks here are achieving.

PR #3785, opened 2026-02-20
planning: Split the new experimental planning phase into two passes

apparentlymart:

Earlier work on the planning engine tried to do both the discovery of the resource instance graph and the construction of the execution graph in a single pass. We reached the limits of that in my previous PR https://github.com/opentofu/opentofu/pull/3747, where it became clear that we don't have enough information in a single pass to calculate correct dependencies between "delete" actions in the execution graph, since those need to be the inverse of the dependencies discovered by the evaluator from the configuration.

Therefore the planning engine now works in two passes:

1. Use the evaluator to do localized analysis and planning of each resource instance object in isolation, discovering the resource instance graph along the way and now materializing it into an intermediate representation that we send to the second pass.

    For the sake of this PR the intermediate representation only describes the relationships between resource instance objects themselves, but I'm expecting to extend that to also describe the dependencies between provider instance configurations and the upstream resource instances they rely on in a future PR, to fill in the currently-missing execution graph dependencies for those. (**Edit:** Done in https://github.com/opentofu/opentofu/pull/3808 .)
3. Once we have the resource instance graph fully constructed, we can lower it into an execution graph that describes the individual changes we need to make during the apply phase and what order they need to happen in.

    This pass first decides what order of operations each resource instance object must use if it's going to be replaced, then inserts a subgraph of operations for each object, and then finally fills in the dependency edges between them so that their operations will all happen in a correct order relative to each other.

    (I suppose we could think of this pass as actually being three separate passes in their own right, but I'm not thinking of it that way right now because when I say "pass" I intend it to mean a self-contained process that produces a well-defined, self-contained artifact for the next pass to consume, whereas the three parts of this second pass are more tightly coupled together, sharing some mutable data structures.)

This is quite a significant change to the overall internal architecture of the planning engine, and so this PR got pretty big. I've intentionally stopped as soon as I reached a point where it was complete enough to support all of the same situations it previously supported, but that means I've left the details of the execution graph builder in a very messy and hard-to-read state because I didn't want to disturb the implementation of those methods _too_ much as part of this initial reworking.

This PR therefore is therefore mainly just to demonstrate that splitting the planning engine into two passes is sufficient to construct a correct execution graph when "delete" actions are involved. After this is merged I intend to work on tidying up some of the details in subsequent smaller PRs.

Unfortunately the earlier stubbing of handling ephemeral resource instances has not survived this first pass of refactoring. That was a risk we accepted in working on that stuff before we completed the walking skeleton milestone. We're simultaneously thinking about different approaches for how to handle that, but [our project plan](https://github.com/opentofu/opentofu/issues/3414) was to complete the walking skeleton phase _before_ working on the details of ephemeral resources.

As is often the case with these exploratory PRs, it might be easier to review it [on a commit-by-commit basis](https://github.com/opentofu/opentofu/pull/3785/commits).

### DIFF
--- a/internal/engine/planning/execgraph_managed_test.go
+++ b/internal/engine/planning/execgraph_managed_test.go
@@ -47,14 +47,14 @@ func TestExecGraphBuilder_ManagedResourceInstanceSubgraph(t *testing.T) {
 			providerClientRef execgraph.ResultRef[*exec.ProviderClient],
 		) (
 			execgraph.ResourceInstanceResultRef,
-			execgraph.AnyResultRef,
+			execgraph.ResourceInstanceResultRef,
 			func(execgraph.AnyResultRef),
 			func(execgraph.AnyResultRef),
 		)
 		WantRepr string
 	}{
 		"create": {
-			func(b *execGraphBuilder, providerClientRef execgraph.ResultRef[*exec.ProviderClient]) (execgraph.ResourceInstanceResultRef, execgraph.AnyResultRef, func(execgraph.AnyResultRef), func(execgraph.AnyResultRef)) {
+			func(b *execGraphBuilder, providerClientRef execgraph.ResultRef[*exec.ProviderClient]) (execgraph.ResourceInstanceResultRef, execgraph.ResourceInstanceResultRef, func(execgraph.AnyResultRef), func(execgraph.AnyResultRef)) {
 				return b.ManagedResourceInstanceSubgraph(
 					&plans.ResourceInstanceChange{
 						Addr:        instAddr,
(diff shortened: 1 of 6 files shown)


internal/engine/planning/execgraph_managed_test.go
```diff
@@ -189,14 +189,14 @@ func TestExecGraphBuilder_ManagedResourceInstanceSubgraph(t *testing.T) {
 				r[1] = ResourceInstanceDesired(test.placeholder, await());
 				r[2] = ManagedFinalPlan(r[1], nil, v[0], nil);
 				r[3] = ManagedFinalPlan(nil, r[0], v[1], nil);
-				r[4] = ManagedApply(r[3], nil, nil, await(r[2]));
+				r[4] = ManagedApply(r[3], nil, nil, await());
```
apparentlymart, line 192: This extra `await` dependency was actually an incorrect one that I didn't notice previously because we weren't yet inserting the dependencies of the delete step when building a graph containing changes more than one resource instance object.

It was incorrect because it caused a cyclic dependency between the create and delete steps as soon as two `DeleteThenCreate` changes got chained together in the same graph. In that case the correct order is "delete B, delete A, create A, create B", whereas this edge specified that the planning part of "create B" also needed to happen before "delete B", and that's impossible.

Issue #3773, opened 2026-02-17
A policy for when to use `go fix`, and when to perform other similar broad fixups


apparentlymart:
[Go v1.26 has expanded the scope of `go fix`](https://go.dev/blog/gofix) to include a number of "modernizers".

There is a tension with using tools like this in a codebase like OpenTofu:

- We are quite an old codebase that has "lived through" a number of changes in Go idiom. There's lots of older code that could be improved by this sort of automatic fixup, but...
- ...we also typically maintain at least three active release branches at a time and want to be able to easily backport bug fixes between them when warranted, and it's very annoying and disruptive when backports fail due to conflicts with purely cosmetic code changes.

I'd like to propose that we adopt the following broad compromise:

- We will _not_ just indiscriminately run modernizer tools like `go fix`, or other similar bulk updates, on code where we're not imminently changing the behavior of the code. Most of the time, code that has the same behavior in different versions should have identical source code between versions.

    Relatedly, we will not typically merge community PRs that consist _only_ of cosmetic modernizer updates, unless it comes in conjunction with another change as described in the following item.
- However, each time we complete some work that significantly changes existing code, as follow-up work _in the same release series_ we'll apply `go fix` to the packages that were affected by that work and submit the subset of changes that are in the files that were already significantly changed as a separate PR.

    The assumption here is that if we've already significantly changed something then we're not going to be trivially backporting nearby changes to earlier branches _anyway_, and so that's a good time to bundle in some modernizer changes. But doing it in a separate followup PR, rather than as part of the PR that changed the behavior, will keep the main PR relatively clean. The followup PR should then contain _only_ cosmetic changes, where any change in behavior is unexpected.

- The above are broad guidelines rather than hard-and-fast rules. Maintainers may sometimes propose making different tradeoffs in specific cases, while keeping the same general concerns in mind.

    In particular, this proposal focuses mainly on `go fix` in particular as a good baseline set of modernizers for routine use, but in certain cases we might choose to run other third-party tools where there are more unusual problems with existing code that we want to correct, but would probably still try to make such changes only in the same release series as other significant nearby code changes.

Accepting this only represents changing our policy and the maintainers agreeing to use the modernizers when possible in future as described above. It would not cause or require any immediate code changes.

[...]

PR #3788, opened 2026-02-20
command: `go fix` on various files we've changed recently anyway

apparentlymart:
This is a first round of changes under the policy we adopted in https://github.com/opentofu/opentofu/issues/3773. 

I've started with the `command` package here because the work in https://github.com/opentofu/opentofu/issues/3765 has already made a lot of changes in this package, and so the likelihood of successful automatic backporting to an earlier release branch is low already.

My rules for what to include or not are admittedly quite "vibes-based", but the general idea was:

 - If the existing diff in a file is already significantly larger than the changes the fixer proposed to make, or if the fixer is proposing to change a line that was already changed in this development period.
 - More willing to include "_test.go" files than non-test files, even if they hadn't changed as much already, just because backports from test files for bug fixes tend to be entirely new test cases more than they are modifications to existing test cases, and so the risk of conflicts is lower there.

Everything here was changed by `go fix`, except that I manually removed the `ptrTo` function from `oci_credentials_test.go` because all of the calls to it got replaced with the `new` builtin and so the linter was complaining about it now being dead code. :grinning: 

I expect to get a stronger intuition of what changes to include or not as we practice more with applying this policy, but I think this is a relatively safe set of fixes to start with.

### DIFF
--- a/internal/command/show_test.go
+++ b/internal/command/show_test.go
@@ -957,10 +957,10 @@ func TestShow_json_output_state(t *testing.T) {
 
 			// compare ui output to wanted output
 			type state struct {
-				FormatVersion    string                 `json:"format_version,omitempty"`
-				TerraformVersion string                 `json:"terraform_version"`
-				Values           map[string]interface{} `json:"values,omitempty"`
-				SensitiveValues  map[string]bool        `json:"sensitive_values,omitempty"`
+				FormatVersion    string          `json:"format_version,omitempty"`
+				TerraformVersion string          `json:"terraform_version"`
+				Values           map[string]any  `json:"values,omitempty"`
+				SensitiveValues  map[string]bool `json:"sensitive_values,omitempty"`
 			}
 			var got, want state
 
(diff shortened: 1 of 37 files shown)


internal/command/show_test.go
```diff
@@ -1343,21 +1343,21 @@ func showFixturePlanFile(t *testing.T, action plans.Action) string {
 // to avoid needing to constantly update the expected output; as a potential
 // TODO we could write a jsonplan compare function.
 type plan struct {
-	FormatVersion   string                 `json:"format_version,omitempty"`
-	Variables       map[string]interface{} `json:"variables,omitempty"`
-	PlannedValues   map[string]interface{} `json:"planned_values,omitempty"`
-	ResourceDrift   []interface{}          `json:"resource_drift,omitempty"`
-	ResourceChanges []interface{}          `json:"resource_changes,omitempty"`
-	OutputChanges   map[string]interface{} `json:"output_changes,omitempty"`
-	PriorState      priorState             `json:"prior_state,omitempty"`
```
lbordowitz, line 1352: Will removing `omitempty` change anything here?

apparentlymart: `omitempty` isn't meaningful for a field of a struct type, because it's unclear what "empty" would mean in this case. So I believe `go fix` removed this just to be explicit that it does nothing.

apparentlymart: This change was made by [the `omitzero` modernizer](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_omitzero), which finds useless `omitempty` like this example and can either remove it (as in this case) or change it to `omitzero`, which is a more modern alternative that skips serializing a struct field if it contains the zero value of the type.

The `go fix` tool prefers the option of removing it because that does not cause a behavior change. Swapping for `omitzero` could potentially cause a behavior change, if there's any situation where this field would contain the zero value of `priorState`. For the sake of this PR I think the no-behavior-change option is the better choice.

PR #3796, opened 2026-02-24
Documentation of a bug report process

apparentlymart:
The initial motivation for this document was to introduce the "fast path" as a way to hopefully streamline our handling of simple, uncontroversial bug reports so that we can reserve our synchronous discussion time for more complicated or ambiguous situations.

However, I also took the opportunity to describe the bug process more broadly, including a description of how the maintainers currently interact with bug reports outside of the proposed new "fast path", in case that's helpful to reporters who want to know what to expect or to understand where they are in the process.

[Rendered Version](https://github.com/opentofu/opentofu/blob/doc-bug-report-process/BUG_REPORTS.md)

### DIFF
--- a/BUG_REPORTS.md
+++ b/BUG_REPORTS.md
@@ -1,0 +1,145 @@
+# OpenTofu Bug Report Process
+
+This document describes this project's typical process for posting, triaging,
+and responding to bug reports.
+
+The main purpose of this documentation is to help contributors know what they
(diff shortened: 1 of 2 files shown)


apparentlymart:
FWIW, https://github.com/opentofu/opentofu/issues/3803 was our first real example of using the "fast path". :grinning:

PR #3802, opened 2026-02-25
Refactor `graph` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu graph` command to the View abstraction and to the `arguments` package.
* ⚠️ I skipped adding the JSON view to this since it makes no sense having json output for this command. If you disagree, please let me know and I can add it.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/graph.go
+++ b/internal/command/arguments/graph.go
@@ -60,7 +60,6 @@ func ParseGraph(args []string) (*Graph, func(), tfdiags.Diagnostics) {
 		return arguments, closer, diags
 	}
 
-	// The graph command accepts an optional path argument
 	if len(cmdFlags.Args()) > 0 {
 		diags = diags.Append(tfdiags.Sourceless(
 			tfdiags.Error,


internal/command/views/graph.go
```diff
@@ -0,0 +1,39 @@
...
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package views
+
+import (
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+type Graph interface {
```
yottta, line 12: Looking at the number of methods in this, I wanted to extract these into a common view implementation and have that embedded in the structs, but I held myself back from doing so because I want to finish all the commands and in the future, if we really want, we can tidy up based on the patterns that will be visible when each command will have its own view.

internal/command/arguments/graph.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// we only parse but do not register the views flags since this command does not need it
+	closer, moreDiags := arguments.ViewOptions.Parse()
+	diags = diags.Append(moreDiags)
+	if diags.HasErrors() {
+		return arguments, closer, diags
+	}
+
+	// The graph command accepts an optional path argument
```
cam72cam: I don't understand what this comment is trying to convey?

yottta: Good point. Most probably was a leftover that I forgot about after I played around here 🤷 .
I am confused by that too 😅 

Removed in 661d194aaa72e195596dcd671b5ba708237fb5fd

PR #1105, opened 2024-01-11
Allow known references in import blocks

cam72cam:
Imports are attached to their respective modules, but must execute expressions in the root module context.  This PR fixes that and makes sure that refererences to the root module are properly described via the graph.

It is recommended to review this with hide-whitespace (?w=true) turned on as it makes the diff easier to read.

Resolves #1084 

## Target Release

1.7.0, 1.6.1

PR #3804, opened 2026-02-25
Ensure that Provider Defined Functions in import blocks do not error

### LINKED ISSUE #3803: `Uninitialized function provider` when importing azurerm resources using provider function
### Community note

> [!TIP]

### OpenTofu Version

```shell
1.11.5
```

```hcl
...tofu config...

import {
  id = provider::azurerm::normalise_resource_id("/subscriptions/${var.subscription_id}/resourceGroups/${var.resource_group_name}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/${local.identity_aks_name}")
  to = azurerm_user_assigned_identity.aks
}

resource "azurerm_user_assigned_identity" "aks" {
  name                = local.identity_aks_name
  resource_group_name = var.resource_group_name
  location            = local.primary_region.location

  lifecycle {
    prevent_destroy = true
  }
}
```

### Debug Output

[N/A](https://gist.github.com/theonlyames/393b9f8e560f314a4de8a0900957c95b)

### Expected Behavior

Expectation is for the resource to be imported

### Actual Behavior

Results in an unexpected error. 
"Error BUG: Uninitialized function provider"

Image

### Steps to Reproduce

`tofu init -reconfigure`
`tofu plan -var-file .\vars.tfvars`

### Additional Context

Confirmed working with Terraform 1.13.x, if that matters.

### References

_No response_

Yantrio:
The `ProviderFunctionTransformer` only inspected `References()` when tracking which providers need to be initialized for function calls. 

Import block expressions are returned via `RootReferences()` (introduced in #1105) since import blocks are always evaluated in the root module context, so provider functions in import IDs were never tracked.

This meant that the provider was never set up correctly for provider defined functions if they were only referenced in a `import` block.

I broke this into 2 commits.

1. **Extract method to track provider function references**: Pure refactor, extracts the inner loop body of hooking up the graph into a `trackProviderFunction` helper.
2. **Also look at `RootReferences()` when collating references**: Adds a check to see if the referencer is a `GraphNodeRootReferencer` and iterates across those references.

Resolves #3803

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/transform_provider.go
+++ b/internal/tofu/transform_provider.go
@@ -11,6 +11,7 @@ import (
 	"log"
 
 	"github.com/hashicorp/hcl/v2"
+
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/configs"
 	"github.com/opentofu/opentofu/internal/dag"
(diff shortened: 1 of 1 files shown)


Yantrio:
I'm moving this to draft as I'm about to finish for the day and I need to add a unit test for this to ensure we dont regress, and I need to add a changelog entry for this too. If someone else wants to do this then please feel free to assign yourself and pick it up, otherwise I'll tackle this in my work hours tomorrow.

cam72cam:
@Yantrio, why not consider a simpler diff:
```diff
diff --git a/internal/tofu/transform_provider.go b/internal/tofu/transform_provider.go
index f4f77d0d0b..3ba903e6ae 100644
--- a/internal/tofu/transform_provider.go
+++ b/internal/tofu/transform_provider.go
@@ -377,7 +377,13 @@ func (t *ProviderFunctionTransformer) Transform(_ context.Context, g *Graph) err
        for _, v := range g.Vertices() {
                // Provider function references
                if nr, ok := v.(GraphNodeReferencer); ok && t.Config != nil {
-                       for _, ref := range nr.References() {
+
+                       refs := nr.References()
+                       if nr, ok := v.(GraphNodeRootReferencer); ok {
+                               refs = append(refs, nr.RootReferences()...)
+                       }
+
+                       for _, ref := range refs {
                                if pf, ok := ref.Subject.(addrs.ProviderFunction); ok {
                                        refPath := nr.ModulePath()
 
```

That works as GraphNodeRootReferencer is guaranteed to be a GraphNodeReferencer.

Yantrio:
I've taken that feedback onboard, thanks @cam72cam . Please see 3f57d50 or https://github.com/opentofu/opentofu/pull/3804/changes#diff-2b903eb2dfe957cbd9f12f13c5d55dfc8ef718d021e97d3865e4cdf945a24f4d.

For background, this was not as simple as you mentioned because we have to handle that RootReferencer references should always use the root module. This small reason is why i introduced the internal `NodeReference` struct.

apparentlymart:
This is the first time I've really noticed the `GraphNodeRootReferencer` interface, so I'm inferring from the discussion here that it is like `GraphNodeReferencer` except that all of the references returned are treated as belonging to the root module rather than to the module returned by the `ModulePath` method of `GraphNodeModulePath`.

It seem like that's really just a special case of `GraphNodeReferenceOutside` then, right? If the relevant graph nodes implemented that interface and returned `addrs.RootModule` for the `referencePath` then the results from `GraphNodeReferencer.References` would get interpreted in the root module.

I don't think we should try to significantly rework this as part of fixing the bug -- the smallest possible change is better here I think, and the current proposal seems like a good fit -- but I am curious if we could simplify this further by relying on `GraphNodeReferenceOutside` completely. Happy to defer discussing this until another day, though!

apparentlymart (approved):

yottta (approved):

Yantrio:
I think that you're right, but the difference between the two implementations is that the scope of the references is different. I agree its not worth re-working here but I think as things go on I would like to try and find a way to help converge these two (Or ditch them with the new engine execution!)

PR #3805, opened 2026-02-26
Refactor `fmt` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu fmt` command to the View abstraction and to the `arguments` package.
* ⚠️ I skipped adding the JSON view to this since it makes no sense having json output for this command. If you disagree, please let me know and I can add it.

## Checklist

### Go checklist

 


internal/command/views/fmt.go
```diff
@@ -0,0 +1,42 @@
...
+func (v *FmtHuman) Diagnostics(diags tfdiags.Diagnostics) {
+	v.view.Diagnostics(diags)
+}
+
+// UserOutputWriter returns a [io.Writer] that uses the [FmtHuman.view] as a proxy to write
+// the user facing information during formatting.
+func (v *FmtHuman) UserOutputWriter() io.Writer {
+	return &writer{
+		writeFn: func(msg string) {
+			_, _ = v.view.streams.Println(msg)
```
cam72cam: Why not use v.view.streams.Stdout.File as the io.Writer instead?

yottta: Because we want to call the formatting logic inside `Println`:
```go
func (s *Streams) Println(a ...interface{}) (n int, err error) {
	return fmt.Fprintln(s.Stdout.File, a...)
}
```

cam72cam: Why?  I patched the code and it worked just as expected.

```diff
diff --git a/internal/command/views/fmt.go b/internal/command/views/fmt.go
index f833e3005d..d10dc288ae 100644
--- a/internal/command/views/fmt.go
+++ b/internal/command/views/fmt.go
@@ -34,9 +34,5 @@ func (v *FmtHuman) Diagnostics(diags tfdiags.Diagnostics) {
 // UserOutputWriter returns a [io.Writer] that uses the [FmtHuman.view] as a proxy to write
 // the user facing information during formatting.
 func (v *FmtHuman) UserOutputWriter() io.Writer {
-       return &writer{
-               writeFn: func(msg string) {
-                       _, _ = v.view.streams.Println(msg)
-               },
-       }
+       return v.view.streams.Stdout.File
 }
```

I don't see a reason not to use the stream directly.

yottta: Thanks for pushing to look closer to this.
Indeed, the `fmt.go#processFile` already [uses `fmt.Fprintln`](https://github.com/opentofu/opentofu/blob/ff2e9b234b3faea756d4ad38ede0cc57cc218cdd/internal/command/fmt.go#L201C15-L201C23) to add a new line.
Then the `Println` called inside the `writeFn` was adding another new line which was removed by the [`writer` implementation](https://github.com/opentofu/opentofu/blob/ff2e9b234b3faea756d4ad38ede0cc57cc218cdd/internal/command/views/writer.go#L17).

I followed blindly the pattern that was used before without checking the actual logic behind.

Removed the `writer` and changed the `UserOutputWriter` implementation in a8bfb6332bfa4313e1fb24f5d7fff8ef59c74d89

PR #3806, opened 2026-02-26
Refactor `version` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu version` command to the View abstraction and to the `arguments` package.
* ⚠️ The JSON view for this command does not follow the general approach and instead it uses part of the existing view abstraction arguments to generate the version in the same format it was generating before the migration.

## Checklist

### Go checklist

 


internal/command/arguments/version.go
```diff
@@ -0,0 +1,44 @@
...
+
+// ParseVersion processes CLI arguments, returning a Version value, a closer function, and errors.
+// If errors are encountered, a Version value is still returned representing
+// the best effort interpretation of the arguments.
+func ParseVersion(args []string) (*Version, func(), tfdiags.Diagnostics) {
+	var diags tfdiags.Diagnostics
+	ret := &Version{}
+
+	cmdFlags := defaultFlagSet("version")
+	// Enable but ignore the global version flags. In main.go, if any of the
```
cam72cam, line 25: maybe this warrants a patch to main.go instead of persisting this hack here?

yottta: Will do, in a future step. Wanted to do it too, but is a little bit out of scope.

internal/command/views/version.go
```diff
@@ -0,0 +1,106 @@
...
+// NewVersion returns an initialized Version implementation for the given ViewType.
+// This view behaves differently from the general approach since the JSON format is not meant to follow
+// the general JSON format.
+// Instead, the view that is returned will always print diagnostics in human format while
+// [Version.PrintVersion] will return different results based on the [arguments.ViewOptions#ViewType].
+func NewVersion(args arguments.ViewOptions, view *View) Version {
+	return &VersionHuman{view: view, json: args.ViewType == arguments.ViewJSON}
+}
+
+type VersionHuman struct {
```
cam72cam: Sug: rename to VersionCustom

yottta: Good point. Though, I renamed it to `VersionMixed`. That sounds more appropriate to me, but feel free to disagree.

Changes in 67cd68822362e09fce42515955fe4a96129c3f93

PR #3809, opened 2026-02-26
eval/tofu2024: Handle for_each for resources, etc

apparentlymart:

In earlier work I skipped implementing for_each because I was just using count for initial testing, so now it's time to resolve that TODO so that all of the usual instance-selection arguments are now implemented.

This also includes a few small retroactive fixes of details missed in the implementations of the other arguments before, which I discovered during some manual testing.

As usual with these codepaths that are still quite in flux, there aren't any unit tests here yet but this code is written with the goal of being relatively easy to unit test in isolation later.


cam72cam (approved):
This works reasonably well as a standalone implementation.  Long term, I'd like to at least pull in the tests that we have for the existing for_each and count evaluation.

apparentlymart:
FWIW, I'm intending to add unit tests in various places like this as part of completing the "walking skeleton" milestone, but I'm intentionally deferring it for the moment because I've already been bitten several times by writing a bunch of tests and then having to delete them again (or heavily modify them in a confusing way) because of refactoring whatever they were testing.

Just want to get to a point where the functionality seems relatively complete and stable before spending too much more time on that.

PR #3811, opened 2026-02-27
Refactor `providers` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu providers` command to the View abstraction and to the `arguments` package.
* ⚠️ I skipped adding the JSON view since that requires a separate amount of work where it's needed to be decided on the format we want to print the providers in that format.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/providers_test.go
+++ b/internal/command/providers_test.go
@@ -9,24 +9,24 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/mitchellh/cli"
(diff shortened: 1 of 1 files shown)


yottta:
> Approving, with the idea that we will add a json output mode to this post-refactor. Maybe create an issue so we don't loose track of it?

Added a note at the bottom of [#3595](https://github.com/opentofu/opentofu/issues/3595).

PR #3814, opened 2026-02-27
lang/eval: Propagate provider instances between modules

apparentlymart:

Earlier work dealt with the problem of evaluating the references in a "providers" argument in a module call, but we didn't actually plumb it through the code that "compiles" the config representation into our internal configgraph representation.

Filling in this gap means that it's now possible to declare a provider instance in the root module and then use it in a child module, at the expense of unfortunately spreading around this "providers side-channel" special case to some more parts of the system. That seems inevitable though, since that special case is part of the user-facing language.

### DIFF
--- a/internal/lang/eval/internal/configgraph/module_call_instance.go
+++ b/internal/lang/eval/internal/configgraph/module_call_instance.go
@@ -66,8 +66,14 @@ type ModuleCallInstance struct {
 	// are no cycles in the dependency chain within the module.
 	InputsValuer *OnceValuer
 
-	// TODO: Also something for the "providers side-channel", as represented
-	// by the "providers" meta-argument in the current language.
+	// ProvidersFromParentValuers is our representation of the weird
+	// "side-channel" that allows providers to pass between modules, which
+	// is separate from the concept of input variables despite being
+	// conceptually similar.
+	//
+	// Each valuer in this map is expected to evaluate to a value of a type
+	// returned by [ProviderInstanceRefType].
+	ProvidersFromParentValuers map[addrs.LocalProviderConfig]*OnceValuer
 
 	validatedInputs grapheval.Once[cty.Value]
 }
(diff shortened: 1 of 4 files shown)


apparentlymart:
The reason this diff seems incomplete is because a bunch of the needed logic was already in place here.

What was missing was that the map of provider instances was not being propagated to child module calls during evaluation and so they were previously being given an empty map of provider instances in all cases. The logic to check that empty map was already there, but of course it was always returning an error saying that the needed provider wasn't present. This PR now populates that map so that the existing logic can successfully look up the providers.

As of this PR this only supports explicit passing of providers using the `providers` argument and doesn't yet handle the implicit inheritance behavior properly, because the new codepath uses the `configs` package in a different way that bypasses the handling of implicit inheritance, but I was intending to deal with that in a separate PR. This one is an incremental step that makes it possible to use resources in child modules _at all_.

apparentlymart:
The most relevant part of the existing code to this PR is the function here for "compiling" the table of provider instances available in a module:

https://github.com/opentofu/opentofu/blob/05c6507615f3548706b288b14080c3322f86d988/internal/lang/eval/internal/tofu2024/providers_sidechannel.go#L50-L60

Before this PR, the `fromParent` map was always empty. After this PR it contains the references described by the `providers` argument in the `module` block, so that looking up a provider instance that came from a parent module will now return an `exprs.Valuer` that came from the parent module instead of one from the current module.

The `providers` argument in a `module` block is also resolved using this same table, and so provider instance references can pass through multiple levels of call while still "remembering" where they were originally defined.

As noted in the previous comment, this doesn't yet handle the implicit inheritance of default provider instances to child modules when there is no `providers` block present in a `module` block. That's because in the current system that's handled by [`configs.BuildConfig`](https://github.com/opentofu/opentofu/blob/05c6507615f3548706b288b14080c3322f86d988/internal/configs/config_build.go#L22-L50), which doesn't run in the new codepath because the new evaluator implements config tree building itself, dynamically during evaluation. In a future commit I'm intending to add equivalent logic to `package tofu2024` to handle that, but I need to study the equivalent code in `package configs` more closely first to remember all of the weird edge-cases it needs to handle.

cam72cam:
That makes sense, thank you for explaining!

PR #3818, opened 2026-03-02
Refactor `taint` && `untaint` commands to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu taint` and `tofu untaint` commands to the View abstraction and to the `arguments` package.
* Put both commands in the same abstractions since the differences are really slim and doesn't really worth to duplicate the code just for the sake of separation.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/taint.go
+++ b/internal/command/taint.go
@@ -107,6 +107,8 @@ func (c *TaintCommand) Run(rawArgs []string) int {
 	if diags.HasErrors() {
 		return 1
 	}
+	// since we already printed the diagnostics above, we can discard the possible warnings
+	diags = tfdiags.Diagnostics{}
 
 	// Get the state
 	stateMgr, err := b.StateMgr(ctx, workspace)
--- a/internal/command/untaint.go
+++ b/internal/command/untaint.go
@@ -103,6 +103,8 @@ func (c *UntaintCommand) Run(rawArgs []string) int {
 	if diags.HasErrors() {
 		return 1
 	}
+	// since we already printed the diagnostics above, we can discard any possible warnings
+	diags = tfdiags.Diagnostics{}
 
 	// Get the state
 	stateMgr, err := b.StateMgr(ctx, workspace)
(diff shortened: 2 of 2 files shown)


internal/command/taint.go
```diff
@@ -25,118 +27,133 @@ type TaintCommand struct {
...
+
+	// Instantiate the view, even if there are flag errors, so that we render
+	// diagnostics according to the desired view
+	view := views.NewTaint(args.ViewOptions, c.View)
+	// ... and initialise the Meta.Ui to wrap Meta.View into a new implementation
+	// that is able to print by using View abstraction and use the Meta.Ui
+	// to ask for the user input.
+	c.Meta.configureUiFromView(args.ViewOptions)
 
-	if addr.Resource.Resource.Mode != addrs.ManagedResourceMode {
```
yottta, line 62: Validation moved to the `arguments` package

internal/command/taint_test.go
```diff
@@ -381,27 +380,28 @@ func TestTaint_missingAllow(t *testing.T) {
...
-	if code := c.Run(args); code != 0 {
-		t.Fatalf("bad: %d\n\n%s", code, ui.ErrorWriter.String())
+	code := c.Run(args)
+	output := done(t)
+	if code != 0 {
+		t.Fatalf("bad: %d\n\n%s", code, output.Stderr())
 	}
 
 	// Check for the warning
-	actual := strings.TrimSpace(ui.ErrorWriter.String())
```
yottta, line 405: Because the default `cli.MockUi` was configured differently from how the UI that tofu uses is (write warns to stdout and not stderr), I had to change this to reflect the actual real life behaviour.

internal/command/untaint.go
```diff
@@ -25,87 +27,103 @@ type UntaintCommand struct {
...
+	view.Diagnostics(diags)
 	if diags.HasErrors() {
 		return 1
 	}
 
 	// Get the state
 	stateMgr, err := b.StateMgr(ctx, workspace)
 	if err != nil {
-		c.Ui.Error(fmt.Sprintf("Failed to load state: %s", err))
+		view.Diagnostics(diags.Append(tfdiags.Sourceless(
```
cam72cam, line 112: This might print the warning diags twice given the line above.  This should be a new diags object

cam72cam: Though given existing bugs seen below, the safest option would be to set diags = nil after the view.Diagnostics call above.

cam72cam: Same applies to the taint command

yottta: Thanks. Great catch! Fixed in d79beb302d7b32e70d7b8015b43e8d8631edf837

PR #3821, opened 2026-03-02
resources: A new package for resource-related operations

apparentlymart:

The original motivation here was to implement the special case of making a synthetic plan without calling into the provider when the desired state is absent and the provider hasn't opted in to planning to destroy objects. This case needs to be opted in because older providers will crash if asked to plan with the config value or proposed value set to null. Resolving this deals with one of the current problems in how the new runtime deals with deleting existing objects, getting us a little closer to the "walking skeleton" milestone.

However, there was already far too much code duplicated across both the planning engine and the apply engine's "final plan" operation, and so this seemed like a good prompt to finally resolve the existing TODO comments by factoring out the shared planning logic into a separate place that both can depend on.

The new package "resources" is intended to grow into the home for all of these resource-related operations that each wrap a lower-level provider call but also perform preparation or followup actions such as making sure the provider's response is valid according to the requirements of the plugin protocol.

For now it only includes config validation and planning for managed resources, but is designed to grow to include other similar operations in future. The most likely next step is to add a method for applying a previously-created plan, which would replace a bunch of the code in the apply engine's implementation of the "ManagedApply" operation.

Currently the way this integrates with the rest of the provider-related infrastructure is a little awkward. We can hopefully improve on that in future, but for now the priority was to avoid this becoming yet another sprawling architecture change that would be hard to review. Once we feel like we've achieved the "walking skeleton" milestone for the new runtime we can think about how we want to tidy up the rough edges between the different components of this new design.

As has become tradition for these new codepaths, this is designed to be independently testable but not yet actually tested because we want to wait and see how all of this code settles before we start writing tests that would then make it harder to refactor as we learn more.

### DIFF
--- a/internal/resources/managed.go
+++ b/internal/resources/managed.go
@@ -1,0 +1,100 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resources
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/providers"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// ManagedResourceType represents a named resource type in a specific provider,
+// and also carries a client for interacting with that provider.
+//
+// Most methods of this type relate to managed-resource-related operations in
+// the underlying provider protocol, but also include additional OpenTofu-level
+// logic such as verifying that the provider is correctly implementing the
+// protocol's constraints on how objects are allowed to change.
+type ManagedResourceType struct {
+	// providerAddr is the provider that this resource type belongs to.
+	providerAddr addrs.Provider
+
+	// typeName is the resource type name as expected by the associated provider.
+	typeName string
+
+	// client is the client to use to interact with the provider that this
+	// resource type belongs to.
+	client providers.Interface
+}
+
+var _ ResourceType = (*ManagedResourceType)(nil)
+
+// NewManagedResourceType constructs a new [ManagedResourceType] for the
+// given resource type name in the provider whose client is provided.
+//
+// It's the caller's responsibility to make sure that the given client is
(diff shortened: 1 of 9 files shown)


yottta (approved):

internal/resources/managed.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/providers"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// ManagedResourceType represents
```
Yantrio: Nit: Missing the rest of this comment?

internal/resources/managed_plan.go
```diff
@@ -0,0 +1,325 @@
...
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Provider produced inconsistent final plan",
+			fmt.Sprintf(
+				"When expanding the plan for %s to include new values learned so far during apply, provider %q produced an invalid new value for %s.\n\nThis is a bug in the provider, which should be reported in the provider's own issue tracker.",
+				dispAddr, rt.providerAddr, tfdiags.FormatError(err),
+			),
+		))
+	}
+	return nil
```
Yantrio:
```suggestion
	return diags
```

cam72cam:
We should also consider moving the temporary "shared" package under this banner as well.

Mid-term we could consider pulling more tofu logic into here, once we have a more "locked-in" design for the new engine.

apparentlymart:
I agree that the similar functionality for ephemeral resources would be a good fit for this package too, but I'm intentionally leaving that as-is for the moment because ephemeral resources are not in the "walking skeleton" scope.

I have some mixed feelings about modifying the current runtime to use these new codepaths because that would reduce the amount of "known-good" unmodified code we have to compare behavior with as we get further along, but if the old and new implementations end up coexisting in stable releases for a while then I think that would make a good compromise so that we only need to do maintenance in one place. We can wait and see how the "draw the rest of the owl" part of the project will be shaped before we make a final decision about that, I think.

cam72cam (approved):

PR #3822, opened 2026-03-03
Refactor providers schema command to use views and arguments packages

AndrewCharlesHay:
Helps with #3765 

## Checklist

### Go checklist

 


internal/backend/remote-state/s3/client_test.go
```diff
@@ -921,7 +921,7 @@ func TestS3ChecksumsHeaders(t *testing.T) {
 // TestS3LockingWritingHeaders is double-checking that the configuration for writing the lock object is the same
 // with the state writing configuration
 func TestS3LockingWritingHeaders(t *testing.T) {
-	ignoredValues := []string{"Amz-Sdk-Invocation-Id", "Authorization", "X-Amz-Checksum-Sha256"}
+	ignoredValues := []string{"Amz-Sdk-Invocation-Id", "Authorization", "X-Amz-Checksum-Sha256", "X-Amz-Date"}
```
apparentlymart, line 924: This does seem like a reasonable thing to fix, but let's fix it in a separate PR since it's independent of what you set out to fix here.

If you did this because you saw test runs fail on your PR, then don't worry about that: maintainers can re-run the flakes on this PR until they pass, and then we can fix the flake on a separate PR merged either before or after this one.

AndrewCharlesHay: Yeah it was a flaky test. It is undone now

AndrewCharlesHay:
@yottta I apologize. It won't happen again. I've added tests

Here is the breakdown by function: 

internal/command/arguments/providers_schema.go
ParseProvidersSchema: 92.9%

internal/command/views/providers_schema.go
NewProvidersSchema: 100.0%

(*ProvidersSchemaHuman).UnsupportedLocalOp: 100.0%

(*ProvidersSchemaHuman).Output: 100.0%

(*ProvidersSchemaJSON).UnsupportedLocalOp: 100.0%

(*ProvidersSchemaJSON).Output: 100.0%

internal/command/arguments/providers_schema.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) The OpenTofu Authors
```
yottta, line 1: For all the new files, please add all the copyright headers. I know that it sounds weird, but a long time ago was decided to do it this way.

internal/command/arguments/providers_schema.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	closer, moreDiags := schema.ViewOptions.Parse()
+	diags = diags.Append(moreDiags)
+
+	if schema.ViewOptions.ViewType != ViewJSON {
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Output only in json is allowed",
+			"The `tofu providers schema` command requires the `-json` flag.\n",
```
yottta:
```suggestion
			"The `tofu providers schema` command requires the `-json` flag.",
```

internal/command/arguments/providers_schema_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+func TestParseProvidersSchema_jsonValidation(t *testing.T) {
```
yottta: Tests look good, but because there are a lot of PRs with similar code, would be nicer to have the same format for the tests.
Would be great if you could apply the structure that is used in [console_test.go](https://github.com/opentofu/opentofu/blob/0ffecc7a14d6c507e548841e62cf76bb9d2ceffa/internal/command/arguments/console_test.go#L18)

internal/command/views/providers_schema.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (v *ProvidersSchemaHuman) Diagnostics(diags tfdiags.Diagnostics) {
+	v.view.Diagnostics(diags)
+}
+
+func (v *ProvidersSchemaHuman) HelpPrompt() {
+	v.view.HelpPrompt("providers schema")
+}
+
+func (v *ProvidersSchemaHuman) UnsupportedLocalOp() {
```
yottta: To maintain backwards compatibility, we want to print the information as similar as possible with the one before the changes.
Check the implementation from [console.go](https://github.com/opentofu/opentofu/blob/0ffecc7a14d6c507e548841e62cf76bb9d2ceffa/internal/command/views/console.go#L124C24-L124C42) for how this should be done.

[views/views.go](https://github.com/opentofu/opentofu/blob/0ffecc7a14d6c507e548841e62cf76bb9d2ceffa/internal/command/views/views.go#L10) already contains the message. Please use that instead.

internal/command/arguments/providers_schema.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+
+package arguments
+
+import (
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// ProvidersSchema represents the command-line arguments for the providers schema command.
```
yottta:
```suggestion
// ProvidersSchema represents the command-line arguments for the 'providers schema' command.
```

internal/command/providers_schema.go
```diff
@@ -29,57 +31,68 @@ func (c *ProvidersSchemaCommand) Synopsis() string {
...
+
+	args, closer, diags := arguments.ParseProvidersSchema(rawArgs)
+	defer closer()
+
+	view := views.NewProvidersSchema(args.ViewOptions, c.View)
+
+	// Since the output format is JSON and not the json ui, we don't want to
+	// initialize the json ui.
+	if args.ViewOptions.ViewType != arguments.ViewJSON {
+		c.Meta.configureUiFromView(args.ViewOptions)
```
yottta: I understand why you wouldn't want to do it, but if the `ViewType` is not json, the next check will return because of the validations in `arguments.ParseProvidersSchema`.
Therefore, there is no need for the `if` statement.

internal/command/providers_schema_test.go
```diff
@@ -23,18 +22,19 @@ import (
...
-			Ui:               ui,
+			View:             view,
 		},
 	}
 
 	if code := c.Run(nil); code != 1 {
-		fmt.Println(ui.OutputWriter.String())
-		t.Fatalf("expected error: \n%s", ui.OutputWriter.String())
+		output := done(t)
+		fmt.Println(output.Stdout())
```
yottta:
```suggestion
```

Please remove that since will be printed by the `t.Fatalf` call.

internal/command/providers_schema.go
AndrewCharlesHay: That makes sense since ParseProvidersSchema already adds an error diagnostic when ViewType != ViewJSON, the HasErrors() check immediately after handles that case and returns early. The if guard was redundant. I've removed it, thanks for the explanation!

internal/command/views/providers_schema.go
AndrewCharlesHay: Done. I've updated both `ProvidersSchemaHuman.UnsupportedLocalOp()` and `ProvidersSchemaJSON.UnsupportedLocalOp()` to use `v.view.errorln(errUnsupportedLocalOp)`, matching the pattern in console.go

internal/command/arguments/providers_schema_test.go
AndrewCharlesHay: Done

internal/command/arguments/providers_schema.go
AndrewCharlesHay: Done

internal/command/arguments/providers_schema_test.go
```diff
@@ -0,0 +1,98 @@
...
+			if tc.wantDiags {
+				if len(diags) == 0 {
+					t.Fatal("expected diagnostics but got none")
+				}
+				for _, want := range tc.wantContain {
+					if !strings.Contains(diags.Err().Error(), want) {
+						t.Fatalf("wrong diags\n got: %s\nwant: %s", diags.Err().Error(), want)
+					}
+				}
+				return
```
yottta, line 84: Even though the `ParseProvidersSchema` returns diagnostics, we still should check for the returned args structure. These methods are meant to do their best because the arguments are still used to create a view later that will be able to show the diagnostics generated from it.

Therefore, please ensure that we do the diff of the `tc.want` also when the diagnostics exist.
If you rebased, you can find an example in [`arguments/taint_test.go`](https://github.com/opentofu/opentofu/blob/2c86fe8f85fc9c65954d277f7e468764ccbb9746/internal/command/arguments/taint_test.go#L110-L125)

internal/command/views/providers_schema.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// UnsupportedLocalOp is used to output a message to the user that the current operation is unsupported locally.
+	UnsupportedLocalOp()
+
+	// Output is used to display the final JSON output of the providers schema map.
+	Output(json string)
+}
+
+// NewProvidersSchema creates a new ProvidersSchema view based on the view type specified in the options.
+func NewProvidersSchema(opts arguments.ViewOptions, v *View) ProvidersSchema {
```
yottta, line 29: This is a similar situation with the implementation in [`views/version.go`](https://github.com/opentofu/opentofu/blob/2c86fe8f85fc9c65954d277f7e468764ccbb9746/internal/command/views/version.go#L34).
That being said, I would suggest to remove the `ProvidersSchemaJSON` and rename `ProvidersSchemaHuman` to something like `ProvidersSchemaMixed` (or `ProvidersSchemaCustom`) and treat it as a "human format" implementation.
The implementation should be really simple because the `Output` does nothing special to print the information but just gets a string and prints it.

Even though it's executed with the `-json` flag, as the old implementation requires it, the errors and warnings should be still shown in human form.

Just want to note that this is not my "personal opinion". This whole refactor is trying to get rid of some ancient structures (like the `Ui` struct) but we want to keep the functionality as much as possible equivalent with the current behavior.

AndrewCharlesHay:
I've updated the PR with the requested changes:
- Renamed `ProvidersSchemaHuman` to `ProvidersSchemaMixed` and unified the view logic.
- Updated tests in `arguments/providers_schema_test.go` to verify the arguments struct even when diagnostics are returned.

Let me know if there's anything else needed!

AndrewCharlesHay:
> Thanks for resolving the comments above. Besides the inline comments, please update [this](https://github.com/AndrewCharlesHay/opentofu/blob/6464b258b41662538cc028e55fe769aa7aa9fe5e/internal/command/providers_schema.go#L110) too. We need to use the `args.ViewOptions` instead of that new instance of `ViewOptions`.


Done! Thanks for pointing that out, args.ViewOptions is now being used.

AndrewCharlesHay:
> This looks good. I will wait till one more maintainer will have another look over this. Thanks a lot for the help.

That was quick! Thank you!

yottta:
Please check the tests, there are some failures on the output.

diofeher:
You need to sign DCO as well (git commit --amend --no-edit --signoff does the trick well when you forget it :)

internal/command/arguments/providers_schema.go
```diff
@@ -0,0 +1,64 @@
...
+func ParseProvidersSchema(args []string) (*ProvidersSchema, func(), tfdiags.Diagnostics) {
+	var diags tfdiags.Diagnostics
+
+	schema := &ProvidersSchema{
+		Vars: &Vars{},
+	}
+
+	cmdFlags := extendedFlagSet("providers schema", nil, nil, schema.Vars)
+
+	schema.ViewOptions.AddFlags(cmdFlags, false)
```
yottta:
```suggestion
	schema.ViewOptions.AddGranularFlags(cmdFlags, false, false)
```
It should be configured like this to just register the `-json` flag. Since this command works **only** with the `-json` flag, there is no use for the `-json-into`.

internal/command/views/providers_schema.go
```diff
@@ -0,0 +1,52 @@
...
+	v.view.HelpPrompt("providers schema")
+}
+
+func (v *ProvidersSchemaMixed) UnsupportedLocalOp() {
+	v.view.errorln(errUnsupportedLocalOp)
+}
+
+func (v *ProvidersSchemaMixed) Output(json string) {
+	// The provider schema output is expected to just be the raw JSON
+	v.view.streams.Print(json)
```
yottta:
```suggestion
	v.view.streams.Println(json)
```
Considering that before, in `command/providers_schema.go` was used `c.Ui.Output()` to print the json schema, if you look into that implementation, it adds a new line at the end. Please do the same here.

internal/command/providers_schema.go
```diff
@@ -29,57 +31,67 @@ func (c *ProvidersSchemaCommand) Synopsis() string {
...
-	if !jsonOutput {
-		c.Ui.Error(
-			"The `tofu providers schema` command requires the `-json` flag.\n")
-		cmdFlags.Usage()
+	view := views.NewProvidersSchema(args.ViewOptions, c.View)
+
+	// The actual schema data is printed directly to stdout at the end of the command
+	// and doesn't rely on the UI layer. Configuring JSON UI drops a Version payload
+	// into stdout which invalidates the raw schema JSON output.
+	// c.Meta.configureUiFromView(args.ViewOptions)
```
yottta: I don't agree with this.

We have the validation for this in the `arguments` package.

I know that it could be seen as counter-intuitive, but the `-json` flag in this command tries to get only the output in JSON format, but not the rest of the output. I know, a bummer. But this is how it was designed, and this is how we should keep it for the moment.

Instead of ignoring the configuration of the Meta.Ui (which could create other errors down the line), we have other options to do so:
* In the arguments package, where the condition for ViewJSON is, you could set the human view type once the validation passes, forcing this configuration to go on human form.
* In this `configureUiFromView`, you can pass a new ViewOptions with the Human view type.

Personally, I would go with the first, since if ever there will be another functionality added here that would need the ViewOptions, might tend to use the `args.ViewOptions`, which would put the output in a weird state. Furthermore, the first option is better because this new structure (with the `arguments` package) is meant to separate the concern of parsing the flags and arguments into a separate component, we should rely on what that returns.

yottta: Any news on this @AndrewCharlesHay?
I would like to merge this soon.

AndrewCharlesHay: Sorry for the delay @yottta  it should be good now

yottta: Even though I am not sure about the fix with the `configureUiFromView`, I will let this slide for the moment and will revisit it later if needed.
Thanks for the changes.

AndrewCharlesHay:
The `windows_amd64` unit test failure (TestJsonIntoStream/plan in internal/command/e2etest) is unrelated to the changes in this PR. None of the files modified here touch the e2etest package or anything in the state locking path.

The failure is a timing-sensitive comparison between two sequential plan runs on Windows. The second run (using -json-into) crosses the threshold that triggers the state_lock_acquire JSON message, while the first run (using -json) does not — resulting in a one-line diff between the two outputs. This is a pre-existing flakiness on the Windows runner and not a regression introduced by this PR

PR #3828, opened 2026-03-04
Refactor `state list` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state list` command to the View abstraction and to the `arguments` package.
* The view implementation will be used for all of the other state commands since there are outputs shared between all of these commands. Similar to how it has been done in #3770 for the `tofu workspace *` commands.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/state_list_test.go
+++ b/internal/command/state_list_test.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/mitchellh/cli"
 	"github.com/opentofu/opentofu/internal/command/arguments"
 	"github.com/opentofu/opentofu/internal/command/workdir"
 )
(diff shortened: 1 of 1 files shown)


internal/command/views/state.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+var _ State = (*StateJSON)(nil)
+
+func (v *StateJSON) Diagnostics(diags tfdiags.Diagnostics) {
+	v.view.Diagnostics(diags)
+}
+
+func (v *StateJSON) StateLoadingFailure(baseError string) {
+	cleanedUp := strings.ReplaceAll(
```
cam72cam, line 104: This is fine for now, but it would probably make sense to create a new function that formats the error properly.

Though out of curiosity, how would this look as a sourceless error diag in Human/JSON?

yottta: With the current implementation:
Screenshot 2026-03-04 at 13 42 33

and with sourceless diagnostics:
Screenshot 2026-03-04 at 13 37 54
Kind of weird, right? 😅 

And JSON since you asked:
Now:
```shell
{"@level":"error","@message":"Error loading the state: this is the base error. Please ensure that your OpenTofu state exists and that you've configured it properly. You can use the \"-state\" flag to point OpenTofu at another state file.",
"@module":"tofu.ui","@timestamp":"2026-03-04T13:43:07.660866+02:00"}
```
and with diagnostic:
```shell
{"@level":"error","@message":"Error: This is the summary","@module":"tofu.ui","@timestamp":"2026-03-04T13:43:07.660884+02:00","diagnostic":{"severity":"error","summary":"This is the summary","detail":"Error loading the state: this would be the base error\n\nPlease ensure that your OpenTofu state exists and that you've\nconfigured it properly. You can use the \"-state\" flag to point\nOpenTofu at another state file."},"type":"diagnostic"}
```

---

There is a note at the bottom of the #3765's description that I took from a conversation that I had on another PR: https://github.com/opentofu/opentofu/pull/3764#discussion_r2850215802

I plan on getting rid of these `errorln` and `warnln` once all commands are migrated.
Together with that, most of these methods that print a full block colored in one color, will be replaced with proper sourceless diagnostics in such a way to keep as much of the format they have today but colored consistently with the rest of diagnostics in OpenTofu.

That's why I have this shortcut on formatting the message. Once migrated, the JSON error will also be reworked to provide equivalent messages.

cam72cam: I was thinking that the first line of the error would be the diagnostic description and the rest the summary.

cam72cam: Image

cam72cam: We spoke offline and @yottta convinced me that it's easier to leave this as is for now as the summary should be static.  That would not be the case in a 1-1 translation of this code.

PR #3831, opened 2026-03-04
applying: Update state correctly after deleting an object

apparentlymart:

The methods of `states.State` and `states.SyncState` related to saving resource instance objects have some quirky API design where some resource-level and resource-instance-level properties also get passed in as some additional arguments and then it opportunistically updates those higher-level fields in the state structure at the same time as updating the specific resource instance object.

That quirky design caused me to originally try to write this in a way where we always pass in a state object but sometimes the actual value is null to represent that the object has been deleted, but that doesn't actually work properly: it causes there to be a broken object in the state which still has partial metadata but is incomplete and cannot be properly decoded again later.

Ideally we'd fix this at the root by making the main methods more robust and have a less confusing API design, but at our current phase of work on the new runtime we have a preference for leaving existing codepaths unchanged and compensating in our new experimental callers instead, so that's the strategy I followed here: the experimental method for writing a "full" object into the state relies on the underlying method's implementation detail of ignoring the metadata fields whenever the main object is `nil`, just passing zero-values for those to get the needed behavior.

With this change in place, deletion of resource instance objects now causes them to get properly removed from the state, without corrupting the state in a way that makes it unusable for future plan/apply rounds.

### DIFF
--- a/internal/states/instance_object_full.go
+++ b/internal/states/instance_object_full.go
@@ -48,6 +48,9 @@ type ResourceInstanceObjectFull = resourceInstanceObjectRepr[cty.Value]
 type ResourceInstanceObjectFullSrc = resourceInstanceObjectRepr[ValueJSONWithMetadata]
 
 func DecodeResourceInstanceObjectFull(src *ResourceInstanceObjectFullSrc, ty cty.Type) (*ResourceInstanceObjectFull, error) {
+	if src == nil {
+		return nil, nil
+	}
 	v, err := src.Value.Decode(ty)
 	if err != nil {
 		return nil, err
@@ -56,6 +59,9 @@ func DecodeResourceInstanceObjectFull(src *ResourceInstanceObjectFullSrc, ty cty
 }
 
 func EncodeResourceInstanceObjectFull(obj *ResourceInstanceObjectFull, ty cty.Type) (*ResourceInstanceObjectFullSrc, error) {
+	if obj == nil {
+		return nil, nil
+	}
 	vSrc, err := EncodeValueJSONWithMetadata(obj.Value, ty)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


internal/states/instance_object_full.go
```diff
@@ -167,41 +173,55 @@ func (s *SyncState) SetResourceInstanceObjectFull(addr addrs.AbsResourceInstance
...
+				smallerObj.AttrSensitivePaths[i] = cty.PathValueMarks{
+					Path:  path,
+					Marks: marks,
+				}
 			}
 		}
 	}
 	if deposedKey == NotDeposed {
-		ms.SetResourceInstanceCurrent(addr.Resource, smallerObj, providerConfigAddr, obj.ProviderInstanceAddr.Key)
+		ms.SetResourceInstanceCurrent(addr.Resource, smallerObj, providerConfigAddr, providerInstanceKey)
```
cam72cam, line 222: Given how we store the providerConfigAddr and providerInstanceKey, is it possible for this to get in a state where the state contains a deposed resource instance, but the provider config has been overridden by the current being deleted?

I'm thinking of a scenario where the Current object is deleted and the Deposed object fails to apply.

apparentlymart: Your comment made me notice a more fundamental problem with what I did here:

I didn't look closely enough at the equivalent functionality in `ModuleState.SetResourceInstanceDeposed`, and that one doesn't have the same structure of completely ignoring the `providerConfigAddr` and `providerInstanceKey` arguments when `smallerObj` is nil, so that would leave the resource-level and instance-level state objects broken in a _different_ way that would cause the same effect of the state being invalid for subsequent plan/apply rounds.

I'm going to abandon this PR and make another attempt more similar to what I was originally intending this to be like: where the `obj` argument to `SetResourceInstanceObjectFull` is always non-nil and instead it's the value inside that is unset, so that we can still pass in metadata like the provider instance address even when the data itself is absent.

(That is still awkward in some different ways than this first PR was, but I don't think there's any approach to this that isn't at least _slightly_ awkward because the `ModuleState` API is pretty awkward itself, and I'm trying to avoid changing that here.)

apparentlymart:
This was closed in favor of https://github.com/opentofu/opentofu/pull/3840.

PR #3835, opened 2026-03-05
Refactor `state mv` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state mv` command to the View abstraction and to the `arguments` package.
* The handling of the flags is weirder than in other commands. Because the `-backup` have a different default value than the [State.Backup](https://github.com/opentofu/opentofu/blob/b992aa6582b3aa7995d3998172eeb1181cfab934/internal/command/arguments/extended.go#L325) flag. That default value is actually critical in the way the command logic is implemented so we had to define all the flags for `arguments.State` to be able to have this custom default value. There is a note in the code and I would like to tackle it eventually.
## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/state_mv_test.go
+++ b/internal/command/state_mv_test.go
@@ -13,7 +13,6 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/mitchellh/cli"
 	"github.com/opentofu/opentofu/internal/command/workdir"
 	"github.com/opentofu/opentofu/internal/tfdiags"
 
@@ -60,14 +59,12 @@ func TestStateMv(t *testing.T) {
 	statePath := testStateFile(t, state)
 
 	p := testProvider()
-	ui := new(cli.MockUi)
-	view, _ := testView(t)
+	view, done := testView(t)
 	c := &StateMvCommand{
 		StateMeta{
 			Meta: Meta{
 				WorkingDir:       workdir.NewDir("."),
 				testingOverrides: metaOverridesForProvider(p),
-				Ui:               ui,
 				View:             view,
 			},
 		},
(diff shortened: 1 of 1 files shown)


internal/command/arguments/state_mv.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// StateMv represents the command-line arguments for the 'state list' command.
+type StateMv struct {
+	// RawSrcAddr represents a resources address that is requested by the user to be moved
+	RawSrcAddr string
+	// RawSrcAddr represents a resources address that is requested by the user to be used to move the
```
diofeher:
```suggestion
	// RawDestAddr represents a resources address that is requested by the user to be used to move the
```

internal/command/arguments/state_mv.go
```diff
@@ -0,0 +1,89 @@
...
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package arguments
+
+import (
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// StateMv represents the command-line arguments for the 'state list' command.
```
diofeher:
```suggestion
// StateMv represents the command-line arguments for the 'state mv' command.
```

yottta: Fixed. Sadly I rebased instead of pushing a new commit 🤦

internal/command/arguments/state_mv.go
yottta: Fixed. Sadly I rebased instead of pushing a new commit 🤦

PR #3837, opened 2026-03-05
Refactor `state pull` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state pull` command to the View abstraction and to the `arguments` package.
* This command already prints the state in JSON format so the it cannot have the `-json` or `-json-into` flags. Therefore, the `StateJSON` newly added method prints something different in case a future maintainer will start using it. If any of you thinks that we should panic there instead, please let me know.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/state_pull_test.go
+++ b/internal/command/arguments/state_pull_test.go
@@ -6,7 +6,6 @@
 package arguments
 
 import (
-	"fmt"
 	"strings"
 	"testing"
 
@@ -38,7 +37,7 @@ func TestParseStatePull_basicValidation(t *testing.T) {
 	}
 
 	cmpOpts := cmp.Options{
-		cmpopts.IgnoreUnexported(Vars{}, ViewOptions{}, State{}),
+		cmpopts.IgnoreUnexported(Vars{}, ViewOptions{}),
 		cmpopts.IgnoreFields(ViewOptions{}, "JSONInto"), // We ignore JSONInto because it contains a file which is not really diffable
 	}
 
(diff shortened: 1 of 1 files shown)


internal/command/arguments/state_pull_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		"unknown flag": {
+			args:        []string{"-unknown"},
+			want:        statePullArgsWithDefaults(nil),
+			wantErrText: "Failed to parse command-line flags",
+		},
+	}
+
+	cmpOpts := cmp.Options{
+		cmpopts.IgnoreUnexported(Vars{}, ViewOptions{}, State{}),
```
diofeher:
```suggestion
		cmpopts.IgnoreUnexported(Vars{}, ViewOptions{}),
```

There's no State{} in this code path, so it can be removed from cmpOpts

internal/command/arguments/state_pull_test.go
```diff
@@ -0,0 +1,128 @@
...
+		"multiple vars mixed": {
+			args:      []string{"-var", "a=1", "-var-file", "f.tfvars", "-var", "b=2"},
+			wantCount: 3,
+			wantEmpty: false,
+		},
+	}
+
+	for name, tc := range testCases {
+		for _, isTaint := range []bool{true, false} {
+			tcName := fmt.Sprintf("%s_%t", name, isTaint)
```
diofeher: I think this inner loop can be removed. IsTaint is only used on the name of the test, but the same test is run twice here.

diofeher (changes requested):

internal/command/arguments/state_pull_test.go
yottta: Thanks. Copy paste leftovers 🤦 . Fixed in 76d0267ff94f7d7f42a5eacc6f2086397e0ad22e

internal/command/arguments/state_pull_test.go
yottta: Thanks. Copy paste leftovers 🤦 . Fixed in 76d0267ff94f7d7f42a5eacc6f2086397e0ad22e

diofeher (approved):

PR #3842, opened 2026-03-06
Refactor `state push` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state push` command to the View abstraction and to the `arguments` package.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/state_push.go
+++ b/internal/command/arguments/state_push.go
@@ -37,6 +37,7 @@ func ParseStatePush(args []string) (*StatePush, func(), tfdiags.Diagnostics) {
 	ret.Backend.AddIgnoreRemoteVersionFlag(cmdFlags)
 	ret.Backend.AddStateFlags(cmdFlags)
 	cmdFlags.BoolVar(&ret.Force, "force", false, "")
+	ret.ViewOptions.AddFlags(cmdFlags, false)
 
 	if err := cmdFlags.Parse(args); err != nil {
 		diags = diags.Append(tfdiags.Sourceless(
@@ -57,8 +58,6 @@ func ParseStatePush(args []string) (*StatePush, func(), tfdiags.Diagnostics) {
 		ret.StateSrc = args[0]
 	}
 
-	// we only parse but do not register the views flags since this command does not need it because it already
-	// prints the state in json format
 	closer, moreDiags := ret.ViewOptions.Parse()
 	diags = diags.Append(moreDiags)
 


internal/command/state_push.go
```diff
@@ -203,10 +218,37 @@ Options:
...
 
+  -json               Produce output in a machine-readable JSON format, 
+                      suitable for use in text editor integrations and other 
+                      automated systems. Always disables color.
+
+  -json-into=out.json Produce the same output as -json, but sent directly
+                      to the given file. This allows automation to preserve
+                      the original human-readable output streams, while
+                      capturing more detailed logs for machine analysis.
+
```
diofeher, line 229: Since we're adding these arguments, should we update the website docs?

I see in another part that you wrote: 

```
// we only parse but do not register the views flags since this command does not need it because it already
	// prints the state in json format
```

So the idea is just to have these as "common flags" but don't do anything with them since they are not important here as `-json` is the default?

yottta: Great catch!
Yeah, I need to get done whole migration carefully because I pay less and less attention since the work is repetitive 🤦 
Yes, you're right. I copied this from `state pull` which prints the pulled state that is already json formatted. That's why that command cannot support the `-json` flag because in human format it prints the state as json already.
But `state push` doesn't do it, so that comment does not apply here. It can use the `-json` flag even though it will be used strictly for diagnostics.
Fixed in 1011f41857686c046b150609e46d5c5aca625384

internal/command/state_push.go
```diff
@@ -84,53 +100,53 @@ func (c *StatePushCommand) Run(args []string) int {
 		c.Close()
 	}
 	if err != nil {
-		c.Ui.Error(fmt.Sprintf("Error reading source state %q: %s", args[0], err))
+		view.Diagnostics(diags.Append(fmt.Errorf("Error reading source state %q: %s", args.StateSrc, err)))
```
diofeher, line 103:
```suggestion
		view.Diagnostics(diags.Append(fmt.Errorf("Error reading source state %q: %w", args.StateSrc, err)))
```

I just noticed there are multiple `fmt.Errorf` calls using `"%s", err`, while we should be using `"%w", err`

internal/command/state_push.go
diofeher: No problem, I totally feel that is super repetitive ^^ But you're doing a great work on it

internal/command/state_push.go
yottta: I agree that it's the correct way to do it, when you return the error to be able to unwrap it, but in this case it's just for printing and nothing else.
In a future step (noted in #3765 all the way to the bottom of the description) I plan to handle multiple things like this:
> We need to check if we want to get rid by the newly added warnln and errorln that colors the whole given text and replace that with actual diagnostics. There is a conversation about that in https://github.com/opentofu/opentofu/pull/3764#discussion_r2850215802.
> * This point should also handle the view.Diagnostics(diags.append(fmt.Errorf(...))) from the command by replacing those with tfdiags.Sourceless diagnostics..
The idea is to also rework some of these diagnostics, to use `Sourceless` diagnostics instead and in there we will add also the error in the `Details` of the diagnostic instead.

That being said, do you think that I should still handle this right here right now?

AndrewCharlesHay: I agree with @yottta. leaving it as `%s` is fine since it's just for stringification right now. Deferring the actual diagnostic wrapping to a dedicated follow-up PR #3765  using `tfdiags.Sourceless` will keep this PR more focused on the view migration.

diofeher: @yottta Hmm, that's a good point. I can see pros and cons of doing that now or later, but I'm good with leaving the way it's rn

PR #3844, opened 2026-03-06
Refactor `state replace-provider` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state replace-provider` command to the View abstraction and to the `arguments` package.
* Because the command was using `Ui` for the input too, I had to change that to the other component that is meant to do this better, [`UIInput`](https://github.com/opentofu/opentofu/blob/315d0287675f5f312d234296a22585b52ba8034f/internal/command/ui_input.go#L36). Checked its implementation comparatively with the logic inside Ui and are equivalent in logic but the `UiInput` provides more customisation. When it comes to interruption and kill signals, both are working the same. In terms of UX, the `UiInput` looks even better:
  image

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/views/state.go
+++ b/internal/command/views/state.go
@@ -320,7 +320,7 @@ func (v *StateJSON) ReplaceProviderOverview(from, to addrs.Provider, willReplace
 		replacedResources[i] = resource.Addr.String()
 	}
 	msg := fmt.Sprintf("OpenTofu will replace provider from %s to %s for %d resources", from, to, len(willReplace))
-	v.view.log.Info(msg, "resources", replacedResources, "type", "replace_provider")
+	v.view.log.Info(msg, "resources", replacedResources, "type", "replace_provider", "from", from.String(), "to", to.String())
 }
 
 func (v *StateJSON) ReplaceProviderCancelled() {
--- a/internal/command/views/state_test.go
+++ b/internal/command/views/state_test.go
@@ -217,6 +217,8 @@ func TestStateViews(t *testing.T) {
 					"resources": []any{"res.foo"},
 					"@module":   "tofu.ui",
 					"type":      "replace_provider",
+					"from":      "registry1.org/ns/prov",
+					"to":        "registry2.org/ns/prov",
 				},
 			},
 			wantStdout: `OpenTofu will perform the following actions:


internal/command/arguments/state_replace_provider.go
```diff
@@ -0,0 +1,92 @@
...
+	ret.Backend.AddIgnoreRemoteVersionFlag(cmdFlags)
+	ret.Backend.AddStateFlags(cmdFlags)
+	cmdFlags.BoolVar(&ret.AutoApprove, "auto-approve", false, "skip interactive approval of replacements")
+	// NOTE: because the `-backup` flag needs a different default value than usual,
+	// we cannot use the [State] flags extension to register and parse these.
+	// Therefore, we need to have those flags redefined here.
+	// TODO meta-refactor: we might want to have a separate function on the [arguments.State] to register flags,
+	//  where default value can be provided by the caller. This way we could still use those flags instead of redefining
+	//  everything again.
+	cmdFlags.StringVar(&ret.BackupPath, "backup", "-", "backup-path")
```
yottta, line 56: This is not **truly** needed but since it has that default value I want to be safe than sorry. I will dig more into a later iteration (due to the TODO above) to be sure that we can move to the `State` extension instead.

internal/command/views/state.go
```diff
@@ -253,6 +311,27 @@ func (v *StateJSON) PrintPulledState(_ string) {
...
+	v.view.log.Info("No matching resources found")
+}
+
+func (v *StateJSON) ReplaceProviderOverview(from, to regaddr.Provider, willReplace []*states.Resource) {
+	replacedResources := make([]string, len(willReplace))
+	for i, resource := range willReplace {
+		replacedResources[i] = resource.Addr.String()
+	}
+	msg := fmt.Sprintf("OpenTofu will replace provider from %s to %s for %d resources", from, to, len(willReplace))
+	v.view.log.Info(msg, "resources", replacedResources, "type", "replace_provider")
```
yottta: @apparentlymart, you had really good takes recently on json output of the commands. Could you please take a quick look at this particular method?
There is a comparison of human vs json in the tests of this file, if that helps

internal/command/state_replace_provider.go
```diff
@@ -137,38 +156,29 @@ func (c *StateReplaceProviderCommand) Run(args []string) int {
...
 
 	// Confirm
-	if !autoApprove {
-		c.Ui.Output(colorize.Color(
-			"\n[bold]Do you want to make these changes?[reset]\n" +
-				"Only 'yes' will be accepted to continue.\n",
-		))
-		v, err := c.Ui.Ask("Enter a value:")
+	if !args.AutoApprove {
+		v, err := c.UIInput().Input(ctx, &tofu.InputOpts{
```
yottta, line 171: Since `Ui` was used before to ask for user input, and we want to move away from that, I included this change too.
Left some details in the description of the PR about this.


_I wrote this comment to raise awareness to whoever will review this._
_Please let me know if you see any possible issues with this._

internal/command/views/state.go
apparentlymart: For the list of resource addresses this seems like a reasonable approach!

I noticed that the "human" equivalent of this function also includes the old and new provider addresses. Do you think we should include those as part of this object too, so that an alternative UI implemented in terms of the JSON output would have access to all of the same information?

internal/command/views/state.go
```diff
@@ -11,7 +11,9 @@ import (
 
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/command/arguments"
+	"github.com/opentofu/opentofu/internal/states"
 	"github.com/opentofu/opentofu/internal/tfdiags"
+	regaddr "github.com/opentofu/registry-address/v2"
```
apparentlymart: I don't think we've discussed this explicitly anywhere yet, but FWIW the trend so far has been to keep using `addrs.Provider` even though it's an alias for `regaddr.Provider`, treating the use of parts of an external library here as an implementation detail.

I'd personally prefer to keep doing that instead of exposing so many different packages in this codebase directly to the upstream package, given that our address types are a _very_ cross-cutting concern that almost every package ends up interacting with. For me ideally only `package addrs` would actually import `github.com/opentofu/registry-address/v2` directly, and everything else would interact with it only indirectly through `github.com/opentofu/opentofu/internal/addrs` so that we're consistent about all of our address types being defined in a single central place.

yottta: Good catch! Didn't even realise that's pointing to the same type. It makes total sense! Thanks for the details.
Fixed in 2eb7418760f996f4edc676c1523eefc3ae72b83b

internal/command/views/state.go
yottta: Sounds good! Like so b31317457ef3f957c8833bffb639f288c0440187?

apparentlymart: Yeah, that looks good to me. Thanks!

PR #3849, opened 2026-03-09
Refactor `state rm` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state rm` command to the View abstraction and to the `arguments` package.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/state_rm.go
+++ b/internal/command/state_rm.go
@@ -264,10 +264,3 @@ func (c *StateRmCommand) GatherVariables(args *arguments.Vars) {
 	}
 	c.Meta.variableArgs = flags.RawFlags{Items: &items}
 }
-
-// TODO meta-refactor: remove this once we migrated all `state` commands to the view abstraction
-const errStateRmPersist = `Error saving the state: %s
-
-The state was not saved. No items were removed from the persisted
-state. No backup was created since no modification occurred. Please
-resolve the issue above and try again.`
--- a/internal/command/meta.go
+++ b/internal/command/meta.go
@@ -602,19 +602,6 @@ func (m *Meta) defaultFlagSet(n string) *flag.FlagSet {
 	return f
 }
 
-// ignoreRemoteVersionFlagSet add the ignore-remote version flag to suppress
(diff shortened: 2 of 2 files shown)


internal/command/state_rm.go
```diff
@@ -25,111 +26,125 @@ type StateRmCommand struct {
...
+	view := views.NewState(args.ViewOptions, c.View)
+	// ... and initialise the Meta.Ui to wrap Meta.View into a new implementation
+	// that is able to print by using View abstraction and use the Meta.Ui
+	// to ask for the user input.
+	c.Meta.configureUiFromView(args.ViewOptions)
+	if diags.HasErrors() {
+		view.Diagnostics(diags)
+		if args.ViewOptions.ViewType == arguments.ViewJSON {
+			return 1 // We don't want to print the help of the command in JSON view
+		}
```
diofeher, line 60: I just noticed that we haven't done the same at https://github.com/opentofu/opentofu/blob/9e0d54f007681c5de9baeae47e06e6a4c4d303f5/internal/command/state_mv.go

yottta: Thanks for catching it 🙌 

Possibly there might be other commands with the same issue. That's why in #3765 I have a note all the way at the bottom of the description for it:
> Double check that the return of cli.RunResultHelp is consistent across all commands and ignored when -json flag is active. We don't want to mix the human output with the json one. The error printed in json format through diagnostics is enough.
> * And must be discussed with the maintainers on what level, changing the previous exit code is a violation of the backwards compatibilty. We might need to move all the commands to a HelpPrompt approach instead and leave the exist code as it was before.

This is something I've realised while did this refactor. There is no consistency regarding this part. Will check all the commands again once all commands are migrated.

diofeher: Oh sorry, I missed this comment!

yottta: No reason to be sorry. I keep updating that description. Don't need for everybody to stay on top of it. It's just a reminder for the work still to be done.

diofeher (approved):

PR #3857, opened 2026-03-09
go.mod: Upgrade to Go 1.25.8

lbordowitz:
Resolves the same issues as #3856, for OpenTofu 1.11.

## Checklist

### Go checklist

 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/opentofu/opentofu
 
-go 1.25.7
+go 1.25.8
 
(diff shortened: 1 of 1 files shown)


apparentlymart:
My analysis of the relevant advisories suggests that this is not significant enough to justify immediately issuing a new patch release for the v1.11.x series, but we I think we should still merge this into the `v1.11` branch anyway so that it'll be picked up in any subsequent patch releases we make for that series for other reasons later.

Given that none of the associated advisories are significant to OpenTofu, I don't think we need changelog entries covering any of them.

PR #3865, opened 2026-03-10
Refactor `state show` command to use View instead of Ui and to use the `arguments` package

yottta:

Part of #3765  

This PR migrates the `tofu state show` command to the View abstraction and to the `arguments` package.
* It propagates the `arguments.StateShow.ShowSensitive` into the `Meta.View` that is used later by the `views.StateHuman` to control weather the sensitive values are shown or not.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/views/state.go
+++ b/internal/command/views/state.go
@@ -8,6 +8,7 @@ package views
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/opentofu/opentofu/internal/addrs"
@@ -64,7 +65,7 @@ func NewState(args arguments.ViewOptions, view *View) State {
 	var ret State
 	switch args.ViewType {
 	case arguments.ViewJSON:
-		ret = &StateJSON{view: NewJSONView(view, nil)}
+		ret = &StateJSON{view: NewJSONView(view, nil), output: view.streams.Stdout.File}
 	case arguments.ViewHuman:
 		ret = &StateHuman{view: view}
 	default:
@@ -72,7 +73,7 @@ func NewState(args arguments.ViewOptions, view *View) State {
 	}
 
 	if args.JSONInto != nil {
-		ret = &StateMulti{ret, &StateJSON{view: NewJSONView(view, args.JSONInto)}}
+		ret = &StateMulti{ret, &StateJSON{view: NewJSONView(view, args.JSONInto), output: args.JSONInto}}
 	}
 	return ret
 }
@@ -361,7 +362,8 @@ func (v *StateHuman) ShowResourceState(_ context.Context, stateFile *statefile.F
 }
 
 type StateJSON struct {
-	view *JSONView
+	view   *JSONView
+	output *os.File
 }
 
 var _ State = (*StateJSON)(nil)
@@ -512,7 +514,8 @@ func (v *StateJSON) ShowResourceState(_ context.Context, stateFile *statefile.Fi
 		v.view.Info("no state")
 		return 0
 	}
-	state, err := jsonstate.MarshalForLog(stateFile, schemas)
+
+	rawState, err := jsonstate.Marshal(stateFile, schemas)
 	if err != nil {
 		v.Diagnostics(tfdiags.Diagnostics{}.Append(tfdiags.Sourceless(
 			tfdiags.Error,
@@ -521,7 +524,7 @@ func (v *StateJSON) ShowResourceState(_ context.Context, stateFile *statefile.Fi
 		)))
 		return 1
 	}
-	v.view.log.Info("resource state", "state", state)
+	_, _ = fmt.Fprintln(v.output, string(rawState))
 	return 0
 }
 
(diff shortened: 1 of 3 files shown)


internal/command/views/state.go
```diff
@@ -398,6 +483,48 @@ func (v *StateJSON) RemoveFinalStatus(count int) {
...
+	state, err := jsonstate.MarshalForLog(stateFile, schemas)
+	if err != nil {
+		v.Diagnostics(tfdiags.Diagnostics{}.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Failed to marshal state to json",
+			fmt.Sprintf("Error while marshalling state to json: %s", err),
+		)))
+		return 1
+	}
+	v.view.log.Info("resource state", "state", state)
```
yottta: The output here right now looks like this:
```json
{"@level":"info","@message":"resource state","@module":"tofu.ui","@timestamp":"...","state":{"format_version":"1.0","terraform_version":"1.12.0","values":{"root_module":{...}}}
```

I cannot decide if we should format this differently, something like:
```json
{"@level":"info","@message":{"format_version":"1.0","terraform_version":"1.12.0","values":{"root_module":{...}},"@module":"tofu.ui","@timestamp":"...","type":"resource_state"}}
```
Where the actual json state is in the `@message` field and there is a `type=resource_state` field.

WDYT?

internal/command/state_show.go
```diff
@@ -31,135 +29,147 @@ type StateShowCommand struct {
...
+	view := views.NewState(args.ViewOptions, c.View)
+	// ... and initialise the Meta.Ui to wrap Meta.View into a new implementation
+	// that is able to print by using View abstraction and use the Meta.Ui
+	// to ask for the user input.
+	c.Meta.configureUiFromView(args.ViewOptions)
+	if diags.HasErrors() {
+		view.Diagnostics(diags)
 		return cli.RunResultHelp
 	}
+	c.View.SetShowSensitive(args.ShowSensitive)
```
yottta, line 63: Propagates the show sensitive to be used later in `view.ShowResourceState`

internal/command/views/state.go
diofeher: Do you think this could be a breaking change for people consuming that output and turning it into actionable (I'm thinking orchestrators or platform teams here)?

internal/command/arguments/state_show_test.go
```diff
@@ -0,0 +1,108 @@
...
+		"no arguments": {
+			args:        []string{},
+			want:        stateShowArgsWithDefaults(nil),
+			wantErrText: "Invalid number of arguments",
+		},
+		"too many arguments": {
+			args:        []string{"resource_address", "extra"},
+			want:        stateShowArgsWithDefaults(nil),
+			wantErrText: "Invalid number of arguments",
+		},
```
diofeher, line 65: Can you add one extra test for invalid arguments? Like `--invalid`

yottta: Added in 3964a5be8292af9e3d391d9fc782304b9eed8aa1

internal/command/views/state.go
yottta: Right now, it definitely isn't because the json format does not even exist right now for this command.
This is more of the contract that we agree on right now and commit to go with for the foreseeable future.
After we ship this, yeah, if we decide to change this, we might break some flows that people built around this.
So I would like to hear more from the maintainers around this.

cam72cam: I think I prefer the first option. "@message" is always a string in the current format.

cam72cam: Actually, should this behave more like `tofu output` for the -json mode?

yottta: Thanks a lot @cam72cam! Fixed in 545a3020d9f419f7a6066043f94579e0b35c7843.
That's it!

apparentlymart: Perhaps we can all agree that there are two "flavors" of command in OpenTofu:

1. Those which run an long-running process that is described using a stream of events emitted gradually over time, like `tofu plan` and `tofu init`. These follow the conventions of [machine-readable UI](https://opentofu.org/docs/internals/machine-readable-ui/) when in `-json` mode.
2. Those which just quietly do some work and then produce a single result at the end, like `tofu output` and `tofu show`. These _don't_ follow "machine-readable UI" and instead produce a command-specific JSON schema that's either directly documented as part of the command itself, or refers to [JSON Output Format](https://opentofu.org/docs/internals/json-format/#state-representation) for cross-cutting concerns like describing the current state.

From this discussion it seems like this was originally written as flavor 1, but has now been rewritten as flavor 2. In which case, I agree with that and am only posting this comment to see if we can turn this into a durable policy we'll follow for other commands in future, rather than just a one-off decision for `tofu state show`.

yottta: I took a note and will propose some documentation later to have it clearly stated in our docs as a policy.

Issue #3858, opened 2026-03-09
RFC Tracker: Improving interactions with provider locking


cam72cam:
This is the [RFC tracking issue](https://github.com/opentofu/opentofu/blob/main/rfc/20241023-rfc-tracking-issues.md) for [Improving interactions with provider locking](https://github.com/opentofu/opentofu/blob/main/rfc/20251027-provider-registry-hashes.md), which was originally discussed in https://github.com/opentofu/opentofu/pull/3434

**Please do not post bug reports or other feedback on the features here!** Instead, please [open a new issue](https://github.com/opentofu/opentofu/issues/new/choose) using our normal Bug Report or Feature Request templates. We'll update this tracking issue to link to other issues that represent work we intend to complete as part of this project.

Tasks:
* [x] Update registry repo to support calculating and storing additional provider metadata 
   - https://github.com/opentofu/registry/pull/3744
* [x] Update registry repo to automatically backfill additional provider metadata
  - https://github.com/opentofu/registry/pull/3751
* [x] Update opentofu's provider registry source to [process and record the additional registry metadata](https://github.com/opentofu/opentofu/blob/main/rfc/20251027-provider-registry-hashes.md#proposed-solution-registry)
  - https://github.com/opentofu/opentofu/pull/3868
* [x] Update opentofu's provider mirror source to [allow additional trust to be given to provider mirrors](https://github.com/opentofu/opentofu/blob/main/rfc/20251027-provider-registry-hashes.md#proposed-solution-mirrors)
  - https://github.com/opentofu/opentofu/pull/3885
* [x] Tweak opentofu's provider installer to [omit registry requests when possible](https://github.com/opentofu/opentofu/blob/main/rfc/20251027-provider-registry-hashes.md#proposed-solution-registry-is-down-but-cache-is-valid)
  - https://github.com/opentofu/opentofu/pull/3906

cam72cam:
Linking comment here: https://github.com/opentofu/opentofu/pull/3873#issuecomment-4055253964

cam72cam:
I've started digging into [omit registry requests when possible](https://github.com/opentofu/opentofu/blob/main/rfc/20251027-provider-registry-hashes.md#proposed-solution-registry-is-down-but-cache-is-valid), but have run into a bit of  a snag.

Our current package installer has logic to double check that the registry still lists a locked package as a valid version, even before checking if it's installed. Given that existing logic, I think we have three options:

[...]

PR #3868, opened 2026-03-11
Utilize new registry platform metadata

cam72cam:
Part of #3858 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/getproviders/package_authentication.go
+++ b/internal/getproviders/package_authentication.go
@@ -371,7 +371,7 @@ func (m matchingChecksumAuthentication) AuthenticatePackage(location PackageLoca
 
 type RegistryPlatformData struct {
 	Hashes      []Hash
-	PackageSize int
+	PackageSize int64
 }
 
 type registryPackageAuthentication struct {
@@ -423,7 +423,7 @@ func (a *registryPackageAuthentication) AuthenticatePackage(localLocation Packag
 	if err != nil {
 		return nil, err
 	}
-	if stat.Size() != int64(platformData.PackageSize) {
+	if stat.Size() != platformData.PackageSize {
 		return nil, fmt.Errorf("registry response indicates a package of size %v, but recieved a package of size %v", platformData.PackageSize, stat.Size())
 	}
 
--- a/internal/getproviders/registry_client.go
+++ b/internal/getproviders/registry_client.go
@@ -228,8 +228,8 @@ func (c *registryClient) PackageMeta(ctx context.Context, provider addrs.Provide
 
 		Packages map[string]struct {
 			Hashes      []string `json:"hashes"`
-			PackageSize int      `json:"package_size"`
-		}
+			PackageSize int64    `json:"package_size"`
+		} `json:"packages"`
 	}
 	var body ResponseBody
 


internal/getproviders/package_authentication.go
```diff
@@ -368,6 +369,100 @@ func (m matchingChecksumAuthentication) AuthenticatePackage(location PackageLoca
...
+
+type registryPackageAuthentication struct {
+	Meta        PackageMeta
+	SHA256Sum   string
+	PackageData map[Platform]RegistryPlatformData
+}
+
+// NewRegistryPackageAuthentication returns a PackageAuthentication
+// implementation that ensures the SHA256SUM (verified elsewhere) is
+// matches the respective registry package data entry. We assume
```
apparentlymart: There's an unfinished sentence here. What do we assume? 🤔 

One possible completion I could imagine is "We assume that another authenticator will check that `sha256sum` is included in the signed checksums for this provider", but that could come under "(verified elsewhere)" so I wonder if you were intending to document something different here.

website/docs/language/files/dependency-lock.mdx
```diff
@@ -112,15 +112,17 @@ provider version.
...
-
-  In this case, the `tofu init` output will include the fingerprint of
-  the key that signed the checksums, with a message like
+- If you install a provider from an origin registry which provides checksums,
+  OpenTofu will treat all of the checksums as valid as long as one checksum matches
+  the installed package. The lock file will therefore include checksums for both
+  the package you installed for your current platform _and_ any other packages
+  that might be available for other platforms. If you wish to restrict this
+  behavior to only providers that are signed with a cryptographic signature,
+  you can set [OPENTOFU_ENFORCE_GPG_VALIDATION](../../cli/plugins/signing.mdx)
```
apparentlymart:
```suggestion
  you can set [`OPENTOFU_ENFORCE_GPG_VALIDATION`](../../cli/plugins/signing.mdx#environment-variables)
```

Monospace just for consistency with our typical style, and a fragment identifier on the URL just because the scope of that page is broader than only the environment variables and so I expect this link is more likely to remain useful if that signing page grows to include additional content in future.

website/docs/internals/provider-registry-protocol.mdx
```diff
@@ -324,6 +340,11 @@ A successful result is a JSON object with the following properties:
 
     * `ascii_armor` (required): an "ascii-armor" encoding of the **public key**
       associated with this GPG key.
+* `packages` (recommended): platform specific package metadata to assist in
+  provider locking.
+  * `hashes` (required): List of valid hashes for the specific provider package.
+  * `package_size` (required): Size in bytes of the specific provider package.
+    Used to help prevent hash collision attacks.
```
apparentlymart, line 347:
```suggestion
* `packages` (recommended): platform specific package metadata to assist in
  provider locking.
  * `hashes` (required): List of valid hashes for the specific provider package.
  * `package_size` (required): Size in bytes of the specific provider package.
    Used to help prevent hash collision attacks.
    
    If `packages` is included then the information for all platforms must be included regardless of which platform was requested, because the included checksums will typically be copied into a dependency lock file for verifying packages installed on other platforms in future. The entry for each platform must include at least a `zh:` hash matching the value that would be returned in the top-level `shasum` property when requesting that platform.
```

internal/getproviders/package_authentication.go
cam72cam: Ah, this was a dangling comment from before (verified elsewhere) was added.

internal/getproviders/registry_client.go
```diff
@@ -220,10 +220,16 @@ func (c *registryClient) PackageMeta(ctx context.Context, provider addrs.Provide
...
+		// Information that should be IDENTICAL between all platform downloads of a given version
 		SHA256SumsURL          string `json:"shasums_url"`
 		SHA256SumsSignatureURL string `json:"shasums_signature_url"`
 
 		SigningKeys SigningKeyList `json:"signing_keys"`
+
+		Packages map[string]struct {
+			Hashes      []string `json:"hashes"`
+			PackageSize int      `json:"package_size"`
+		}
```
Yantrio:
```suggestion
		} `json:"packages"`
```

internal/getproviders/package_authentication.go
```diff
@@ -368,6 +369,100 @@ func (m matchingChecksumAuthentication) AuthenticatePackage(location PackageLoca
...
+	}
+	archivePath, err := filepath.EvalSymlinks(localLocation.String())
+	if err != nil {
+		return nil, err
+	}
+	stat, err := os.Stat(archivePath)
+	if err != nil {
+		return nil, err
+	}
+	if stat.Size() != int64(platformData.PackageSize) {
```
Yantrio: Hmm, what do you think about just making the PackageSize property int64 instead of having to cast here?

internal/getproviders/registry_client.go
cam72cam: Good catch, fixed in 03443cb988

internal/getproviders/package_authentication.go
cam72cam: In practice, I don't think we will hit it.  A 2GB provider is quite a lot.

For safety against overflow attacks though, I've bumped it in 03443cb988.

PR #3871, opened 2026-03-12
configs: New-style core version constraints in `language` block, etc

apparentlymart:
This is for https://github.com/opentofu/opentofu/issues/3300.

Previously we interpreted a `required_version` argument in a `terraform` block as if it were specifying an OpenTofu version constraint, but in practice most modules use this to represent a version constraint for OpenTofu's predecessor instead.

The primary effect of this commit is to introduce a new top-level block type called `language` which describes language and implementation compatibility metadata in a way that intentionally differs from what's used by OpenTofu's predecessor.

This also causes OpenTofu to ignore the `required_version` argument unless it appears in an OpenTofu-specific file with a `.tofu` suffix, and makes OpenTofu completely ignore the language edition and experimental feature opt-in options from OpenTofu's predecessor on the assumption that those could continue to evolve independently of changes in OpenTofu.

We retain support for using `required_versions` in `.tofu` files as a bridge solution for modules that need to remain compatible with OpenTofu versions prior to v1.12. Module authors should keep following the strategy of having both a `versions.tf` and a `versions.tofu` file for now, and wait until the OpenTofu v1.11 series is end-of-life before adopting the new `language` block type.

---

I also took this opportunity to simplify how we handle these parts of the configuration, since the OpenTofu project has no immediate plans to use either multiple language editions or language experiments and so for now we can reduce our handling of those language features to just enough that we'd return reasonable error messages if today's OpenTofu is exposed to a module that was written for a newer version of OpenTofu that extends these language features.

The cross-cutting plumbing for representing the active experiments for a module is still present elsewhere in the codebase so that we could reactivate it later if we need to, but for now that set will always be empty.


apparentlymart:
This is currently implementing the modified proposal from https://github.com/opentofu/opentofu/pull/3870, and so I'll keep it in draft until we've finished discussing that amendment.

It'll also need a changelog entry before merging, which I'll add once we have a final idea of what the functionality is.

**Edit:** Everything mentioned in this comment is now done.

apparentlymart:
One additional note I just realized I forgot to mention in the initial writeup:

A big part of the simplification I did here was to make version compatibility checking be part of the module loading process, rather than being a separate step run afterwards. For most of the system that makes no externally-observable difference whatsoever, but it does cause a small behavior difference for `tofu init`.

The logic for that command was previously written to collect three different sets of diagnostics and then return only the first one that had at least one error, in order of preference where the earlier items are more preferred:

1. Version constraint mismatch
2. Backend configuration invalid
3. Any other general errors from decoding the configuration

Because "version constraint mismatch" is now handled at the same time as "any other general errors from decoding the configuration", and the `configs` package itself implements a rule that it'll hide any other diagnostics for a module if at least one version constraint mismatch is reported, the effective new preference order is:

1. Version constraint mismatch from the root module (since that's where the backend configuration comes from)
2. Any other general errors from decoding the root module (since that's where the backend configuration comes from)
3. Backend configuration invalid
4. Version constraint mismatch from non-root modules
5. Any other general errors from non-root modules

I don't personally think this change is a big deal for OpenTofu. I strongly suspect that the original ordering was motivated by handling a pretty narrow edge-case that was potentially more important for our predecessor than for OpenTofu:

- The root module contains a `cloud` or `backend "remote"` block that refers to a remote workspace that implements the "remote operations" part of HCP Terraform's API.
- The version of OpenTofu CLI running locally is older than the one configured in the remote workspace, and the configuration uses features only available in the newer version of OpenTofu in the remote workspace, _but_ nothing in the configuration specifies an OpenTofu version constraint that would block the older version running locally.
- The operator running the local OpenTofu CLI has forgotten to run `tofu login`, or otherwise configure credentials for accessing the remote API.

In that narrow situation, the previous behavior would have `tofu init` prefer to report the credentials-related error from the backend instead of complaining about the unrecognized elements elsewhere in the configuration. After this change, unrecognized elements elsewhere in the configuration would prevent `tofu init` from even attempting backend configuration. However, the operator would still have eventually needed to deal with other errors in the root module anyway once they've got their credentials configured correctly, since this just about which category of errors gets reported _first_.

Remote operations is a much less commonly used feature in OpenTofu than its predecessor, so I'm arguing that this is a reasonable sacrifice to make in return for not having to separately call the version constraint verification in several other parts of the system, and also makes the `tofu init` logic (marginally) easier to follow and hopefully slightly easier to adapt as we continue working towards modernizing the CLI layer.

internal/configs/parser_config.go
```diff
@@ -328,14 +304,23 @@ var configFileSchema = &hcl.BodySchema{
...
 // a configuration file.
 var terraformBlockSchema = &hcl.BodySchema{
 	Attributes: []hcl.AttributeSchema{
+		// This argument is accepted in any file, but ignored unless it appears
+		// in a file named with a ".tofu" or similar suffix that indicates
+		// it's intended for OpenTofu rather than its predecessor.
 		{Name: "required_version"},
+
+		// The following two are included for compatibility with modules
+		// written by OpenTofu's predecessor, but are ignored when present.
```
cam72cam: Is "but are ignored when present." true? That seems to conflict with my understanding of the backwards compatibility documentation below?

apparentlymart: What I meant by "ignored" here is that I removed all of the code that reacts to the `language` and `experiments` attributes from `loadConfigFileBody` and what was formerly `sniffCoreVersionRequirements` and `sniffActiveExperiments`, so they will be accepted and captured into the `hcl.BodyContent` but then ignored after that. My goal was to avoid OpenTofu failing with an error when they are present, but to avoid making any decisions based on what's assigned to them because we can't predict what an arbitrary future Terraform language edition or language experiment keyword might mean.

Now I look closer at how this appeared in the GitHub diff I guess it's not super clear _as a diff_ because the code that handles the `terraform` block was already designed to ignore both of these anyway, so there's no diff there.

I'm not sure exactly what part of the documentation you're referring to, but if part of seems to imply that OpenTofu is going to act on these two arguments then I'd love to revise it!

cam72cam: Ah, right. I saw that and it just didn't connect in my brain.  I'm ok resolving this as is, but perhaps mention that these fields are instead handled by the `checkVersionRequirements` function and it's more specific schemas in this comment?

apparentlymart: These _specific_ arguments are not handled anywhere in this new implementation.

There are similar arguments in the `language` block which _are_ minimally handled -- `edition` and `experiments` -- but only as simple syntax validation in the `validateLanguageBlock` function.

`checkVersionRequirements` only cares about `required_version`, which is intentionally listed above the comment this thread is attached to because that one is certainly not "ignored" in any defensible sense.

cam72cam: 🤦 I somehow mixed those up in my head, sorry for the confusion!

apparentlymart: Cool, I figured that might be the case. No worries! :grinning:

website/docs/language/settings/index.mdx
```diff
@@ -1,130 +1,128 @@
...
-intended for modules distributed by the same vendor as the associated provider.
+This page is focused on specifying OpenTofu version compatibility and language
+edition settings, but a `terraform` block accepts a number of other unrelated
+settings described elsewhere:
 
-For more information, see [Provider Metadata](../../internals/provider-meta.mdx).
+- `backend` blocks represent [backend configuration](../../language/settings/backends/configuration.mdx).
+- `cloud` blocks represent [cloud configuration](../../language/settings/tf-cloud.mdx).
+- `required_providers` blocks represent [provider requirements](../../language/providers/requirements.mdx).
+- `provider_meta` blocks configure [provider metadata](../../internals/provider-meta.mdx).
```
yottta:
```suggestion
- `backend` blocks represent [backend configuration](./backends/configuration.mdx).
- `cloud` blocks represent [cloud configuration](./tf-cloud.mdx).
- `required_providers` blocks represent [provider requirements](../providers/requirements.mdx).
- `provider_meta` blocks configure [provider metadata](../../internals/provider-meta.mdx).
```

A particular reason not to use the shorter version of paths? 🤔

yottta (approved):
Just a small nitpick.
Did not test it, but it looks great! 🙌

website/docs/language/settings/index.mdx
apparentlymart: I honestly just copied the URLs from the links to the same places that were already in the previous version of this page, without thinking too hard about how they are written.

I'm not sure what our typical preference for this would be. I like the idea of using shorter relative paths because it means we wouldn't need to rewrite these URLs if we moved this whole directory somewhere else later, but I'm not sure if there's some Docusaurus-related reason why these longer paths are useful. 🤔

diofeher: We tested this change in other PRs previously, if I'm not mistaken, and it worked fine ^^ 

I prefer these shorter relative paths as well

apparentlymart: Okay... I'll change this to the shorter paths before this is merged!

I'm going to wait for one more review before merging it though, since this is a pretty significant change.

PR #3873, opened 2026-03-13
Additional hash filter for registry package source

cam72cam:
This handles the scenario for if we introduce a new hashing scheme to the registry response that old versions of OpenTofu may not recognise.

Part of #3858
Follow up to #3868 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/getproviders/package_authentication.go
+++ b/internal/getproviders/package_authentication.go
@@ -448,6 +448,12 @@ func (a *registryPackageAuthentication) AuthenticatePackage(localLocation Packag
 	hashes := HashDispositions{}
 	for _, meta := range a.PackageData {
 		for _, hash := range meta.Hashes {
+			switch hash.Scheme() {
+			case HashScheme1, HashSchemeZip:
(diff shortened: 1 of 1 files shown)


apparentlymart:
This seems to overlap with (and currently have a very similar implementation to) [`getproviders.PreferredHashes`](https://pkg.go.dev/github.com/opentofu/opentofu/internal/getproviders#PreferredHashes).

The doc comment on that function is a little stale and doesn't quite match the current behavior, but the intention behind it is that if new hash schemes were added in future then previous versions would ignore the ones they don't understand while newer versions (with a modified implementation of this function) would potentially ignore older schemes when newer schemes are present, if for example we find that `h1:` is flawed in some way and want to use it only as a last resort.

I wonder if it would make sense to use the same function in the code you added here, for similar reasons. Then we'd be able to just update that one central function if our hashing preferences change in a future version. 🤔

cam72cam:
I was hesitant to use that function, given the docstring. I'm open to a PR that fixes that and updates the comment.

PR #3883, opened 2026-03-16
Improve resource identity rendering in plan output and import hooks

Yantrio:
This will pretty-print identity json in plan comments instead of a minified one-liner. This displays a best effort key/value (eg [name=foo]) in import progress messages instead of dumping the entire entity object and it will also show the json representation of the resource identity in the plan ui too.

Part of #3671

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonformat/plan_test.go
+++ b/internal/command/jsonformat/plan_test.go
@@ -12,8 +12,6 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/mitchellh/colorstring"
-	"github.com/zclconf/go-cty/cty"
-
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/command/jsonformat/differ"
 	"github.com/opentofu/opentofu/internal/command/jsonformat/structured"
@@ -27,6 +25,7 @@ import (
 	"github.com/opentofu/opentofu/internal/states"
 	"github.com/opentofu/opentofu/internal/terminal"
 	"github.com/opentofu/opentofu/internal/tofu"
+	"github.com/zclconf/go-cty/cty"
 )
 
 func TestRenderHuman_EmptyPlan(t *testing.T) {
@@ -158,16 +157,16 @@ Plan: 1 to import, 0 to add, 0 to change, 0 to destroy.
 						ProviderName: "test",
 						Change: jsonplan.Change{
 							Actions: []string{"no-op"},
-							Before: marshalJson(t, map[string]interface{}{
+							Before: marshalJson(t, map[string]any{
 								"id":    "1D5F5E9E-F2E5-401B-9ED5-692A215AC67E",
 								"value": "Hello, World!",
 							}),
-							After: marshalJson(t, map[string]interface{}{
+							After: marshalJson(t, map[string]any{
 								"id":    "1D5F5E9E-F2E5-401B-9ED5-692A215AC67E",
 								"value": "Hello, World!",
 							}),
 							Importing: &jsonplan.Importing{
-								Identity: marshalJson(t, map[string]interface{}{
+								Identity: marshalJson(t, map[string]any{
 									"name": "my-resource",
 								}),
 							},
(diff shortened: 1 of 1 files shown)


internal/command/views/hook_ui.go
```diff
@@ -318,41 +318,25 @@ func (h *UiHook) PostImportState(addr addrs.AbsResourceInstance, imported []prov
...
-			addr, importing.ID,
-		))
-	} else {
-		h.println(fmt.Sprintf(
-			h.view.colorize.Color("[reset][bold]%s: Importing... [identity=%s]"),
-			addr, importing.String(),
-		))
-	}
-
+	h.println(fmt.Sprintf(
```
yottta, line 328: Since nothing failed in the checks, I can safely assume that we have no tests for these. Could you add something to have a more clear understanding of the final output?

Yantrio: Sure, I'll add something now

Yantrio: I just pushed 798d4eb, this should be enough coverage. but if im missing anything please let me know

internal/command/format/object_id.go
```diff
@@ -157,3 +160,48 @@ func ObjectValueIDOrName(obj cty.Value) (k, v string) {
...
 }
+
+// ObjectValueBest is a wrapper around ObjectValueIDOrName that attempts to extract the best possible human-friendly
+// string value for an object by first trying the ID and Name heuristics, and then falling back to any other string attribute if those fail.
+// This is intended to be used in cases where we want to display some sort of human-friendly context about an object,
+// but we don't have a specific expectation about which attribute will be the most useful for that purpose.
+//
+// Just as with the functions it wraps, it may return two empty strings if no suitable attribute can be found for a given object.
+func ObjectValueBest(obj cty.Value) (k, v string) {
+	// First,attempt to fetch it from ir or name attributes, which are the most likely to be unique and human-friendly, respectively.
```
cam72cam: `s/ir/id`

internal/command/format/object_id.go
```diff
@@ -157,3 +160,48 @@ func ObjectValueIDOrName(obj cty.Value) (k, v string) {
...
 	return
 }
+
+// ObjectValueBest is a wrapper around ObjectValueIDOrName that attempts to extract the best possible human-friendly
+// string value for an object by first trying the ID and Name heuristics, and then falling back to any other string attribute if those fail.
+// This is intended to be used in cases where we want to display some sort of human-friendly context about an object,
+// but we don't have a specific expectation about which attribute will be the most useful for that purpose.
+//
+// Just as with the functions it wraps, it may return two empty strings if no suitable attribute can be found for a given object.
+func ObjectValueBest(obj cty.Value) (k, v string) {
```
cam72cam: ObjectValueBestGuess?

internal/command/format/object_id.go
```diff
@@ -157,3 +160,48 @@ func ObjectValueIDOrName(obj cty.Value) (k, v string) {
...
+	if obj.IsNull() || !obj.IsKnown() {
+		return "", ""
+	}
+
+	// Now we know we have _some_ information, we go through the rest of the attributes and
+	// This is a bit of a Hail Mary, but it is better than nothing.
+	attributeTypes := obj.Type().AttributeTypes()
+
+	// Fetch whichver string attribute is first in the list that is not marked as unusable, and return it as the best effort description of this object.
+	// We itearte over the attributes in sorted order to ensure that we get a deterministic result, even if the underlying map iteration order is not deterministic.
```
diofeher:
```suggestion
	// We iterate over the attributes in sorted order to ensure that we get a deterministic result, even if the underlying map iteration order is not deterministic.
```

internal/command/views/hook_ui.go
```diff
@@ -539,3 +523,35 @@ func truncateId(id string, maxLen int) string {
...
+// for use in import progress messages. Returns an empty string if no useful
+// description can be determined.
+func describeImport(id string, identity cty.Value) string {
+	if id != "" {
+		return fmt.Sprintf(" [id=%s]", id)
+	}
+	if k, v := format.ObjectValueBest(identity); k != "" {
+		return fmt.Sprintf(" [%s=%s]", k, v)
+	}
+	return ""
```
cam72cam, line 537: When can this happen?

internal/command/format/object_id.go
```diff
@@ -157,3 +160,48 @@ func ObjectValueIDOrName(obj cty.Value) (k, v string) {
...
+	// If we cant get anything, just return "", "" to indicate that we have no useful information about this object.
+	if obj.IsNull() || !obj.IsKnown() {
+		return "", ""
+	}
+
+	// Now we know we have _some_ information, we go through the rest of the attributes and
+	// This is a bit of a Hail Mary, but it is better than nothing.
+	attributeTypes := obj.Type().AttributeTypes()
+
+	// Fetch whichver string attribute is first in the list that is not marked as unusable, and return it as the best effort description of this object.
```
diofeher:
```suggestion
	// Fetch whichever string attribute is first in the list that is not marked as unusable, and return it as the best effort description of this object.
```

internal/command/views/hook_ui.go
Yantrio: This can happen if ObjectValueBest fails to return a key.

For example, if the resource identity is populated but all items in the resource identity are non-strings. This means that we cannot easily display any of the items.

Or, there are strings but they are ephemeral or sensitive..

diofeher:
Not sure if this is worth it, but I saw there's a `truncateId` function on `stillApplying` being used around this code: https://github.com/opentofu/opentofu/blob/42c144403ffb93f1e5af011abafd84fa665195b8/internal/command/views/hook_ui.go#L495

I'm wondering if it's worth it to wrap it on `ObjectValueBest` to have a truncated ID since this code is for outputting ID/values as well

Yantrio:
>I'm wondering if it's worth it to wrap it on ObjectValueBest to have a truncated ID since this code is for outputting ID/values as well

I don't think its worth it right now, we can investigate and do this at a later date if we need to.

internal/command/jsonformat/plan_test.go
```diff
@@ -355,6 +400,53 @@ OpenTofu will perform the following actions:
...
+			},
+			output: `
+OpenTofu used the selected providers to generate the following execution
+plan. Resource actions are indicated with the following symbols:
+  ~ update in-place (current -> planned)
+
+OpenTofu will perform the following actions:
+
+  # test_resource.resource will be updated in-place
+  # imported using resource identity: {
```
yottta, line 500: Could we add a test with also a more complex object for the identity?
Something that has a nested object, or a map, or something that creates another level of indentation.

Yantrio: Sure, I'll add that now.

Yantrio: Added in 7d141577c38f28e5813c09d214a5a4e123c4beb9

yottta: Thanks. Crazy how much space it takes after rendering such an identity. Hopefully not a lot of resources will use such a big identity object.

yottta (approved):

PR #3885, opened 2026-03-16
`tofu providers mirror`: Add option to trust hashes from a mirror source

cam72cam:
Add `trusted_source` to `network_mirror` cli config.

Write additional discovered hashes as part of `tofu providers mirror`

Part of #3858

## Checklist

### Go checklist

 


website/docs/cli/config/config-file.mdx
```diff
@@ -301,6 +301,28 @@ will be used for each method that it is configured with. For the methods
 with no such configuration, the environment variable will be used instead.
 :::
 
+The `network_mirror` method supports an additional argument
+`trusted_source`. When set to `true`, any hashes the network mirror
+reports will be included in the lockfile, similar to how registry hashes
+are handled.
```
apparentlymart, line 307: The general name `trusted_source` suggests to me that we might want to extend this argument to cover other behaviors that require trust in future. Is that true?

If so, we should probably mention that here so that folks who configure it are prepared for the meaning to broaden in future.

However, for security-related things like this I'd personally favor a more specific option those meaning we will never expand, such as `trust_hashes`. `trust_hashes = true` has the advantage of also being a little clearer on the page to an uninformed reader who encounters it in a CLI Configuration file that was written by one of their coworkers, without them necessarily having to consult the documentation.

internal/command/providers_mirror.go
```diff
@@ -255,6 +260,7 @@ func (c *ProvidersMirrorCommand) Run(rawArgs []string) int {
 					))
 					continue
 				}
+				computedHashes[provider][platform] = result.HashesWithDisposition(func(d *getproviders.HashDisposition) bool { return d.Platform != nil && *d.Platform == platform })
```
apparentlymart: I'm feeling ambivalent about whether filtering by platform here is the best behavior or not. 🤔 

`tofu providers mirror` offers the ability to explicitly list platforms because each additional platform listed means another remote package to download to calculate hashes locally.

But additional hash information from the origin registry comes essentially "for free" here as a result of fetching those packages, so I find myself thinking that we should just save the full set regardless of platform, and then `tofu providers lock`  without any arguments would _typically_ be enough to get the full set of hashes without having to list any additional platforms at all.

Keeping all of the hashes where either `VerifledLocally` or `ReportedByRegistry` is set, regardless of platform, seems consistent with this command's original mission as a way to make sure your lock file contains the "official" checksums even when you've configured things so that `tofu init` will use a mirror, but may be less defensible for other things someone might be using this command for. I'm not sure.

If we decide (on the other hand) that the scope of this command is limited only to checksums calculated locally for the platforms that were explicitly requested, shouldn't this filter also require that `d.VerifiedLocally` is set so that we're only saving hashes we calculated locally, and discarding any that came from the registry? That would be more consistent with the potential position that `tofu providers lock` is focused only on local checksum calculations, if that's the way we'd prefer to frame it.

internal/getproviders/hash.go
```diff
@@ -574,7 +579,17 @@ func MergeHashDisposition(a, b *HashDisposition) *HashDisposition {
...
 	}
 	ret.ReportedByRegistry = a.ReportedByRegistry || b.ReportedByRegistry
+	ret.ReportedByTrustedMirror = a.ReportedByTrustedMirror || b.ReportedByTrustedMirror
 	ret.VerifiedLocally = a.VerifiedLocally || b.VerifiedLocally
+	if a.Platform != nil {
+		ret.Platform = a.Platform
+	}
+	if b.Platform != nil {
+		if ret.Platform != nil && *ret.Platform != *b.Platform {
+			panic("BUG: conflicting platforms")
```
Yantrio: Nitpick: it might be nice to print the platforms here for those who report it so we get a little bit more information. (If it ever gets reported)

internal/command/providers_mirror.go
Yantrio: I think that keeping all hashes regardless of platform feels right to me.

website/docs/cli/config/config-file.mdx
Yantrio: +1 for `trust_hashes`. 

Forconfig like this, I prefer a more granularly scoped, and self-explanatory name. I worry that this naming now may might grow in scope later.

internal/command/providers_mirror.go
cam72cam: Sounds reasonable to me, and reduces the scope of this PR :+1:

cam72cam: In practice, this feels a bit weird:
```json
{
  "archives": {
    "linux_amd64": {
      "hashes": [
        "h1:2GHkrTPEdnKKp8br3GjnGhUihG1qE0vpQiE8fqosbbE=",
        "h1:8bmSGsP2lTMEtLe8BBHXvM6uwaLQlu3cSWg2/fmSuWI=",
        "h1:FmYk21WMiQtragnqrIamTcJ4jJoMN1Vlzrd9sW8GjZM=",
        "h1:Q1NTszPWmiltpevj7SqfEL3GnYby/13i4oKGEWkJCKM=",
        "h1:U6ggw46oEBrMvMz9TrI0e+RS8U2JAmzUkjH2EM87hZw=",
        "h1:Ux/POa6alRr9mnZzlL5tS0Tm4z5rkmsncznkU0frQr4=",
        "h1:X+/UQSKTFeYNkLYvdyOJgQtCwEeueAT2Ci5TPvFm6K4=",
        "h1:s1jlMgF/MJil/OZ6BB5AKCablpwAdVJGj0nfVEUjP6Y=",
        "h1:sHLoGGUC68uEqqYwRyW7VwqSE4boAiiDaSXRrYtk9I0=",
        "h1:yZ6W9xVRg/J+CsYC5zefHGnaygOxVVCk34uidFslHV4=",
        "zh:0ec304b02c2651cb52fec062e82a3b8c720d28f8ec985a7ae4f4086b2b8e0317",
        "zh:11e7bbdc0d7e460f2b15e9daf63d37d8d25c0937b266cbfc108b165ce198a297",
        "zh:245c445029d5a778110e48d1c783e54bef53e3da8bb452abfa18d7017a9a3ad0",
        "zh:7bdfd893487c7e1231680cd50d2d328d993a6f3fb113ee7fa90b1ede6b82a293",
        "zh:88538e326c01db6f8b6011b4ffbcbdea8483fe430ffda7883fb63bca48cfdd11",
        "zh:89b598555f517f09ad5df03007c85e648ccb7e6fe020accaf8a093f36155796f",
        "zh:b4798fb4b1171ee66435212dc633c1836ed38314566c67a2a29d2d426b020a77",
        "zh:d221d5160e397f0c45f4afe87021af39ef987d1be40318e7e35c7ab15f599362",
        "zh:d8e93fb787d4376ff9dba120c7d5070ffb5c760770ede3f8da5afafc644b1596",
        "zh:fa36a9bfe2f1532af36af30ab607f2858aef6ae73754349123fe140f9bef0701"
      ],
      "url": "terraform-provider-tfcoremock_0.5.0_linux_amd64.zip"
    },
    "linux_arm64": {
      "hashes": [
        "h1:2GHkrTPEdnKKp8br3GjnGhUihG1qE0vpQiE8fqosbbE=",
        "h1:8bmSGsP2lTMEtLe8BBHXvM6uwaLQlu3cSWg2/fmSuWI=",
        "h1:FmYk21WMiQtragnqrIamTcJ4jJoMN1Vlzrd9sW8GjZM=",
        "h1:Q1NTszPWmiltpevj7SqfEL3GnYby/13i4oKGEWkJCKM=",
        "h1:U6ggw46oEBrMvMz9TrI0e+RS8U2JAmzUkjH2EM87hZw=",
        "h1:Ux/POa6alRr9mnZzlL5tS0Tm4z5rkmsncznkU0frQr4=",
        "h1:X+/UQSKTFeYNkLYvdyOJgQtCwEeueAT2Ci5TPvFm6K4=",
        "h1:s1jlMgF/MJil/OZ6BB5AKCablpwAdVJGj0nfVEUjP6Y=",
        "h1:sHLoGGUC68uEqqYwRyW7VwqSE4boAiiDaSXRrYtk9I0=",
        "h1:yZ6W9xVRg/J+CsYC5zefHGnaygOxVVCk34uidFslHV4=",
        "zh:0ec304b02c2651cb52fec062e82a3b8c720d28f8ec985a7ae4f4086b2b8e0317",
        "zh:11e7bbdc0d7e460f2b15e9daf63d37d8d25c0937b266cbfc108b165ce198a297",
        "zh:245c445029d5a778110e48d1c783e54bef53e3da8bb452abfa18d7017a9a3ad0",
        "zh:7bdfd893487c7e1231680cd50d2d328d993a6f3fb113ee7fa90b1ede6b82a293",
        "zh:88538e326c01db6f8b6011b4ffbcbdea8483fe430ffda7883fb63bca48cfdd11",
        "zh:89b598555f517f09ad5df03007c85e648ccb7e6fe020accaf8a093f36155796f",
        "zh:b4798fb4b1171ee66435212dc633c1836ed38314566c67a2a29d2d426b020a77",
        "zh:d221d5160e397f0c45f4afe87021af39ef987d1be40318e7e35c7ab15f599362",
        "zh:d8e93fb787d4376ff9dba120c7d5070ffb5c760770ede3f8da5afafc644b1596",
        "zh:fa36a9bfe2f1532af36af30ab607f2858aef6ae73754349123fe140f9bef0701"
      ],
      "url": "terraform-provider-tfcoremock_0.5.0_linux_arm64.zip"
    }
  }
}
```

I keep going back and forth on this.

diofeher: Why do you feel it's a bit weird, @cam72cam ?

internal/command/cliconfig/testdata/provider-installation
```diff
@@ -12,6 +12,7 @@ provider_installation {
     include = ["registry.opentofu.org/*/*"]
     exclude = ["registry.OpenTofu.org/foobar/*"]
     download_retry_count = 2
+    trusted_source = true
```
diofeher: While testing locally, I noticed that we don't have any validation for the arguments used in these blocks. I could use `trusted_source1`, for example, without receiving any errors. I was doing this test to see if `trusted_source` could be used with other arguments in `.tofurc`, like `oci_mirror`. I think we should add some validation around this, maybe in another PR? (but that would be a breaking change).

internal/command/providers_mirror.go
apparentlymart: Just wanted to note here for the record that when we discussed this today I realized that I incorrectly thought that I was commenting on changes to _`tofu providers lock`_, rather than `tofu providers mirror`. Sorry about that!

So I'm retracting this feedback. Filtering by platform makes sense for `tofu providers mirror` because we _want_ to segregate the hashes by platform in the mirror metadata so that a client fetching from this mirror can know which subset of the hashes are intended to match the package given in `"url"`.

---

The question of which subset of hash dispositions we ought to include here might still be relevant.

My initial thought was that we should include only hashes where `ReportedByRegistry` is true because this command knows that it's making a copy of packages from the origin registry and so it's reasonable to expect that all of those packages should be covered by the official set of checksums, but this command currently _effectively_ uses only the `VerifiedLocally` hashes (though it achieves that result in a different way) and so we probably do need to also include the `VerifiedLocally` checksums so that we don't break mirroring from registries that don't yet report `h1:` checksums. `ReportedByRegistry` and `VerifiedLocally` together is not much different than just keeping everything, and what you originally wrote here is a simpler way to get that effect.

internal/command/cliconfig/testdata/provider-installation
apparentlymart: The `cliconfig` package often allows unknown arguments because the CLI Configuration is usually a "global" thing in a user's home directory and they might need to run different versions of OpenTofu under that same user account. It could be annoying if adding this new argument for OpenTofu v1.12 made it harder to run OpenTofu v1.11.

It's definitely a tradeoff, but it seems valid in this case where it's okay for older versions of OpenTofu to silently ignore this new setting and just continue ignoring the hashes reported by network mirrors when building the dependency lock file. :thinking:

cam72cam: I got a similar question from @yhakbar when discussing this last night. It allows TG to set this flag universally instead of having to swap the config for each instance based on version.

diofeher: Makes sense, thanks for the explanation!

website/docs/cli/config/config-file.mdx
cam72cam: I feel that `trust_hashes` is too specific.  If we decided to extend the registry protocol to include support for `package_size`, would we want to also add a `trust_package_size` attribute?

Can you think of a scenario in which we would add more to the protocol and want to have a more fine grained approach for trust?  I feel like a mirror is either trusted or it's not.

internal/command/providers_mirror.go
cam72cam: What do you think of the changes in cde918c25f?  I've also included ReportedByTrustedMirror.

website/docs/cli/config/config-file.mdx
apparentlymart: `package_size` is an interesting case because it's not an authentication mechanism _directly_, but it does support the hashes by making it harder to manufacture collisions with them.

My initial instinct is that if we added `package_size` to the mirror protocol then I'd expect the mirror to report the size of whatever package it is serving, regardless of whether that matches what the origin registry would report, and then we'd verify it whenever present without requiring an additional opt-in.

This does mean that a malicious mirror could try to serve a different-sized package with a manufactured collision to one of the "official" checksums, but that's no worse than the current situation where that protocol has no package size property at all, and so someone wanting to perform that attack could just publish metadata without that new property.

---

Addressing the more general point: I think mirrors are _always_ "trusted" in the sense of "I trust that it will not send me malicious packages", as the documentation warns:

> **Warning**
>
> Don't configure `network_mirror` URLs that you do not trust. Provider mirror servers are subject to TLS certificate checks to verify identity, but a network mirror with a TLS certificate can potentially serve modified copies of upstream providers with malicious content.

This new argument represents an additional level of trust: "I _also_ trust that it will return accurate official hashes, so there's no need to cross-check with the origin registry or calculate all the checksums locally". It's an additional opt-in because it propagates potentially-incorrect information from the mirror directly into the dependency lock file without any local verification, which could therefore cause a maliciously-generated hash to get committed into VCS and propagated even to systems where the malicious mirror isn't configured.

So the name I proposed was intended to communicate that specific additional level of trust, rather than trust _in general_. I think we should be careful about anything that makes it seem like a network mirror configured without this argument is safe to use without having an existing trust relationship with its administrators.

(Note that this kind of argument doesn't apply to `package_size` unless we start also recording that in dependency lock file and other similar places. As long as it's only being used transiently during verification of a specific package downloaded from the same place that made the assertion about the size -- without saving that information to another place for future use -- it's not "additional trust" in this sense.)

internal/getproviders/hash.go
cam72cam: Fixed in 23ea3cea9d

website/docs/cli/config/config-file.mdx
cam72cam: Fair enough, I ended up changing it to `trust_all_hashes` in 32eb1c0892 to be a bit more explicit.  What do you think about that rename?

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ UPGRADE NOTES:
 
     If you run OpenTofu in a context where an environment variable of that name is already set, it may cause OpenTofu to now open a web browser in a different way than previous versions would have. Unsetting that environment variable will restore the previous platform-specific behavior.
 
+- If you are installing providers from the registry (most users), you should expect to see additional `h1:value` provider hashes in your `.terraform.lock.hcl` file.
+
+    We have improved the OpenTofu registry to serve both `zh:value` and `h1:value` hashes, as well as instructing OpenTofu in how to integrate this data into it's existing provider trust chain. Including these additional hashes will reduce friction in cross-platform environments. These and other related changes below should subsume the need to use `tofu providers lock` in most scenarios, simplifying many existing cross-platform workflows. For more information, see the [corresponding RFC](rfc/20251027-provider-registry-hashes.md) and [discussion](https://github.com/opentofu/opentofu/pull/3434)
```
Yantrio:
```suggestion
    We have improved the OpenTofu registry to serve both `zh:value` and `h1:value` hashes, as well as instructing OpenTofu in how to integrate this data into its existing provider trust chain. Including these additional hashes will reduce friction in cross-platform environments. These and other related changes below should subsume the need to use `tofu providers lock` in most scenarios, simplifying many existing cross-platform workflows. For more information, see the [corresponding RFC](rfc/20251027-provider-registry-hashes.md) and [discussion](https://github.com/opentofu/opentofu/pull/3434)
```

internal/command/providers_mirror.go
apparentlymart: I don't think it's actually possible for `ReportedByTrustedMirror` to be set here because `tofu providers mirror` is hard-coded to always use the registry source to ensure that it's mirroring whatever the origin registry reports.

But that also makes it harmless to include and perhaps less likely to get missed if this command later gets an option to mirror from another mirror, which would be an unusual thing to do but someone might still ask for it anyway. :man_shrugging:

internal/command/cliconfig/provider_installation.go
```diff
@@ -209,6 +210,7 @@ func decodeProviderInstallationFromConfig(hclFile *hclast.File) ([]*ProviderInst
 					Include         []string `hcl:"include"`
 					Exclude         []string `hcl:"exclude"`
 					DownloadRetries *int     `hcl:"download_retry_count"`
+					Trusted         *bool    `hcl:"trust_all_hashes"`
```
apparentlymart: It seems like `download_retry_count` was defined as a pointer because leaving it unset has a different meaning than setting it to zero, which `retriesF` below reports using a second return value.

For `trust_all_hashes` it looks like leaving it unset is exactly equivalent to setting it to `false` -- which seems intuitive to me -- so maybe we could simplify by making this just `bool` and letting the HCL decoder leave it set to `false` when the argument isn't set at all. 🤔

internal/getproviders/package_authentication.go
```diff
@@ -247,10 +255,26 @@ func (a packageHashAuthentication) AuthenticatePackage(localLocation PackageLoca
...
 		}
 	}
 
 	if len(hashes) > 0 {
+		if a.TrustAllHashes {
+			// The user has explicitly marked this source as trusted. We will add
+			// all of the reported hashes.
+			//
+			// We use RequiredHashes here as we know they are compatible with this version
+			// of OpenTofu. If PreferredHashes ever changes, we may need to change this loop.
```
apparentlymart, line 268: Someone who is modifying `PreferredHashes` in future presumably won't think to check here to learn that there's something else they might need to update, so maybe we should echo this comment in the body of the `PreferredHashes` function to give that future maintainer a hint that they ought to consider if a change is needed here.

CHANGELOG.md
```diff
@@ -25,6 +30,9 @@ ENHANCEMENTS:
 - Module registries can now specify that package downloads should use the same credentials as the registry's API calls, without needing to configure credentials separately in a `.netrc` file. This approach is helpful when the module packages are served by the registry itself, rather than when the registry just links to an external location such as a GitHub repository. ([#3313](https://github.com/opentofu/opentofu/issues/3313))
 - Provider installation now makes concurrent requests to download provider packages, which may allow `tofu init` to complete faster. ([#2729](https://github.com/opentofu/opentofu/pull/2729))
 - Provider checksum verification and schema loading are now better optimized, including no longer verifying checksums for providers that are present in the local cache but will not be used by a particular command. ([#2730](https://github.com/opentofu/opentofu/pull/2730))
+- `tofu init` now accepts additional provider platform hash data from the registry, which should simplify managing lockfiles in cross-platform environments. ([3868](https://github.com/opentofu/opentofu/pull/3868))
```
```suggestion
- `tofu init` now includes a full set of checksums for all supported platforms when updating a dependency lock file, using additional information now reported by the provider registry. This should remove the need to run `tofu providers lock` in many situations where it was previously required. ([#3868](https://github.com/opentofu/opentofu/pull/3868))
```

My goal with this edit is to try to describe it in a way that matches with how people described the problem to us, so that folks skimming the changelog are more likely to realize this is intended to solve their problem.

---

Also, since we've already done the copyedit/reordering of the changelog for v1.12, maybe let's insert this one alongside the items about `tofu destroy`, `tofu console`, and `tofu login` above so that all of the command-related changes are grouped together.

Subjectively, I think this change is likely to be more impactful than any of the others already listed and so I'd insert it first in that part of the list, just before the existing item about `tofu destroy`.

internal/command/cliconfig/provider_installation.go
cam72cam: Changed in 27a5812b80

internal/getproviders/package_authentication.go
cam72cam: Added a warning to check refs when modifying PreferredHashes in a251f9cc7b

PR #3886, opened 2026-03-17
Add support for Customer Provided Keys and Customer managed keys in AzureRM backend

### LINKED ISSUE #2911: Add support for customer-provided keys in AzureRM backend
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v.1.9.1
on windows_amd64
```

Currently OpenTofu supports client-side encryption. That is great, and offers better security than having a plaintext statefile. 

However, in both the gcs and s3 backends we have the option to either use a Customer-Managed or a Customer-supplied encryption key. These features, especially the latter, even prevent the statefile from being downloaded when the user trying to download the statefile does not provide the encryption key. However, this type of functionality is completely lacking for the AzureRM backend, even though Azure Blob Storage does offer such functionality under the name of 'customer-provided keys'.

Exactly this feature request has been raised earlier for [terraform](https://github.com/hashicorp/terraform/issues/30944)

### Attempted Solutions

The client-side encryption is an improvement, but still feels tricky since the encrypted payload can still easily be downloaded in various ways (also depending on the security settings of the storage account). Additionally, the only ingredient required to decrypt it, seems to be the passphrase. 

### Proposal

To create feature parity with the two other main cloud providers it would be great if:
1. Customer-managed keys (managed in KeyVault) 
2.  Customer-provided encryption keys 

could be supported on the AzureRM backend. One attempt for the customer-managed keys, has already been made in [this PR](https://github.com/hashicorp/terraform/pull/29289) which unfortunately was never merged due to Hashicorp having other priorities. Even though I'm not an OpenTofu developer, it seems a similar approach could be used to implement Customer-provided keys.

The alternative I used for now is the client-side encryption feature. As stated before, it's an improvement over having no encryption at all, but it doesn't offer the same level of server-side security, that some of the other major cloud providers offer (see above). Besides that, using the customer-provided keys feature, is usually easier, than setting up the client-side encryption, as it can usually be done using these steps:
1. Generate an AES-256 encryption key
2. Provide it through an environment variable
3. (Optional) if on an exising project: perform a one-time copy to encrypt the statefile
4. Run `terraform init` once because of the backend changes...

Usually everything works after these steps, and it does not require a variety of code changes, or runs of `terraform apply` which is the case for client-side encryption...

### References

- https://github.com/hashicorp/terraform/issues/30944

Legin-ML:
Resolves #2911
Adds two new optional arguments to the backend, customer provided key, which expects an AES-256 key OR an encryption scope with a customer managed key defined.

## Checklist

### Go checklist

 


Legin-ML:
@lbordowitz I would like to request you to look at this when you're available. Thanks so much!

lbordowitz:
@Legin-ML This looks like it's on the right track! Variables are where they should be, and I like what I'm seeing with the CMK module in the meta-test. Let me know when it's ready to review, and I'll give it a closer look, including double-checking the details of usage in the `azure-sdk-for-go`.

One thing that this PR is currently missing is documentation. You'll want to add the variable to the [azurerm docs](https://github.com/opentofu/opentofu/blob/main/website/docs/language/settings/backends/azurerm.mdx), and you'll probably want to add a line to [the changelog](https://github.com/opentofu/opentofu/blob/main/CHANGELOG.md) specifying this new feature.

By the way: [this PR](https://github.com/opentofu/opentofu/pull/3890) was just merged. You'll probably want to rebase against main and run `go fix` against your work in that package.

Legin-ML:
@lbordowitz I have completed the rebase and run `go fix` .  I have also added notes to both CHANGELOG as well as the website and tested the website running. I think this is ready for review.

benvdh:
@Legin-ML perhaps a new rebase is required (I see some conflicts)
@lbordowitz perhaps you can start the initial review?

cam72cam:
A lot of our team is at KubeCon this week, so reviews may be a bit delayed.  Thank you for your patience!

Legin-ML:
> @Legin-ML perhaps a new rebase is required (I see some conflicts)

Done @benvdh, It was minor changelog stuff, seems like a new feature for azurerm got added.

> A lot of our team is at KubeCon this week, so reviews may be a bit delayed. Thank you for your patience!

I understand, Take your time!

internal/backend/remote-state/azure/backend.go
```diff
@@ -251,6 +252,30 @@ func New(enc encryption.StateEncryption) backend.Backend {
...
+					}
+					return nil, nil
+				},
+				ConflictsWith: []string{"encryption_scope"},
+			},
+
+			"encryption_scope": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "Encryption scope which should be used to encrypt the blob.",
```
lbordowitz: This is going to be user-visible. The end user would think about it as "the state" rather than "the blob", even if that's technically how it's implemented.

I also recommend you reword this. The phrasing "should be used" implies some sort of ideal, or maybe a step that might not happen in certain scenarios. The encryption scope "is used" to encrypt the state.

internal/backend/remote-state/azure/client.go
```diff
@@ -279,7 +294,9 @@ func (c *RemoteClient) Unlock(ctx context.Context, id string) error {
 func (c *RemoteClient) getBlobProperties(ctx context.Context) (blob.GetPropertiesResponse, error) {
 	ctx, ctxCancel := c.getContextWithTimeout(ctx)
 	defer ctxCancel()
-	resp, err := c.blobClient.GetProperties(ctx, &blob.GetPropertiesOptions{AccessConditions: c.leaseAccessCondition()})
+	resp, err := c.blobClient.GetProperties(ctx, &blob.GetPropertiesOptions{
+		AccessConditions: c.leaseAccessCondition(),
+		CPKInfo:          c.cpkInfo})
```
lbordowitz, line 299: This is fine. This comment is just kinda wondering out loud.

It's really weird that `GetPropertiesOptions` for some reason doesn't have `CPKScopeInfo`... I wonder why. Is it because it's a metadata operation? Then why have CPK info in the first place?

internal/backend/remote-state/azure/backend.go
```diff
@@ -285,6 +312,8 @@ func (b *Backend) configure(ctx context.Context) error {
 	b.keyName = data.Get("key").(string)
 	b.snapshot = data.Get("snapshot").(bool)
 	b.timeout = time.Duration(data.Get("timeout_seconds").(int)) * time.Second
+	b.cpkInfo, _ = newCPKInfo(data.Get("customer_provided_key").(string))
```
lbordowitz: I'm going back and forth on whether or not to handle the error. I know it's handled in validation, but I think it would put my mind at ease if it were also handled here, too.

internal/backend/remote-state/azure/meta-test/cmk/version.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "4.35.0"
```
lbordowitz: Since you made this PR, the version of `azurerm` used got upgraded to `4.62.1`.

internal/backend/remote-state/azure/client.go
Legin-ML: It's because I'm pretty sure Azure implicitly uses the scope with which it was created to decrypt the contents. I came to this conclusion, because according to the [docs](https://learn.microsoft.com/en-us/azure/storage/common/customer-managed-keys-overview#blob-storage-operations-that-fail-after-a-key-is-revoked), GetBlobMetadata will fail if the key is revoked.

internal/backend/remote-state/azure/backend.go
Legin-ML: Modifed the text.

internal/backend/remote-state/azure/backend.go
Legin-ML: I was debating this too, but I have added the error handling anyway, better safe than sorry.

internal/backend/remote-state/azure/meta-test/cmk/version.tf
Legin-ML: Updated!

Legin-ML:
Thanks for the detailed review!

> That's pretty bad. A user might lose their state if they remove CPK encryption. I'm not sure how this would properly be handled...

Me too, Apart from having migration instructions to cleanly remove CPK and rewrite state, I'm at a loss for what else can be done. CPK is based on the assumption that clients don't lose their keys, so I think this is pretty on brand for what is expected here, please let me know if you have any ideas!

> A minor, but annoying issue. If I run this in a workspace where I provide an encryption scope that does not exist, we get a big, clunky Azure error instead of a more precise, pointed error directly from OpenTofu

I thought it was explicit and to the point, but if a more consice error is needed, I can provide that as well. I am thinking it would be similar how `notFound()` is in the current codebase.

> It may even be more trouble than it's worth to check the existence of the encryption scope when constructing the backend. Maybe. But probably worth it, anyway.

We would have to work with ARM APIs for it, but only Azure-AD based access methods are allowed to call this API, the alternative would be to test a blob creation with scope and it will fail if the scope does not exist. [Source](https://stackoverflow.com/questions/79296616/how-can-i-obtain-or-create-encryption-scopes-for-an-azure-storage-account-by-usi)

benvdh:
> Thanks for the detailed review!
> 
> > That's pretty bad. A user might lose their state if they remove CPK encryption. I'm not sure how this would properly be handled...
> 
> Me too, Apart from having migration instructions to cleanly remove CPK and rewrite state, I'm at a loss for what else can be done. CPK is based on the assumption that clients don't lose their keys, so I think this is pretty on brand for what is expected here, please let me know if you have any ideas!
> 

@Legin-ML @lbordowitz The behaviour you are seeing is indeed in line with both how AWS and GCP handle the case: You cannot download the encrypted statefile when you do not supply the encryption key. Personally, I see that as a security feature (and it was one of the main reason I wrote the original feature request), because an attacker without the encryption key would not be able to download the encrypted blob (making it less vulnerable to store-now-decrypt-later attacks). 

The [official docs](https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys) confirm this 409:

> When a client creates or updates a blob using a customer-provided key on the request, then subsequent read and write requests for that blob must also provide the key. If the key isn't provided on a request for a blob that has already been encrypted with a customer-provided key, then the request fails with error code 409 (Conflict).

Regarding the scenario where you might want to move away from CPK encryption (and for example replace it with a different security mechanism), you could consider following a protocol like this (which sort of aligns with the [key rotation mechanism in the official docs](https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys#rotate-customer-provided-keys)) :
1. Download the state file (still encrypted) using a `tofu state pull > state.json` or the AZ CLI (still using the current encryption key)
2. Replace the state file on blob storage with an unencrypted version, using the azure CLI or Web portal
3. Change your backend configuration
4. Apply the new encryption mechanism you prefer.

Regarding losing keys and thereby the statefile: that can be a concern, though I think most setups would store the encryption key in the secret store of their CI/CD pipeline, Azure Key Vault (as Microsoft suggests), and probably in a password manager of one or more admins. So that already makes at least 2 to 3 copies. But I think a fair warning for using this feature could be part of the docs, perhaps also describing a procedure like the one above somewhere in the docs.

Legin-ML:
@benvdh I have already added a warning in the docs for CPK. Thanks for the detailed breakdown and the steps to move away from CPK as well!

internal/backend/remote-state/azure/backend.go
```diff
@@ -285,6 +312,15 @@ func (b *Backend) configure(ctx context.Context) error {
 	b.keyName = data.Get("key").(string)
 	b.snapshot = data.Get("snapshot").(bool)
 	b.timeout = time.Duration(data.Get("timeout_seconds").(int)) * time.Second
+	cpkKey := data.Get("customer_provided_key").(string)
+	if cpkKey != "" {
```
lbordowitz: Could you please add a `key == nil -> return nil, nil` case in the `newCPKInfo` instead?

Legin-ML: @lbordowitz Made the change and ran the tests locally with success.

benvdh:
@lbordowitz can you re-review as LeginML made the changes required by your previous review...

Legin-ML:
I have resolved the merge conflict with the `CHANGELOG.md`. Thanks @lbordowitz and @benvdh for the help! Please let me know if anything else is required from my end.

benvdh:
> I have resolved the merge conflict with the `CHANGELOG.md`. Thanks @lbordowitz and @benvdh for the help! Please let me know if anything else is required from my end.

@Legin-ML Glad I could help on the functional side. I also want to thank you for all the hard work on building this functionality into OpenTofu and following up well on the comments you got! Well done 👍🏻!

PR #3892, opened 2026-03-17
flock: `go fix`

apparentlymart:
This is just the result of running the `go fix` modernizers against this package.

It seems that there were some lines with trailing whitespace previously, which also got removed here because `go fix` includes an implicit `go fmt`.

### DIFF
--- a/internal/flock/filesystem_lock_test.go
+++ b/internal/flock/filesystem_lock_test.go
@@ -314,15 +314,12 @@ func TestConcurrentLocking(t *testing.T) {
 
 	// Launch multiple goroutines trying to acquire locks
 	for i := range numGoroutines {
-		wg.Add(1)
-		go func(id int) {
-			defer wg.Done()
-
+		wg.Go(func() {
 			successes := 0
 			for range iterations {
 				f, err := os.OpenFile(lockFile, os.O_RDWR, 0644)
 				if err != nil {
-					t.Errorf("Goroutine %d: Failed to open file: %v", id, err)
+					t.Errorf("Goroutine %d: Failed to open file: %v", i, err)
 					continue
 				}
 
@@ -339,7 +336,7 @@ func TestConcurrentLocking(t *testing.T) {
 				time.Sleep(1 * time.Millisecond)
 			}
 			successCount <- successes
-		}(i)
+		})
 	}
 
 	wg.Wait()


internal/flock/filesystem_lock_test.go
```diff
@@ -313,21 +311,21 @@ func TestConcurrentLocking(t *testing.T) {
...
 	var wg sync.WaitGroup
 	successCount := make(chan int, numGoroutines)
-	
+
 	// Launch multiple goroutines trying to acquire locks
-	for i := 0; i < numGoroutines; i++ {
+	for i := range numGoroutines {
 		wg.Add(1)
 		go func(id int) {
 			defer wg.Done()
```
lbordowitz: Interesting that `go fix` made the change to use `wg.Go` above, but not here.

cam72cam: My guess is because of the `(i)` below.  IIRC that pattern is no longer needed in modern go versions in the first place.

apparentlymart: This is interesting indeed!

My guess is that this happens the content of this function is far more complex than the example above, and so maybe the static analysis can't prove this transformation is safe here. 

However, I'll try to track down exactly what modernizer implements this rule to see if I can explain the difference more precisely.

apparentlymart: The passing of the iterator variable to the function is also a plausible explanation, indeed! (I've not looked at the modernizer to try to prove that either way yet, since I'm on my way back from lunch.)

If that is the explanation then that'd be kinda funny because there's another modernizer that tries to remove unnecessary variable shadows in loops (since modern Go does that automatically for each iteration) but perhaps this particular combination of those two situations is just too complex for the analyser to understand.

Hopefully more interesting information to share once I've found the relevant modernizer code!

apparentlymart: @cam72cam was correct that it's the extra argument to the function that made the difference: [the modernizer requires a function matching the signature `func ()` exactly](https://github.com/golang/tools/blob/ae30076ef44657efd4920d7180a1abbf3e343c0a/go/analysis/passes/modernize/waitgroupgo.go#L97-L102).

The other modernizer that removes unnecessary shadowing of the loop variables is only applicable when the shadowing appears as the very first statement in the loop and must be `x := x` where `x` is one of the iteration variable names. So this situation is too complex for that modernizer to deal with it.

Since I've now gone to the effort of learning how these two modernizers work, I'm going to push a second commit where I'll tweak it enough to make this transform get applied, just so we can have this code following a consistent style across both loops.

Thanks for asking about this so we could learn more about how these modernizers work.

apparentlymart: I've pushed a second commit here now, where I manually removed the argument from the function and just let it capture `i` from the loop directly. I then ran `go fix` again and it was indeed able to rewrite it to use `wg.Go` after that.

I also noticed while I was doing it that as soon as I made this rewrite possible `gopls` offered to apply a localized fix directly inside my editor, which is not something I'd noticed before but could be a good way to notice modernization opportunities in code we're already directly editing, so that there's less need to perform these bulk `go fix` runs.

apparentlymart:
In https://github.com/opentofu/opentofu/pull/3892#pullrequestreview-3963451375 @lbordowitz pointed out that one of the waitgroup loops hadn't been modernized in the same way as the other one, which caused us to learn that that modernizer only works when the goroutine function has the signature `func ()` (no arguments).

Therefore there's now a second commit here where I made a small manual change to remove the function argument and let the the loop variable be captured into the closure automatically, because that's safe in modern Go. After I did that, I ran `go fix` again and it was able to complete the modernization to use `wg.Go` in that case too.

Other than that manual change to the signature of the closure, these changes are all just `go fix` results.

PR #3894, opened 2026-03-17
CHANGELOG: Copyedit in preparation for the v1.12 prerelease period

apparentlymart:
This rewrites and reorders various entries added to the changelog during the v1.12 development period to use consistent terminology and phrasing between similar items and to present the changes in a very subjective ordering by "relevance".

As usual, "relevance" is pretty loosely defined but the general guideline is: configuration-related changes, then CLI-related changes, then changes that affect integrations with other software like providers and backends. Within each of those categories I went mainly by "vibes" based on my impression of how much interest there had been in each of the features or fixes in earlier discussion.

[Rendered version](https://github.com/opentofu/opentofu/blob/changelog-v1.12-copyedit/CHANGELOG.md)


yottta (approved):
Since you reordered these, I crossed check these to be sure that we still have all the entries we had before the rework.
Didn't check the rewording specifically.

PR #3897, opened 2026-03-18
Rework ephemeral resources apply phase

### LINKED ISSUE #3799: Explore possible approaches to alleviate the divergence of ephemeral resources design
> [!NOTE]
> This issue is meant to be a conversation which we would like to result in another (possible) issue describing the actionable item(s) to counteract the current divergence in OpenTofu's ephemeral implementation compared with the predecessor project.
>
> This issue was created as a response to https://github.com/opentofu/opentofu/issues/3781.

## Divergence context
_TL;DR_: OpenTofu diverge from its predecessor because it stores in the plan file some stub information of the planned ephemeral resources to be able to rebuild the apply graph, this way, avoiding building and executing another subgraph during the apply graph build process.

Citing from the [ephemeral RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20250317-ephemeral-resources.md#plan-file):
> During the implementation of the ephemeral resources execution, we had to make a decision on how we can include ephemeral resources
> in the execution graph for the `apply` phase, especially when `tofu` is executed with a plan file.
> There were two immediate options available:
> * Add a new graph transformer (or enhance the expansion one from the apply graph steps) to create a subgraph responsible with the creation of concrete nodes for the ephemeral resources in the configuration.
> * Use the already planned ephemeral resources and store only a stub in the changes to enable the already existing `DiffTransformer` to create concrete graph nodes based on the changelist.
> 
> We went with the second to ensure that the flow is consistent between `tofu apply -auto-aprove` and `tofu plan -out planfile && tofu apply planfile`.
> 
> The information from the ephemeral resources that is stored in the plan file is limited to only the address of the resource and the action (ie: OPEN).
> Anything else related to this type of changes, is trimmed out and must never be stored in the plan file.

## Side effects
As seen in the submitted #3781, when an acceptance test generates a plan file based on a configuration that contains ephemeral resources:
* In the predecessor project, the ephemeral resources are not in the change list from the plan file;
* In OpenTofu, the ephemeral resources appear in the change list from the plan file;

This forces the maintainers of providers, that run their acceptance tests against both tools, to create different assertions based on the tool the acceptance tests are executed against.

## Possible solutions
As discussed with the maintainers, we would like to discuss in-depth the following (and any other proposed) possible solutions to fix this long-term:
1. Do not store the ephemeral information in the `resource_changes` field of the plan, but create another field (eg: `opened_ephemerals`) in the plan that will be OpenTofu specific. 
    * This will keep the current behaviour as it is and will avoid breaking the tools that work directly on the plan content
3. Update `tofu show` command to exclude the ephemeral resources from the changes before printing those.
    * This will work for the situation presented in #3781, because the [testing framework](https://github.com/hashicorp/terraform-plugin-testing/blob/00a2cce7fdd86ad45eda9669fbc2c252cdc353c4/internal/plugintest/working_dir.go#L452) uses the output of the `show` command to read the state.
    * Even though OpenTofu does not recommend such an approach, might still be an issue for other tools out there that work directly on the plan file.

One additional thing that we might want to think and debate about, is the capability of the accepted approach to be backported to v1.11.
For the 2 approaches listed above, both could be backported, with a question mark on the first one. That might not be able if the compatibility promises on the plan file block it.

yottta:
Resolves #3799

This branch reworks how ephemeral resources are handled during the apply phase in OpenTofu. 
With this, we move away from storing ephemeral resource stubs in the plan to expanding and evaluating them dynamically during the apply phase.

Questions:
* Checked also https://github.com/opentofu/opentofu/pull/3209 (ephemerality documentation) and found no place that would use updates. If you think that we should update some parts of it, please let me know in the reviews.

---

This is a quite big PR so I recommend reviewing it commit by commit.
Each commit already has context in the commit message, but here is an overview of the changes:
* 259224d899a4c2c384608fc9069101d171f3b086 - Detach ephemeral resource instance creation based on plan changes
  * It breaks the dependency between ephemeral resource instance creation and plan changes by removing logic from the DiffTransformer.
At this point the ephemeral resources are in a broken state. The subsequent commits will add expansion of these instead.

* 2e8860f0e4c62b238759b7b796ac9a0bdd1c34e0 - Expand for ephemeral resources
  * Adds dynamic expansion logic for ephemeral resources during the apply phase, implementing similar patterns to 
`nodeExpandPlannableResource.DynamicExpand` but tailored for ephemeral resources. 
  * Chose to leave the `nodeExpandApplyableResource.Execute` to allow the existing logic for managed resources and data sources to follow
the current flow.
Added `nodeExpandApplyableResource.DynamicExpand` for the ephemeral resources only.

* 1e2e37497de3469ab45b56d28b850149be272082 - (Bug with preconditions on ephemeral resources) Add test to confirm the issue with missing preconditions on ephemeral
  * During working on testing the expansion, realised that some checks for ephemeral resources
remained in the "unknown" status.
The test added in this commit confirmed the bug.

* df52dac4399ab6b2cf66078224e6d44c798a3e33 - (Bug with preconditions on ephemeral resources) Fix the precondition bug in ephemeral resources
  * Fixes the bug identified in the previous commit where preconditions were not being executed for ephemeral resources, resulting in "unknown" values for checks.

* ba02e50ae6d8fdcd3afb3d294f6718969d8341d3 - **Renewal was not triggered when renewAt was not nil**
  * Thanks to the `TestEphemeralWorkflowAndOutput` test, realised that the renew was not executed at all since [#3710](https://github.com/opentofu/opentofu/pull/3710/changes#diff-d643c4682ab985e4ff40bf48b7ec7523a14dd944e02a6a5e13e9073d2faea45eR159-R163).

* 30d9ef8f024e678733df52682edc635645bb5f91 - Update tests with the new behavior where ephemeral is not opened during apply if not referenced
  * Ephemeral resources are now only opened during apply if they are actually referenced by other resources.
  * Because the expansion happens only during the graph walk, `pruneUnusedNodesTransformer` now removes the 
node because has no other dependencies on it:
    * **Before:** Ephemeral resources were opened during apply regardless of whether they were referenced  
    * **After:** Ephemeral resources are only opened if referenced by other constructs

> [!WARNING]
> From a conceptual point of view, this makes total sense: do not open the ephemeral if its value is not used.
> The concern I have about this is that the opening of it could be implemented in such way to create side-effects on 
> the platform, side effect that the other resources (or the inner-working of the platform) might rely on. 
> 
> Ephemeral resources are not meant for such things. And if the users really would like to do so, can create another
> dummy block with a `depends_on` the ephemeral resource.

* 69a9832d258869357db9556bb767abe9109fbb42 - Remove the change that ephemeral resources save into the plan and update evaluator
  * Removes ephemeral resources from plan changesets and moves to using `workingState` for temporary storage to be able
to have access to their value during evaluating other expressions pointing to their values.

  * To be able to do so, I had to introduce a deferred attribute in the resource change object, attribute that, during evaluation,
is used to return unknown values for the null attributes inside the ephemeral state.

  * The way I plugged the ephemeral value in `evaluationStateData.GetResource`, was to avoid duplicating the logic for 
the repetition information at the end of the `evaluationStateData.GetResource` method.

* b174d3bcf3a537b421e64a7720e5a1249a1f4291 - Remove the Open plan action
  * Cleanup commit that removes the `plans.Open` action and all references to it.
  * This action is no longer needed since:
    1. We don't store the ephemeral resource in plan anymore
    2. We don't generate changes for it, and instead use the plan (that uses no action)

>[!NOTE]
> This reverts the changes from #3776 that are no longer needed.

* 77165326a5a4e111b8effb7ba3e26f6838906e99 - Rework ephemeral workflow test
  * Comprehensive rework of the ephemeral workflow test to properly test dependency chains and deferral behavior across modules.

  _Check the [commit message](https://github.com/opentofu/opentofu/pull/3897/changes/6f3ce7d35f05fe5e3288dec6c8175712bbab3ae9) for more details on this._

* 440a768c0636edec9fabec29b23c558679bd885d - Ensure that there are no changes for ephemeral resources saved into the plan
  * Add error if ephemeral resources are encountered in the plan (writing/reading).

_If there are any other commits after this one, are just for cleanup or to tackle PR reviews_

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/transform_destroy_edge.go
+++ b/internal/tofu/transform_destroy_edge.go
@@ -366,15 +366,10 @@ func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) err
 							return
 						}
 					}
-					// TODO ephemeral - cleanup this after review
-					// if c, ok := n.(*nodeExpandApplyableResource); ok && len(up) == 0 {
-					// 	if c.Addr.Resource.Mode == addrs.EphemeralResourceMode {
-					// 		// TODO ephemeral - do we really want to have the ephemeral nodes executed during apply if they are not referenced?
-					// 		//   or should we just allow this since are referenced?
-					// 		log.Printf("[TRACE] pruneUnusedNodes: expand node for ephemeral not pruned")
-					// 		return
-					// 	}
-					// }
+					// NOTE: Decided here to not add additional logic to skip pruning of the ephemeral resources
+					//  expansion nodes. This will remove any ephemeral resources expansion nodes that are not referenced
+					//  in any other place. Therefore, there is no need for an ephemeral to be opened during apply
+					//  if its value will not be used.
 
 				case GraphNodeProvider:
 					// Only keep providers for evaluation if they have


internal/tofu/evaluate.go
```diff
@@ -1146,6 +1156,40 @@ func (d *evaluationStateData) GetCheckBlock(_ context.Context, addr addrs.Check,
...
+	}
+	// If we would return cty.UnknownVal(ty), then whatever other ephemeral resources using the config
+	// values of the defered instance would be deferred too.
+	// Instead, we want to return as much information as possible for the requested ephemeral resource.
+	// But, during the encoding of the deferred value, the unknown fields are converted to nil and are not converted back
+	// because of https://github.com/opentofu/opentofu/blob/cba3902c0bf20531ee27d6c76e907fa7348b74e6/internal/states/instance_object.go#L116-L118.
+	// Therefore, we use the same function to convert null values to unknowns and keep as much of the configuration
+	// of the resource in the returned value.
+	if v.Deferred {
+		return objchange.PlannedUnknownObject(schema, v.Value).Mark(marks.Ephemeral), nil
```
yottta, line 1187: Note [this](https://github.com/opentofu/opentofu/blob/cba3902c0bf20531ee27d6c76e907fa7348b74e6/internal/states/instance_object.go#L116-L118) from `ResourceInstanceObject.Encode`:
> 	// Our state serialization can't represent unknown values, so we convert
	// them to nulls here. This is lossy, but nobody should be writing unknown
	// values here and expecting to get them out again later.

Because of that, we need to use the schema to set the null values as unknowns where the schema indicates so.

internal/tofu/context_plan.go
```diff
@@ -386,13 +386,10 @@ func (c *Context) refreshOnlyPlan(ctx context.Context, config *configs.Config, p
 	// to refresh only, the set of resource changes should always be empty.
 	// We'll safety-check that here so we can return a clear message about it,
 	// rather than probably just generating confusing output at the UI layer.
-	// Because the ephemeral resources changes in the plan are meant to be used
-	// later to build the apply graph, those shouldn't be counted when we are
-	// doing this check.
-	if changes := plan.Changes.ActionableResources(); len(changes) != 0 {
```
yottta, line 392: This was introduced in https://github.com/opentofu/opentofu/pull/3776 because during refresh, this was showing an error because the plan contained ephemeral changes. Because of that, we added this new `ActionableResources` to return any change that is not ephemeral.

Now, that we don't store the ephemeral resource values in the plan changes, this is no longer needed, so we revert what was introduced in the aforementioned PR.

internal/tofu/transform_destroy_edge.go
```diff
@@ -371,6 +366,15 @@ func (t *pruneUnusedNodesTransformer) Transform(_ context.Context, g *Graph) err
...
 					}
+					// TODO ephemeral - cleanup this after review
+					// if c, ok := n.(*nodeExpandApplyableResource); ok && len(up) == 0 {
+					// 	if c.Addr.Resource.Mode == addrs.EphemeralResourceMode {
+					// 		// TODO ephemeral - do we really want to have the ephemeral nodes executed during apply if they are not referenced?
+					// 		//   or should we just allow this since are referenced?
+					// 		log.Printf("[TRACE] pruneUnusedNodes: expand node for ephemeral not pruned")
+					// 		return
+					// 	}
+					// }
```
yottta: Before this PR, during the apply graph building, DiffTransformer added the ephemeral resources as instance nodes and linking those to the expansion nodes created earlier, during ConfigTransformer.
Now, we only add the expansion nodes during the ConfigTransformer and nothing else in DiffTransformer, since there are no more changes in the plan for the transformer to act on.

Check also the removed comment from the DiffTransformer:
> 	// Ensure that the ephemeral resource instance node connects to
>  // the resource node. This is needed to ensure that the ephemeral
> // expansion node will not get pruned due to having no connections

Because of that, if there is no other dependency on the ephemeral resource, the node gets pruned here.

After some offline conversations, we agreed that from a conceptual POV, this is expected: we shouldn't open again an ephemeral that has no usage.

There is a warning in the description of this PR about the concerns around this.

WDYT? Should we allow it to work like this? Or should we integrate a logic similar to the one above? (I have other ideas in mind on how to do it, but the end result is the same: do not prune the expansion node for ephemeral resource).

internal/command/e2etest/testdata/ephemeral-workflow/main.tf
```diff
@@ -57,24 +57,19 @@ resource "simple_resource" "test_res" {
   // NOTE: value_wo cannot be used in a provisioner because it is returned as null by the provider so the interpolation fails
 }
 
-data "simple_resource" "test_data2" {
```
yottta, line 60: Removed this data source since this was meant to test the ephemeral usage in a precondition, but that has been moved to the child module of this configuration.

internal/command/e2etest/primary_test.go
```diff
@@ -254,7 +254,6 @@ func TestPrimaryChdirOption(t *testing.T) {
 func TestEphemeralWorkflowAndOutput(t *testing.T) {
 	t.Parallel()
 
-	skipIfCannotAccessNetwork(t)
```
yottta, line 257: Considering the issue introduced recently with the renew process, I would like to have this run in the pipeline, all the time.

internal/command/e2etest/primary_test.go
```diff
@@ -437,15 +436,17 @@ Changes to Outputs:
...
 				outputEntry{[]string{"simple_resource.test_res_second_provider: Creating..."}, true},
 				outputEntry{[]string{"simple_resource.test_res_second_provider: Creation complete after"}, true},
-				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[0]: Renewing..."}, false},
-				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[0]: Renew complete after"}, false},
-				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[1]: Renewing..."}, false},
-				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[1]: Renew complete after"}, false},
+				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[0]: Renewing..."}, true},
+				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[0]: Renew complete after"}, true},
+				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[1]: Renewing..."}, true},
+				outputEntry{[]string{"ephemeral.simple_resource.test_ephemeral[1]: Renew complete after"}, true},
```
yottta, line 449: This now requires to have the renew updates in the logs.
Technically speaking, this is safe and stable to do so.
The only thing that could create some flakiness (especially in the CI) is that the renew might not trigger in some runs. If so, we can adjust the logic around this in our simple-providers.

internal/plans/changes.go
```diff
@@ -73,23 +66,6 @@ func (c *Changes) Empty() bool {
 	return true
 }
 
-// ActionableResources returns all the [Changes.Resources] that are changes that would actually
-// update the resources.
-// This method's main purpose is to exclude from [Changes.Resources] the changes that are
-// in the plan strictly for building the graph and are not going to change the resource state.
-// In case of [Open] actions, these are needed to build the required ephemeral nodes
-// in [DiffTransformer].
-func (c *Changes) ActionableResources() []*ResourceInstanceChangeSrc {
```
yottta, line 82: Reverts the changes added in https://github.com/opentofu/opentofu/pull/3776

internal/provider-simple-v6/provider.go
```diff
@@ -164,8 +164,8 @@ func (s simple) PlanResourceChange(_ context.Context, req providers.PlanResource
 	}
 
 	m := req.ProposedNewState.AsValueMap()
-	_, ok := m["id"]
-	if !ok {
+	idVal, ok := m["id"]
+	if !ok || idVal.IsNull() {
```
yottta, line 168: This was a bug that gave me big time headaches. `id` is marked as computed. Therefore, during planning, this should be returned back as unknown. It was returned as null and that created some weird behaviour during `TestEphemeralWorkflowAndOutput`.

_The same fix for the v5 simple-provider_

internal/tofu/evaluate.go
```diff
@@ -841,6 +842,12 @@ func (d *evaluationStateData) GetResource(ctx context.Context, addr addrs.Resour
 
 		instAddr := addr.Instance(key).Absolute(d.ModulePath)
 
+		if instAddr.Resource.Resource.Mode == addrs.EphemeralResourceMode {
```
yottta, line 845: I plugged this in here to make use of the rest of the logic from this method.
I recommend to you all to check the whole function and let me know if you would like to 
have the whole ephemeral evaluation logic in one method together with the rest of the logic in here.

internal/tofu/evaluate_test.go
```diff
@@ -751,14 +721,37 @@ func TestEvaluatorGetResource_Ephemeral(t *testing.T) {
...
 		},
-		"with object ready state and no changes": {
-			plans.BuildChanges(func(sync *plans.ChangesSync) {}).SyncWrapper(),
+		"with defered state": {
+			states.BuildState(func(state *states.SyncState) {
+				state.SetResourceInstanceCurrent(
+					rc.Addr().Instance(addrs.NoKey).Absolute(addrs.RootModuleInstance),
+					&states.ResourceInstanceObjectSrc{
+						Status: states.ObjectPlanned,
+						// NOTE: during decoding, null values will be converted to unknown
```
yottta, line 730: This comment is what actually matters here. We want to be sure that null values gets converted correctly into unknown values.

internal/tofu/transform_destroy_edge.go
cam72cam: I think it is safe to prune the unused ephemeral.  We can leave a comment here describing the conversations we have had as a team and the conclusion we came to.

yottta: Changed in 718810514cbfa8d6eafd92ef2898a7cd76974d1c

PR #3899, opened 2026-03-18
Added deprecation notice for winrm provisioner connections

### LINKED ISSUE #3825: Produce warning when `type = "winrm"` is used in a `connection` block
This is the first step for eventually removing WinRM support. Refer to the parent issue for more information.

Yantrio:
Resolves #3825 

## Checklist

### Go checklist

 

### DIFF
--- a/website/docs/language/resources/provisioners/connection.mdx
+++ b/website/docs/language/resources/provisioners/connection.mdx
@@ -1,13 +1,13 @@
 ---
 description: >-
-  The connection block allows you to manage provisioner connection defaults for
-  SSH and WinRM.
+  The connection block describes connection settings for the remote-exec and file provisioners.
 ---
 
 # Provisioner Connection Settings
 
-Most provisioners require access to the remote resource via SSH or WinRM and
-expect a nested `connection` block with details about how to connect.
+The `remote-exec` and `file` provisioners access a remote system using the
+Secure Shell (SSH) protocol. These provisioners must therefore be used with
+a `connection` block that describes how to connect.
 
 :::warning Important
 Use provisioners as a last resort. There are better alternatives for most situations. Refer to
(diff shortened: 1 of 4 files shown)


internal/tofu/node_resource_validate.go
```diff
@@ -158,9 +158,42 @@ func (n *NodeValidatableResource) validateProvisioner(ctx context.Context, evalC
 		// configuration keys that are not valid for *any* communicator, catching
 		// typos early rather than waiting until we actually try to run one of
 		// the resource's provisioners.
-		_, _, connDiags := n.evaluateBlock(ctx, evalCtx, p.Connection.Config, shared.ConnectionBlockSupersetSchema)
+		connVal, _, connDiags := n.evaluateBlock(ctx, evalCtx, p.Connection.Config, shared.ConnectionBlockSupersetSchema)
 		diags = diags.Append(connDiags)
+
+		if connVal == cty.NilVal {
```
cam72cam: `|| connDiags.HasErrors()`?

cam72cam: Is it possible for connVal to be unknown here?

Yantrio: I think it's better to be safe than sorry and do a small guard here. We don't always know how the evalcontext is going to change over time. Today this cannot be NilVal as far as I can see, but I usually prefer having these kinds of guards in here just incase.

If you feel strongly about this im happy to remove the nilval check. I will add the connDiags error check though, you're right that I missed that

cam72cam: I'm ok with keeping the Nil check, but wanted to ask if we should also check for unknowns here.

cam72cam:
CHANGELOG.md entry?

internal/tofu/node_resource_validate.go
apparentlymart: Assuming this function is doing what its name suggests it is doing -- using the HCL API to decode the content of a block body -- the object representing the block would not be able to be unknown itself. It would only be possible for the nested attribute values in the object to be unknown.

(I did not verify that's actually what it's doing.)

cam72cam:
Also realized that we are missing website changes for the deprecation

apparentlymart:
I've added a changelog entry and a reorganization of the relevant documentation to this PR in the hope that we can merge it soon, so we can get things finalized for the imminent v1.12.0 feature freeze.

PR #3906, opened 2026-03-19
Reduce registry requests when global provider cache is enabled

cam72cam:
Where possible, we skip making any registry requests when the global provider cache is enabled and correctly populated.

The major tradeoff of this change is that we no-longer detect when versions have been redacted from the registry. This is a very uncommon occurance and is probably not worth the effort in this particular scenario. A debug message has been added to the log to help trace if this ever is a problem.

Init `-upgrade` still works as expected and fetches the full list of versions from the registry.

This is the last part of #3858

## Checklist

### Go checklist

 

### DIFF
--- a/internal/providercache/installer_test.go
+++ b/internal/providercache/installer_test.go
@@ -37,14 +37,15 @@ func TestEnsureProviderVersions(t *testing.T) {
 	// permutations that it ends up being more concise to express them as
 	// normal code.
 	type Test struct {
-		Source     getproviders.Source
-		Prepare    func(*testing.T, *Installer, *Dir)
-		LockFile   string
-		Reqs       getproviders.Requirements
-		Mode       InstallMode
-		Check      func(*testing.T, *Dir, *depsfile.Locks)
-		WantErr    string
-		WantEvents func(*Installer, *Dir) map[addrs.Provider][]*testInstallerEventLogItem
+		Source      getproviders.Source
+		Prepare     func(*testing.T, *Installer, *Dir)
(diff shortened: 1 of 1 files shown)


internal/providercache/installer.go
```diff
@@ -626,6 +639,21 @@ func (i *Installer) ensureProviderVersionInDirectory(
 				return nil, lock.AllHashes(), nil
 			}
 		}
+
+		// If the global cache already contains the installed provider and it matches existing hashes, install it.
+		// Don't add any additional hashes to the lockfile. This allows us to prevent spurious registry requests
+		// when using the global provider cache
```
apparentlymart: This comment says that we'll "install it" if it's already in the cache, but it doesn't seem like that agrees with the code below which I understand more like "if we're installing to the global cache and the package is already there, do nothing".

Am I understanding that correctly, or am I missing something?

---

(It's unfortunate that this function has a signature that appears like it's generic over installing to either the local or global cache directory but then has quite different behavior based on which one was selected, but that's certainly not a _new_ problem with this code that we ought to fix as part of this PR, just something that made me feel unsure about whether I was understanding this correctly today...)

cam72cam: Ah, I wrote this comment before I fully implemented the function.  This actually just marks it as already installed and let's the linking action in the caller happen.  I'll update the comment.

cam72cam:
I think I found a way to alter the existing tests to cover at least part of this scenario, let me give it a try.

Issue #3877, opened 2026-03-13
Upgrade dependencies in preparation for v1.12.0


apparentlymart:
Our usual policy is to upgrade as many dependencies as possible to latest before we begin a new minor release series, because that increases the chance that we'll be able to selectively adopt only patch releases if any security advisories or other critical bugs are discovered during the series.

What follows is a snapshot of all of the possible upgrades we could run at the time of opening this issue. We do not necessarily need to upgrade _all_ of these in order for this issue to be "done", and we also might not select exactly the versions shown here if e.g. the latest available release contains something problematic for OpenTofu. This is just an overview to show roughly how many potential upgrades we have.

The list is ordered topologically so that we can more easily upgrade one module at a time, or notice collections of modules that are worth upgrading together to amortize the testing and review cost if the importers are similar and the upstream changeset is small. Links to changesets in GitHub repositories are included where possible, but that only really works properly for the simple case where a GitHub repository contains only a single Go module.

Note that a a number of the available upgrades are related to functionality of state storage backends or state encryption integrations which are not covered by our normal unit test runs, and so anyone upgrading those dependencies must manually run their acceptance tests. For example, upgrading the AWS SDK requires running the `s3` backend and AWS KMS key provider acceptance tests.

- **`github.com/Azure/azure-sdk-for-go/sdk/azcore`**
    - [x] `go get github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.21.0` (from `v1.20.0`)
- **`github.com/Azure/azure-sdk-for-go/sdk/storage/azblob`**
    - [x] `go get github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.6.4` (from `v1.6.3`)
- **`github.com/ProtonMail/go-crypto`**
    - [x] `go get github.com/ProtonMail/go-crypto@v1.4.1` ([from `v1.4.0`](https://github.com/ProtonMail/go-crypto/compare/v1.4.0...v1.4.1))
- **`github.com/aws/smithy-go`**
    - [x] `go get github.com/aws/smithy-go@v1.24.2` ([from `v1.24.0`](https://github.com/aws/smithy-go/compare/v1.24.0...v1.24.2))
- **`github.com/bmatcuk/doublestar/v4`**
    - [x] `go get github.com/bmatcuk/doublestar/v4@v4.10.0` ([from `v4.9.1`](https://github.com/bmatcuk/doublestar/compare/v4.9.1...v4.10.0))
[...]

PR #3908, opened 2026-03-19
go.mod: Upgrade protobuf and grpc libraries

apparentlymart:
This is a routine upgrade for https://github.com/opentofu/opentofu/issues/3877.

OpenTofu directly uses these libraries as part of implementing the provider plugin protocol and plan file format, but they are also used indirectly by the libraries we use to interact with OpenTelemetry collectors and some Google APIs and so the impact of upgrading these is always quite tricky to evaluate.

However, I've reviewed the changes between these releases and it seems like there are no behavior changes that would be significant for the use-cases OpenTofu cares about. The gRPC library in particular has had a lot of commits since we last upgraded it, but many of them are performance improvements in the implementation details rather than significant behavior changes.

Detailed changes for each library:

- [`google.golang.org/grpc`](https://github.com/grpc/grpc-go/compare/v1.76.0...v1.79.3)
- [`google.golang.org/protobuf`](https://github.com/protocolbuffers/protobuf-go/compare/v1.36.10...v1.36.11)


apparentlymart:
One thing I do want to note here, mainly just to add some keywords that would come up if we try to search for this in response to a bug report in future, is that the gRPC library previously supported an environment variable for disabling [ALPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation) at runtime: `GRPC_ENFORCE_ALPN_ENABLED`.

We never advertised OpenTofu as supporting that environment variable, and for our interactions with provider plugins the server side of all commonly-used plugins is the same gRPC implementation that is ALPN-compatible, so during my review I classified this as not significant for what OpenTofu cares about.

If we _do_ come to learn that someone had for some reason been quietly depending on this without telling us, then hopefully we'll easily find this comment using GitHub issue search and know what happened!

PR #3935, opened 2026-03-24
Added complementary output message to tofu plan when -target is in effect

baa-ableton:
Please note since the title of the issue has not been updated, that this PR applies the compromise solution offered by the maintainer

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonformat/plan.go
+++ b/internal/command/jsonformat/plan.go
@@ -184,6 +184,12 @@ func (plan Plan) renderHuman(renderer Renderer, mode plans.Mode, opts ...plans.Q
 					"OpenTofu has compared your real infrastructure against your configuration and found no differences, so no changes are needed.",
 					renderer.Streams.Stdout.Columns(),
 				))
+				if checkOpts(plans.Targeted) {
+					renderer.Streams.Println(format.WordWrap(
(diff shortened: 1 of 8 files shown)


AndrewCharlesHay (commented):
Implementation looks correct and matches the compromise @apparentlymart described in the issue. A couple of things worth noting:

**Missing third heuristic condition**

The issue comment specified three conditions before showing the hint: (1) at least one `-target` address, (2) no changes, (3) at least one resource block in configuration. The PR implements 1 and 2 but not 3. Without it, an empty configuration run with `-target` would also show the message. Edge case, but worth a follow-up at minimum.

**Message text**

The approach here (appending a `Note:` paragraph after the standard "No changes." message) differs slightly from the maintainer's proposal, which was to modify the message itself. The additive approach is arguably less disruptive so probably fine — just flagging in case it should be aligned with the issue.

**Changelog**

The bot reminded about this and it's not in the diff.

Otherwise the implementation is clean — `Targeted` quality added correctly, `operation.go` sets it from `TargetAddrs`, and both positive/negative test cases are covered.

baa-ableton:
I made a mess of this branch while rebasing. I simply cherry picked the changes I wanted into a new PR:
https://github.com/opentofu/opentofu/pull/3943

PR #3941, opened 2026-03-25
Unify patterns across the refactored commands

yottta:
Part of #3765 
While migrating all the commands to the `views` and `arguments` packages, in the end, some patterns emerged. 

This PR contains one commit per command. You can review it commit-by-commit for a more isolated check.

The fixes that this PR tackle are mostly summarised in the description of #3765, but here's a more detailed breakdown:
* When calling `NewStateLocker()` and `Meta.Operation()`, we need to use the `arguments` created `arguments.ViewOptions` object and not create a new one.
* If a command supports and is ran with the `-json` flag, it needs to exit with code 1 and not `cli.RunResultHelp` if the arguments parsing fails. Otherwise, might print the diagnostics in machine readable format and it will print the help of the command in human format.
* When `-json` is not used, if arguments parsing fails, we should return `cli.RunResultHelp`. This is not changing the final exit code of the command, because when the [cli library encounters](https://github.com/mitchellh/cli/blob/e73bd003a976863797bccc7d5630e83969f20bb8/cli.go#L263-L266) the `cli.RunResultHelp` exit code from a command, it prints the help text of the command and returns exit code 1, which is in line with the previous behavior.
* Add comments to all the struct fields in the `arguments` package
* Removes the usage of `warnln` and `errorln` added in [#3764](https://github.com/opentofu/opentofu/pull/3764/changes#diff-3245de91c1fe1707374ceee15a6a17693ceb15bb1a3569a6c19f63c0d16bfc74)
  * This resolves [this conversation](https://github.com/opentofu/opentofu/pull/3764#discussion_r2850215802)
* Get rid of `diags.Append(err)` calls and replace those with `tfdiags.Sourceless` diagnostics instead
* Add `-json`/`-json-into` flags into the help text of the commands that support those.
* Update documentation for all commands refactored by this issue.
  * This also adds a new section about the `Raw JSON output` in the machine-readable UI documentation. This solves the conversation had [here](https://github.com/opentofu/opentofu/pull/3865#discussion_r2966527325).

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/metadata_functions.go
+++ b/internal/command/arguments/metadata_functions.go
@@ -39,6 +39,10 @@ func ParseMetadataFunctions(args []string) (*MetadataFunctions, func(), tfdiags.
 		return arguments, closer, diags
 	}
 
+	// The 'metadata functions' command just forces the user to use the `-json` flag but any of the diagnostics should
+	// be printed as human format. This makes it clear that the success output of this command will be in json and
+	// that it needs to be processed accordingly.
+	// The print of the functions will be in JSON all the time.
 	if arguments.ViewOptions.ViewType != ViewJSON {
 		diags = diags.Append(tfdiags.Sourceless(
 			tfdiags.Error,
@@ -46,10 +50,7 @@ func ParseMetadataFunctions(args []string) (*MetadataFunctions, func(), tfdiags.
 			"The `tofu metadata functions` command requires the `-json` flag.",
 		))
 	}
-	// The 'metadata functions' command just forces the user to use the `-json` flag but any of the diagnostics should
-	// be printed as human format. This is make it clear that the success output of this command will be in json and
-	// that it needs to be processed accordingly.
-	// The print of the functions will be in JSON all the time.
 	arguments.ViewOptions.ViewType = ViewHuman
+	
 	return arguments, closer, diags
 }


internal/command/arguments/providers_schema.go
```diff
@@ -40,15 +40,6 @@ func ParseProvidersSchema(args []string) (*ProvidersSchema, func(), tfdiags.Diag
 		))
 	}
 
-	args = cmdFlags.Args()
-	if len(args) > 0 {
```
cam72cam, line 44: Why was this removed?

internal/command/views/console.go
```diff
@@ -76,53 +69,8 @@ func (v *ConsoleHuman) Diagnostics(diags tfdiags.Diagnostics) {
...
-  -json-into=out.json    All the evaluation results returned back to the user
-                         are streamed in json format in the given file.
-
-`
-	v.view.errorln(helpText)
-}
-
 func (v *ConsoleHuman) UnsupportedLocalOp() {
-	v.view.errorln(errUnsupportedLocalOp)
+	v.Diagnostics(tfdiags.Diagnostics{diagUnsupportedLocalOp})
```
cam72cam, line 73: Why not move these diags back to the command itself as part of standard error handling?

internal/command/arguments/providers_schema.go
yottta: Because [before the refactor](https://github.com/opentofu/opentofu/blob/6280691025e9e0cb331faee9b84d27658d8b18f5/internal/command/providers_schema.go#L41-L61), it was not checking the number of arguments.
I missed this during reviewing https://github.com/opentofu/opentofu/pull/3822.
Even though the validation makes sense, I don't want to change the general behavior of the command.

internal/command/views/console.go
yottta: I was thinking to do it, but this particular one is used in multiple commands so I wanted to keep it here.
If you think that would be clearer to move it back to the commands package, I can do it.

internal/command/arguments/providers_schema.go
cam72cam: Let's at least open an issue so we don't forget this.

internal/command/views/console.go
cam72cam: That's reasonable. I don't have a strong opinion on this pattern in the grand scheme of things.

internal/command/arguments/providers_schema.go
yottta: Doesn't really worth an issue. After some offline exchange, I agree that since this is not documented, people should not send random arguments that are not used. 
Reinstated in 2c6484730ef146b088e9f3b3052061f0830a37c1

PR #3943, opened 2026-03-25
Added to tofu plan a complementary output when -target flag is provided

### LINKED ISSUE #3729: If the target supplied by --target does not exist, opentofu exits 0
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.4
on darwin_arm64
```

```hcl
resource "null_resource" "real_resource" {
  ...
}
```

### Debug Output

N/A

### Expected Behavior

I expect OpenTofu to fail if any resource specified in --target does not exist and take no action at all.

### Actual Behavior

OpenTofu does nothing and exits 0.

### Steps to Reproduce

tofu apply --target null_resource.fake_resource

### Additional Context

I know behavior changes like this are always controversial.

We have a use care where we specify a very large number of targets. 

If any of these do not exist, something has gone wrong and tofu should not continue. I can't imagine a scenario when this isn't the case. I'm sure someone has some strange edge case they have built off this behavior though (threads like this always surface a few).

### References

_No response_

baa-ableton:
Resolves #3729 

Copy of PR 3935 since I rebased poorly and made a mess out of that branch :)

Please note since the title of the issue has not been updated, that this PR applies the compromise solution offered by the maintainer

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/jsonformat/plan_test.go
+++ b/internal/command/jsonformat/plan_test.go
@@ -93,7 +93,7 @@ func TestRenderHuman_NoChanges_Targeted(t *testing.T) {
 	plan.renderHuman(renderer, plans.NormalMode, plans.NoChanges, plans.Targeted)
 
 	got := done(t).Stdout()
-	if !strings.Contains(got, "-target") {
+	if !strings.Contains(got, "You used the -target option") {
 		t.Errorf("expected output to contain a -target hint when Targeted quality is set, got:\n%s", got)
 	}
 }
@@ -108,7 +108,7 @@ func TestRenderHuman_NoChanges_NotTargeted(t *testing.T) {
 	plan.renderHuman(renderer, plans.NormalMode, plans.NoChanges)
 
 	got := done(t).Stdout()
-	if strings.Contains(got, "-target") {
+	if strings.Contains(got, "You used the -target option") {
 		t.Errorf("expected output to NOT contain a -target hint when Targeted quality is not set, got:\n%s", got)
 	}
 }
--- a/internal/command/jsonformat/plan_test.go
+++ b/internal/command/jsonformat/plan_test.go
@@ -8,6 +8,7 @@ package jsonformat
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -79,6 +80,36 @@ found no differences, so no changes are needed.
 	got := done(t).Stdout()
 	if diff := cmp.Diff(want, got); len(diff) > 0 {
 		t.Errorf("unexpected output\ngot:\n%s\nwant:\n%s\ndiff:\n%s", got, want, diff)
+	}
+}
+
+func TestRenderHuman_NoChanges_Targeted(t *testing.T) {
+	color := &colorstring.Colorize{Colors: colorstring.DefaultColors, Disable: true}
+	streams, done := terminal.StreamsForTesting(t)
+
+	plan := Plan{}
+
+	renderer := Renderer{Colorize: color, Streams: streams}
+	plan.renderHuman(renderer, plans.NormalMode, plans.NoChanges, plans.Targeted)
+
+	got := done(t).Stdout()
+	if !strings.Contains(got, "-target") {
+		t.Errorf("expected output to contain a -target hint when Targeted quality is set, got:\n%s", got)
+	}
+}
+
+func TestRenderHuman_NoChanges_NotTargeted(t *testing.T) {
+	color := &colorstring.Colorize{Colors: colorstring.DefaultColors, Disable: true}
+	streams, done := terminal.StreamsForTesting(t)
+
+	plan := Plan{}
+
+	renderer := Renderer{Colorize: color, Streams: streams}
+	plan.renderHuman(renderer, plans.NormalMode, plans.NoChanges)
+
+	got := done(t).Stdout()
+	if strings.Contains(got, "-target") {
+		t.Errorf("expected output to NOT contain a -target hint when Targeted quality is not set, got:\n%s", got)
 	}
 }
 
--- a/internal/command/jsonformat/plan.go
+++ b/internal/command/jsonformat/plan.go
@@ -184,6 +184,12 @@ func (plan Plan) renderHuman(renderer Renderer, mode plans.Mode, opts ...plans.Q
 					"OpenTofu has compared your real infrastructure against your configuration and found no differences, so no changes are needed.",
 					renderer.Streams.Stdout.Columns(),
 				))
+				if checkOpts(plans.Targeted) {
+					renderer.Streams.Println(format.WordWrap(
+						"\nNote: You used the -target option, so OpenTofu only checked resources matching those targets and may have skipped others. If you expected changes to appear, verify that the addresses passed to -target are correct.",
+						renderer.Streams.Stdout.Columns(),
+					))
+				}
 			}
 		}
 	}
--- a/internal/command/views/operation.go
+++ b/internal/command/views/operation.go
@@ -188,6 +188,9 @@ func (v *OperationHuman) Plan(plan *plans.Plan, schemas *tofu.Schemas) {
 	if plan.Errored {
 		opts = append(opts, plans.Errored)
 	}
+	if len(plan.TargetAddrs) > 0 {
+		opts = append(opts, plans.Targeted)
+	}
 
 	renderer.RenderHumanPlan(jplan, plan.UIMode, opts...)
 }
--- a/internal/plans/quality.go
+++ b/internal/plans/quality.go
@@ -19,4 +19,9 @@ const (
 	// semantically meaningful updates to state. They can sometimes still affect
 	// the format of state if applied.
 	NoChanges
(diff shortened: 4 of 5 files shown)


internal/command/jsonformat/plan_test.go
```diff
@@ -82,6 +83,36 @@ found no differences, so no changes are needed.
...
+	color := &colorstring.Colorize{Colors: colorstring.DefaultColors, Disable: true}
+	streams, done := terminal.StreamsForTesting(t)
+
+	plan := Plan{}
+
+	renderer := Renderer{Colorize: color, Streams: streams}
+	plan.renderHuman(renderer, plans.NormalMode, plans.NoChanges, plans.Targeted)
+
+	got := done(t).Stdout()
+	if !strings.Contains(got, "-target") {
```
apparentlymart: Matching the human-readable output in tests is always a bit of a tradeoff in how much of the text we want to expect not to change, but I'd suggest we compromise with the full "You used the -target option" message here just because it seems likely that the substring "-target" would appear somewhere else in the output when that option is being used, and so this test could end up becoming a false-positive under future maintenance.

apparentlymart (commented):
This seems reasonable to me! Thanks for working on it.

I've left one comment inline about a minor thing in the tests, but I think it's worth doing just because I've been bitten in the past by assertions against the human-readable output accidentally matching parts of the output they weren't intended to. 😀

diofeher (commented):
Thanks for your contribution @baa-ableton !

While testing and seeing how it looked, I noticed that this would be better as a real warning:

Screenshot 2026-03-26 at 16 15 34

The message can be overshadowed by other warnings, making it easy to miss, but I'm not super strong about it.

diofeher:
Now that I'm reading the existing warnings, I feel this added information is kind redundant:

```
╷
│ Warning: Resource targeting is in effect
│
│ You are creating a plan with either the -target option or the -exclude option,
│ which means that the result of this plan may not represent all of the changes
│ requested by the current configuration.
│
│ The -target and -exclude options are not for routine use, and are provided
│ only for exceptional situations such as recovering from errors or mistakes, or
│ when OpenTofu specifically suggests to use it as part of an error message.
╵
╷
│ Warning: Applied changes may be incomplete
│
│ The plan was created with the -target or the -exclude option in effect, so
│ some changes requested in the configuration may have been ignored and the
│ output values may not be fully updated. Run the following command to verify
│ that no other changes are pending:
│     tofu plan
│
│ Note that the -target and -exclude options are not suitable for routine use,
│ and are provided only for exceptional situations such as recovering from
│ errors or mistakes, or when OpenTofu specifically suggests to use it as part
│ of an error message.
╵
```

What do you think it wasn't covered by these existing warnings, as opposed to the newly-added one? @baa-ableton

apparentlymart:
I think the discussion in https://github.com/opentofu/opentofu/issues/3729 implies that the "Resource targeting is in effect" warning wasn't enough to hint to some operators that an incorrect address in `-target` might be the cause of there being no changes proposed, so I think we can assume that this additional messaging _isn't_ redundant.

I had originally proposed it being a modification of the main result line that starts with "No changes", because I imagined that was the part of the output that someone would most likely be reading when they are surprised that there were no changes. However, I feel okay with the compromise of adding a new paragraph as implemented here because it still places the note _relatively_ near to the "No changes" message.

I don't think we should make this new message be a warning _in addition to_ the "Resource targeting is in effect" warning because when we return more than one warning they compete in the "visual hierarchy" for which one is the most important to read, and that's extra confusing when both are discussing aspects of the same problem. Maybe we could incorporate this new note about `-target` possibly eliminating all of the options into the warning we already have, but I think that's architecturally inconvenient because that warning comes from the main language runtime (`package tofu`) rather than the CLI layer and we normally want this sort of UI decision to live in the CLI layer. :thinking:

internal/command/jsonformat/plan_test.go
baa-ableton: Right gotcha 👍

baa-ableton:
@diofeher 
The output you shared definitely have semantically identical (pretty much) parts served in 2 chunks which I think is problematic and it deserves a looking into.

I agree that positioning and formatting are really import like @apparentlymart said (which makes the point I just made more important). I don't have a preference whether or not the new piece of log information that I introduce is formatted as Warning or bare output - as long as it is grouped with the most relevant pieces of information.

I kind of think that the above two are related issues. Maybe we should review what @diofeher shared plus what I'm about to add as one whole and see what part would make most sense cutting away with

baa-ableton:
```
╷
│ Warning: Resource targeting is in effect
│
│ You are creating a plan with either the -target option or the -exclude option,
│ which means that the result of this plan may not represent all of the changes
│ requested by the current configuration.
│
│ The -target and -exclude options are not for routine use, and are provided
│ only for exceptional situations such as recovering from errors or mistakes, or
│ when OpenTofu specifically suggests to use it as part of an error message.
╵
╷
│ Warning: Applied changes may be incomplete
│
│ The plan was created with the -target or the -exclude option in effect, so
│ some changes requested in the configuration may have been ignored and the
│ output values may not be fully updated. Run the following command to verify
│ that no other changes are pending:
│     tofu plan
│
│ Note that the -target and -exclude options are not suitable for routine use,
│ and are provided only for exceptional situations such as recovering from
│ errors or mistakes, or when OpenTofu specifically suggests to use it as part
│ of an error message.
```

into 

```
╷
│ Warning: Resource targeting is in effect
│
│ You are creating a plan with either the -target option or the -exclude option,
│ which means that the result of this plan may not represent all of the changes
│ requested by the current configuration. Applied changes may be incomplete
│
│ The -target and -exclude options are not for routine use, and are provided
│ only for exceptional situations such as recovering from errors or mistakes, or
│ when OpenTofu specifically suggests to use it as part of an error message.
╵
```

seems better to me. I just moved the subject line of the 2nd warning into the body of the 1st one.

apparentlymart:
I believe the two warnings that @diofeher mentioned actually appear quite far apart in the real output:

- The "Resource targeting is in effect" warning is the plan phase warning that the plan might not include everything that ought to have been updated, before the operator applies the plan.
- The "Applied changes may be incomplete" warning appears only after the entire apply phase has completed, to remind the operator that they've probably just applied an incomplete plan and so they ought to run another plan/apply round to deal with anything they skipped on the first round.

So while they do seem redundant when written out together in the GitHub comment here, I don't think they are so redundant in the real UI.

diofeher:
@apparentlymart @baa-ableton 

Oh, in case it wasn't clear, the first screenshot I took and the messages I shared here appeared together.
I created a test repository from the issue, and while testing, I was a bit confused by seeing these messages together **(while running apply without using a plan output file)**, even though they're meant to appear at different points/phases of the cycle. That's why I'm suggesting reviewing them, as it looked kind of redundant to see them together ^^ 

Screenshot 2026-03-27 at 14 40 56

I totally understand your point, though, @apparentlymart, as these messages are meant to appear in different phases, and we cannot merge them because if we're using a different workflow (like saving to a plan file, then applying it), this warning wouldn't be shown if we're removing from one of the places.

apparentlymart:
Thanks for clarifying what you were looking at, @diofeher.

I think you've arguably found a minor bug there in that we probably shouldn't be returning that "Applied changes may be incomplete" warning unless there was at least one change in the (partial) plan, because indeed in this case it _is_ completely redundant[^1].

The "Applied changes may be incomplete" warning is intended for the case where there _was_ something included in the partial plan, and the operator chose to apply that partial set of actions. We produce it just because otherwise a lot of time may have passed between approving the plan and the apply phase completing and so the operator might've forgotten that they were applying a partial plan and so not remember to run an untargeted plan again afterwards.

I think that's something separate from what this PR is trying to solve though, and so I'd suggest we discuss improving that in a different place. (I hope it would just be as simple as skipping the "Applied changes may be incomplete" warning unless there's at least one entry in the set of planned changes.)

[^1]: I think this is actually exposing an implementation detail: as far as the end-user is concerned, a "No changes" plan doesn't get applied at all. But we do actually run the apply phase anyway in this case -- with an empty plan -- just to deal with some administrative work like updating the relationships between resources and provider instances in the state. But those are not "applied changes" in the sense that warning is talking about.

baa-ableton:
@apparentlymart @diofeher 
shall I also wrap the new message as a warning then? Sorry I got lost in terms of if there are any outstanding tasks.

apparentlymart:
Hi @baa-ableton,

I personally would vote for keeping the compromise you started with of making this an additional "Note:" message to start, since UI output is not subject to compatibility promises and so we can always make more changes to it if we learn that it wasn't obvious enough.

I'm going to wait to see if @diofeher agrees before we merge this, but for now I suggest leaving this you currently have it until he has a chance to respond. Thanks!

diofeher:
Sorry, guys, I completely missed this one! I probably missclicked on the ones I've read in my notifications :(

I agree with @apparentlymart. We can do it this way!

apparentlymart (approved):

diofeher (approved):

apparentlymart:
Sorry for the delay, @baa-ableton.

I've merged this now, and so it should be included in the forthcoming v1.3.0 release. Thanks!

PR #3946, opened 2026-03-25
Support const variables in static eval

### LINKED ISSUE #3945: Add support for `const` to variable block
When we originally added support for static/const variables and locals in OpenTofu, we decided against requiring an additional attribute to declare variables as "const".  This decision was made to make cross-compatibility with terraform simpler at the time.

With the latest terraform beta adding support for variables and locals in module sources, we wish to make it easy for users to enjoy the benefits of this feature in both tools. The main caveat being the `const = true` requirement now present in terraform.

I propose that we add support for const as opt-in. If you enjoy using OpenTofu's static evaluation feature as-is, you will not need to change anything.  If a "const" attribute is present on any variables within a module, the whole module is treated as opting-in to requiring const on any variables used in a static context.

The timing on this feature is not ideal as we are headed into our beta period shortly after terraform has entered theirs. There is a small risk that terraform decides to drop "const" before release, in which case we would want to back out these changes.

cam72cam:
This allows OpenTofu to understand and respond to the `const` attr on variables in a way that balances compatibility with terraform and the early evaluation that came before.

Resolves #3945

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/module_call_test.go
+++ b/internal/configs/module_call_test.go
@@ -409,7 +409,7 @@ variable "path" {
 	}, {
 		name:  "failure, post-const",
 		files: []string{"mod.tf", "path_notconst.tf"},
-		err:   "The variable \"path\" does not have `const = true` set and can not be used in a static context.",
+		err:   `The variable "path" cannot be used in a static context, because it is declared as "const = false`,
 	}}
 
 	for _, tc := range cases {


internal/configs/module.go
```diff
@@ -57,6 +58,9 @@ type Module struct {
 
 	Tests map[string]*TestFile
 
+	// If variable const support has been detected
```
yottta: I would enhance this comment for posterity.

internal/configs/module.go
```diff
@@ -232,6 +236,25 @@ func NewModuleUneval(primaryFiles, overrideFiles []*File, sourceDir string, load
...
+			break
+		}
+	}
+	if mod.ConstEnabled {
+		// Opt into new functionality
+		log.Printf("[DEBUG] Enabling const variable support for module %s", sourceDir)
+		for _, v := range mod.Variables {
+			if !v.ConstSet {
+				// Reset default to false
+				v.Const = false
```
yottta: I don't get it... If `v.ConstSet` is `false`, how can `v.Const` be not `false`? 🤔

yottta: I see that this is linked to the parsing logic:
```go
		// Default to true for backwards compat, which is subject to change
		// in a future language edition.
		// This default may be overwritten if other variables specify const
		v.Const = true
```
But why?

Wouldn't be easier to reason about this logic if:
* Default for constSet is false (aka, remove the `else` branch from the parsing of the config)
* In `static_scope.go` there is already logic for "error if module contains at least one `const` variable and the variable used in the expression is not constant"
* So then, this `if mod.ConstEnabled` block, could be removed entirely.

I might be wrong and I might have read this wrong, so please correct me if it's the case.

cam72cam: From the issue:

I propose that we add support for const as opt-in. If you enjoy using OpenTofu's static evaluation feature as-is, you will not need to change anything. If a "const" attribute is present on any variables within a module, the whole module is treated as opting-in to requiring const on any variables used in a static context.

When parsing a .tofu or .tf file, our initial assumption is that we are using the original "variables in module sources scheme" which does not take "const" into consideration.

If we determine that the module is using "const" in any way, we assume that they are opting-in to the new "const must be set" functionality that matches terraform.

internal/configs/module.go
```diff
@@ -57,6 +58,9 @@ type Module struct {
 
 	Tests map[string]*TestFile
 
+	// If variable const support has been detected
+	ConstEnabled bool
```
apparentlymart: I feel pretty ambivalent about this module-level behavior, which I'm understanding as "if you use `const = true` for any one variable then that implicitly disables early eval support for every other variable in the same module".

I can see where you are coming from with that: it effectively makes the presence of `const = true` represent a module-wide opt-in for "treat early eval more like like Terraform does".

But I think there's an alternative take which seems easier to implement and easier to explain: setting `const = true` on a particular variable means that the expression used to define that variable _must_ be early-eval-compatible. Leaving `const` unset means that the early-eval-compat is decided dynamically based on how the variable was defined, just as OpenTofu has done since this feature was implemented.

If we want to hedge for someone wanting the ability to say "this variable must not be used in early eval contexts regardless of how it's defined" then we could represent that as explicitly setting `const = false`, which I'm assuming (without testing) Terraform would accept and also treat it as meaning "this variable cannot be used in early eval contexts". Leaving `const` completely unset would still retain the "we infer it automatically based on definition" behavior.

To be clear, my main reservation here is that having the presence of an argument in one place cause implicit behavior changes for other declarations that don't set that argument seems confusing and complicated. I'd prefer to find a compromise that keeps this as a per-declaration local decision, and I suggested one such compromise above but I'm open to others too.

cam72cam: That is a fair point. What I added is "magical" from an end user perspective, and I try to avoid that as much as possible.

We could always make this part of a language edition change in the future if there is enough ask for the terraform-style default.

cam72cam: Changed in [9ff9001](https://github.com/opentofu/opentofu/pull/3946/commits/9ff90012c8ca4a1bf6f3f083153b688f84062b7e)

cam72cam:
@apparentlymart I kept all of the additional docs as name-agnostics as possible in regards to static/early/const. The good news is that the website does not use any of that terminology from what I can tell.

It's mostly our blogs that are a bit mixed up.

website/docs/language/values/variables.mdx
```diff
@@ -346,7 +347,7 @@ instructions on how to migrate away from usage of this variable. Here is an exam
 configuration:
 
 ```hcl
-variable "example" {
+variable "examle" {
```
Yantrio, line 350:
```suggestion
variable "example" {
```

cam72cam: This was an intentional change, I noticed that this was wrong when updating the docs

Yantrio: Ignore me, examle is right

internal/configs/module.go
```diff
@@ -272,6 +272,21 @@ func NewModule(primaryFiles, overrideFiles []*File, call StaticModuleCall, sourc
...
+	// Ensure const variables are actually const
+	var constRefs []*addrs.Reference
+	for _, variable := range mod.Variables {
+		if variable.Const {
+			constRefs = append(constRefs, &addrs.Reference{
+				Subject: addrs.InputVariable{Name: variable.Name},
+			})
+		}
+	}
+	_, vDiags := mod.StaticEvaluator.EvalContext(context.TODO(), StaticIdentifier{
```
diofeher, line 284: Should we call that only if `constRefs` is not empty? From what I'm reading, it would work fine even if it is empty, but it wouldn't return anything useful

cam72cam: IMO it's (effectively) zero cost to call this function with empty refs and minimizes the cognitive overhead of parsing this snippet.

cam72cam: I don't have a terribly strong opinion however and am happy to be overruled.

internal/configs/static_scope.go
```diff
@@ -260,6 +260,15 @@ func (s staticScopeData) GetInputVariable(_ context.Context, ident addrs.InputVa
 		})
 	}
 
+	if variable.ConstSet && !variable.Const {
+		return cty.NilVal, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Unable to use variable in static context",
+			Detail:   fmt.Sprintf("The variable %q does not have `const = true` set and can not be used in a static context.", variable.Name),
```
apparentlymart: I believe this error appears only if the author explicitly wrote `const = false`, because if `const` were not set at all then `variable.ConstSet` would be `false`, and if `const = true` were set then `variable.Const` would be `true`.

Assuming I understood that correctly, perhaps the following error message would be clearer:


```suggestion
			Detail:   fmt.Sprintf("The variable %q cannot be used in a static context, because it is declared as \"const = false\".", variable.Name),
```

(I used quotes rather than backticks here just because I don't think there's precedent for using Markdown-style formatting in our diagnostic messages.)

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ ENHANCEMENTS:
 - The `s3` backend now automatically discovers and uses AWS credentials issued using [the `aws login` command](https://docs.aws.amazon.com/cli/latest/reference/login/) in AWS CLI. ([#3767](https://github.com/opentofu/opentofu/pull/3767))
 - The `azurerm` backend now supports authentication using Azure DevOps and Azure Pipelines workload identity federation. ([#3820](https://github.com/opentofu/opentofu/pull/3820))
 - Most of the commands now prints the usage text when arguments parsing fails, prints previously legacy error messages as regular diagnostics and received support JSON output. ([#3941](https://github.com/opentofu/opentofu/pull/3941))
+- A new `const` attribute (optional) has been added to variable blocks to declare if they are allowed to be used in static evaluation. ([#3946](https://github.com/opentofu/opentofu/pull/3946))
```
```suggestion
- Input variables can now be declared as `const = true` to require that the assigned value is compatible with static evaluation. ([#3946](https://github.com/opentofu/opentofu/pull/3946))
```

(I also suggest moving this further up the list so it's after the item for https://github.com/opentofu/opentofu/issues/3300, so that all of the language-related items are still grouped together.)

cam72cam: Moved in 3c60ceeec0

internal/configs/module.go
diofeher: Since you and Martin are against it, I'm fine with it!

PR #3949, opened 2026-03-26
Add backend view and migrate outputs

### LINKED ISSUE #3765: Migrate all commands to the `arguments` and `views` packages

yottta:
Resolves #3765
Changes summary:
* Add `views.Backend` with all the messages needed in `meta_backend.go`, `meta_backend_migrate.go` and `meta_config.go`.
* Move some of the methods added initially in the `views.Init` to the new `views.Backend` since this where their scope is actually.
* Wherever the BackendOpts are instantiated, add also the ViewOptions.
* Add `viewOptions()` method to BackendOpts and backendMigrateOpts that returns the ViewOptions in a safe manner, with the default agreed before this refactor.
* Wherever needed, create `views.NewBackend` based on the ViewOptions from the arguments.
* There were some missed migrations to the new `views` package in `state_meta.go` so moved those too.
* Extract the logic that prints workspaces to be used in the backend view too.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/meta_backend.go
+++ b/internal/command/meta_backend.go
@@ -62,20 +62,14 @@ type BackendOpts struct {
 
 	// ViewOptions will set console output format for the
 	// initialization operation (JSON or human-readable).
-	ViewOptions arguments.ViewOptions
+	View views.Backend
 }
 
-func (b *BackendOpts) viewOptions() arguments.ViewOptions {
-	var viewOptions arguments.ViewOptions
-	if b != nil {
-		viewOptions = b.ViewOptions
+func (b *BackendOpts) backendView(base *views.View) views.Backend {
+	if b != nil && b.View != nil {
+		return b.View
 	}
-	// Set default viewtype if none was set as the StateLocker needs to know exactly
-	// what viewType we want to have.
-	if viewOptions.ViewType != arguments.ViewHuman && viewOptions.ViewType != arguments.ViewJSON {
-		viewOptions.ViewType = arguments.ViewHuman
-	}
-	return viewOptions
+	return views.NewBackendHuman(base)
 }
 
 // BackendWithRemoteTerraformVersion is a shared interface between the 'remote' and 'cloud' backends
@@ -448,7 +442,7 @@ func (m *Meta) backendCLIOpts(ctx context.Context) (*backend.CLIOpts, error) {
 // This prepares the operation. After calling this, the caller is expected
 // to modify fields of the operation such as Sequence to specify what will
 // be called.
-func (m *Meta) Operation(ctx context.Context, b backend.Backend, vt arguments.ViewOptions, enc encryption.Encryption) *backend.Operation {
+func (m *Meta) Operation(ctx context.Context, b backend.Backend, view views.Backend, enc encryption.Encryption) *backend.Operation {
 	schema := b.ConfigSchema()
 	workspace, err := m.Workspace(ctx)
 	if err != nil {
@@ -468,8 +462,7 @@ func (m *Meta) Operation(ctx context.Context, b backend.Backend, vt arguments.Vi
 
 	stateLocker := clistate.NewNoopLocker()
 	if m.stateLock {
-		view := views.NewStateLocker(vt, m.View)
-		stateLocker = clistate.NewLocker(m.stateLockTimeout, view)
+		stateLocker = clistate.NewLocker(m.stateLockTimeout, view.StateLocker())
 	}
 
 	depLocks, diags := m.lockedDependencies()
@@ -923,8 +916,7 @@ func (m *Meta) backend_c_r_S(
 	c *configs.Backend, cHash int, sMgr *clistate.LocalState, output bool, opts *BackendOpts, enc encryption.StateEncryption) (backend.Backend, tfdiags.Diagnostics) {
 
 	var diags tfdiags.Diagnostics
-	viewOpts := opts.viewOptions()
-	view := views.NewBackend(viewOpts, m.View)
+	view := opts.backendView(m.View)
 
 	s := sMgr.State()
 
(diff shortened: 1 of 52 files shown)


internal/command/meta_backend.go
```diff
@@ -65,6 +65,19 @@ type BackendOpts struct {
...
+	if b != nil {
+		viewOptions = b.ViewOptions
+	}
+	// Set default viewtype if none was set as the StateLocker needs to know exactly
+	// what viewType we want to have.
+	if viewOptions.ViewType != arguments.ViewHuman && viewOptions.ViewType != arguments.ViewJSON {
+		viewOptions.ViewType = arguments.ViewHuman
+	}
+	return viewOptions
+}
```
diofeher: I see that the implementation of this function is slightly different from that of backendMigrateOpts. The other initializes with ViewHuman as default, while this one instantiates with `arguments.ViewOptions`.

Is this the desired behavior? Or should we instantiate ViewOptions here with ViewHuman as the default too?

https://github.com/opentofu/opentofu/pull/3949/changes#diff-beb9f757a39340f477d9384b2344a3caea2b790c0deb6b044127fd9565e613caR42

yottta:
> This looks good to me.
> 
> While reading I did wonder if we could reduce the number of calls like `views.NewBackend(viewOptions, m.View)` by passing around the `views.Backend` value directly instead of passing around the opts, but I'm assuming that's harder than it seems due to how complicated and kinda-messy these backend-related codepaths are.
> 
> I think it's fine to start with this and then hopefully continue to improve it later. I think that addressing the code organization concerns with backends is a separate project that we shouldn't try to do at the same time as we adopt `views`/`arguments`. 😀


<details>
<summary> Message typed this morning 😅  </summary>
That was my initial attempt. But the issue is that out there are places which calls `views.NewStateLocker()` and that requires `arguments.ViewOptions`.

I tried to make something similar with my previous attempt [here](https://github.com/opentofu/opentofu/pull/3952), but that highlighted another issue: calling [`NewJSONView`](https://github.com/opentofu/opentofu/blob/866b067c0daf18165ced10879992a1d49d3de8df/internal/command/views/json_view.go#L26-L41) multiple times, prints the version log entry multiple times and I am not sure how hard of a requirement is to have only just occurrence of that. _See the encountered issues in the PR description_.
At this point, because we are calling `NewJSONView` so many times, I think that the work I did here broke the way [`JSONView`](https://github.com/opentofu/opentofu/blob/1324f8a06ce10edf966b8bc94fe64c4e73e0613c/internal/command/views/json_view.go#L26) is meant to be used.
</details>

_Evening message_ 😅 
After a lot of tinkering around and changing the whole mental model of the views package, I came up with https://github.com/opentofu/opentofu/pull/3956.
This does several things:
* Separates the views in 2 categories:
  * "root level" or "command specific" views (`Apply`, `Plan`, etc)
  * "child" or "specific non-command functionality" views (`Backend`, `StateLocker`, etc)
* Any "root level" view that needs to work with the backend logic, needs to expose a `Backend()` method that will create the `views.Backend` by using the inner properties. This ensures that we don't create new loggers or new `JSONView` instances, but instead, it's using the same reference in any "child" view.
* Added tests for the `views.StateLocker` since those were missing, and changed its implementation to match the other JSON views structure of the log entries (it was formatting the timestamp in a different way)

With this, all the issues exposed by the reviews, the ones highlighted in my "morning message" and the ones explained in #3952 are now solved and everything seems to work just fine 😍

internal/command/meta_backend.go
yottta: This was just copy pasted from the place(s) where now `viewOptions` is called.
The end result is the same:
* if the opts are not nil, use the viewOptions inside
* if the viewOptions.ViewType is default (aka not human either json), set is as human format
* return

These are 2 different ways to express the same idea.

Anyway, could you please check #3956 since this method was replaced with something else there?
I intend to merge that PR in this one and then go to main.

diofeher: Sure!

PR #3952, opened 2026-03-27
Attempt to propagate the backend view instead of the view options

yottta:
_Attempt to improve based on [this comment](https://github.com/opentofu/opentofu/pull/3949#pullrequestreview-4015579025)_
This works fine but it fails one of the tests that has a hard assertion about having a version entry, though, not being clear if it's required to have just one entry or there is no issue in having multiple version entries.

To counteract the multiple versions log entries, tried the following:
* use `sync.Once` in `views/json_view.go` to print the version only once. Works for the general `tofu` process, but fails some tests, since the `go test` is a single process and some tests assert for that version entry, which is an assertion that we want to meet. And since the `sync.Once` is per process, not per test, first test that runs it, gets the version log entry, but the tests that run after, don't get that log entry.
* Instead of calling `NewJSONView` for the non-command views instantiation (eg: `views.Backend`, `views.StateLocker`), tried to create methods in the command views implementation to return the `views.Backend` (similar to the `views.Backend.StateLocker()` one), but that's also painful because some command's views are configured differently based on the needs, and to be able to create a `views.Backend`, the `*os.File` for the output should be carried around through the views (because `views.StateLocker` needs it in the end), which just makes things more confusing.

Resolves # 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/apply.go
+++ b/internal/command/apply.go
@@ -241,8 +241,8 @@ func (c *ApplyCommand) PrepareBackend(ctx context.Context, planFile *planfile.Wr
 		}
 
 		be, beDiags = c.Backend(ctx, &BackendOpts{
-			Config:      backendConfig,
-			ViewOptions: viewOptions,
+			Config: backendConfig,
+			View:   views.NewBackend(viewOptions, c.View),
 		}, enc)
 	}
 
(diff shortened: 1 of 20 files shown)


yottta:
Closing this in favor of #3956.

PR #3954, opened 2026-03-27
Update plugin protos to 5.10 and 6.10

cam72cam:
This is from the MPL2.0 licensed source

Precursor to #3476 

## Checklist

### Go checklist

 


docs/plugin-protocol/tfplugin5.10.proto
```diff
@@ -0,0 +1,953 @@
+// Copyright IBM Corp. 2014, 2026
+// SPDX-License-Identifier: MPL-2.0
```
diofeher, line 2: My only question here is: should we add our OpenTofu headers to these files as well?

cam72cam: We have not done that historically.

PR #3956, opened 2026-03-27
Non commands views configurable only from package level

yottta:
While exploring the possible options for the inconvenience highlighted by [this comment](https://github.com/opentofu/opentofu/pull/3949#pullrequestreview-4015579025), I had to change the mental model of the views package, landing on the fact that the package contains 2 types of views:
* Root level, or command specific views, like `Apply`, `Plan`, `Workspace`, etc
* Specific isolated functionality views, like `Backend`, `StateLocker`, etc

Had to change this mental model because by calling [`NewJSONView`](https://github.com/opentofu/opentofu/blob/866b067c0daf18165ced10879992a1d49d3de8df/internal/command/views/json_view.go#L26-L41) multiple times, generates multiple version log entries and it might interfere with external tools that might rely on sole existence of that version log entry.

The way I structured the relations between views here forces a "command view" to intentionally expose a `Backend()` method that will return a properly configured `views.Backend`.

ℹ️ One thing that I had to change to make this work, was the implementation of the [`StateLockerJSON`](https://github.com/opentofu/opentofu/blob/1324f8a06ce10edf966b8bc94fe64c4e73e0613c/internal/command/views/state_locker.go#L83-L94). The changes affect **only** the format of the timestamp, which IMO, was mostly an inconsistency which had to be fixed. Checked the whole project, and that was the only view that used [time.RFC3339](https://pkg.go.dev/time#pkg-constants) to format the log entry timestamp. As you can see in the commits history of this PR, I added tests first for the `StateLockerJSON` and only after changed the implementation, ensuring that the rest of the fields are there as expected.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/meta_backend.go
+++ b/internal/command/meta_backend.go
@@ -65,7 +65,7 @@ type BackendOpts struct {
 	View views.Backend
 }
 
-func (b *BackendOpts) view(base *views.View) views.Backend {
+func (b *BackendOpts) backendView(base *views.View) views.Backend {
 	if b != nil && b.View != nil {
 		return b.View
 	}
@@ -916,7 +916,7 @@ func (m *Meta) backend_c_r_S(
 	c *configs.Backend, cHash int, sMgr *clistate.LocalState, output bool, opts *BackendOpts, enc encryption.StateEncryption) (backend.Backend, tfdiags.Diagnostics) {
 
 	var diags tfdiags.Diagnostics
-	view := opts.view(m.View)
+	view := opts.backendView(m.View)
 
 	s := sMgr.State()
 
@@ -985,7 +985,7 @@ func (m *Meta) backend_c_r_S(
 func (m *Meta) backend_C_r_s(ctx context.Context, c *configs.Backend, cHash int, sMgr *clistate.LocalState, opts *BackendOpts, enc encryption.StateEncryption) (backend.Backend, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 
-	view := opts.view(m.View)
+	view := opts.backendView(m.View)
 	// Grab a purely local backend to get the local state if it exists
 	localB, localBDiags := m.Backend(ctx, &BackendOpts{ForceLocal: true, Init: true, View: view}, enc)
 	if localBDiags.HasErrors() {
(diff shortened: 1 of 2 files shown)


yottta:
@diofeher, @apparentlymart , this is the follow-up for #3949. Since your comments there really helped with pushing for more exploration on how to integrate this new backend view better in the workflow, I came up with this final PR that I intend to merge into #3949 before having that merged to main.
I would appreciate your review on this, if you have time for it.

yottta:
> This seems like a nice compromise to me.
> 
> I do still hope we can simplify the backend-related codepaths in the CLI layer more in later work -- maybe as part of eventually removing all of the legacy backend code and exclusively relying on "state storage as plugins" -- but that's a much bigger project far outside of what you are doing here, so I'll just wait patiently for a better time to think about that. 😀

Totally agree that it would be nice to do so, but until then, I plan to extract the whole backend logic in its own component (soon), and for that I want to have as clear dependencies as possible. And the changes in this PR actually help tons with that 😄

internal/command/views/apply.go
```diff
@@ -200,6 +221,7 @@ type ApplyJSON struct {
 	destroy bool
 
 	countHook *countHook
+	output    *os.File
```
cam72cam: This is unused, why add it?

yottta: Good catch! I was experimenting with passing it down to the stateLocker view and most probably forgot to remove it. Afterwards, decided to change the stateLocker view implementation to use the actual view implementation instead of using the output directly.
Thanks for catching it.
Removed in 29a7be20da221395625f8cfc9ea646882327c391

internal/command/meta_backend.go
```diff
@@ -62,20 +62,14 @@ type BackendOpts struct {
...
 	// initialization operation (JSON or human-readable).
-	ViewOptions arguments.ViewOptions
+	View views.Backend
 }
 
-func (b *BackendOpts) viewOptions() arguments.ViewOptions {
-	var viewOptions arguments.ViewOptions
-	if b != nil {
-		viewOptions = b.ViewOptions
+func (b *BackendOpts) view(base *views.View) views.Backend {
```
cam72cam: Sug: rename to viewBackend or backendView?

yottta: Renamed to `backendView` in d399b08e63465540b9cdf97b1850bebad48245d2

PR #3958, opened 2026-03-27
Remove Meta not needed functionality

yottta:
* Removes most of the methods and attributes from `Meta` related to coloring. This implied also to remove code from most of the commands. The only thing that was replaced, was the `Colorize()` one, that now is used from the `m.View.Colorize()` instead. This a temporary solution, since this should not be exposed from the view, but wherever it is called now, we should replace that with a view usage.
* Updated tests for colorize, and removed one case since it made no sense anymore. Details in the `NOTE` left in the code.
* Remove all the flag related methods from Meta (`defaultFlagSet`, `extendedFlagSet`, etc). This meant that I had to update some tests that were relying on those. Therefore, added some new `TODO`s to remove also the assumed state of `Meta` during testing
* Removed some attributes that were related to `showDiagnostics`, and I missed those in the [previous PR](https://github.com/opentofu/opentofu/pull/3950) (consolidate and compact warnings/errors)
* Removed the `Meta.process()` method, since the logic in there now is managed by `views.ParseView`
* Updated `Meta.configureUiFromView()` to get rid of the `Meta.oldUI`

The `Meta.UI` will have to stay for the moment, since that is passed over to the remote/cloud backends. There is additional work needed to be able to remove this too.

Part of  https://github.com/opentofu/opentofu/issues/3765
## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/meta.go
+++ b/internal/command/meta.go
@@ -226,14 +226,6 @@ type Meta struct {
 	variableArgs flags.RawFlags
 	input        bool
 
-	// Targets for this context (private)
-	targets     []addrs.Targetable
-	targetFlags []string
-
-	// Excludes for this context (private)
-	excludes     []addrs.Targetable
-	excludeFlags []string
-
 	// The fields below are expected to be set by the command via
 	// command line flags. See the Apply command for an example.
 	//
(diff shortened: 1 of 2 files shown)


diofeher (approved):
🚀

PR #3961, opened 2026-03-29
Failed to load backend configuration

### LINKED ISSUE #3643: Legacy state reader incorrectly triggered when TF_DATA_DIR contains terraform.tfstate

KrishnaSindhur:
Resolves #3643 (2nd part updated error message

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/clistate/cli_state.go
+++ b/internal/command/clistate/cli_state.go
@@ -103,7 +103,6 @@ type jsonVersionOnly struct {
 	Version int `json:"version"`
 }
 
-
 // ReadState reads the CLI state file format written by WriteState.
 func ReadState(src io.Reader) (*CLIState, error) {
 	if f, ok := src.(*os.File); ok && f == nil {
--- a/internal/command/clistate/cli_state.go
+++ b/internal/command/clistate/cli_state.go
@@ -103,6 +103,7 @@ type jsonVersionOnly struct {
 	Version int `json:"version"`
 }
 
+
 // ReadState reads the CLI state file format written by WriteState.
 func ReadState(src io.Reader) (*CLIState, error) {
 	if f, ok := src.(*os.File); ok && f == nil {
@@ -129,6 +130,14 @@ func ReadState(src io.Reader) (*CLIState, error) {
 	}
 
 	if ver.Version != StateVersion {
+		if bytes.Contains(jsonBytes, []byte(`"terraform_version"`)) {
+			return nil, fmt.Errorf(
+				"found a resource state file (version %d) where a backend configuration "+
+					"file (version %d) was expected; if TF_DATA_DIR is set, ensure it does "+
+					"not point to a directory containing a terraform.tfstate resource state file",
+				ver.Version, StateVersion,
+			)
+		}
 		return nil, fmt.Errorf(
 			"opentofu %s does not support CLI state version %d, please update",
 			tfversion.SemVer.String(),
--- a/internal/command/clistate/cli_state.go
+++ b/internal/command/clistate/cli_state.go
@@ -104,7 +104,9 @@ type jsonVersionOnly struct {
 }
 
 // ReadState reads the CLI state file format written by WriteState.
-func ReadState(src io.Reader) (*CLIState, error) {
+// dataDirOverridden should be true when the data directory was explicitly
+// overridden (e.g. via TF_DATA_DIR).
+func ReadState(src io.Reader, dataDirOverridden bool) (*CLIState, error) {
 	if f, ok := src.(*os.File); ok && f == nil {
 		return nil, ErrNoState
 	}
@@ -129,12 +131,12 @@ func ReadState(src io.Reader) (*CLIState, error) {
 	}
 
 	if ver.Version != StateVersion {
-		if bytes.Contains(jsonBytes, []byte(`"terraform_version"`)) {
+		if dataDirOverridden && bytes.Contains(jsonBytes, []byte(`"terraform_version"`)) {
 			return nil, fmt.Errorf(
-				"found a resource state file (version %d) where a backend configuration "+
-					"file (version %d) was expected; if TF_DATA_DIR is set, ensure it does "+
-					"not point to a directory containing a terraform.tfstate resource state file",
-				ver.Version, StateVersion,
+				"found a resource state file (version %d) where a backend configuration cache "+
+					"file was expected; if TF_DATA_DIR is set, ensure it does "+
+					"not point to a directory containing your main terraform.tfstate resource state file",
+				ver.Version,
 			)
 		}
 		return nil, fmt.Errorf(
--- a/internal/command/workdir/dir.go
+++ b/internal/command/workdir/dir.go
@@ -66,6 +66,10 @@ type Dir struct {
 	// directory named ".terraform" within mainDir, but users may
 	// override it.
 	dataDir string
+
+	// dataDirOverridden is true when OverrideDataDir has been called,
+	// which typically means the TF_DATA_DIR environment variable was set.
+	dataDirOverridden bool
 }
 
 // NewWorkdir returns a [*Dir] instance configured with the following:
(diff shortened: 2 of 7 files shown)


KrishnaSindhur:
Error message in CLI 
Before:
`Error: Failed to load state: opentofu 1.12.0 does not support CLI state version 4, please update`

After this PR change 
`Error: Failed to load backend configuration from /tmp/tofu-repro/data/terraform.tfstate: found a resource state file (version 4) where a backend configuration file (version 3) was expected; if TF_DATA_DIR is set, ensure it does not point to a directory containing a terraform.tfstate resource state file`

internal/command/clistate/cli_state.go
```diff
@@ -129,6 +129,14 @@ func ReadState(src io.Reader) (*CLIState, error) {
 	}
 
 	if ver.Version != StateVersion {
+		if bytes.Contains(jsonBytes, []byte(`"terraform_version"`)) {
```
yottta: We should, somehow, plug the check requested [here](https://github.com/opentofu/opentofu/issues/3643#issuecomment-3739547627):
> If we do this then I think the special error message should be returned only if the TF_DATA_DIR environment variable is set, because otherwise we'd risk misleading someone who has an invalid CLI state file for some other reason that's unrelated to the TF_DATA_DIR variable.

Though, considering that the env variable is defined and used in the [`workdir`](https://github.com/opentofu/opentofu/blob/6e2250e050d10a4e2bd82d717bae594da845395d/internal/command/workdir/dir.go#L15) package, I would suggest one of the following approaches:
* Export a new function from that package that returns a `bool` if that environment variable has a value defined. 
* Export a function for the `workdir.Dir` that will return a stored `bool` that will be set to `true` only when [`workdir.Dir.OverrideDataDir`](https://github.com/opentofu/opentofu/blob/6e2250e050d10a4e2bd82d717bae594da845395d/internal/command/workdir/dir.go#L145) is called. And then pass the value returned by that into the `clistate.LocalState` whenever is initialised.

Personally, I would suggest to go with the second option, since it goes well with the initialisation of the `LocalState`, since it already uses information from the `workdir.Dir` configured object.

Before applying the changes, let's wait a couple of days, maybe other maintainers will like to have this done differently.

Other than that, I am not sure that the `bytes.Contains` check should be kept once we add the env var check 🤔

yottta (changes requested):
Thanks for the help here.
There is one thing that is still to be tackled. Details inline.

internal/command/clistate/cli_state.go
```diff
@@ -129,6 +129,14 @@ func ReadState(src io.Reader) (*CLIState, error) {
...
+			return nil, fmt.Errorf(
+				"found a resource state file (version %d) where a backend configuration "+
+					"file (version %d) was expected; if TF_DATA_DIR is set, ensure it does "+
+					"not point to a directory containing a terraform.tfstate resource state file",
+				ver.Version, StateVersion,
+			)
```
apparentlymart, line 140: This is a tricky situation because this error message needs to mention concepts we've never needed to mention in the UI or docs before, so we don't have established terminology to use.

But I do have some small suggested adjustments here:

```suggestion
			return nil, fmt.Errorf(
				"found a resource state file (version %d) where a backend configuration cache "+
					"file was expected; if TF_DATA_DIR is set, ensure it does "+
					"not point to a directory containing your main terraform.tfstate resource state file",
				ver.Version,
			)
```

Here I introduced the new terminology "backend configuration _cache_ file", because we already use the term "backend configuration" to refer to a `backend` block in a normal configuration file. The `$TF_DATA_DIR/terraform.tfstate` file behaves as a _sort of_ cache of the backend configuration settings that were present during `tofu init`, so that seemed like a reasonable way to distinguish it.

I also removed the part that would've said "(version 3)" here because the fact that the "backend configuration cache file" format was derived from version 3 of the resource state file format is an implementation detail that I think would make this situation appear more confusing, because it suggests that it might be possible to "upgrade" the backend configuration cache file format into the resource state format, which isn't true.

I finally added "your main" before mentioning `terraform.tfstate` by name, because I worry that someone who isn't already familiar with this problem could have trouble understanding what we mean given that both of these files are named `terraform.tstate`. I'm not sure that "your main terraform.tfstate" is enough to distinguish them either, but I'm just hoping that including that extra modifier would make the reader realize that we're trying to distinguish between two different files of the same name here. 🤷‍♂️

internal/command/clistate/cli_state.go
KrishnaSindhur: I have kept both the checks,  `if dataDirOverridden && bytes.Contains(jsonBytes, []byte(`"terraform_version"`))` is the second check redundant?

internal/command/workdir/dir.go
```diff
@@ -143,6 +147,13 @@ func (d *Dir) OverrideOriginalWorkingDir(originalPath string) {
 // working directory.
 func (d *Dir) OverrideDataDir(dataDir string) {
 	d.dataDir = filepath.Clean(dataDir)
+	d.dataDirOverridden = true
```
yottta, line 150: I know that this is quite dumb to ask for, but can we add a unit test for this with some comments on that test to explain why that test exists and why that expectancy is like that?
This is mostly for the future contributors that might want to change anything around this flag.

yottta (approved):
Other than the suggestion left by @apparentlymart and the one small thing that I left during this review, this looks pretty good.
Thanks for the help! 🙌

internal/command/workdir/dir.go
yottta: Thanks for this. Added in 51ae09271a99e0893c8c271c9b55775603d52c6f

apparentlymart (approved):

PR #3962, opened 2026-03-30
Remove `GatherVariables` methods

yottta:
Part of #3765

Ever since [var flags parsing has been unified into a single struct](https://github.com/opentofu/opentofu/pull/3725) these methods were mostly useless, but we wanted to keep the same pattern until we had all the commands using the same pattern.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/views/backend.go
+++ b/internal/command/views/backend.go
@@ -31,8 +31,11 @@ type Backend interface {
 }
 
 // NewBackendHuman returns a new Backend instance that will print in human format.
-// This functions is meant to be used only for testing purposes. In actual flows,
-// Backend should be acquired from another view.
+// This particular function is meant to be used only in special cases, where the
+// Backend view cannot be acquired from a command related view (eg: Apply.Backend).
+// At the moment of writing this comment, this function is meant to be used only to
+// create this view in cases where it is not initialised correctly, which are paths
+// that are only reachable from incomplete configured tests.
 func NewBackendHuman(view *View) Backend {
 	return &BackendHuman{view: view}
 }


internal/command/command_test.go
```diff
@@ -1226,3 +1226,101 @@ func testHangServer(t testing.TB) (server *httptest.Server, reqs <-chan *http.Re
...
+		expectedContent []string
+	}{
+		"plan": {
+			cmdBuilder: func(m Meta) cli.Command {
+				return &PlanCommand{m}
+			},
+		},
+		"apply": {
+			cmdBuilder: func(m Meta) cli.Command {
+				return &PlanCommand{m}
```
diofeher:
```suggestion
				return &ApplyCommand{m}
```

internal/command/command_test.go
```diff
@@ -1226,3 +1226,101 @@ func testHangServer(t testing.TB) (server *httptest.Server, reqs <-chan *http.Re
...
+			},
+		},
+		"output": {
+			cmdBuilder: func(m Meta) cli.Command {
+				return &PlanCommand{m}
```
diofeher:
```suggestion
				return &OutputCommand{m}
```

internal/command/command_test.go
```diff
@@ -1226,3 +1226,101 @@ func testHangServer(t testing.TB) (server *httptest.Server, reqs <-chan *http.Re
...
+			},
+		},
+		"show": {
+			cmdBuilder: func(m Meta) cli.Command {
+				return &PlanCommand{m}
```
diofeher:
```suggestion
				return &ShowCommand{m}
```

internal/command/command_test.go
```diff
@@ -1226,3 +1226,101 @@ func testHangServer(t testing.TB) (server *httptest.Server, reqs <-chan *http.Re
...
+			},
+		},
+		"refresh": {
+			cmdBuilder: func(m Meta) cli.Command {
+				return &PlanCommand{m}
```
diofeher:
```suggestion
				return &RefreshCommand{m}
```

internal/command/views/backend.go
diofeher:
```suggestion
// This function is meant to be used only for testing purposes. In actual flows,
```

internal/command/meta.go
```diff
@@ -223,7 +223,7 @@ type Meta struct {
 	backendState *clistate.BackendState
 
 	// Variables for the context (private)
-	variableArgs flags.RawFlags
+	variableArgs []flags.RawFlag
```
diofeher, line 226: Any reason for changing this type? I see that RawFlags still exists

diofeher (approved):
Looks good! I found some minor stuff, but I'm approving it already

internal/command/command_test.go
yottta: Silly me... Had these tests configured in individual sub-tests and converted to table test, managed to screw that up 🤦 
Thanks for catching it.
Fixed in 56be41ecf80ee68c50e590282f512b393ddee43f

internal/command/command_test.go
yottta: Fixed in 56be41ecf80ee68c50e590282f512b393ddee43f

internal/command/command_test.go
yottta: Fixed in 56be41ecf80ee68c50e590282f512b393ddee43f

internal/command/command_test.go
yottta: Fixed in 56be41ecf80ee68c50e590282f512b393ddee43f

internal/command/views/backend.go
yottta: Revisiting that comment, thanks to your comment, realised that it was confusing indeed.
Rewrote it in caa57ec7a13da5b07c50413f12013bd529efee70.
Please let me know if it's better.

internal/command/meta.go
yottta: TL;DR: To ease the flag values passing and avoid introducing a new method in the `arguments.Vars`.

In #3725, we removed a duplicate implementation of the `flags.RawFlags`.
As a result, we changed the [`arguments.Vars.All()`](https://github.com/opentofu/opentofu/pull/3725/changes/07db852af10bb7ebcd69752aa72594bad0bb6a25#diff-e5860969219089a51442a23f7ffb95009835c83347327e3ce491a4ce237fe165L296) method return type.
[`GatherVariables`](https://github.com/opentofu/opentofu/blob/07db852af10bb7ebcd69752aa72594bad0bb6a25/internal/command/apply.go#L304-L310) method responsibility was to convert from the `arguments.Vars.All()` type (`flags.FlagNameValueSlice`) to the type used in `Meta.variableArgs` (`flags.RawFlags`). 

So after that change in #3725, that conversion was pretty much useless (converting from `flags.RawFlags` to `flags.RawFlags`).

With all that context, the usage of `flags.RawFlags` for `Meta.variableArgs` brings no additional benefits. And I could have kept it as it is, but then, in the `Run()` methods we would have:
```go
c.Meta.variableArgs = flags.RawFlags{Items: new(args.Vars.All())}
```
instead of what we have now:
```go
c.Meta.variableArgs = args.Vars.All()
```

or a similar approach that could have involved adding a new method in the `arguments.Vars` to return a new `flags.RawFlags`.

Personally, I consider that changing that type of `Meta.variableArgs` is more cleaner than creating other types or expose new methods just for this. 🤷 

Does it make sense? 

_this should have been a quick reply and ended up in a novel, sorry for that_ 😅

internal/command/views/backend.go
diofeher: Thanks, makes sense!

internal/command/meta.go
diofeher: Thanks for the novel haha It helped to clarify

PR #3965, opened 2026-03-31
Always return null value for write-only attributes from the mock provider for 'tofu test'

### LINKED ISSUE #3964: test cleanup fails with mocked provider with a resource type that has a write-only attribute
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.4
on linux_amd64
+ provider registry.opentofu.org/hashicorp/helm v3.1.1
```

```hcl
resource "helm_release" "foo" {
  name       = "nothing"
  repository = "https://github.com/stuff"
  chart      = "broken"
  version    = "1.0.0"
  namespace  = "bar"
  set_wo = [
    {
      name  = "thing"
      value = "dont-tell-anyone"
    }
  ]
}

terraform {
  required_version = "~> 1.11"
  required_providers {
    helm = {
      source  = "hashicorp/helm"
      version = ">= 3.0"
    }
  }
}
```

### Debug Output

Here's the errored state: https://gist.github.com/grimm26/a5b5e1599b99df0e9cfa4437352fdc3b

### Expected Behavior

Just work :). Handle ephemeral/wo attributes like they should be.

### Actual Behavior

Didn't work :). Wrote wo attributes to state.

### Steps to Reproduce

1. tofu init
2. tofu test

### Additional Context

_No response_

### References

_No response_

yottta:
Resolves #3964

This fixes the way values are generated during 'tofu test' for write-only attributes.
A write-only attribute should always be null in the response of the provider, which matches the [behavior of the
terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework/blob/4f243a8faa2716d4c34abd6e8b26766ecc920821/internal/fwserver/write_only_nullification.go#L17-L21).

The fix provided here ignores any of the other rules to generate values because a write-only attribute, regardless if it's required/optional/computed, should be null in the returned state (planned state during `plan` or final state during `apply`).

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/hcl2shim/mock_value_composer.go
+++ b/internal/configs/hcl2shim/mock_value_composer.go
@@ -156,7 +156,7 @@ func (mvc MockValueComposer) composeMockValueForAttributes(attrs map[string]*con
 				diags = diags.Append(tfdiags.WholeContainingBody(
 					tfdiags.Warning,
 					fmt.Sprintf("Encountered override for write-only attribute `%v`", k),
-					"Override values for write-only attributes has no impact since the write-only attributes will always be null in the provider response.",
+					"Override values for write-only attributes have no impact since the write-only attributes will always be null in the provider response.",
 				))
 			}
 		} else if attr.Required {


internal/configs/hcl2shim/mock_value_composer.go
```diff
@@ -148,7 +150,9 @@ func (mvc MockValueComposer) composeMockValueForAttributes(attrs map[string]*con
 		}
 
 		// Determine the value
-		if attr.Required {
+		if attr.WriteOnly {
+			mockAttrs[k] = cty.NullVal(attr.ImpliedType())
```
cam72cam, line 154: Should we error if the user tries to set an override for this attr?

yottta: Good question, Since write-only promise stands in having it returned as null from the provider, setting such an override will do nothing, is useless.
I don't think that it worths adding an error for such cases.
Maybe we could add something in the docs?

cam72cam: When useless attrs are set, we typically warn users why it won't work in the mocking system.

yottta: WDYT? fd522945ab6478dacf214b904f36067edcd67195

cam72cam: Looks reasonable to me.  Would be good to have @Yantrio double check as we built this truth table together.

Yantrio: It looks reasonable to me too. I like the approach

internal/configs/hcl2shim/mock_value_composer.go
```diff
@@ -148,7 +150,16 @@ func (mvc MockValueComposer) composeMockValueForAttributes(attrs map[string]*con
...
+			if hasOverride {
+				diags = diags.Append(tfdiags.WholeContainingBody(
+					tfdiags.Warning,
+					fmt.Sprintf("Encountered override for write-only attribute `%v`", k),
+					"Override values for write-only attributes has no impact since the write-only attributes will always be null in the provider response.",
```
Yantrio:
```suggestion
					"Override values for write-only attributes have no impact since the write-only attributes will always be null in the provider response.",
```

yottta: Fixed in 39bf32816ec66168aaf0b0ee347c292cfe00d741

PR #3966, opened 2026-03-31
go.mod: Upgrade to Go 1.25.9

apparentlymart:
The Go team has preannounced that they intend to release Go 1.25.9 on April 7 2026, and that it will contain changes to address the following security advisories that are not yet public:

- CVE-2026-32282
- CVE-2026-32289
- CVE-2026-33810
- CVE-2026-27144
- CVE-2026-27143
- CVE-2026-32288
- CVE-2026-32283
- CVE-2026-27140

We don't know yet if any of these advisories is significant for OpenTofu. I'm opening this PR early so that we can hopefully preapprove it so that we can act quickly on April 7 if at least one of the advisories turns out to be highly-sensitive for OpenTofu.

**Update:** These advisories were not announced as part of this release after all. More information in https://github.com/opentofu/opentofu/pull/3966#issuecomment-4201975382.

---

The checks will, of course, fail until the Go 1.25.9 release is actually published. Once that release is out we should use the GitHub Actions "re-run failed jobs" feature to verify that the checks begin passing.

If we _do_ find that one of these advisories is relevant to OpenTofu then we will also need to add a CHANGELOG entry here about it before merging and releasing. I'm going to add an unresolved comment inline to remind us to think about whether we need to do that.

If we find that none of these advisories are significant for OpenTofu then we would still merge this PR so it would be included in the next patch release for the v1.11 series, but we would not _immediately_ produce that patch release and would instead wait until there's something more significant to release. This is consistent with our policy for upstream security advisories where we only produce patch releases for problems that actually affect OpenTofu's behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/opentofu/opentofu
 
-go 1.25.8
+go 1.25.9
 
 // At the time of adding this configuration, the new Go feature introduced here https://github.com/golang/go/issues/67061,
 // was having a good amount of issues linked to, affecting AWS Firewall, GCP various services and a lot more.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/opentofu/opentofu
 
-go 1.25.8
+go 1.25.9
```
apparentlymart, line 3: Once this release is available we should verify whether any of the advisories covered by this release are relevant to OpenTofu, and if so we'll need to add a changelog entry before merging.

This unresolved comment is there just to remind us not to immediately merge this without a changelog entry, until we've verified that all of the advisories are insignificant for OpenTofu.

apparentlymart:
Go 1.25.9 appears to have been released, but the issues representing the various security advisories were moved to a later milestone before the release was cut and so I assume they are being held back for the next release in each series.

The GitHub Action we use to install Go doesn't seem to be able to download the new release yet -- it's trying to fetch `https://storage.googleapis.com/golang/go1.25.9.linux-amd64.tar.gz`[^1] and getting a `403 Forbidden` error -- but since there doesn't appear to be anything super urgent in [the 1.25.9 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.25.9) I'm just going to hold for now and see if that starts working in the next hour or so.

**Edit:** Ah, we are using an old version of that action which doesn't have https://github.com/actions/setup-go/pull/665. I think it was only working before because the hard-coded URL is only used as a fallback for releases that aren't known to [this action's own manifest](https://github.com/actions/go-versions/blob/main/versions-manifest.json) yet, and this one is too new to be known yet. I'm going to propose updating to a newer version of this action in a separate PR, but I also assume the new Go release will be added to that manifest eventually.

**Edit again:** I've opened https://github.com/opentofu/opentofu/pull/4006 to update the `actions/setup-go` reusable action on the v1.11 branch. After that is merged I intend to rebase this onto it, unless the maintainers of `actions/setup-go` get their manifest updated first. (Their process seems to involve a twice-daily scheduled GitHub Action which opens a PR, and then humans review that PR.)

[^1]: I note that this isn't the same URL that the main download page links to -- the URL on the download page redirects to `https://dl.google.com/go/go1.25.9.linux-amd64.tar.gz` instead -- so I wonder if this URL scheme has changed recently. :thinking:

apparentlymart:
Some time after we merged this there was a release announcement on `golang-dev` which reveals that this _did_ in fact reveal those CVEs, but the issues in the milestone related to them were not annotated in that way and the issues that _had_ been marked in that way were removed from the milestone.

Therefore it seems like we have now got those fixes in the v1.11 branch after all. :man_facepalming: 

I'm going to review them retroactively to determine if any of them are relevant to OpenTofu. If any are then I'll open another pull request to update the changelog so we can make a v1.11 patch release.

PR #3968, opened 2026-03-31
states: go fix various files that changed recently

apparentlymart:
We already changed these files considerably in opentofu/opentofu#3671, so we'll take this opportunity to modernize the rest of the code while the likelihood of successful clean backporting from this file is pretty low anyway, following our policy from https://github.com/opentofu/opentofu/issues/3773.

In the case of `instance_object_src_test.go` some additional editing was required because `uint64Ptr` becomes unused when we adopt the built-in `new` function, which would cause this file to fail the dead code lint.

### DIFF
--- a/internal/states/state_deepcopy.go
+++ b/internal/states/state_deepcopy.go
@@ -62,15 +62,14 @@ func (ms *Module) DeepCopy() *Module {
 	for k, v := range ms.OutputValues {
 		outputValues[k] = v.DeepCopy()
 	}
-	localValues := make(map[string]cty.Value, len(ms.LocalValues))
-	// cty.Value is immutable, so we don't need to copy these.
-	maps.Copy(localValues, ms.LocalValues)
 
 	return &Module{
 		Addr:         ms.Addr, // technically mutable, but immutable by convention
 		Resources:    resources,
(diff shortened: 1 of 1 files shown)


internal/states/state_deepcopy.go
```diff
@@ -61,10 +63,8 @@ func (ms *Module) DeepCopy() *Module {
 		outputValues[k] = v.DeepCopy()
 	}
 	localValues := make(map[string]cty.Value, len(ms.LocalValues))
-	for k, v := range ms.LocalValues {
-		// cty.Value is immutable, so we don't need to copy these.
-		localValues[k] = v
-	}
+	// cty.Value is immutable, so we don't need to copy these.
```
diofeher: I know this comment is not related to your changes, but it sounds confusing to me because we're basically calling Copy after it

diofeher (approved):

internal/states/state_deepcopy.go
apparentlymart: I was intending to avoid making any changes other than what `go fix` proposed, but since the linter already forced me to make some manual changes here anyway I guess I might as well improve this comment at the same time too. :man_shrugging:

PR #3973, opened 2026-04-01
Deprecated resource blocks attrs

### LINKED ISSUE #3971: Support for Resource Schema DeprecatedMessage field in updated Provider Protocol
With the latest provider protocol versions (v5.10, v6.10), a new field DeprecatedMessage has been added to resource schema Atttributes and Body.

We already support the deprecated concept for variables and outputs, but held of on implementing support for Deprecated on resources, primarily because the feature was not fully fleshed out in the provider protocol at the time.

This issue covers pulling this data from the proto into providers.Schema, and then ensuring it makes it's way to the correct locations in the evaluation chain to show up in the areas it's needed.

cam72cam:
Resolves #3971

## Checklist

### Go checklist

 


internal/configs/configschema/implied_type.go
```diff
@@ -61,6 +61,30 @@ func (b *Block) ContainsSensitive() bool {
 	return false
 }
 
+// ContainsDeprecated returns true if any of the attributes of the receiving
+// block or any of its descendent blocks are marked as deprecated.
+//
+// Blocks themselves cannot be deprecated as a whole -- sensitivity is a
+// per-attribute idea -- but sometimes we want to include a whole object
+// decoded from a block in some UI output, and that is safe to do only if
+// none of the contained attributes are deprecated.
```
apparentlymart: I can see that this comment is just an adaptation of the one from `ContainsSensitive` above, which is reasonable given how similar they are, but the second paragraph of this doesn't really seem relevant to deprecation because it's talking about the "safety" of displaying something in the UI, whereas deprecation marks are instead about helping us generate deprecation warnings.

I think we could just drop the second paragraph completely here. The first one seems sufficient documentation for what this function does.

internal/configs/configschema/marks.go
```diff
@@ -21,30 +23,61 @@ func copyAndExtendPath(path cty.Path, nextSteps ...cty.PathStep) cty.Path {
...
 
+func deprecatedBy(by addrs.Referenceable, path cty.Path) string {
+	return by.String() + tfdiags.FormatCtyPath(path)
+}
+
 // ValueMarks returns a set of path value marks for a given value and path,
 // based on the sensitive flag for each attribute within the schema. Nested
 // blocks are descended (if present in the given value).
-func (b *Block) ValueMarks(val cty.Value, path cty.Path) []cty.PathValueMarks {
+func (b *Block) ValueMarks(val cty.Value, path cty.Path, by addrs.Referenceable, isFromRemote bool) []cty.PathValueMarks {
```
apparentlymart: It took me some time reading the function to understand what `by` means here.

Currently it literally just means "the address to use if the block is deprecated", but I guess a more general way to think about it is that it's the address of the symbol that this value will become part of. So maybe it could be called `symbol` and then add something to the docstring above saying that this argument should be an address that a user would understand the given value as belonging to, which will be used for deprecation marks.

internal/configs/configschema/unknown_value.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/zclconf/go-cty/cty"
+)
+
+// UnknownValue returns the "unknown value" for the receiving block, which for
+// a block type is a non-null object where all of the attribute values are
+// the unknown values of the block's attributes and nested block types.
+//
+// In other words, it returns the value that would be returned if an unknown
+// block were decoded against the receiving schema, assuming that no required
+// attribute or block constraints were honored.
```
apparentlymart:
```suggestion
// In other words, it returns the value that would be returned if an unknown
// block were decoded against the receiving schema.
```

(This part about "required attribute constraints" seems to be talking about the fact that an attribute with `Required: true` would not normally be allowed to be null but `Block.EmptyValue` doesn't respect that rule. That situation doesn't matter for unknown values though, because all attributes are allowed to be unknown.)

internal/configs/configschema/validate_traversal_test.go
```diff
@@ -228,7 +228,7 @@ func TestStaticValidateTraversal(t *testing.T) {
 		},
 		{
 			`obj.deprecated`,
-			`Deprecated attribute: The attribute "deprecated" is deprecated. Refer to the provider documentation for details.`,
+			``, // Now handled in the evaluator
```
apparentlymart:
```suggestion
			``, // Handled dynamically in the evaluator, not statically
```

Mainly I just wanted to remove "now" because someone reading this code in the future won't know that we'd been checking this statically in an earlier release. But just "handled in the evaluator" alone didn't seem clear what it was distinguishing from, so I also added "not statically" in the hope of making it easier for a future reader to understand this.

internal/lang/marks/marks.go
```diff
@@ -69,11 +69,11 @@ const TypeType = valueMark("TypeType")
 
 var (
 	_ diagnosticExtraDeprecationCause = DeprecationCause{}
+	_ tfdiags.Keyable                 = DeprecationCause{}
 )
 
 type DeprecationCause struct {
-	By      addrs.Referenceable
-	Key     string
+	By      string
```
apparentlymart, line 76: Do you remember why we'd previously chosen to distinguish the coalescing key from the UI-visible key? I have a feeling we had a reason for doing that at the time, but I don't remember what it was. 🤔 If possible it'd be nice to confirm that whatever our reason had been is no longer valid.

apparentlymart (commented):
You asked (in a different place) for some early reactions to this, so I've left some inline.

Overall this approach seems reasonable to me. My inline comments are probably not really in the spirit of "feedback on the general approach" so feel free to ignore them for now if you don't want to get into these weeds yet, but I hope at least some are useful. 😀

internal/configs/configschema/marks.go
```diff
@@ -114,7 +147,9 @@ func (o *Object) ValueMarks(val cty.Value, path cty.Path) []cty.PathValueMarks {
 
 	for name, attrS := range o.Attributes {
 		// Skip attributes which can never produce sensitive path value marks
```
yottta:
```suggestion
		// Skip attributes which can never produce sensitive or deprecate path value marks
```

internal/configs/configschema/marks.go
```diff
@@ -123,16 +158,28 @@ func (o *Object) ValueMarks(val cty.Value, path cty.Path) []cty.PathValueMarks {
 			// Create a path to this attribute
 			attrPath := copyAndExtendPath(path, cty.GetAttrStep{Name: name})
 
+			var attrMarks []any
 			if attrS.Sensitive {
 				// If the entire attribute is sensitive, mark it so
+				attrMarks = append(attrMarks, marks.Sensitive)
+			}
+			if attrS.Deprecated {
+				// If the entire attribute is sensitive, mark it so
```
yottta:
```suggestion
				// If the entire attribute is deprecated, mark it so
```

internal/configs/configschema/marks.go
```diff
@@ -148,16 +195,28 @@ func (o *Object) ValueMarks(val cty.Value, path cty.Path) []cty.PathValueMarks {
 				// representing multiple collection elements.
 				attrPath := copyAndExtendPath(path, cty.IndexStep{Key: idx}, cty.GetAttrStep{Name: name})
 
+				var attrMarks []any
 				if attrS.Sensitive {
 					// If the entire attribute is sensitive, mark it so
+					attrMarks = append(attrMarks, marks.Sensitive)
+				}
+				if attrS.Deprecated {
+					// If the entire attribute is sensitive, mark it so
```
yottta:
```suggestion
					// If the entire attribute is deprecated, mark it so
```

internal/tofu/evaluate.go
```diff
@@ -795,27 +795,19 @@ func (d *evaluationStateData) GetResource(ctx context.Context, addr addrs.Resour
 			}
 
 		default:
+			if config.Count != nil || config.ForEach != nil {
+				return cty.DynamicVal, diags
```
yottta: This seems wrong 🤔 
If this gets the information from an ephemeral resource, not having the `DynamicVal` marked as ephemeral will not flag the misusage of it during validation.

Check also the big removed comment.

internal/configs/configschema/unknown_value.go
```diff
@@ -0,0 +1,45 @@
...
+func (b *Block) UnknownValue() cty.Value {
```
yottta, line 18: Isn't this method purpose, arguably, the same with the purpose of [`objchange.PlannedUnknownObject`](https://github.com/opentofu/opentofu/blob/8a7ba57fef580e67adeb2823a160d3686c35c9ce/internal/plans/objchange/objchange.go#L75) called with a nil config?

internal/configs/configschema/empty_unknown_value_test.go
```diff
@@ -16,12 +16,14 @@ import (
 
 func TestBlockEmptyValue(t *testing.T) {
```
yottta:
```suggestion
func TestBlockEmptyAndUnknownValue(t *testing.T) {
```

internal/lang/marks/marks.go
yottta: Might be because of the removed comment several lines below:
> // Used to identify the output on the consolidation diagnostics and
>  // make sure they are consolidated correctly. We use:
>  // output value + deprecated message as the key

Maybe we should still keep a comment like that somewhere.

internal/lang/marks/marks.go
```diff
@@ -160,10 +162,11 @@ func ExtractDeprecationDiagnosticsWithBody(val cty.Value, body hcl.Body) (cty.Va
 		diag := tfdiags.AttributeValue(
 			tfdiags.Warning,
 			"Value derived from a deprecated source",
-			fmt.Sprintf("This value is derived from %v, which is deprecated with the following message:\n\n%s", cause.By, cause.Message),
+			fmt.Sprintf("This value is derived from %s, which is deprecated with the following message:\n\n%s", cause.By, cause.Message),
 			path,
 		)
 		diags = diags.Append(tfdiags.Override(diag, tfdiags.Warning, func() tfdiags.DiagnosticExtraWrapper {
+			println(cause.By)
```
yottta:
```suggestion
```

internal/lang/marks/marks.go
```diff
@@ -197,10 +200,11 @@ func ExtractDeprecatedDiagnosticsWithExpr(val cty.Value, expr hcl.Expression) (c
 			source += "'s attribute " + attr
 		}
 		cause := dm.Cause
+		println(cause.By)
```
yottta:
```suggestion
```

internal/tofu/context_plan.go
```diff
@@ -1016,7 +1017,7 @@ func (c *Context) driftedResources(ctx context.Context, config *configs.Config,
 				var oldVal, newVal cty.Value
 				oldVal = oldObj.Value
 				if newObj != nil {
-					newVal = newObj.Value
+					newVal = marks.RemoveDeepDeprecated(newObj.Value)
```
yottta: This is ok assuming that reaching this point, we already generated the diagnostics for the deprecation marks.
Is this way of understanding this complete and correct?

internal/tofu/evaluate.go
yottta: I [updated](https://github.com/opentofu/opentofu/commit/57841103fb33cea34c856fc55eaf32abca0b1944) one of the tests to validate this and on `main` works as expected.

I added that test into a branch based on yours, so if you want to incorporate those changes, [here is the diff](https://github.com/opentofu/opentofu/compare/deprecated_resource_blocks_attrs...enhance-get-resource-test?expand=1) against your branch for easy merge.

internal/lang/marks/marks.go
cam72cam: You can see some of the ramifications of this change in the e2e test changes I made.

From what I can tell / remember, we chose different Keys in different areas to help with the coalescing of different deprecated diagnostics.  In practice however, I think that it was a mistake and led to inconsistent behavior.

For example, the output deprecation was just the output variable name. This was intended to handle the scenario where you are using multiple instances of the same module under different names.  The way it was implemented however, caused any deprecated outputs that shared a name in any modules (including root) to be coalesced together, regardless of actual source.

This change makes deprecated diagnostics a bit more verbose, but I think the consistency and correctness trade off is worth it.   If we were to re-add Key to enable more aggressive coalescing, I would insist that we make it represent { module_source, addrs.Referenceable }.

yottta (changes requested):
~~Something might be missing and don't know exactly where: `configschema.Block.ValueMarks` is always called with `false` for the `isRemote` argument.
The only places where that argument receives a variable is when it recurses.
Please double check.~~

EDIT: That is mentioned in the TODOs from the description of this PR. Missed it.

internal/configs/configschema/unknown_value.go
cam72cam: No, this generates a stubbed out unknown value, the code you linked just returns `cty.UnknownVal(schema.ImpliedType())`

yottta: True, didn't see the `config == nil` check. You're right.

internal/tofu/context_plan.go
cam72cam: That is correct, this code just handles detecting drift and should not care about deprecated marks.

internal/tofu/evaluate.go
cam72cam: fixed in 5d2af34fd6

internal/configs/configschema/implied_type.go
cam72cam: Fixed in aa8c4db676

internal/configs/configschema/marks.go
cam72cam: Renamed in eb0f918222

internal/lang/marks/marks.go
cam72cam: I've captured my current thoughts in https://github.com/opentofu/opentofu/pull/3996.  I'm hesitant to do that large of a rework right before a release.

website/docs/language/resources/syntax.mdx
```diff
@@ -257,3 +257,37 @@ for configuration, if any.
 Many managed resources now offer write-only attributes as an alternative to older attributes which existed before the concept of ephemeral was added.
 As an example, see [aws_secretsmanager_secret_version](https://search.opentofu.org/provider/hashicorp/aws/v6.11.0/docs/resources/secretsmanager_secret_version) and its `secret_string_wo` alternative for `secret_string`).
 Additional documentation is available for [write-only attributes](../ephemerality/write-only-attributes.mdx)
+
+## Deprecation of blocks and attributes
+
+Over time, provider authors may need to change the structure of resources provided. Ideally, they will mark attributes and blocks as deprecated before removing them completely. This notifies consumers of the deprecated
+fields, providing time to migrate. OpenTofu utilizes the provider's schema to track usage of any deprecated attributes or blocks and will produce a clear warning.
```
apparentlymart:
```suggestion
Over time, provider authors may need to change the schema of some of their resource types. Ideally they will mark attributes and blocks as deprecated before removing them completely. This notifies consumers of the deprecated
features, providing time to migrate. OpenTofu produces a warning each time it finds a reference to a deprecated provider feature.
```

Just some terminology fiddling: "resource types" are what has schema, and "schema" is how we discuss the attributes and blocks of a resource type in other parts of UI and documentation. We don't normally use the word "field" to refer to parts of a provider schema, but saying "attributes and blocks" repeatedly seemed too verbose and so I used "features" here as a general (non-jargon) way to refer to them both collectively.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ ENHANCEMENTS:
...
 - Comparing an object or other complex-typed value to `null` using the `==` operator now returns a sensitive boolean result only if the object as a whole is sensitive, and not when the object merely contains a sensitive value nested inside one of its attributes. This means that comparisons to null can now be used in parts of the configuration where sensitive values are not allowed, such as in the `enabled` meta-argument on resources and modules. ([#3793](https://github.com/opentofu/opentofu/pull/3793))
 - Resources using `replace_triggered_by` in their `lifecycle` block are now replaced when a resource they refer to is itself being replaced, whereas before this triggered only when it was being updated. ([#3714](https://github.com/opentofu/opentofu/issues/3714))
+- Deprecated attributes and blocks in resources now produce warnings when utilized. They can be filtered/disabled with the `-deprecation=` argument. ([#3973](https://github.com/opentofu/opentofu/pull/3973/))
```
```suggestion
- OpenTofu now produces warnings for any references to attributes or blocks of a resource type that are marked as deprecated in the provider schema, unless disabled by the `-deprecation=` option. ([#3973](https://github.com/opentofu/opentofu/pull/3973/))
```

apparentlymart (approved):
This looks good to me!

My inline comments are just prose nits. As usual they are subjective, so you can ignore or modify them as you see fit.

yottta (approved):
This looks good. Looking forward to check https://github.com/opentofu/opentofu/pull/3996 too, once will be ready.

internal/plugin/convert/schema.go
```diff
@@ -23,24 +23,26 @@ import (
...
-			WriteOnly:       a.WriteOnly,
+			Name:               name,
+			Description:        a.Description,
+			DescriptionKind:    protoStringKind(a.DescriptionKind),
+			Optional:           a.Optional,
+			Computed:           a.Computed,
+			Required:           a.Required,
+			Sensitive:          a.Sensitive,
+			Deprecated:         a.Deprecated,
+			DeprecationMessage: b.DeprecationMessage,
```
Yantrio: I think this should be

```suggestion
			DeprecationMessage: a.DeprecationMessage,
```

to match what we're doing with plugin6

cam72cam: Gah, good catch!  Fixed in 14f04bdcd7

Yantrio (approved):

internal/tofu/deprecation_level.go
```diff
@@ -29,35 +29,51 @@ const (
 	DeprecationWarningLevelNone
 )
 
+type DeprecationDiagnosticAllowedSeen map[any]any
```
yottta, line 33: `suggestion`
Would be nice to have a comment on this type to make clear what its purpose is.

yottta (approved):

internal/tofu/deprecation_level.go
cam72cam: Added in 483e59cf34

apparentlymart:
FWIW a similar feature in Terraform got some feedback in https://github.com/hashicorp/terraform/issues/38521. (EDIT: and again in https://github.com/hashicorp/terraform/issues/38554)

If we wanted to compromise to avoid that situation in OpenTofu, one way we could consider is to make output values only check for _shallow_ deprecation marks and then wait to see if the leaf value ends up getting used in some other place outside the module before warning about it. That means that the author of the module containing the output value would not necessarily get warned during their own testing that they are exporting something that contains a deprecated value somewhere inside it, but in this case that seems desirable since presumably the intent of the module author was to just export the entirety of whatever the provider returned and delegate the details of how to use it to the caller.

Doing that naively would also fail to catch a situation where the deprecated value is _explicitly_ placed deeply in a result, though:

```hcl
output "example" {
  value = {
    fine       = data.whatever.whatever.fine
    # arguably the following line should still generate deprecation warning
    # because there's a direct reference to the deprecated thing.
    deprecated = data.whatever.whatever.deprecated
  }
}
```

So, some other potential compromises/heuristics:

- Do the shallow check for deprecation marks only when the `value` expression is literally just a direct reference to a resource or resource instance, treating that as evidence that the author's intent was just to return whatever the provider returned and let the calling module deal with any deprecations.

    This would narrowly support this specific situation without changing any other behavior, but it's annoying to have a weird special case like this.
- Combine the marks-based deprecation check with a secondary check based on static analysis of references, so that `data.whatever.whatever.deprecated` would get detected as directly referenced in the above even though it appears deeply inside the output value.

    (I don't think this particular one is really workable though, because it wouldn't be able to deal with `data.whatever.whatever[anything_dynamic].deprecated` due to the limitations of our static analysis.)

I don't know if it's actually necessary to do _any_ of what I described here, but just mainly wanted to mention the feedback for now in case we get similar feedback after shipping our own flavor of it.

PR #3974, opened 2026-04-01
Add warning when provider symlinks fail

### LINKED ISSUE #3972: Windows users using the provider cache directory can experience confusion due to the silent fallback to copying instead of symlinking
### Community note

> [!TIP]

### OpenTofu Version

```shell
Latest
```

# Any configs

### Debug Output

The lack of debug/warn output is the issue.

### Expected Behavior

When OpenTofu fails to symlink a provider and falls back to copying (due to being in particular Windows setups), it should emit some sort of warning, etc. to let users know that they are not going to save any disk space (barring having administrator privileges, being in developer mode or using WSL).

### Actual Behavior

OpenTofu should signal to the user (perhaps on an opt-in debugging basis) that it has fallen back to copying instead of symlinking when symlinking fails.

### Steps to Reproduce

Use the provider cache directory on Windows on a machine with a user that doesn't have symlink permissions.

### Additional Context

This was reported as a bug in Terragrunt (https://github.com/gruntwork-io/terragrunt/issues/5061).

Chatted with Christian about whether there's a silent fallback in OpenTofu when symlinking fails, and he was able to confirm there is. He promised I didn't need to make this bug report very fleshed out to track the work to address it.

### References

- https://github.com/gruntwork-io/terragrunt/issues/5061

cam72cam:
Resolves #3972 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/getproviders/package_location_local_dir.go
+++ b/internal/getproviders/package_location_local_dir.go
@@ -131,10 +131,11 @@ func (p PackageLocalDir) InstallProviderPackage(_ context.Context, meta PackageM
 	// If we get down here then symlinking failed and we need a deep copy
 	// instead. To make a copy, we first need to create the target directory,
 	// which would otherwise be a symlink.
-	log.Printf("[WARN] Unable to create provider symlink: %s. Falling back to making a copy.", err)
+	warning := fmt.Sprintf("[WARN] Failed to create provider symlink: %s. Falling back to copying the package.", err)
 	if runtime.GOOS == "windows" {
-		log.Print("[WARN] In certain circumstances, the operating system or your organization may prevent this as a policy. See https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/create-symbolic-links for more details")
+		warning += " Refer to https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/create-symbolic-links for limitations of symlink support on Windows."
 	}
+	log.Print(warning)
 
 	//nolint:mnd // magic number predates us using this linter
 	err = os.Mkdir(absNew, 0755)


cam72cam:
@yhakbar I give you permission to pull this solution verbatim into the MIT licensed terragrunt project.

Edit: this is for the original implementation found here: https://github.com/opentofu/opentofu/commit/8a3a8f76cb00ec16b17d6d4c7d66a0d567eec7c1

internal/command/init.go
```diff
@@ -926,6 +936,35 @@ func (c *InitCommand) getProviders(ctx context.Context, config *configs.Config,
...
+				diags = diags.Append(tfdiags.Sourceless(tfdiags.Warning, "Unable to check for symlink support", err.Error()))
+				return
+			}
+
+			err = os.Symlink(source, dest)
+			if err != nil {
+				diags = diags.Append(tfdiags.Sourceless(
+					tfdiags.Warning,
+					"Unable to create symlinks",
+					fmt.Sprintf("TF_PLUGIN_CACHE_DIR is set, but unable to be fully utilized. The disk space savings are prevented due to inability to create symlinks on this system: %s", err.Error()),
```
yhakbar: This should probably call out that Windows has limitations on creating symlinks and offer users a link to docs to investigate further.

apparentlymart: This message is also assuming that the cache directory was enabled by the environment variable, but it's also possible to enable it with a CLI configuration setting. Maybe we should just say "The global plugin cache is enabled" without saying _how_ it's enabled, since code in this location can't distinguish between those two ways of enabling it. 🤔

cam72cam: Per the discussion in the issue, I've reworked this PR to emit a log warning within the getproviders package.

internal/getproviders/package_location_local_dir.go
```diff
@@ -129,6 +131,11 @@ func (p PackageLocalDir) InstallProviderPackage(_ context.Context, meta PackageM
 	// If we get down here then symlinking failed and we need a deep copy
 	// instead. To make a copy, we first need to create the target directory,
 	// which would otherwise be a symlink.
+	log.Printf("[WARN] Unable to create provider symlink: %s. Falling back to making a copy.", err)
+	if runtime.GOOS == "windows" {
+		log.Print("[WARN] In certain circumstances, the operating system or your organization may prevent this as a policy. See https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/create-symbolic-links for more details")
```
apparentlymart: I'd suggest reworking this to produce just one warning containing all of the information, but generating that warning differently if we're running on Windows.

I suggest this because I believe there can now (after some recent work) be multiple operations happening concurrently here and so these two log lines might end up getting split apart by log lines generated by other goroutines, and this second warning just says "this" instead of talking specifically about symlinks, so if it were separated it would be hard to understand what it's talking about.

Maybe something like this?

```go
var warningExtra string
    warningExtra = " Refer to https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/create-symbolic-links for limitations of symlink support on Windows."
}
log.Printf("[WARN] Failed to create provider package symlink from global plugin cache: %s. Making a copy of the package instead.%s.", err, warningExtra)
```

cam72cam: Changed in d383986e49

PR #3981, opened 2026-04-02
rfc: prerelease provider version constraints

mcanevet:
For #1651

This RFC proposes adopting the RubyGems `Dependency#match?` opt-in convention for provider version constraints: prereleases are included in matching if and only if at least one version boundary in the constraint set is itself a prerelease (e.g. `>= 1.0.0-0`).

The change is limited to provider version constraints. Module constraints are deferred pending resolution of #2147.

## Checklist

### DIFF
--- a/rfc/20260402-prerelease-provider-version-constraints.md
+++ b/rfc/20260402-prerelease-provider-version-constraints.md
@@ -1,0 +1,194 @@
+# Prerelease Provider Version Constraints
+
+Issue: [https://github.com/opentofu/opentofu/issues/1651](https://github.com/opentofu/opentofu/issues/1651)
+
+Currently, prerelease versions (e.g. `1.0.0-beta.1`, `1.0.0-rc.1`) can only be selected by
+an exact version constraint (`= 1.0.0-beta.1` or `1.0.0-beta.1`). Inexact operators such as
+`>=`, `~>`, `!=` silently exclude prerelease versions even when a prerelease suffix is
+included in the constraint boundary. For example, `>= 1.0.0-0` behaves identically to
+`>= 1.0.0`: no prerelease versions are ever matched.
+
+This makes it impossible to track a rolling prerelease channel (e.g. always use the latest
+`1.0.x` beta) without manually updating an exact pin on every new prerelease.
+
+The version constraint syntax OpenTofu uses is modeled on RubyGems. RubyGems itself has a
+well-defined opt-in convention for prerelease matching, but OpenTofu never implemented it.
+
+Issue [#2147](https://github.com/opentofu/opentofu/issues/2147) also touches on prerelease
+version quirks, but specifically in the module installer, where two competing parsers create
+additional complexity. This RFC addresses only provider version constraints, where the code
+path is simpler and the risk of unintended side effects is lower. Module constraints are left
+for a follow-up once #2147 is resolved.
+
+## Proposed Solution
+
+Adopt the RubyGems `Dependency#match?` opt-in convention for provider version constraints:
+prereleases are included in matching if and only if at least one version boundary in the
+constraint set is itself a prerelease. The choice of operator does not affect this decision
+(see Open Questions for edge cases around `!=`).
+
+### User Documentation
+
+By default, version constraints only match stable (non-prerelease) releases:
+
+```hcl
+version = ">= 1.0.0"   # matches 1.0.0, 1.0.1, 2.0.0, ... (no prereleases)
+version = "~> 1.0"     # matches 1.x stable releases only
+```
+
+To opt in to prerelease versions, include a prerelease suffix on at least one boundary. The
+conventional way to do this is the `-0` suffix, which is the lowest possible prerelease
+identifier and therefore matches all prereleases of that version:
+
+```hcl
+version = ">= 1.0.0-0"      # matches 1.0.0-alpha, 1.0.0-beta.1, 1.0.0-rc.1, 1.0.0, 1.0.1, ...
+version = "~> 1.0.0-0"      # matches prereleases and stable releases in the 1.0.x range
+```
+
+Semver ordering applies within prereleases, so more specific lower bounds work as expected:
+
+```hcl
+version = ">= 1.0.0-beta.2"  # matches 1.0.0-beta.2, 1.0.0-rc.1, 1.0.0, ... but NOT 1.0.0-beta.1
+version = ">= 1.0.0-beta.1"  # matches 1.0.0-beta.1, 1.0.0-beta.2, 1.0.0-rc.1, 1.0.0, ...
+```
+
+> **⚠️ Upper bound warning:** Because semver defines prerelease versions as _less than_ their
+> corresponding release (e.g. `2.0.0-beta.1 < 2.0.0`), an upper bound without a prerelease
+> suffix does **not** exclude prereleases of that version. For example:
+>
+> ```hcl
+> version = ">= 1.0.0-0, < 2.0.0"   # also matches 2.0.0-beta.1, 2.0.0-rc.1, etc.!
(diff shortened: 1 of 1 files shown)


rfc/20260402-prerelease-provider-version-constraints.md
```diff
@@ -0,0 +1,194 @@
...
+**Known limitation:** For `~> X.Y.Z-pre`, the go-versions library computes the upper bound
+as `OlderThan(X.Y+1.0)`. Because semver defines prerelease < release, versions like
+`X.Y+1.0-beta.1` are numerically less than `X.Y+1.0` and thus fall inside the range. In
+RubyGems, the `~>` upper bound uses `v.release < r.bump`, which excludes such versions.
+This edge case only arises when using `~>` with a prerelease boundary, which is an unusual
+pattern in practice.
+
+### Open Questions
+
+1. **Upper bound behavior:** When a constraint opts in (e.g. `>= 1.0.0-0, < 2.0.0`), the
```
yottta: As long as this will be documented properly, I would be ok with going with the `< 2.0.0-0` for this.

rfc/20260402-prerelease-provider-version-constraints.md
```diff
@@ -0,0 +1,194 @@
...
+   `2.0.0-beta.1 < 2.0.0`. This is consistent with RubyGems but likely surprising to users
+   who expect `< 2.0.0` to mean "strictly below the 2.0.0 stable release". Should we
+   deviate from RubyGems here and apply `Released` filtering to upper-bound constraints even
+   when the lower bound opts in? Or is documenting the `< 2.0.0-0` pattern sufficient?
+
+2. **`!=` with prerelease boundary:** A constraint like `!= 1.0.0-bad` opts in globally,
+   enabling all other prereleases — likely not the intent. Should `!=` be excluded from the
+   opt-in rule, or is it acceptable as documented behavior?
+
+3. Should OpenTofu emit a **warning** when a prerelease-opting-in constraint is used, to
```
yottta: TBH, considering the risk this change brings in, I would suggest rolling this out in 3 phases:
* When introduced, make it accessible only if a specific env variable is configured accordingly. Defaults to the current behavior
* In the next development cycle, make it default and use that previous env var to be used to go back to the previous version
* In the next development cycle, remove the old implementation, default the new one and remove the env var.

rfc/20260402-prerelease-provider-version-constraints.md
```diff
@@ -0,0 +1,194 @@
...
+### Open Questions
+
+1. **Upper bound behavior:** When a constraint opts in (e.g. `>= 1.0.0-0, < 2.0.0`), the
+   upper bound `< 2.0.0` also includes `2.0.0-beta.1` because semver defines
+   `2.0.0-beta.1 < 2.0.0`. This is consistent with RubyGems but likely surprising to users
+   who expect `< 2.0.0` to mean "strictly below the 2.0.0 stable release". Should we
+   deviate from RubyGems here and apply `Released` filtering to upper-bound constraints even
+   when the lower bound opts in? Or is documenting the `< 2.0.0-0` pattern sufficient?
+
+2. **`!=` with prerelease boundary:** A constraint like `!= 1.0.0-bad` opts in globally,
```
yottta: Well, I would say that this is quite annoying use case.
What about `>= 1.0.0-0, != 1.0.0-bad`? If it's the case, then yeah, we need to be able to select prereleases, right? Because of the first constraint.
So I would say that the `!=` operator should just exclude the given version and that's it.

rfc/20260402-prerelease-provider-version-constraints.md
```diff
@@ -0,0 +1,194 @@
...
+   enabling all other prereleases — likely not the intent. Should `!=` be excluded from the
+   opt-in rule, or is it acceptable as documented behavior?
+
+3. Should OpenTofu emit a **warning** when a prerelease-opting-in constraint is used, to
+   make the behavior change visible to users upgrading from older versions?
+
+4. The `~> X.Y.Z-pre` upper bound edge case (see Technical Approach): is the divergence from
+   RubyGems acceptable, or should it be addressed?
+
+5. Should the opt-in eventually be extended to **module** version constraints as a follow-up
```
yottta: Maybe, if this RFC will be accepted, we can add a note in [#2147](https://github.com/opentofu/opentofu/issues/2147) to take note of a similar improvement there.

yottta (commented):
Thanks for starting this and sorry for the delay.

mcanevet:
Thanks for the review, @yottta!

I've updated the RFC to address your feedback:

Upper bound behavior (Q1): Agreed — documenting the `< 2.0.0-0` convention is sufficient and keeps us aligned with RubyGems. Removed the open question and tightened the warning box to make the guidance more direct.

`!=` opt-in behavior (Q2): Makes sense. `!=` is now explicitly excluded from the opt-in check — only boundary operators (`>=`, `>`, `<=`, `<`, `~>`) can trigger prerelease matching. A `!= 1.0.0-bad` constraint by itself does nothing to the opt-in; it just excludes that version from whatever set the remaining constraints produce. Your example (`>= 1.0.0-0, != 1.0.0-bad`) works exactly as expected: opt-in driven by the first constraint, exclusion applied on top.

Rollout plan: Added a three-phase env-var-gated rollout as you suggested (`OPENTOFU_EXPERIMENT_PRERELEASE_CONSTRAINTS`). I used a generic experiment variable name — happy to discuss the name if you have a preference.

Module follow-up (#2147): Added a note in Future Considerations to file a cross-reference on #2147 once this RFC is accepted.

The two remaining open questions (whether to emit a warning on opt-in, and the `~> X.Y.Z-pre` upper bound edge case) are lower priority — any thoughts welcome.

rfc/20260402-prerelease-provider-version-constraints.md
```diff
@@ -141,30 +158,20 @@ pattern in practice.
...
-   who expect `< 2.0.0` to mean "strictly below the 2.0.0 stable release". Should we
-   deviate from RubyGems here and apply `Released` filtering to upper-bound constraints even
-   when the lower bound opts in? Or is documenting the `< 2.0.0-0` pattern sufficient?
-
-2. **`!=` with prerelease boundary:** A constraint like `!= 1.0.0-bad` opts in globally,
-   enabling all other prereleases — likely not the intent. Should `!=` be excluded from the
-   opt-in rule, or is it acceptable as documented behavior?
-
-3. Should OpenTofu emit a **warning** when a prerelease-opting-in constraint is used, to
+1. Should OpenTofu emit a **warning** when a prerelease-opting-in constraint is used, to
```
yottta, line 161: We could have 2 things here:
1. A log statement (maybe a debug or info level)
2. A warning that could be linked to the rollout plan: in the 1st phase that will be shown only if the proposed env var will be enabled. In the 2nd phase it could be issued all the time and skipped if the env var will be turned off. In the 3rd phase we could remove it altogether and leave only the log statement

But let's wait for other maintainers takes on this.

Yantrio:
Hey @mcanevet, thanks for taking the time to put this RFC together and apologies that it took us a while to circle back to you on this.

We discussed this during the triage meeting this week amongst the maintainers and the decision we made is that we've decided to not go in this direction. We do not feel that the amount of work required to add and maintain such a feature would be worth the maintenance cost in the long term.

One other thing to note for the future: for issues that carry the `needs-rfc` tag we would really prefer if you checked with the maintainers first before putting effort into contributing. We're always happy to have such conversations up front.

I noticed also that you did not check the "I have no used an AI coding assistant to create this PR" box above in the template. Right now this repository is not accepting AI-assisted contributions so this is something to be aware of too.

Thanks again for the RFC and engaging with our process 🙏

PR #3982, opened 2026-04-02
Migrate `cloud` and `remote` backends to the `views` package

yottta:
Part of #3765

Several things to note:
* Because there is no json output already for the remote/cloud backends, I added no such implementation.
  * Therefore, the human implementation can be created from outside the `views` package, despite the rule that I want to enforce on the `views` package (ie: only command related views can be create from outside of the `views` package, while the non-command views [like backend, stateLocker, backendRemote, etc] can be created only by command specific views). Added a TODO for later on this.
* Because I removed the colorisation object from the `Remote` and `Cloud` objects, for some tests I had to add a sanitizer function to remove those markers from the produced and expected outputs.
* Due to the fragmented nature of these backends' outputs, I had to add in the view generic methods to allow printing any string content, since there are a lot of places in the backends where these are built dynamically.
* The tests updated with `UIInput` answers, were not failing previously, because the UI into the meta was not configured, Ui that was passed down the line into the backend Operation where it was used to determine if confirmation can be asked or not. Now, with the changes from the commits in this PR, the UI is always there so the confirmation is asked now. In day by day operations, this is not a valid situation, since there is always an UI object configured (View now, after migration).
 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/cloud/backend_apply_test.go
+++ b/internal/cloud/backend_apply_test.go
@@ -103,7 +103,7 @@ func TestCloud_applyBasic(t *testing.T) {
 	}
 
 	output := voutput.Stdout()
-	if !strings.Contains(output, "Running apply in the cloud backend") {
+	if !strings.Contains(output, "Running apply in cloud backend") {
 		t.Fatalf("expected TFC header in output: %s", output)
 	}
 	if !strings.Contains(output, "1 to add, 0 to change, 0 to destroy") {
@@ -698,7 +698,7 @@ func TestCloud_applyWithRequiredVariables(t *testing.T) {
 
 	voutput := done(t)
 	output := voutput.Stdout()
-	if !strings.Contains(output, "Running apply in the cloud backend") {
+	if !strings.Contains(output, "Running apply in cloud backend") {
 		t.Fatalf("unexpected TFC header in output: %s", output)
 	}
 }
@@ -850,7 +850,7 @@ func TestCloud_applyAutoApprove(t *testing.T) {
 	}
 
 	output := voutput.Stdout()
-	if !strings.Contains(output, "Running apply in the cloud backend") {
+	if !strings.Contains(output, "Running apply in cloud backend") {
 		t.Fatalf("expected TFC header in output: %s", output)
 	}
 	if !strings.Contains(output, "1 to add, 0 to change, 0 to destroy") {
@@ -919,7 +919,7 @@ func TestCloud_applyApprovedExternally(t *testing.T) {
 	}
 
 	output := voutput.Stdout()
-	if !strings.Contains(output, "Running apply in the cloud backend") {
+	if !strings.Contains(output, "Running apply in cloud backend") {
 		t.Fatalf("expected TFC header in output: %s", output)
 	}
 	if !strings.Contains(output, "1 to add, 0 to change, 0 to destroy") {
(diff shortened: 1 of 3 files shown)


internal/command/meta_backend.go
```diff
@@ -479,7 +477,6 @@ func (m *Meta) Operation(ctx context.Context, b backend.Backend, view views.Back
 		Encryption:      enc,
 		PlanOutBackend:  planOutBackend,
 		UIIn:            m.UIInput(),
-		UIOut:           m.Ui,
```
yottta, line 482: The `UIOut` was used strictly for checks to verify if the user has an UI or not. No calls are done on this.
Instead, the `View` field inside (of type `views.Operation`) is what is used to print information to the user.

internal/command/apply_destroy_test.go
```diff
@@ -438,6 +438,7 @@ func TestApply_destroySkipInConfigAndState(t *testing.T) {
 		},
 	}
 
+	t.Cleanup(testInputMap(t, map[string]string{"approve": "yes"}))
```
yottta, line 441: Because now the `CLIOpts.View` value is not relying on how the test is configured (ie: before this PR, having null `command.Meta.Ui` lead to having null `UI` in the `CLIOpts`), once the execution reaches the confirmation step, now it asks for it, so we had to add this for some failing tests.

internal/cloud/backend_taskStage_taskResults_test.go
```diff
@@ -26,17 +27,22 @@ func (s *testIntegrationOutput) End() {
...
-	s.output.WriteString(s.ctx.B.Colorize().Color("[reset]│ ") + str + "\n")
+	s.output.WriteString(removeColoriseMarks(str) + "\n")
 }
 
 func (s *testIntegrationOutput) OutputElapsed(message string, maxMessage int) {
 	s.output.WriteString("PENDING MESSAGE: " + message)
 }
 
+func removeColoriseMarks(in string) string {
+	re := regexp.MustCompile(`\[.*?\]`)
```
cam72cam, line 42: Perhaps we should use a more specific color matcher like in https://github.com/opentofu/opentofu/pull/3606/changes#diff-6f7b657aca11193775757e97ade76b180e4846bcdaf948174814806f63365e0fR29-R38?

yottta:
> To be honest I'm pretty nervous about changing this code in such a significant way since from our perspective it's close to being a legacy system: it's an implementation of an API we don't know much about, and as far as we know it's not very commonly used and so if we're introducing subtle regressions here it'll probably be a long time before we notice them.

Thanks for saying this! 🙌  Having another one feeling similarly about this, just made me revisit everything.

I checked again the code paths and there is indeed something that I might have missed: `Cloud` and `Remote` don't have the `View` given during initialisation ([`Cloud`](https://github.com/opentofu/opentofu/blob/c1b7229bbd9886c8cf3414ee716f8e868d3bae81/internal/cloud/backend.go#L121) and [`Remote`](https://github.com/opentofu/opentofu/blob/c1b7229bbd9886c8cf3414ee716f8e868d3bae81/internal/backend/remote/backend.go#L108)), but during the `CLIInit` ([`Cloud`](https://github.com/opentofu/opentofu/blob/c1b7229bbd9886c8cf3414ee716f8e868d3bae81/internal/cloud/backend_cli.go#L14) and [`Remote`](https://github.com/opentofu/opentofu/blob/c1b7229bbd9886c8cf3414ee716f8e868d3bae81/internal/backend/remote/cli.go#L13)).
Therefore, there is a slim chance for the logic that involved `View` calls, to be called before the `CLIInit` call, so I restored the nil checks just to be sure. Pushed in bf8d382d2622f80d3bf25aa79380650f82311036.

EDIT: Reverted one more change that could have altered the output: 8d8098a6147c01f99c2614ab954e35df942da79a

internal/cloud/backend_taskStage_taskResults_test.go
yottta: It doesn't work because the output is never colorised, therefore the squared brackets marks are never replaced.

PR #3984, opened 2026-04-02
go.mod: Bump google deps

cam72cam:
Ensured tests were run and pass with

`TF_ACC=1 GOOGLE_PROJECT=<project> go test -v ./internal/backend/remote-state/gcs/`

Part of #3877 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/gcs/backend.go
+++ b/internal/backend/remote-state/gcs/backend.go
@@ -178,6 +178,10 @@ func (b *Backend) configure(ctx context.Context) error {
 			return fmt.Errorf("the string provided in credentials is neither valid json nor a valid file path")
 		}
 
+		//nolint:staticcheck // [option.WithCredentialsJSON] is deprecated
+		// upstream, but we're waiting to see how the hashicorp/google provider
+		// reacts to that deprecation so we can then mimic their strategy as
+		// closely as possible.
 		credOptions = append(credOptions, option.WithCredentialsJSON([]byte(contents)))
 	}
 
--- a/internal/backend/remote-state/gcs/backend_test.go
+++ b/internal/backend/remote-state/gcs/backend_test.go
@@ -458,6 +458,10 @@ func testGetClientOptions(t *testing.T) ([]option.ClientOption, error) {
 	if !json.Valid([]byte(contents)) {
 		return nil, fmt.Errorf("the string provided in credentials is neither valid json nor a valid file path")
 	}
+	//nolint:staticcheck // [option.WithCredentialsJSON] is deprecated
+	// upstream, but we're waiting to see how the hashicorp/google provider
+	// reacts to that deprecation so we can then mimic their strategy as
+	// closely as possible.
 	credOptions = append(credOptions, option.WithCredentialsJSON([]byte(contents)))
 	opts = append(opts, credOptions...)
 	opts = append(opts, option.WithUserAgent(httpclient.OpenTofuUserAgent(version.Version)))
--- a/internal/encryption/keyprovider/gcp_kms/config.go
+++ b/internal/encryption/keyprovider/gcp_kms/config.go
@@ -105,6 +105,10 @@ func (c Config) Build() (keyprovider.KeyProvider, keyprovider.KeyMeta, error) {
 			return nil, nil, &keyprovider.ErrInvalidConfiguration{Message: "the string provided in credentials is neither valid json nor a valid file path"}
 		}
 
+		//nolint:staticcheck // [option.WithCredentialsJSON] is deprecated
+		// upstream, but we're waiting to see how the hashicorp/google provider
+		// reacts to that deprecation so we can then mimic their strategy as
+		// closely as possible.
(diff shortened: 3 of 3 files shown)


apparentlymart:
I just quickly did the `go mod tidy` myself here in the hope that would allow us to merge and move on, but unfortunately the new version of at least one of these libraries has some symbols marked as deprecated and we have a lint rule that requires us not to use deprecated symbols:

```
Error: internal/backend/remote-state/gcs/backend.go:181:37: SA1019: option.WithCredentialsJSON is deprecated: This function is being deprecated because of a potential security risk. (staticcheck)
		credOptions = append(credOptions, option.WithCredentialsJSON([]byte(contents)))
		                                  ^
Error: internal/backend/remote-state/gcs/backend_test.go:461:36: SA1019: option.WithCredentialsJSON is deprecated: This function is being deprecated because of a potential security risk. (staticcheck)
	credOptions = append(credOptions, option.WithCredentialsJSON([]byte(contents)))
	                                  ^
Error: internal/encryption/keyprovider/gcp_kms/config.go:108:37: SA1019: option.WithCredentialsJSON is deprecated: This function is being deprecated because of a potential security risk. (staticcheck)
		credOptions = append(credOptions, option.WithCredentialsJSON([]byte(contents)))
		                                  ^
```

...so seems like we'll either need to adapt this to use a different approach or add a `nolint` to opt out of `staticcheck` for these statements.

(Honestly it seems quite counterproductive to have a lint rule that prevents using deprecated symbols because the whole point of deprecating something -- rather than just immediately removing it -- is to give callers like us the opportunity to separate upgrading the dependency from rewriting the code. But of course that's a topic to discuss in another place, since a dependency upgrade PR is not the right time to re-litigate our linter policy! :upside_down_face: )

---

The docstring for [`option.WithCredentialsJSON`](https://pkg.go.dev/google.golang.org/api@v0.274.0/option#WithCredentialsJSON) says:

> Deprecated: This function is being deprecated because of a potential security risk. 
>
> This function does not validate the credential configuration. The security risk occurs when a credential configuration is accepted from a source that is not under your control and used without validation on your side. 
>
> If you know that you will be loading credential configurations of a specific type, it is recommended to use a credential-type-specific option function. This will ensure that an unexpected credential type with potential for malicious intent is not loaded unintentionally. You might still have to do validation for certain credential types. Please follow the recommendation for that function. For example, if you want to load only service accounts, you can use [WithAuthCredentialsJSON](https://pkg.go.dev/google.golang.org/api@v0.274.0/option#WithAuthCredentialsJSON) with [ServiceAccount](https://pkg.go.dev/google.golang.org/api@v0.274.0/option#ServiceAccount):
>
> ```go
> option.WithAuthCredentialsJSON(option.ServiceAccount, json)
> ```
>
> If you are loading your credential configuration from an untrusted source and have not mitigated the risks (e.g. by validating the configuration yourself), make these changes as soon as possible to prevent security risks to your environment.
>
> Regardless of the function used, it is always your responsibility to validate configurations received from external sources. 

I believe our intention here _is_ to just accept whatever we were given, because configuration provided by an OpenTofu user is assumed to be trusted in most cases. So I'm not sure what to suggest we should do here, since the suggested `option.WithAuthCredentialsJSON` would force us to specify a specific kind of principal that must be used (service account, authorized user, impersonated service account, or external account), but it doesn't seem like we _want_ to force the use of any one of those in particular. :thinking: 

Commentary in the `gcp_kms` implementation suggests that our use of this was inspired by similar usage in the `hashicorp/google` provider, so maybe we can study what _they_ did in response to this deprecation.

**Update:** It seems like [`hashicorp/google` is currently still using the deprecated function](https://github.com/hashicorp/terraform-provider-google/blob/213cb7ef5435f519864756da4ea9d9b4192f99af/google/transport/config.go#L2576-L2593), which I guess they can do because they don't have a linter preventing the use of deprecated functions. :grinning: 

Perhaps we should "just" add `nolint` comments on this for now and wait until we see whether and how the provider team responds to this deprecation, so that we can match whatever they do and thus preserve the goal of the GCS backend and GCP KMS key provider following the same behavior as the `hashicorp/gcp` provider does.

apparentlymart:
I was not originally intending to do anything that isn't completely mechanical tool-based updates here since this wasn't my PR, but since I already had the context loaded anyway from the research in my previous comment I've just gone ahead and pushed another commit here that adds `nolint:staticcheck` comments to temporarily opt out of the lint check on the three affected statements.

I suggest that we move forward with this compromise and then revisit only once `hashicorp/google` has responded to the same deprecation, so we can make our implementation as compatible as possible with theirs. If we accept this compromise then once this is merged I'll make a new issue to represent that followup work.

(There are some recommendations for what applications ought to do in [Validate credential configurations from external sources](https://docs.cloud.google.com/docs/authentication/client-libraries#external-credentials), but that's a non-trivial set of recommended checks and so I think it's better to let the GCP experts who maintain the `hashicorp/google` provider take the lead on how best to interpret them for this use-case.)

**Update:** After merge, I created https://github.com/opentofu/opentofu/issues/3985 to represent the followup work.

PR #3996, opened 2026-04-06
Deprecation diagnostics internal rework

cam72cam:
This PR is a follow up to #3973 that was a bit more involved than would have been reasonable to try to include in v1.12. It changes a few critical things about deprecation diagnostics.

1. Unifies DeprecationCause between variables, outputs, and resources
2. Hides more of the internals in the marks package
3. Moves complex diagnostic processing into views
   a. Deduplicating based on DeprecationCause
   b. RemoteSource filtering in terms of -deprecation cli arg using module source lookup
   c. Enhanced Consolidation using module source lookup (functions as users would expect now)

The only thing I really dislike about this approach is how we wire the module source queries into the view from the command package. We abuse the config loader in a way that works, but is not idea.  Perhaps @yottta 's work on refactoring the CLI package will make this less terrible in the future.

## Checklist

### Go checklist

 


internal/command/views/deprecation_level.go
```diff
@@ -0,0 +1,62 @@
...
+	case arguments.DeprecationWarningLevelAll:
+		return true
+	case arguments.DeprecationWarningLevelLocal:
+		// do not allow deprecation warnings for outputs from remote module calls
+		return !v.isRemoteModuleSource(cause.ModuleInstance().Module())
+	case arguments.DeprecationWarningLevelNone:
+		return false
+	}
+	// unreachable
+	return true
```
diofeher: If it's unreachable, shouldn't we use a panic here? I feel that using `true` will be confusing if we ever reach that point in the future.

internal/lang/marks/marks.go
```diff
@@ -73,18 +83,53 @@ var (
...
+func DeprecationCauseVariable(vaddr addrs.AbsInputVariableInstance, message string) DeprecationCause {
+	return DeprecationCause{
+		module:  vaddr.Module.String(),
+		subject: vaddr.Variable.Name,
+		message: message,
+	}
 }
 
 // ExtraInfoKey returns the key used for consolidation of deprecation diagnostics.
+// This will be enhanced bythe view with module source address
```
diofeher:
```suggestion
// This will be enhanced by the view with module source address
```

internal/lang/marks/diag.go
```diff
@@ -27,40 +27,40 @@ func (c DeprecationCause) diagnosticDeprecationCause() DeprecationCause {
...
-	_ tfdiags.DiagnosticExtraWrapper  = (*deprecatedOutputDiagnosticExtra)(nil)
+	_ diagnosticExtraDeprecationCause = (*deprecatedDiagnosticExtra)(nil)
+	_ tfdiags.DiagnosticExtraWrapper  = (*deprecatedDiagnosticExtra)(nil)
 )
 
-func (c *deprecatedOutputDiagnosticExtra) WrapDiagnosticExtra(inner interface{}) {
+func (c *deprecatedDiagnosticExtra) WrapDiagnosticExtra(inner interface{}) {
 	if c.wrapped != nil {
 		// This is a logical inconsistency, the caller should know whether they have already wrapped an extra or not.
 		panic("Attempted to wrap a diagnostic extra into a deprecatedOutputDiagnosticExtra that is already wrapping a different extra. This is a bug in OpenTofu, please report it.")
```
diofeher:
```suggestion
		panic("Attempted to wrap a diagnostic extra into a deprecatedDiagnosticExtra that is already wrapping a different extra. This is a bug in OpenTofu, please report it.")
```

Since the method parameter's name has changed

internal/command/views/deprecation_level.go
yottta: Agree with this. This path should be reachable only if a new level is added so the contributor should know right away that the implementation is incomplete.

internal/command/views/view.go
```diff
@@ -54,6 +54,9 @@ type View struct {
 	// will be dereferenced as late as possible when rendering diagnostics in
 	// order to access the config loader cache.
 	configSources func() map[string]*hcl.File
+
```
yottta, line 57: Can you add a comment or update the comment above `configSources`? These have similar purposes and come from the same place, so would be good to have these grouped

internal/configs/configload/loader.go
```diff
@@ -159,3 +163,39 @@ func (l *Loader) AllowLanguageExperiments(allowed bool) {
 	// add support here later if we decide to make use of language experiments
 	// in future versions of OpenTofu.
 }
+
+func (l *Loader) IsRemoteModuleSource(path addrs.Module) bool {
```
yottta, line 168: Can you add some unit tests for these 2 new methods please?

internal/command/views/view_test.go
```diff
@@ -121,10 +120,10 @@ func TestView_Configure(t *testing.T) {
 		"module deprecation warning level": {
 
 			viewArgs: &arguments.View{
-				ModuleDeprecationWarnLvl: tofu.DeprecationWarningLevelLocal,
+				ModuleDeprecationWarnLvl: arguments.DeprecationWarningLevelLocal,
 			},
 			validate: func(t *testing.T, v *View) {
-				if v.ModuleDeprecationWarnLvl != tofu.DeprecationWarningLevelLocal {
```
yottta, line 127: In this file, there is a test called [`TestView_Diagnostics`](https://github.com/opentofu/opentofu/blob/cda1588941a59d3549a6b6e0d605e1f227117265/internal/command/views/view_test.go#L153C6-L153C26), which contains a test case named [`diagnostics with sources`](https://github.com/opentofu/opentofu/blob/cda1588941a59d3549a6b6e0d605e1f227117265/internal/command/views/view_test.go#L350C4-L350C28).
Can you add something similar for the newly added `isRemoteModuleSource` and `moduleSourceAddrs`?

internal/configs/configschema/marks_test.go
```diff
@@ -8,7 +8,6 @@ package configschema
 import (
```
yottta, line 8: I see no tests in this file to assert that `ValueMarks` returns also the deprecation ones correctly. Could you add one for that?

internal/lang/marks/diag.go
```diff
@@ -27,40 +27,40 @@ func (c DeprecationCause) diagnosticDeprecationCause() DeprecationCause {
...
 }
 
-func (c *deprecatedOutputDiagnosticExtra) diagnosticDeprecationCause() DeprecationCause {
+func (c *deprecatedDiagnosticExtra) diagnosticDeprecationCause() DeprecationCause {
 	return c.Cause
 }
 
-// DeprecatedOutputDiagnosticOverride is mainly created for unit testing. This is done this way just to avoid
+// DeprecatedDiagnosticOverride is mainly created for unit testing. This is done this way just to avoid
 // exporting deprecatedOutputDiagnosticExtra from this package, which can create confusion when somebody would like to use this package.
```
yottta:
```suggestion
// exporting deprecatedDiagnosticExtra from this package, which can create confusion when somebody would like to use this package.
```

internal/lang/marks/marks.go
```diff
@@ -31,6 +31,16 @@ func Has(val cty.Value, mark valueMark) bool {
 	return val.HasMark(mark)
 }
 
+// HasDeprecated returns true if the cty.Value contains a deprecated mark.
+func HasDeprecated(v cty.Value) bool {
```
yottta, line 35: This is used only in tests. Do you think that would worth moving this to the test file instead? 🤔

internal/lang/marks/marks.go
```diff
@@ -160,10 +189,10 @@ func ExtractDeprecationDiagnosticsWithBody(val cty.Value, body hcl.Body) (cty.Va
 		}
 		cause := dm.Cause
 		var msg string
-		if cause.Message == "" {
-			msg = fmt.Sprintf("This value is derived from %s, which is deprecated.", cause.By)
+		if cause.message == "" {
```
yottta, line 192: This particular function have no unit tests. WDYT adding some to see how the final diagnostics will look like?

yottta (approved):
This looks great! Thanks a lot for tidying up all of this.
The inline comments are mostly about missing unit tests and outdated comments.

And yeah, I totally agree that the way we configure the sources in the view is hacky.
I will take a note on this to check on how this can be improved later.

internal/lang/marks/marks.go
cam72cam: It matches the pattern of other helper utilities within the file. I could also see it being useful for tests in other locations around the OpenTOfu codebase.

internal/command/views/deprecation_level.go
cam72cam: Changed in 56bcaf0527

internal/lang/marks/marks.go
cam72cam: 56bcaf0527

internal/lang/marks/diag.go
cam72cam: 56bcaf0527

internal/command/views/view.go
cam72cam: Done in 56bcaf0527

internal/lang/marks/diag.go
cam72cam: Fixed in 56bcaf0527

internal/lang/marks/marks.go
cam72cam: This is tested e2e and is hard to simulate in isolation `internal/command/e2etest/plan_test.go`

internal/command/views/view_test.go
cam72cam: This is tested e2e and is hard to simulate in isolation `internal/command/e2etest/plan_test.go`

internal/configs/configload/loader.go
cam72cam: Added in e6cbbaf1a2

internal/configs/configschema/marks_test.go
cam72cam: Added in e6cbbaf1a2

internal/command/views/view.go
```diff
@@ -54,6 +54,11 @@ type View struct {
...
+	// These other unfortunate warts is required to enable correct deduplication
+	// and filtering of deprecation diagnostics diagnostics
```
yottta:
```suggestion
	// These other unfortunate warts are required to enable correct deduplication
	// and filtering of deprecation diagnostics
```

internal/configs/configload/loader.go
```diff
@@ -159,3 +163,43 @@ func (l *Loader) AllowLanguageExperiments(allowed bool) {
...
+	current := l.lastLoadedRoot
+	for _, part := range path {
+		child, childOk := current.Children[part]
+		if !childOk {
+			log.Printf("[ERROR] Unable to determine if module source is remote due to missing child")
+			return false
+		}
+		if child.EntersNewPackage() {
+			return true
+		}
```
yottta, line 183: Do I read this correctly? If along the module path there is a module that is pulled from the remote sources, return true here since any other child is transitively a remote module.

I see that this was added in [e6cbbaf1a2f18862e3e69747887bbdf8e0d8cff4](https://github.com/opentofu/opentofu/pull/3996/changes/e6cbbaf1a2f18862e3e69747887bbdf8e0d8cff4#diff-b56d379b07c52805ec95fdf7f8727c53967257ccf8d3d7f50ed2dbb4b0e89447R181), and it looks like it was an issue caught during adding unit tests, right?

internal/configs/configschema/marks_test.go
```diff
@@ -63,26 +74,33 @@ func TestBlockValueMarks(t *testing.T) {
 		return &cp
 	}(schema)
 
-	testCases := map[string]struct {
+	addr := &addrs.AbsResourceInstance{Module: addrs.ModuleInstance{{Name: "name"}}}
+	depMark := deprecationMark(addr, cty.Path{}.GetAttr("deprecated"), "I'm deprecated")
+	depMark0 := deprecationMark(addr, cty.Path{}.GetAttr("list").IndexInt(0).GetAttr("deprecated"), "I'm deprecated")
+	depMark1 := deprecationMark(addr, cty.Path{}.GetAttr("list").IndexInt(1).GetAttr("deprecated"), "I'm deprecated")
+
+	testCases := []struct {
```
yottta, line 82: Why the change from map to struct? 🤔

yottta (approved):
Other than some inline comments, this looks good. 👍

internal/configs/configload/loader.go
cam72cam: That is correct.  We want to be able to filter any modules that are part of a remote chain.

internal/configs/configschema/marks_test.go
cam72cam: Ordering.  It's annoying when the test output changes every time you run it. We have both in the codebase, but the []struct approach is the preferred one.

Issue #2213, opened 2024-11-22
Add first party linting support i.e. `tofu lint`


dragonfleas:
### OpenTofu Version

```shell
N/A
```

### The problem in your OpenTofu project

No linting tools for OpenTofu.

### Attempted Solutions

- Explored many different tools online
- Checked `tflint` which has [indirectly stated they will not support opentofu](https://github.com/terraform-linters/tflint/issues/2037#issuecomment-2094738720)
- Checked for style and quality enforcement from SAST tools, if it exists it's minimal

### Proposal

I have two proposals:

The first of which is full first party linting support using the OpenTofu binary package and running something like `tofu lint`, this will likely be very complicated as the knowledge around specific language specifications is as one in Slack put it *underspecified*. The benefits of this would be a massively increased user experience and ease of adoption for those seeking quality and style enforcement either in editor, or optionally during their plans and applies. There are many opinions about style, primitive usages, built in tofu function use cases, that we could all likely benefit from having some sort of static analysis as a quality gate for those things.

The second alternative proposal is, we go back to the last commit of `tflint` before the dual license [LICENSE_BUSL](https://github.com/terraform-linters/tflint/blob/master/LICENSE-BUSL) was introduced and fork it, creating a fork labeled `tofulint` and take over the project to work on and add the relevant versioned changes that are required from 1.5 era up to the current, I don't know if this would be more of an involved effort or not, but in either of these proposals I'd gladly pledge at least 10 hours a week to put towards this effort.

apparentlymart:
Hi @dragonfleas! Thanks for opening this issue.

In my experience the term "lint" is often used for a wide variety of different kinds of analyses -- basically _anything_ that's more subtle than just "is this valid at all?" -- so I think that along with voting for the general idea of linting it would also be helpful for folks to share what sorts of analyses they are particularly interested in.

---

I have generally been most interested in analyses that might help explain why some particular configuration is not behaving as the author expected. I previously made a list of possible candidates which I will reproduce here just as a starting point, although I am not meaning this to say that any of the following is definitely required for the feature request in this issue to be met.

[...]

PR #3999, opened 2026-04-06
rfc: Built-in Linting

apparentlymart:
This is a proposal for extending OpenTofu with built-in linting features as requested in https://github.com/opentofu/opentofu/issues/2213, starting with just a small set of built-in lint rules so we can get experience with this entirely-new capability.

[Rendered Version](https://github.com/opentofu/opentofu/blob/a7bb6584740c91579c31a984315fcd23534d7872/rfc/20260406-linting.md)

If we accept this then I expect it would _begin_ our work for https://github.com/opentofu/opentofu/issues/2213, but that we would not close that issue until we've been through a few implementation iterations and have a level of functionality that's comparable to (but not necessarily exactly matching) what the third-party `tflint` supports. I've just started small here because this is an entirely new category of functionality for OpenTofu and so I think we should start this cautiously and then learn from feedback.

### DIFF
--- a/rfc/20260406-linting.md
+++ b/rfc/20260406-linting.md
@@ -1,0 +1,269 @@
+# Built-in Linting
+
+(This RFC is primarily for GitHub [issue #2213](https://github.com/opentofu/opentofu/issues/2213).)
+
+OpenTofu does not currently have a fully-functional linting solution. Some folks successfully use the third-party tool `tflint`, but [it's not designed for use with OpenTofu](https://github.com/terraform-linters/tflint/issues/2037#issuecomment-2094738720) and so it tends to fail when used with configurations that use features that are in OpenTofu but not supported in its predecessor.
+
+Maintaining a linter outside of the OpenTofu codebase is also generally challenging, because it involves reimplementing much of the logic OpenTofu uses for analyzing the configuration. An external linter would inevitably lag behind changes to OpenTofu, generating spurious errors about newly-added language features that the linter has not yet been updated for.
+
+This document proposes that we begin adding built-in linting features. This first proposal focuses mainly on enabling linting rules that are written directly inside the OpenTofu codebase and describe concerns related to the core OpenTofu language. It's also intended to leave room for future expansion for lint rules coming from other sources, such as lint rules implemented inside provider plugins that relate to provider-specific features, and totally custom lint rules that an organization could impose on their own code without modifying OpenTofu or any mainstream provider plugins, but those extensions are intentionally not included here because we ought to get some implementation experience with built-in lint rules before we create any new public interfaces that would be subject to compatibility promises and therefore harder to change if we learn of a design problem.
+
+If the initial set of features described here is successful then this will hopefully be the first of a series of proposals that should eventually provide a full-fledged linting system that's built in to OpenTofu. This is an entirely new problem space for OpenTofu to handle though, so this proposal is intentionally just a small initial increment so we can learn from how and whether folks use these features.
+
+## What is "linting" anyway?
+
+For the sake of this document, we'll define "linting" as creating the opportunity for OpenTofu to warn about something that isn't necessarily wrong but may be a cause for concern. A linter therefore produces something similar to warning diagnostics, but with a higher-tolerance for "false positives" where the linter draws attention to something that is not actually a problem in practice.
+
+Because of the higher false-positive rate, linting would not be enabled by default but could be opted in by command line options to certain commands where it is relevant. The higher false-positive rate also means there must be a way for operators to opt-in and opt-out of specific lint rules based on their relevance to each operator's specific situation, whereas for warning diagnostics OpenTofu does not offer such fine-grain control.
+
+For example, producing a deprecation warning is not "linting" because OpenTofu has certainty that a particular feature has been deprecated. An example of a more "linter-like" rule would be to note that one of the addresses written in a `depends_on` argument is unnecessary because the same dependency was already implied by a reference elsewhere in the same block: that's not _wrong_ and so it would be annoying for it to unconditionally generate a warning, but having it be raised in linting mode might help someone who is trying to debug an ordering problem quickly notice that adding a `depends_on` argument is not changing anything about the meaning of the configuration.
+
+Another difference compared to warnings is that "linting" is exclusively used for potential problems relating to _configuration_, whereas normal warning diagnostics can potentially describe other problems such as strange behavior of a remote API, or situations caused by command-line options.
+
+## User Experience
+
+### Lint rule identifiers
+
+To make it possible to enable and disable individual linting rules, each rule would have a terse code associated with it. For our initial round of implementation, all rule identifiers would have the fixed prefix `core:`, followed by a short alphanumeric suffix that is a mnemonic for the rule. This prefix is intended to allow us to introduce other identifier namespaces for provider-defined rules, etc, in later releases. 
+
+For example, the hypothetical lint rule about redundant entries in `depends_on` might have the identifier `core:unuseddeps`. It's not a goal for these names to fully represent exactly what the lint rule does, but rather just to make it easier for someone who is already familiar with a particular lint rule to recognize that rule in a potentially-long list of enabled or disabled rules.
+
+It'd be very tedious to list out each individual rule a team wants to enable though, so we would also support certain identifiers that serve as aliases for groups of other identifiers. For example, the identifier `core:confusing` might represent all of the rules that notice when something has been written in a way that might be misleading to a future reader of the code, which would likely include the previously-mentioned `core:unuseddeps` along with any other similar situation.
+
+Finally, the special identifier `all` represents all available lint rules. When written without a prefix it specifies all lint rules across all namespaces, or it can be used after a prefix (e.g. `core:all`) to specify all of the rules in the "core" namespace. The identifier "all" is therefore reserved across all namespaces and must not be assigned any other meaning, but all other names are expected to vary based on the namespace prefix.
+
+All lint rule identifier prefixes are required to start with a letter or digit, using Unicode's definitions of those categories, but can otherwise contain any printable characters except colon. The part appearing after the colon must be a valid unicode identifier, and must only use lowercase or case-agnostic letters. (If we add provider-defined lint rules in future then their namespaces will presumably contain slashes, following our usual syntax for identifying providers.)
+
+### Enabling lint rules on the command line
+
+For our initial implementation of this feature, we'll require opting in to linting using a new command line option `-lint=...`, where the value is a comma-separated list of lint rule identifiers to include (without any prefix) or to exclude (with the prefix `!`).
+
+The simplest invocation would be `-lint='all'`, which opts in to all available lint rules. If an operator finds that a particular rule is generating only false-positives for a particular situation, they could exclude it by writing an invocation like `-lint='all,!core:unuseddeps'` or (to disable an entire group at once) `-lint='all,!core:confusing'`. Any `-lint=...` option must always include at least one non-negated rule ID.
+
+The new `-lint` option would be available for the following subcommands:
+
+- `tofu validate`
+- `tofu plan`
+- `tofu apply`, and its variants `tofu destroy` and `tofu refresh`
+
+The scope of possible lint rules varies depending on which command `-lint=...` is provided to, because (as usual) each of these commands performs a different subset of OpenTofu's work. As many rules as possible will be implemented in the validation phase so that they can appear for all three of these commands, but some rules might depend on dynamic data that OpenTofu only calculates when planning or applying.
+
+All of the items listed after `-lint=` must match the expected syntax for selecting lint rules, but any syntactically-valid identifier that doesn't match a currently-known rule is just silently ignored, creating the same effect as if an enabled rule were supported but not relevant to the current configuration. This is so that it's easier to just encode a single `-lint=...` option in some automation without having to vary it depending on which version of OpenTofu is in use or which providers the current configuration is using, since the set of rule IDs supported by each component is expected to change over time.
+
+> [!NOTE]  
+> A future extension of this proposal is likely to include a way to select and deselect lint rules for one or more modules as part of the configuration, rather than on the command line.
+>
+> That's intentionally excluded to start because we'd like to get experience with global, command-line-based selections before we design a language extension. For example, it's not clear yet whether lint rules in configuration would be a whole-configuration-level idea specified only alongside the root module and enforced for the whole configuration, if they'd instead be specified on a per-module basis, or something in between.
+>
+> The top-level `language` block is a potential home for such settings, but that would be appropriate only for settings specific to a single module because that block's meaning is to define how _just the current module_ is using the OpenTofu language.
+
+### Lint message rendering
+
+The rendering of a lint message is similar to that of a warning diagnostic, just with a slightly-modified header that includes a lint rule ID:
+
+```
+Warning: Unnecessary explicit dependency [core:unuseddeps]
+
+  on example.tf line 54, in resource "example" "foo":
+   54:   depends_on = [
+   55:     var.whatever,
+   56:   ]
+
+A dependency on var.whatever is already implied by the reference at
+example.tf:51,20, so this explicit dependency is ineffective and potentially
+confusing.
+```
+
+Treating lint messages as a special kind of warning diagnostic means that they can naturally integrate into our existing representation of machine-readable diagnostics, and so any wrapping automation that's collecting diagnostics to render in e.g. a web UI will automatically handle lint messages for any rules enabled on the command line.
+
+It's a goal of this initial proposal to minimize the amount of new integration points that would be covered by compatibility promises so that we can have more opportunity to react to feedback in future releases and so there would not initially be any machine-readable way to distinguish normal warnings from lint message warnings, but if this design is successful then in a later release we could extend the JSON representation of diagnostics to have an additional `"lint_rule_ids"` property mentions both the primary rule ID and any rule group aliases that are associated with the message.
+
+### Initial lint rules
+
+For the first implementation we'd keep this limited to a small set of situations that seem to have commonly caused confusion for those new to the OpenTofu language:
+
+| Rule ID | Rule Groups | Trigger |
+|---------|-------------|---------|
+| `core:unuseddeps` | `core:confusing` | One of the addresses written in a `depends_on` argument is redundant with a dependency implied by a reference elsewhere in the same block. |
+| `core:quotedref` | `core:confusing`, `core:mistake` | A quoted string in the configuration has content matching the address of something else declared in the same module, suggesting that it was probably intended to be a reference to that symbol but should've been written without the quotes. |
+| `core:ineffeq` | `core:mistake` | Evaluated an `==` or `!=` expression where the two operands were not of the same type, making the comparison ineffective. |
+| `core:impurefunc` | `core:noconverge` | An impure function (`timestamp`, `uuid`, or `bcrypt`) was used in a location where convergence is typically expected, such as in a resource argument. |
+
+This initial set is designed to cover just a small set of interesting categories of problem, while also including a mix of different locations in the system where the check would need to be implemented so that we can make sure it's feasible to implement lint rules in all of these locations. The set is intentionally small so that for this first proposal we can focus mainly on just getting the infrastructure into place, and then gradually grow this in future versions based on feedback.
+
+## Implementation details
+
+### Lint diagnostics
+
+To allow us to integrate the lint rules into existing codepaths without making disruptive changes to the internal API, we'll return lint messages as just a special kind of warning diagnostic, passed through the same `tfdiags.Diagnostics` return channel used for other configuration problems.
+
+The `tfdiags` package would therefore gain a new type `tfdiags.LintMessage` to represent a lint message, a type to represent a diagnostic identifier, and a new `Diagnostics` method for filtering out unwanted lint rules:
+
+```go
+package tfdiags
+
+// LintRuleID represents the identifier for an individual lint rule or for a
+// group of lint rules.
+type LintRuleID struct {
+    Namespace string
+    Name      string
+}
+
+// CoreLintRuleID returns a [LintRuleID] value whose namespace is "core" and
+// whose name is the given string.
+func CoreLintRuleID(name string) LintRuleID
+
+// LintMessage represents a message describing a linting problem.
+//
+// Pass pointers to values of this type to [Diagnostics.Append] to include
+// lint messages as part of your returned diagnostics.
+type LintMessage struct {
+    // RuleIDs is a set of all of the rule IDs this message relates to.
+    //
+    // There should always be at least one element in the given slice and
+    // the first element should be the identifier that uniquely identifies
+    // just the one condition being reported. Optional subsequent elements
+    // are the identifiers of any rule groups the main rule ID belongs to.
+    RuleIDs []LintRuleID
+
+    // Summary, Detail, and Subject have their usual meaning for diagnostic
+    // messages.
+    Summary string
+    Detail  string
+    Subject SourceRange
+
+    // Context is an optional additional source range that must enclose
+    // the range in Subject but can include additional characters that provide
+    // relevant context for understanding the problem. The human-oriented
+    // diagnostic renderer uses this to extend the source code snippet to
+    // include additional lines from the input configuration.
+    Context *SourceRange
+}
+
+// FilterLint modifies the backing array of the receiver so that any
+// [Lint] diagnostics which don't match the given include and exclude rules
+// are removed, and then returns a new slice over the valid part of the updated
+// backing array.
+//
+// UI code should call this to exclude any unrequested lint rules before
+// rendering the diagnostics.
+func (diags Diagnostics) FilterLint(include, exclude collections.Set[LintRuleID]) Diagnostics
+```
+
+A codepath that wants to return a lint message would therefore append it to the diagnostics it is going to return in the usual way:
+
+```go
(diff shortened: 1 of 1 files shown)


rfc/20260406-linting.md
```diff
@@ -0,0 +1,277 @@
...
+
+## What is "linting" anyway?
+
+For the sake of this document, we'll define "linting" as creating the opportunity for OpenTofu to warn about something that isn't necessarily wrong but may be a cause for concern. A linter therefore produces something similar to warning diagnostics, but with a higher-tolerance for "false positives" where the linter draws attention to something that is not actually a problem in practice.
+
+Because of the hgher false-positive rate, linting would not be enabled by default but could be opted in by command line options to certain commands where it is relevant. The higher false-positive rate also means there must be a way for operators to opt-in and opt-out of specific lint rules based on their relevance to each operator's specific situation, whereas for warning diagnostics OpenTofu does not offer such fine-grain control.
```
yottta:
```suggestion
Because of the higher false-positive rate, linting would not be enabled by default but could be opted in by command line options to certain commands where it is relevant. The higher false-positive rate also means there must be a way for operators to opt-in and opt-out of specific lint rules based on their relevance to each operator's specific situation, whereas for warning diagnostics OpenTofu does not offer such fine-grain control.
```

rfc/20260406-linting.md
```diff
@@ -0,0 +1,277 @@
...
+   56:   ]
+
+A dependency on var.whatever is already implied by the reference at
+example.tf:51,20, so this explicit dependency is ineffective and potentially
+confusing.
+```
+
+Treating lint messages as a special kind of warning diagnostic means that they can naturally integrate into our existing representation of machine-readable diagnostics, and so any wrapping automation that's collecting diagnostics to render in e.g. a web UI will automatically handle lint messages for any rules enabled on the command line.
+
+It's a goal of this initial proposal to minimize the amount of new integration points that would be covered by compatibility promises so that we can have more opportunity to react to feedback in future releases and so there would not initially be any machine-readable way to distinguish normal warnings from lint message warnings, but if this design is successful then in a later release we could extend the JSON representation of diagnostics to have an additional `"lint_rule_ids"` property mentions both the primary rule ID and any rule group aliases that are associated with the message.
```
yottta, line 79: I am slightly confused by this paragraph: do we want to integrate these in the machine-readable output from the first iteration? Or will we provide only human readable output in the first version?

I am thinking that in order to allow integrations with 3rd party tools and custom parsing, I would like to have this integrated in the first version, but to agree upon a specific `type` field of the log entry to clearly indicate that it's a linting log entry (e.g.: `type`: `linting`).
Additionally, if we enable this from the get-go, we should definitely mark the output format as experimental and avoid covering it by the compatibility promises.

yottta (approved):
In general, looks great.
I know that the types and logic necessary for the linters implementation might be scattered across the project, but I wonder if we could still find a way to store all the implementation in one package.
But that's more of an implementation detail, which should be tackled when the time comes.

rfc/20260406-linting.md
apparentlymart: What I was intending to propose here is that in the first release the lint messages would appear in the JSON output just as normal warning diagnostics, without any machine-readable additions.

This compromise would make sure they still appear in any alternative UI someone has built around the JSON output -- rendering them just like any other warning -- but it would not be possible yet to distinguish them from any other warning based on the JSON data.

Overall the idea is to make lint messages be different from normal warnings only in ways that are not covered by compatibility promises, so that we can easily change how we represent these based on feedback.

rfc/20260406-linting.md
```diff
@@ -0,0 +1,277 @@
...
+
+For our initial implementation of this feature, we'll require opting in to linting using a new command line option `-lint=...`, where the value is a comma-separated list of lint rule identifiers to include (without any prefix) or to exclude (with the prefix `!`).
+
+The simplest invocation would be `-lint='all'`, which opts in to all available lint rules. If an operator finds that a particular rule is generating only false-positives for a particular situation, they could exclude it by writing an invocation like `-lint='all,!core:unuseddeps'` or (to disable an entire group at once) `-lint='all,!core:confusing'`. Any `-lint=...` option must always include at least one non-negated rule ID.
+
+The new `-lint` option would be available for the following subcommands:
+
+- `tofu validate`
+- `tofu plan`
+- `tofu apply`, and its variants `tofu destroy` and `tofu refresh`
```
diofeher, line 47: Hmm, I thought we would be creating a new command `tofu lint` instead of adding this as a command-line option to existing commands. I do agree that it makes sense to bundle it with `validate` due to the command's semantics, but I'm not sure whether the additional complexity of plugging it into other commands, rather than having a separate command, is worth it.

rfc/20260406-linting.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Possible Future Extensions
+
+### Provider-based Lint Rules
+
+In a future version of the provider plugin protocol we could potentially extend provider RPC functions like `ValidateResourceConfig` to be able to return zero or more lint messages alongside the diagnostics already supported, where each one includes a provider-specific lint identifier so that OpenTofu can filter them based on what's specified in the `-lint=` command line option.
+
+For example, a provider might choose to offer lint rules that detect violations of best-practices mentioned in their remote system's documentation, which OpenTofu itelf would have no awareness of.
```
diofeher:
```suggestion
For example, a provider might choose to offer lint rules that detect violations of best-practices mentioned in their remote system's documentation, which OpenTofu itself would have no awareness of.
```

rfc/20260406-linting.md
```diff
@@ -0,0 +1,277 @@
...
+
+All of the items listed after `-lint=` must match the expected syntax for selecting lint rules, but any syntactically-valid identifier that doesn't match a currently-known rule is just silently ignored, creating the same effect as if an enabled rule were supported but not relevant to the current configuration. This is so that it's easier to just encode a single `-lint=...` option in some automation without having to vary it depending on which version of OpenTofu is in use or which providers the current configuration is using, since the set of rule IDs supported by each component is expected to change over time.
+
+> [!NOTE]  
+> A future extension of this proposal is likely to include a way to select and deselect lint rules for one or more modules as part of the configuration, rather than on the command line.
```
diofeher, line 54: For me, that makes sense! Don't you think that, for adoption, it would be better to have that from the beginning?

rfc/20260406-linting.md
```diff
@@ -0,0 +1,277 @@
...
+
+    // Context is an optional additional source range that must enclose
+    // the range in Subject but can include additional characters that provide
+    // relevant context for understanding the problem. The human-oriented
+    // diagnostic renderer uses this to extend the source code snippet to
+    // include additional lines from the input configuration.
+    Context *SourceRange
+}
+
+// FilterLint modifies the backing array of the reciever so that any
```
diofeher:
```suggestion
// FilterLint modifies the backing array of the receiver so that any
```

diofeher (approved):
This is looking really good, thanks for moving forward with this effort @apparentlymart !

I've left a few comments, but I have a general question: how do you see this integration with tofu-ls?

I think just opening the file in a text editor would be enough to trigger these lint hints, but I'm curious if you have thoughts on that.

rfc/20260406-linting.md
apparentlymart: If we don't make it possible to combine linting with `tofu plan` then some lint rules will not be effective, because lots of values are unknown in `tofu validate`.

I've proposed that we implement linting as essentially just some additional warnings that are checked as part of the normal codepaths when linting is enabled, instead of as a separate pass, because I expect that it'll _automatically_ work with all of these commands: the same parts of the language runtime will run for all of them, but during plan and apply there will be more known values and so more opportunity to detect problems. I don't expect that it will cause a lot of extra complexity, though of course that remains to be seen.

rfc/20260406-linting.md
apparentlymart: The text in the two paragraphs below this comment was intended to explain why I suggest to wait to a later phase for this.

diofeher: Oh, sorry, I've read it, but I was commenting while I was in that section and forgot to clean up this comment!

apparentlymart:
Since I intentionally excluded any machine-readable representation of lint rules in this proposal, it would not be possible for `tofu-ls` to give any _special_ treatment of lint rules, but if it were to run `tofu validate -json -lint=all` then it would get normal warning diagnostics in JSON format representing any lint messages.

For the first round perhaps we could add a command line option to `tofu-ls` that specifies which lint rule identifiers should be used [when `tofu-ls` runs validate](https://github.com/opentofu/tofu-ls/blob/7d619d42de488309394d9dcf5de9b1728d34557f/internal/tofu/exec/exec.go#L168-L186), and then the existing functionality for annotating the source code with diagnostics would cause the lint failures to be indicated using the warning color.

If we decide to add a machine-readable way to distinguish normal warnings from lint warnings in a future version then we could extend that by making `tofu-ls` use the LSP "hint" severity instead of "warning" severity, but I want to wait until we get feedback on the first functionality before we introduce any machine-readable output that would be covered by compatibility promises.

dragonfleas:
Forgive me if this isn't the right place to comment, and noting I'm not a collaborator or contributor myself but I have some concerns specifically about the section on in-configuration lint rules.

> ### In-configuration Lint Rules
> In a future release we might want to support custom lint rules implemented directly inside a module, or in some other sort of configuration file that can contain arbitrary expressions to run.
>
> If we do this then we'll need to think about how their namespaces should work, because unlike core and providers there isn't any single concise global addressing scheme for module packages.
>
> The namespacing problem aside, a custom lint rule would presumably have a similar shape to a "policy as code" configuration, as folks already commonly write using languages like Open Policy Agent's "Rego" language, and so we'll probably find that we need to use or design a configuration language of similar expressive power.
>
> As with provider-based lint rules, we'll need to take care with how much we allow custom lint rules to depend on the syntax details of the current OpenTofu language, so that we don't end up making it impossible to evolve the language without breaking existing lint rules. Perhaps custom lint rules would be required to work only with dynamic values from the configuration and not with its raw syntax, but that would mean that these rules could only be used to detect semantic issues and not to enforce stylistic preferences, etc.

One pattern I've seen anecdotally with first-party linting implementations is that the initial iteration often ships without support for custom rules, and it tends to hurt adoption. **Linting is unlikely to be adopted or meaningfully tested if it doesn't deliver immediate value or solve a real problem for the user.** For an open source project like OpenTofu, where value is ultimately defined by the people using it, this matters a great deal.

A useful parallel is Deno, which I mentioned in the original first-party tofu linting issue. They launched with a very small set of basic, low-value rules, and adoption reflected that users show up for an emerging feature expecting it to be useful on day one, and a minimal built-in ruleset rarely clears that bar.

For this reason, I think it's important that the first generally-available release includes support for in-configuration rules. Lint rules should express a team's coding standards and enforce organizational and team-specific invariants. While there's real overlap with diagnostic checks that belong in the language itself, the core purpose of linting is to encode opinions and without a way for teams to define their own, a first-party linter risks shipping without the capability that makes linting worth adopting in the first place.

That, I believe, is the most accurate representation of what linting is *for*. There's a reason ESLint is so widely adopted and it's not just the "too big to die" network effect. It launched with a strong emphasis on user-provided configuration, and that shaped its direction from day one. The scope of ESLint is obviously far larger than what a narrowly-scoped first-party OpenTofu linter would aim for, so the parallel isn't one-to-one, but the underlying lesson that extensibility drives adoption still holds.

To conclude, I'd like to respectfully ask for the collaborators to reconsider the priority level of having in-configuration linting rules.

apparentlymart:
Hi @dragonfleas! Thanks for the feedback.

The primary purpose of this initial iteration for me is to get the plumbing in place and make sure we have a good foundation to build the user-customizable features on top of. The main idea is to understand what it looks like to have any built in linting functionality at all, rather than to ship a complete feature.

I understand that what's described in this RFC is not complete enough to be broadly useful on its own, but a fully-fledged linter for the OpenTofu language has roughly the same design complexity as the main OpenTofu language runtime itself, so we're not going to be able to get there in just one development period. I'm trying to propose something we _can_ complete in a single development period that we would then continue in subsequent development periods, but including a few realistic rules just to help prove that the design is actually reasonable.

rfc/20260406-linting.md
```diff
@@ -0,0 +1,269 @@
+# Built-in Linting
+
+(This RFC is primarily for GitHub [issue #2213](https://github.com/opentofu/opentofu/issues/2213).)
+
+OpenTofu does not currently have a fully-functional linting solution. Some folks successfully use the third-party tool `tflint`, but [it's not designed for use with OpenTofu](https://github.com/terraform-linters/tflint/issues/2037#issuecomment-2094738720) and so it tends to fail when used with configurations that use features that are in OpenTofu but not supported in its predecessor.
+
+Maintaining a linter outside of the OpenTofu codebase is also generally challenging, because it involves reimplementing much of the logic OpenTofu uses for analyzing the configuration. An external linter would inevitably lag behind changes to OpenTofu, generating spurious errors about newly-added language features that the linter has not yet been updated for.
```
lbordowitz, line 7: If the `config` package were made public, would that change anything? As in, if the marshalling of `.tf/.tf.json` etc files, and their corresponding validation were refactored and fit for use as an SDK by external tools (including `tofu-ls`), would that change any thoughts about making the linter internal?

By the way, this question should not block the plans of this RFC. To be sure, I've thought about this a little bit, but I don't have any plans nearly as thought out, and I believe in incremental good changes that we can back out more-or-less cleanly later (this built-in linter being one of those).

rfc/20260406-linting.md
```diff
@@ -0,0 +1,269 @@
...
+
+Finally, the special identifier `all` represents all available lint rules. When written without a prefix it specifies all lint rules across all namespaces, or it can be used after a prefix (e.g. `core:all`) to specify all of the rules in the "core" namespace. The identifier "all" is therefore reserved across all namespaces and must not be assigned any other meaning, but all other names are expected to vary based on the namespace prefix.
+
+All lint rule identifier prefixes are required to start with a letter or digit, using Unicode's definitions of those categories, but can otherwise contain any printable characters except colon. The part appearing after the colon must be a valid unicode identifier, and must only use lowercase or case-agnostic letters. (If we add provider-defined lint rules in future then their namespaces will presumably contain slashes, following our usual syntax for identifying providers.)
```
lbordowitz, line 35: We almost certainly want to require characters be "shell safe." No "|", "&", "[]", "!", spaces, quotation marks, etc.

rfc/20260406-linting.md
```diff
@@ -0,0 +1,269 @@
...
+
+The simplest invocation would be `-lint='all'`, which opts in to all available lint rules. If an operator finds that a particular rule is generating only false-positives for a particular situation, they could exclude it by writing an invocation like `-lint='all,!core:unuseddeps'` or (to disable an entire group at once) `-lint='all,!core:confusing'`. Any `-lint=...` option must always include at least one non-negated rule ID.
+
+The new `-lint` option would be available for the following subcommands:
+
+- `tofu validate`
+- `tofu plan`
+- `tofu apply`, and its variants `tofu destroy` and `tofu refresh`
+
+The scope of possible lint rules varies depending on which command `-lint=...` is provided to, because (as usual) each of these commands performs a different subset of OpenTofu's work. As many rules as possible will be implemented in the validation phase so that they can appear for all three of these commands, but some rules might depend on dynamic data that OpenTofu only calculates when planning or applying.
```
lbordowitz, line 49: > some rules might depend on dynamic data that OpenTofu only calculates when planning or applying.

Okay, that might be relevant to the `config` package thing I was talking about above. Those rules definitely could not be implemented externally. However, I'll note here that this would make using those linting rules in `tofu-ls` quite difficult when also using remote state storage.

rfc/20260406-linting.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Enabling lint rules on the command line
+
+For our initial implementation of this feature, we'll require opting in to linting using a new command line option `-lint=...`, where the value is a comma-separated list of lint rule identifiers to include (without any prefix) or to exclude (with the prefix `!`).
+
+The simplest invocation would be `-lint='all'`, which opts in to all available lint rules. If an operator finds that a particular rule is generating only false-positives for a particular situation, they could exclude it by writing an invocation like `-lint='all,!core:unuseddeps'` or (to disable an entire group at once) `-lint='all,!core:confusing'`. Any `-lint=...` option must always include at least one non-negated rule ID.
```
lbordowitz, line 41: There's an ambiguity here between groups and individual rules. For example, given the proposed lint identifiers below, if I specified `-lint='all,!core:confusing,core:unuseddeps'`, that would clearly mean disabling the entire group of `core:confusing` linters, but enabling `core:unuseddeps` linters which is part of that group. Could you make the precedence more explicit?

On a related note, we should not be allowed to write `-lint='core:unuseddeps,!core:unuseddeps'`; that is, a direct reference to a rule/group cannot be in both the include and exclude list.

lbordowitz (commented):
I love to see this! I've been wanting to pull in a linting tool for OpenTofu for a while, and this is excellent groundwork for starting to do that.

rfc/20260406-linting.md
apparentlymart: `package configs` only really deals with the surface level of decoding the configuration: it finds the top-level declarations and does some shallow decoding of their rigid elements like the `lifecycle` block in a `resource` block.

The sort of linting rules people have discussed wanting to implement[^1] relate to the fine details of how expressions are written and what values they produce, and if we wanted to expose _that_ to the outside then we'd be exposing essentially the entire codebase. I'm sure some folks would love that in the short term, but the usual reservation about that applies: essentially any change we make to any part of the language would be a breaking change for any tool that depends on those packages, and so it could be a huge burden on ongoing development.

Therefore my take is that having it embedded in OpenTofu itself is the best tradeoff: the lint-related logic can live as part of the language runtime code rather than as a caller of the language runtime code, and so we can do linting work at the same time as we're doing other work rather than repeating a bunch of work to generate the data needed for linting, and the lint rules will be able to interact with data from providers and the state that would not be visible to an external linter that's relying only on static analysis. And, most of all, we can evolve the linting code in lockstep with other changes to the system.

(FWIW back in the day I was also an advocate for having the language server embedded in the main executable too, for similar reasons. I was overruled on that and I can certainly see the tradeoffs on both sides, but `tofu-ls` is a good example of both the complexity of and limitations of an external reimplementation of the OpenTofu language that I think would be even worse for an external linter.)

[^1]: I acknowledge that I've not actually made any citations for this claim. I've been hearing feedback about this for many years now but haven't actually documented it in any structured way that I can refer to right now. If others have doubts about this claim then we could choose to hold on this project until we've done some more research first.

rfc/20260406-linting.md
apparentlymart: Yeah, I expect we'd nail this down more once we reach later phases of this work. For this RFC the only thing that can appear before the colon is literally `core`, so this is really just intended to establish a place for future extensions to happen, rather than designing them in detail immediately.

rfc/20260406-linting.md
apparentlymart: This is fair.

It seems like this needs a rule for which inclusions or exclusions "win" when a particular message matches both, and I agree that's an interesting tradeoff to make when multiple groups appear together in the pattern.

In your example `all,!core:confusing,core:unuseddeps` I agree with your assessment of the likely user intent, and so we'll need to decide how to implement the matching so that it agrees with that intent.

One answer is to treat the comma-separated statements as being in a preference order so that the last item relevant to a particular message is the rule that "wins". If a lint message were generated that had the rule ids `core:unuseddeps` and `core:confusing`, all three of those statements would match (because `all` matches everything) and the `core:unuseddeps` statement would "win" because it's the last item in the list there.

Another answer would be to rely on the requirement that the first rule id for each message must be an individual rule ID and all subsequent rule ids must be groups, and then the matching can know which of them was intended to be the individual rule ID and treat it as special during matching so that individual rule IDs always "win" over group ids.

I'm slightly leaning towards the second because it seems simpler for the statements in the `-lint` argument to not be order-sensitive, but I'm curious what you think about this before I commit to something in the main text.

cam72cam:
Overall, this looks like a great start.

The only significant question I have at this point is what your thoughts on an in-config nolint directive might look like? I could see some situations where the example lint rules above might match on *weird* but intentional patterns.

The other question that comes to mind is putting the lint directives in a config file, a common pattern for other linting tools.

yottta:
> Overall, this looks like a great start.
> 
> The only significant question I have at this point is what your thoughts on an in-config nolint directive might look like? I could see some situations where the example lint rules above might match on _weird_ but intentional patterns.

Added in https://github.com/opentofu/opentofu/pull/3999/commits/119b8c8ac61670e7ee1f946ee2026ed1093baed4

yottta:
> The other question that comes to mind is putting the lint directives in a config file, a common pattern for other linting tools.

Added in https://github.com/opentofu/opentofu/pull/3999/commits/cd1760eafa13dce30cb668d05a0003a0eaf22606

rfc/20260406-linting.md
yottta, line 41: What do you think of aliasing the flag without value `-lint` to `-lint='all'`?
And if the user wants to exclude something, then it can go with the comma separated list of rules/groups.

apparentlymart: Go's built-in `flag` package makes that sort of "sometimes has a parameter" flag awkward (but not impossible) to implement, but I'm not opposed to it if we can implement it nicely and have it still work when we adopt Cobra later. 👍🏻

yottta: Ok. Will keep this in mind and if it will not work as intended, will leave it as you proposed here.

apparentlymart: Although: I guess we should think ahead to how this would interact with configuration-file-based linting rules later.

Perhaps `-lint` with no parameter means "do what the linting configuration says", whereas specifying a parameter overrides what the configuration says and forces a particular set of checks (or `all`)? But on the other hand: if a linting configuration file is present then maybe we'd just use its presence as evidence that the operator wants linting enabled, even when the `-lint` argument isn't present. 🤔

yottta: > if a linting configuration file is present then maybe we'd just use its presence as evidence that the operator wants linting enabled, even when the -lint argument isn't present

Not sure that I like it...
This would mean that linting is enabled all the time if a configuration file is present.
And in certain circumstances, linting should not be executed and interfere with the actual apply phase 🤔 
That's why I was thinking of the `-lint` blank flag as a "toggle" on enabling or not the linting process.
And the blank flag would initially be `'all'` and if a configuration file is present, it will be whatever it's in there.

> `'all'` and if a configuration file is present, it will be whatever it's in there.

I would say that this is similar in logic with the cliconfig (IINW), where there is always a default configuration generated by OpenTofu but will use the specified one if it exists.

Maybe we could have a `config.enabled` boolean configuration inside the file and if that is `true` than we would have it running all the time. 🤔

apparentlymart: Yeah, I'm not sure about the "if there's a lint configuration file then linting is always enabled" rule either. I was thinking of it as showing intent for the configured lint rules to always be checked (keeping in mind that lint rules only generate warnings, so they can't _block_ anything) but that's not necessarily true... the file could instead just be a convenient place to write down the default set of rules you want to enforce _only if linting is enabled_. :man_shrugging:

apparentlymart:
This RFC ended up causing some broader conversation about what's in scope, how quickly we could possibly implement different features, etc, and so I'm going to switch this to "draft" for now while we're still having informal discussions about this and then we can switch it back to Ready for Review once we have a stronger idea of exactly what we're proposing to do.

yottta:
Marking this as "ready for review" since maintainers discussed this yesterday and consensus was reached on going forward with this proposal for the first iteration.
Will follow up with an rfc tracker issue.

rfc/20260406-linting.md
yottta: @lbordowitz, considering the consensus maintainers reached recently, do you think that is anything else to be added to this conversation?

rfc/20260406-linting.md
yottta: This is going to be part of the incoming RFC tracker that will be linked to this PR shortly.
Therefore, I will resolve this conversation.
Feel free to unresolve it later if you still think that is more to be discussed here.

rfc/20260406-linting.md
yottta: I agree that the integration with `tofu-ls` will be an interesting part of this, but one option that I think about currently is to avoid implementing linting in `tofu-ls` directly and instead have it rely on the output of `tofu <cmd> -lint="..." -json`.

rfc/20260406-linting.md
yottta: If in a possible future iteration, when an allegedly configuration file will exist for linting, I would argue that the order individual rule IDs or group IDs must not matter. 
Just do a quick exercise: how painful would be for somebody to add a new rule ID in a list of included/excluded rules when that list already contains tens of rules and you need to find the perfect spot in that list to add your new rule in such way to make it work but in the same time not to break the other existing ones.

To conclude: IMO, the second option is the way to go and exactly as @apparentlymart said, the matching of the linting message should start with the individual rule ID which will be a quick return if it's specifically included/excluded.
If not, it will go to the groups level and if not specifically mentioned, it will try to look for the `all` group.
So the matching will be a 3 steps implementation.

Any feedback on this is appreciated since I would like to have something included in the RFC soon.

apparentlymart:
Since the maintainers approved this verbally in a synchronous discussion, and we already have the tracking issue https://github.com/opentofu/opentofu/issues/4310, I'm going to merge this now.

Any future amendments to the RFC text will happen in subsequent PRs associated with that tracking issue, as described in [Amending an RFC](https://github.com/opentofu/opentofu/blob/main/rfc/README.md#amending-an-rfc).

PR #4013, opened 2026-04-08
CHANGELOG: Begin the v1.13 development period

apparentlymart:
The `main` branch now represents v1.13 development, because v1.12 has diverged into its own maintenance branch.


cam72cam:
We should also update `version/VERSION` as  part of this PR

cam72cam (approved):

PR #4014, opened 2026-04-08
local-exec: propagate TRACEPARENT to child processes

### LINKED ISSUE #3936: OpenTelemetry TRACEPARENT propagation to child processes
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.5
```

When OpenTelemetry tracing is enabled, the parent trace context is not propagated to child processes

### Attempted Solutions

Configured OpenTelemetry tracing and used `local-exec` provisioner to check if `TRACEPARENT` is propagated:

  ```hcl
  resource "null_resource" "trace_test" {
    provisioner "local-exec" {
      command = "env | grep -i traceparent"
    }
  }

```

```bash

  export OTEL_TRACES_EXPORTER=otlp
  export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
  export OTEL_EXPORTER_OTLP_INSECURE=true


  tofu apply -auto-approve
```

Output: OTEL config variables are present in the child process environment but TRACEPARENT is missing. 

There is no configuration option or workaround within OpenTofu to inject trace context into child processes

### Proposal

When OpenTelemetry tracing is enabled, and a child process is spawned (`local-exec` /`external` / provider), OpenTofu should inject the `TRACEPARENT` environment variable following the [W3C Trace Context](https://www.w3.org/TR/trace-context/) specification.

N/A

### References

End to end example
https://github.com/denis256/terragrunt-tests/tree/master/tofu-telemetry

Output:
```bash
null_resource.trace_test: Provisioning with 'local-exec'...
null_resource.trace_test (local-exec): Executing: ["/bin/sh" "-c" "bash ./print-env.sh"]
null_resource.trace_test (local-exec): === OpenTelemetry Environment Variables ===
null_resource.trace_test (local-exec): OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
null_resource.trace_test (local-exec): OTEL_EXPORTER_OTLP_INSECURE=true
null_resource.trace_test (local-exec): OTEL_TRACES_EXPORTER=otlp

null_resource.trace_test (local-exec): === TRACEPARENT value ===
null_resource.trace_test (local-exec): TRACEPARENT=not set

null_resource.trace_test: Creation complete after 0s [id=2102028165570962689]
```

ogormans-deptstack:
Resolves #3936

This is PR 2 of 2. PR 1 (#3997, merged) threaded `context.Context` through the provisioner interface. This PR uses that context to propagate trace information to local-exec child processes.

When tracing is active the local-exec provisioner now sets `TRACEPARENT` in the child process environment, formatted per the W3C Trace Context spec (`00-{traceID}-{spanID}-{flags}`). If the user already sets `TRACEPARENT` in the provisioner `environment` block, that value takes precedence.

Uses `trace.SpanContextFromContext(ctx)` from the existing `go.opentelemetry.io/otel/trace` dependency — no new dependencies added, inline implementation as discussed on the issue.

## Checklist

### Go checklist

### DIFF
--- a/internal/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/internal/builtin/provisioners/local-exec/resource_provisioner.go
@@ -17,7 +17,7 @@ import (
 	"github.com/armon/circbuf"
 	"github.com/mitchellh/go-linereader"
 	"github.com/zclconf/go-cty/cty"
-	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/propagation"
 
 	"github.com/opentofu/opentofu/internal/configs/configschema"
 	"github.com/opentofu/opentofu/internal/provisioners"
@@ -117,10 +117,12 @@ func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.Pr
 
 	envVal := req.Config.GetAttr("environment")
 	var env []string
+	var traceparentConfigured bool
 
 	if !envVal.IsNull() {
 		for k, v := range envVal.AsValueMap() {
 			if !v.IsNull() {
+				traceparentConfigured = traceparentConfigured || strings.EqualFold(k, "TRACEPARENT")
 				entry := fmt.Sprintf("%s=%s", k, v.AsString())
 				env = append(env, entry)
 			}
@@ -175,14 +177,11 @@ func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.Pr
 	// variable (per the W3C Trace Context specification), unless the user
 	// has already set TRACEPARENT explicitly in the provisioner's
 	// "environment" block.
-	if !hasEnvVar(env, "TRACEPARENT") {
-		if sc := trace.SpanContextFromContext(ctx); sc.IsValid() {
-			cmdEnv = append(cmdEnv, fmt.Sprintf(
-				"TRACEPARENT=00-%s-%s-%s",
-				sc.TraceID().String(),
-				sc.SpanID().String(),
-				sc.TraceFlags().String(),
-			))
+	if !traceparentConfigured {
+		carrier := make(propagation.MapCarrier)
+		propagation.TraceContext{}.Inject(ctx, carrier)
+		for k, v := range carrier {
+			cmdEnv = append(cmdEnv, strings.ToUpper(k)+"="+v)
 		}
 	}
 
@@ -259,15 +258,4 @@ func copyUIOutput(o provisioners.UIOutput, r io.Reader, doneCh chan<- struct{}) 
 	for line := range lr.Ch {
 		o.Output(line)
 	}
-}
-
-// hasEnvVar checks whether a slice of "KEY=VALUE" strings contains an entry
-// for the given variable name.
-func hasEnvVar(env []string, name string) bool {
-	for _, e := range env {
-		if _, ok := strings.CutPrefix(e, name+"="); ok {
-			return true
-		}
-	}
-	return false
 }
--- a/internal/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/internal/builtin/provisioners/local-exec/resource_provisioner.go
@@ -17,7 +17,7 @@ import (
 	"github.com/armon/circbuf"
 	"github.com/mitchellh/go-linereader"
 	"github.com/zclconf/go-cty/cty"
-	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/opentofu/opentofu/internal/configs/configschema"
 	"github.com/opentofu/opentofu/internal/provisioners"
@@ -178,10 +178,13 @@ func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.Pr
 	// has already set TRACEPARENT explicitly in the provisioner's
 	// "environment" block.
 	if !traceparentConfigured {
-		carrier := make(propagation.MapCarrier)
-		propagation.TraceContext{}.Inject(ctx, carrier)
-		for k, v := range carrier {
-			cmdEnv = append(cmdEnv, strings.ToUpper(k)+"="+v)
+		if sc := trace.SpanContextFromContext(ctx); sc.IsValid() {
+			cmdEnv = append(cmdEnv, fmt.Sprintf(
+				"TRACEPARENT=00-%s-%s-%s",
+				sc.TraceID().String(),
+				sc.SpanID().String(),
+				sc.TraceFlags().String(),
+			))
 		}
 	}
 
--- a/internal/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/internal/builtin/provisioners/local-exec/resource_provisioner.go
@@ -16,6 +16,7 @@ import (
 	"github.com/armon/circbuf"
 	"github.com/mitchellh/go-linereader"
 	"github.com/zclconf/go-cty/cty"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/opentofu/opentofu/internal/configs/configschema"
 	"github.com/opentofu/opentofu/internal/provisioners"
@@ -97,7 +98,7 @@ func (p *provisioner) ValidateProvisionerConfig(req provisioners.ValidateProvisi
 	return resp
 }
 
-func (p *provisioner) ProvisionResource(_ context.Context, req provisioners.ProvisionResourceRequest) (resp provisioners.ProvisionResourceResponse) {
+func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.ProvisionResourceRequest) (resp provisioners.ProvisionResourceResponse) {
 	commandVal := req.Config.GetAttr("command")
 	if commandVal.IsNull() || commandVal.AsString() == "" {
 		resp.Diagnostics = resp.Diagnostics.Append(tfdiags.WholeContainingBody(
@@ -163,6 +164,22 @@ func (p *provisioner) ProvisionResource(_ context.Context, req provisioners.Prov
 	var cmdEnv []string
 	cmdEnv = os.Environ()
 	cmdEnv = append(cmdEnv, env...)
+
+	// If the caller's context carries an active OpenTelemetry trace span,
+	// propagate it to the child process via the TRACEPARENT environment
+	// variable (per the W3C Trace Context specification), unless the user
+	// has already set TRACEPARENT explicitly in the provisioner's
+	// "environment" block.
+	if !hasEnvVar(env, "TRACEPARENT") {
+		if sc := trace.SpanContextFromContext(ctx); sc.IsValid() {
+			cmdEnv = append(cmdEnv, fmt.Sprintf(
+				"TRACEPARENT=00-%s-%s-%s",
+				sc.TraceID().String(),
+				sc.SpanID().String(),
+				sc.TraceFlags().String(),
+			))
+		}
+	}
 
 	// Set up the command
 	cmd := exec.CommandContext(p.ctx, cmdargs[0], cmdargs[1:]...)
@@ -237,4 +254,16 @@ func copyUIOutput(o provisioners.UIOutput, r io.Reader, doneCh chan<- struct{}) 
 	for line := range lr.Ch {
 		o.Output(line)
 	}
+}
+
+// hasEnvVar checks whether a slice of "KEY=VALUE" strings contains an entry
+// for the given variable name.
+func hasEnvVar(env []string, name string) bool {
+	prefix := name + "="
+	for _, e := range env {
+		if len(e) >= len(prefix) && e[:len(prefix)] == prefix {
+			return true
+		}
+	}
+	return false
 }
--- a/internal/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/internal/builtin/provisioners/local-exec/resource_provisioner.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strings"
 
 	"github.com/armon/circbuf"
 	"github.com/mitchellh/go-linereader"
@@ -99,6 +100,10 @@ func (p *provisioner) ValidateProvisionerConfig(req provisioners.ValidateProvisi
 }
 
 func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.ProvisionResourceRequest) (resp provisioners.ProvisionResourceResponse) {
+	// We ignore cancellation of this incoming context because provisioners
+	// currently get cancelled by calling the "Stop" method instead.
+	ctx = context.WithoutCancel(ctx)
+
 	commandVal := req.Config.GetAttr("command")
 	if commandVal.IsNull() || commandVal.AsString() == "" {
 		resp.Diagnostics = resp.Diagnostics.Append(tfdiags.WholeContainingBody(
@@ -259,9 +264,8 @@ func copyUIOutput(o provisioners.UIOutput, r io.Reader, doneCh chan<- struct{}) 
 // hasEnvVar checks whether a slice of "KEY=VALUE" strings contains an entry
 // for the given variable name.
 func hasEnvVar(env []string, name string) bool {
-	prefix := name + "="
 	for _, e := range env {
-		if len(e) >= len(prefix) && e[:len(prefix)] == prefix {
+		if _, ok := strings.CutPrefix(e, name+"="); ok {
 			return true
 		}
 	}
--- a/internal/builtin/provisioners/local-exec/resource_provisioner_test.go
+++ b/internal/builtin/provisioners/local-exec/resource_provisioner_test.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/mitchellh/cli"
 	"github.com/zclconf/go-cty/cty"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/opentofu/opentofu/internal/provisioners"
 )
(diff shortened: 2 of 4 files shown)


internal/builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -164,6 +165,22 @@ func (p *provisioner) ProvisionResource(_ context.Context, req provisioners.Prov
...
+	// has already set TRACEPARENT explicitly in the provisioner's
+	// "environment" block.
+	if !hasEnvVar(env, "TRACEPARENT") {
+		if sc := trace.SpanContextFromContext(ctx); sc.IsValid() {
+			cmdEnv = append(cmdEnv, fmt.Sprintf(
+				"TRACEPARENT=00-%s-%s-%s",
+				sc.TraceID().String(),
+				sc.SpanID().String(),
+				sc.TraceFlags().String(),
+			))
```
apparentlymart, line 187: Out of curiosity I went to see if there was a convenient function for generating this in one of the OpenTelemetry modules we're already using.

The package `go.opentelemetry.io/otel/propagation` has `TraceContext.Inject`, but that seems like a far more complex way to achieve an equivalent result to what you implemented here, and so I prefer the simpler approach you used here.

I notice that the upstream function uses `hex.Encode` (from `encoding/hex`) directly, instead of relying on the `String` method of each of these types, but the `String` methods for these three types all seem to do the same thing just in a more manual way, such as [for `TraceID`](https://github.com/open-telemetry/opentelemetry-go/blob/7ac83c37af755351a0ac5ecf0d5885ac392df185/trace/trace.go#L66-L85), so this should achieve the same result.

ogormans-deptstack: Preferring the more direct cleaner approach but a failing test on windows again... damn golang and its compatibility guarantees!!

apparentlymart:
Do you understand what's going on with this test failure on Windows?

I would've expected what you shared to work, since `%NAME%` is the usual syntax for environment variable syntax in `cmd.exe`, so I must admit that _I'm_ not sure what's going on there. Do you think that the environment variable isn't being set for some reason?

ogormans-deptstack:
> Do you understand what's going on with this test failure on Windows?
> 
> I would've expected what you shared to work, since `%NAME%` is the usual syntax for environment variable syntax in `cmd.exe`, so I must admit that _I'm_ not sure what's going on there. Do you think that the environment variable isn't being set for some reason?

Not immediately obvious, I'll dig tomorrow with a fresh head

ogormans-deptstack:
Figured it out — `cmd.exe` doesn't expand `%TRACEPARENT%` when the variable is undefined, it leaves the literal `%TRACEPARENT%` in the output. On Unix `${TRACEPARENT:-unset}` falls back to "unset", but Windows has no equivalent default-value syntax.

Fixed in 8000ea2 by splitting the assertion per platform — on Windows we check for the unexpanded literal, on Unix we check for "unset". Both confirm the provisioner didn't inject TRACEPARENT.

Second time Windows shell differences have caught me in Go tests. The runtime.GOOS branching for the _command_ was right, I just forgot to branch the _assertion_ too. Something to keep in mind for any future tests that verify env var absence through shell echo — or better yet, avoid shell-based assertions for negative cases entirely when there's a way to inspect the environment directly in Go.

ogormans-deptstack:
Rebased onto main to pick up the v1.13 changelog restructuring (579930a). Conflict was just the CHANGELOG — moved our entry into the new 1.13.0 section.

The Windows fix from the last push is preserved in the rebase. For context on what went wrong there, see the [inline thread](https://github.com/opentofu/opentofu/pull/4014#discussion_r3054355112) — short version is `cmd.exe` silently passes through `%VAR%` as a literal when the variable isn't defined, unlike Unix shells which have fallback syntax.

internal/builtin/provisioners/local-exec/resource_provisioner_test.go
```diff
@@ -338,3 +339,161 @@ func TestResourceProvisioner_nullsInOptionals(t *testing.T) {
 		})
 	}
 }
+
+// testSpanContext returns a known-valid SpanContext for use in tests.
+func testSpanContext() trace.SpanContext {
+	traceID, _ := trace.TraceIDFromHex("4bf92f3577b34da6a3ce929d0e0e4736")
+	spanID, _ := trace.SpanIDFromHex("00f067aa0ba902b7")
```
apparentlymart: I'm sorry that this is a bit of a "nit", but I'd suggest setting these fake ids to something that seems "more fake", like a repeating pattern of `0f` or similar, just so that the expected results in the other test cases below read more clearly as testing against a fixed placeholder instead of against a dynamic trace ID.

internal/builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -238,3 +255,15 @@ func copyUIOutput(o provisioners.UIOutput, r io.Reader, doneCh chan<- struct{})
...
+// for the given variable name.
+func hasEnvVar(env []string, name string) bool {
+	prefix := name + "="
+	for _, e := range env {
+		if len(e) >= len(prefix) && e[:len(prefix)] == prefix {
+			return true
+		}
+	}
+	return false
+}
```
apparentlymart: This is more just a note for me than a review comment, but this seems like another variation that could benefit from the upstream Go proposal https://github.com/golang/go/issues/62038 or https://github.com/golang/go/issues/75934.

It's pretty similar to the examples I discussed in https://github.com/golang/go/issues/62038#issuecomment-1679871335. (In case it isn't obvious, "the main codebase I maintain in my day job" there is referring to OpenTofu. 😀 )

internal/builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -97,7 +98,7 @@ func (p *provisioner) ValidateProvisionerConfig(req provisioners.ValidateProvisi
 	return resp
 }
 
-func (p *provisioner) ProvisionResource(_ context.Context, req provisioners.ProvisionResourceRequest) (resp provisioners.ProvisionResourceResponse) {
+func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.ProvisionResourceRequest) (resp provisioners.ProvisionResourceResponse) {
 	commandVal := req.Config.GetAttr("command")
```
apparentlymart, line 107: Let's start this function with the following just to make sure that cancellation from the context can't possibly propagate to anything in here:

```go
    // We ignore cancellation of this incoming context because provisioners
    // currently get cancelled by calling the "Stop" method instead.
    ctx = context.WithoutCancel(ctx)
```

At the moment this function isn't actually using `ctx` for anything except the trace context (it uses the private `p.ctx` field instead, which is how `Stop` is implemented) but future maintainers might not be aware of that subtlety and would probably just follow the usual pattern of using whatever context is in scope. My hope of including this statement and the associated comment is to help a future maintainer notice that something unusual is going on here.

(Hopefully one day we'll rework provisioners to do cancellation using the modern `context` patterns, but I think we should do that separately from implementing this feature.)

internal/builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -238,3 +255,15 @@ func copyUIOutput(o provisioners.UIOutput, r io.Reader, doneCh chan<- struct{})
 		o.Output(line)
 	}
 }
+
+// hasEnvVar checks whether a slice of "KEY=VALUE" strings contains an entry
+// for the given variable name.
+func hasEnvVar(env []string, name string) bool {
+	prefix := name + "="
+	for _, e := range env {
+		if len(e) >= len(prefix) && e[:len(prefix)] == prefix {
```
apparentlymart: Could we use [`strings.CutPrefix`](https://pkg.go.dev/strings#CutPrefix) here? 🤔

apparentlymart (commented):
Thanks for working on this, and thanks especially for the detailed test cases!

I left some small things inline, but overall this looks great.

ogormans-deptstack:
Addressed the review feedback in 860953a:

- Added `context.WithoutCancel(ctx)` at function entry with the guard comment — good call, that's a subtle trap for future maintainers
- Swapped manual prefix check in `hasEnvVar` for `strings.CutPrefix`
- Replaced the W3C example trace/span IDs with repeating `0f` patterns so the expected values in assertions read as obviously-fake placeholders

Re: the Go proposal note on `hasEnvVar` — interesting thread, I can see how a stdlib `slices.ContainsFunc` variant with early-return semantics would clean up patterns like this.

apparentlymart (commented):
Thanks for the updates!

One final thing I thought about only now (sorry!) is that we should probably mention this in [the documentation for `local-exec`](https://opentofu.org/docs/language/resources/provisioners/local-exec/), or else nobody will know this is even possible. 😀 

We don't currently have an established practice for how to document OpenTelemetry-related behaviors of specific features, so for now I suggest we just add a section to the end that's something like this:

> ## OpenTelemetry Trace Propagation
>
> If you run OpenTofu with [OpenTelemetry Tracing](https://opentofu.org/docs/internals/tracing/) enabled, the `local-exec` provisioner will automatically set the `TRACEPARENT` environment variable when launching the specified program, so that it can contribute its own spans to the overall trace.
>
> Programs launched with this provisioner also inherit environment variables from the parent OpenTofu process by default, so the child program can also access any `OTEL_`-prefixed environment variables you set to enable OpenTofu's tracing functionality.

(The link to the "OpenTelemetry Tracing" page will need to be written in a different way in the website markdown, referring to the source `.md` file rather than its final published URL on the site, but I used the published URL above just so it's clear which page I was intending this to be linked to.)

After that I think this'll be ready to go, but I'll wait for at least one other maintainer to approve before merging.

ogormans-deptstack:
Added the docs section in 00cf64d — no hope of eventually making release notes if we don't document that it exists.

Thanks for sharing the longer-term vision in the review comments here — the context cancellation patterns, Go proposals for env var handling, eventual provisioner modernisation. It gives a clear picture of where this fits in the broader trajectory, which makes it easier to contribute in the right direction.

apparentlymart (approved):

apparentlymart:
The changelog became conflicted again, but that unfortunately just tends to happen more at the start of a new development period where all of the PRs are trying to be the first entry of a particular type. :confounded: 

If you'd like to rebase again to fix that then go ahead, but also don't feel like you need to keep chasing conflicts in the changelog if other things get merged in the meantime since we can always fix it up just before we merge (assuming that you've allowed maintainers to push to the branch in your fork, which is usually the default).

ogormans-deptstack:
Rebased onto main again to pick up the WinRM removal (fb7fc49). Still early in the 1.13 cycle so keeping the changelog clean as we go seems worth the churn.

internal/builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -238,3 +260,14 @@ func copyUIOutput(o provisioners.UIOutput, r io.Reader, doneCh chan<- struct{})
 		o.Output(line)
 	}
 }
+
+// hasEnvVar checks whether a slice of "KEY=VALUE" strings contains an entry
+// for the given variable name.
+func hasEnvVar(env []string, name string) bool {
```
yottta: How about not creating a new function to check this and instead, plug into the [logic that already iterates over the configured environment](https://github.com/ogormans-deptstack/opentofu/blob/bd49502d3a8a3d523aa436cdb4d813e9cc8ae695/internal/builtin/provisioners/local-exec/resource_provisioner.go#L118-L128) of the provisioner?

Something like:
```go
	envVal := req.Config.GetAttr("environment")
	var env []string
	var traceparentConfigured bool

	if !envVal.IsNull() {
		for k, v := range envVal.AsValueMap() {
			if !v.IsNull() {
				traceparentConfigured = traceparentConfigured || strings.EqualFold(k, "TRACEPARENT")
				entry := fmt.Sprintf("%s=%s", k, v.AsString())
				env = append(env, entry)
			}
		}
	}
```

internal/builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -164,6 +170,22 @@ func (p *provisioner) ProvisionResource(_ context.Context, req provisioners.Prov
...
 
+	// If the caller's context carries an active OpenTelemetry trace span,
+	// propagate it to the child process via the TRACEPARENT environment
+	// variable (per the W3C Trace Context specification), unless the user
+	// has already set TRACEPARENT explicitly in the provisioner's
+	// "environment" block.
+	if !hasEnvVar(env, "TRACEPARENT") {
+		if sc := trace.SpanContextFromContext(ctx); sc.IsValid() {
+			cmdEnv = append(cmdEnv, fmt.Sprintf(
+				"TRACEPARENT=00-%s-%s-%s",
```
yottta, line 183: The only concern I have with this, is that if in the future the version of this header (i.e.: `00`) will be bumped up, this will have to be migrated manually and most probably will be missed.
Sadly, as @apparentlymart pointed out in its previous comment, the `version` is part of the `TraceContext.Inject` and not exported from that package.

And if the version will be ever bumped up, the format might change too, which will require additional research and reimplementation for whoever will do the dependencies upgrade, adding even more pressure on the recurrent action we have to do before a new major version release (just to be clear what I mean by "recurrent action", this was the one for v1.12: #3877).

Just for this reason, I would still try and find a clean way of using that otel library provided method instead of building this manually.

internal/builtin/provisioners/local-exec/resource_provisioner_test.go
```diff
@@ -338,3 +339,161 @@ func TestResourceProvisioner_nullsInOptionals(t *testing.T) {
...
+	t.Setenv("TRACEPARENT", "")
+
+	output := cli.NewMockUi()
+	p := New()
+	schema := p.GetSchema().Provisioner
+
+	command := "echo traceparent_is_${TRACEPARENT:-unset}"
+	if runtime.GOOS == "windows" {
+		command = "echo traceparent_is_%TRACEPARENT%"
+	}
```
yottta, line 406:
```suggestion
	command := "echo traceparent_is_${TRACEPARENT:-unset}"
	expectedOutput := "traceparent_is_unset"
	if runtime.GOOS == "windows" {
		command = "echo traceparent_is_%TRACEPARENT%"
		// On Unix, the shell substitutes ${TRACEPARENT:-unset} to "unset".
	        // On Windows, cmd.exe leaves %TRACEPARENT% unexpanded when the
	        // variable is empty, so the literal marker appears in the output.
	        // Either result confirms the provisioner did not inject TRACEPARENT.
		expectedOutput = "traceparent_is_%TRACEPARENT%"
	}
```
See the other suggestion from this file.

internal/builtin/provisioners/local-exec/resource_provisioner_test.go
```diff
@@ -338,3 +339,161 @@ func TestResourceProvisioner_nullsInOptionals(t *testing.T) {
...
+	// Either result confirms the provisioner did not inject TRACEPARENT.
+	if runtime.GOOS == "windows" {
+		if !strings.Contains(got, "traceparent_is_%TRACEPARENT%") {
+			t.Errorf("expected TRACEPARENT to be unset without span context, got %q", got)
+		}
+	} else {
+		if !strings.Contains(got, "traceparent_is_unset") {
+			t.Errorf("expected TRACEPARENT to be unset without span context, got %q", got)
+		}
+	}
```
yottta:
```suggestion
	if !strings.Contains(got, expectedOutput) {
		t.Errorf("expected TRACEPARENT to be unset without span context, got %q", got)
	}
```

ogormans-deptstack:
Thanks for the review. Updated:

- Switched to `propagation.TraceContext{}.Inject(ctx, carrier)` instead of manual TRACEPARENT formatting — good call on the maintainability angle, and this picks up TRACESTATE for free.
- Inlined the TRACEPARENT check into the existing env loop using `strings.EqualFold`, removed the separate `hasEnvVar` function.
- Cleaned up test assertions — extracted `expectedOutput` variable instead of duplicated if/else blocks.

Rebased on main, all tests pass locally.

internal/builtin/provisioners/local-exec/resource_provisioner.go
apparentlymart: Since this environment variable can only be set to one version at a time, this is actually an interesting compatibility constraint: if some later version of OpenTofu suddenly started using `TRACEPARENT=01-...` instead then it would be breaking for anyone who is using this to run a program that only understands version zero.

So I think whatever strategy we take here, OpenTofu should be the one in control of which version it's using so that we can decide when and how to switch to a newer version later. I wouldn't like to get into a situation where we can't upgrade our OpenTelemetry dependencies without it being a breaking change. :thinking:

yottta: Ok, good points. I agree with those.

ogormans-deptstack: Agreed — @apparentlymart's point about keeping explicit version control is the stronger argument. A silent otel dep bump shouldn't be able to change what we emit to child processes — that's unseen technical debt we'd rather not trade off against.

Reverted the Inject approach back to manual formatting in the latest push. Kept the other two changes (inlined env check, test cleanup) since those stand on their own.

Thanks for the rereview @apparentlymart, appreciate the experienced eyes on this.

@yottta if you're happy with where this landed, would you be able to approve?

yottta (approved):

diofeher (approved):

apparentlymart:
Thanks for working on this @ogormans-deptstack!

Now that it's merged, it should be included in the v1.13.0 release later.

PR #4016, opened 2026-04-08
fix: prevent panic in tofu login OAuth flow due to channel race

### LINKED ISSUE #3988: Bug: `tofu login` can panic with "close of closed channel" during OAuth flow
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.12.0-dev on darwin_arm64
```

```hcl
...tofu config...
```
N/A — this bug is in the CLI OAuth login flow, not in `.tf` configuration files.

### Debug Output

The panic can be reproduced with the following minimal test that mirrors the exact goroutine structure of `interactiveGetTokenByCode`. It is **not** a test of `LoginCommand` itself, but isolates the channel mechanics at lines 454/479/527:

```go
// Place inside internal/command/ (package command) and run:
// go test -race -run TestReproDoubleClose ./internal/command/
func TestReproDoubleClose(t *testing.T) {
    for i := 0; i < 200; i++ {
        func() {
            defer func() {
                if r := recover(); r != nil {
                    t.Errorf("panic (no recover in production): %v", r)
                }
            }()

            codeCh := make(chan string)
            triggerServerError := make(chan struct{})

            // close #1 — mirrors HTTP callback handler (line 454)
            go func() {
                codeCh <- "auth-code"
                close(codeCh)
            }()

            // close #2 — mirrors server.Serve() error path (line 479)
            go func() {
                <-triggerServerError
                close(codeCh) // panics if codeCh already closed
            }()

            go func() {
                time.Sleep(time.Microsecond * time.Duration(i%3))
                close(triggerServerError)
            }()

            // close #3 — mirrors ShutdownCh select branch (line 527)
            shutdownCh := make(chan struct{}, 1)
            select {
            case <-shutdownCh:
                close(codeCh)
            case <-codeCh:
            }
        }()
    }
}
```

Running this produces:

```
=== RUN   TestReproDoubleClose
panic: close of closed channel

goroutine 24 [running]:
...internal/command/login_race_repro_test.go:81 +0x80
created by ...interactiveGetTokenByCode.func1
        internal/command/login.go:479
FAIL    github.com/opentofu/opentofu/internal/command   1.020s
```

### Expected Behavior

`tofu login` should handle concurrent shutdown signals and OAuth callback responses gracefully, without panicking.

### Actual Behavior

`tofu login` panics with:
Image

### Steps to Reproduce

**Scenario A — Ctrl+C races with a valid OAuth callback:**

1. Run `tofu login app.terraform.io`
2. Complete the OAuth flow in the browser (click "Approve")
3. Press Ctrl+C at the exact moment the browser redirects back to `localhost`

**Scenario B — server error + Ctrl+C (more reliably reproducible):**

1. Block all TCP ports in the range used by the OAuth callback server (default 10000–10010)
2. Run `tofu login app.terraform.io`
3. Press Ctrl+C while the command is running

---

### Additional Context

- This race was introduced in PR #1074, which added Ctrl+C support to the OAuth login flow by adding the third `close(codeCh)` site (line 527), without accounting for the existing  close sites at lines 454 and 479.
- The bug is a timing race, so it will not occur on every run. It is most likely to surface under slow network conditions or when the OAuth callback and a Ctrl+C signal arrive simultaneously.

### References

- PR #1074 — "shutdown OAuth2 callback server when user hits Ctrl+C" (introduced close #3)

dongjune8931:
Resolves #3988

## Summary

This PR fixes a race condition in \`interactiveGetTokenByCode\` (\`internal/command/login.go\`) that could cause \`tofu login\` to panic with \`close of closed channel\` or deadlock when a shutdown signal (Ctrl+C) arrives concurrently with an OAuth callback.

### Root cause

Multiple goroutines could each call \`close(codeCh)\` independently:

1. HTTP callback handler — after receiving the auth code
2. \`server.Serve()\` error goroutine — when the server fails to start
3. \`case <-c.ShutdownCh\` select branch — on Ctrl+C

If any two of these fired concurrently, the second \`close\` would panic with \`close of closed channel\`.

Additionally, with \`codeCh\` being unbuffered, the HTTP handler could block forever trying to send if the main goroutine had already moved on, causing a deadlock.

### Fix

- Make \`codeCh\` a buffered channel (capacity 1) so the HTTP handler can send without blocking even if the main goroutine has already moved on
- Use a non-blocking \`select\` in the handler to reject duplicate callback requests instead of a \`sync.Mutex\` guard
- Introduce \`serverErrCh\` so the \`server.Serve\` goroutine never writes to shared \`diags\`, eliminating that data race path
- Add a \`sync.WaitGroup\` to track the \`server.Serve\` goroutine lifetime
- Use \`server.Shutdown()\` instead of \`server.Close()\` so that all in-flight HTTP handler goroutines finish before the main goroutine calls \`close(codeCh)\`
- The main goroutine is now the sole owner of \`codeCh\` — it is the only one that closes it, and only after all producers have stopped

## Checklist

### Go checklist

### DIFF
--- a/internal/command/login.go
+++ b/internal/command/login.go
@@ -19,6 +19,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	tfe "github.com/hashicorp/go-tfe"
 	"github.com/mitchellh/cli"
@@ -418,8 +419,14 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 	}
 
 	// codeCh will allow our temporary HTTP server to transmit the OAuth code
-	// to the main execution path that follows.
-	codeCh := make(chan string)
+	// to the main execution path that follows. It is buffered so the handler
+	// can send without blocking even if the main goroutine has already moved on.
+	// Only the main goroutine may close codeCh, after all producers have stopped.
+	codeCh := make(chan string, 1)
+	// serverErrCh carries any unexpected error from server.Serve so that the
+	// main goroutine can handle it without a shared-state race on diags.
+	serverErrCh := make(chan error, 1)
+	var wg sync.WaitGroup
 	server := &http.Server{
 		Handler: http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
 			log.Printf("[TRACE] login: request to callback server")
@@ -442,12 +449,17 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 				return
 			}
 
-			log.Printf("[TRACE] login: request contains an authorization code")
-
-			// Send the code to our blocking wait below, so that the token
-			// fetching process can continue.
-			codeCh <- gotCode
-			close(codeCh)
+			// Non-blocking send: only the first callback request succeeds.
+			// Duplicate or concurrent requests are rejected with 400 so that
+			// the handler never blocks on the already-full buffered channel.
+			select {
+			case codeCh <- gotCode:
+				log.Printf("[TRACE] login: request contains an authorization code")
+			default:
+				log.Printf("[WARN] login: ignoring duplicate callback request")
+				resp.WriteHeader(400)
+				return
+			}
 
 			log.Printf("[TRACE] login: returning response from callback server")
 
@@ -460,21 +472,13 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 		}),
 	}
 	panicHandler := logging.PanicHandlerWithTraceFn()
-	go func() {
+	wg.Go(func() {
 		defer panicHandler()
 		err := server.Serve(listener)
 		if err != nil && err != http.ErrServerClosed {
-			diags = diags.Append(tfdiags.Sourceless(
-				tfdiags.Error,
-				"Can't start temporary login server",
-				fmt.Sprintf(
-					"The login process uses OAuth, which requires starting a temporary HTTP server on localhost. However, no TCP port numbers between %d and %d are available to create such a server.",
-					clientConfig.MinPort, clientConfig.MaxPort,
-				),
-			))
-			close(codeCh)
+			serverErrCh <- err
 		}
-	}()
+	})
 
 	oauthConfig := &oauth2.Config{
 		ClientID:    clientConfig.ID,
@@ -509,7 +513,6 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 	view.WaitingForHostSignal()
 
 	var code string
-	var ok bool
 	select {
 	case <-c.ShutdownCh:
 		diags = diags.Append(
@@ -519,22 +522,36 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 				"Current command was aborted by the calling code.",
 			),
 		)
-		code, ok = "", true
+    if err := server.Shutdown(ctx); err != nil {
+		log.Printf("[WARN] login: callback server shutdown failed: %s", err)
+	}
+		wg.Wait()
 		close(codeCh)
-	case code, ok = <-codeCh:
-	}
-
-	if !ok {
-		// If we got no code at all then the server wasn't able to start
-		// up, so we'll just give up.
 		return nil, diags
+	case serveErr := <-serverErrCh:
+		// The server failed to start up, so we'll just give up.
+		// No need to call Shutdown here: the server never started accepting
+		// requests, so there are no in-flight handler goroutines to wait for.
+		log.Printf("[ERROR] login: callback server error: %s", serveErr)
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Can't start temporary login server",
+			fmt.Sprintf(
+				"The login process uses OAuth, which requires starting a temporary HTTP server on localhost. However, no TCP port numbers between %d and %d are available to create such a server.",
+				clientConfig.MinPort, clientConfig.MaxPort,
+			),
+		))
+		wg.Wait()
+		close(codeCh)
+		return nil, diags
+	case code = <-codeCh:
 	}
 
-	if err := server.Close(); err != nil {
-		// The server will close soon enough when our process exits anyway,
-		// so we won't fuss about it for right now.
+	if err := server.Shutdown(ctx); err != nil {
 		log.Printf("[WARN] login: callback server can't shut down: %s", err)
 	}
+	wg.Wait()
+	close(codeCh)
 
 	if code == "" {
 		// empty code is not possible in happy path as it is validated in the HTTP handler of our callback server
--- a/internal/command/login_test.go
+++ b/internal/command/login_test.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/opentofu/opentofu/internal/command/workdir"
 	"github.com/opentofu/opentofu/internal/terminal"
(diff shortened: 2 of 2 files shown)


dongjune8931:
I have a follow-up question about the remaining close(codeCh) in the server.Serve() error path (line 478).

  I was wondering whether there could theoretically be a race if server.Serve() returns an error while an HTTP handler goroutine is already running — in that case, close(codeCh) in the error goroutine and codeCh <- gotCode in the handler might execute concurrently, potentially causing a "send on closed channel" panic. Though I'm not entirely sure how likely this would be in practice.

  If this is a concern, one option might be to replace the close(codeCh) with a separate errCh chan struct{} for signaling server startup failure, so codeCh is never closed at all.

  Would you like me to make that change as part of this PR, or do you think the current approach is sufficient?

apparentlymart:
Hi @dongjune8931,

I don't think it's possible for the `close(codeCh)` statement to be reached in any case where the `codeCh <- gotCode` statement could be reached, because if the server fails to start up then there's no way for the handler function to run.

Does that make sense for you, or am I missing something? Thanks!

---

Separately: please ignore the request for a changelog entry on this one, and remove the `CHANGELOG.md` changes from this PR. Although it's _technically_ a user-facing change, it's not one we've ever seen an end-user report -- I believe _you_ only found it by analyzing the source code, rather than by actually encountering it, right? -- so I don't think it's worth mentioning in the changelog.

I'm sorry I didn't mention that before. That message just appears unconditionally whenever any PR is opened, so it doesn't take into account subjective questions about what exactly "user-visible change" means. :grinning: 

---

It's the end of my work day now so I'm not going to look at this too closely today, but I'll try to review it tomorrow. Thanks for working on this!

dongjune8931:
Thanks for the clarification! @apparentlymart
 That makes sense — if the server fails to start, the handler can never run, so there's no race possible there.
 I've removed the CHANGELOG.md changes as well. Looking forward to your review!

internal/command/login_test.go
```diff
@@ -369,3 +370,160 @@ func TestLogin(t *testing.T) {
...
+//
+//	go test -race -run TestLoginOAuthCallbackRace ./internal/command/
+func TestLoginOAuthCallbackRace(t *testing.T) {
+	s := httptest.NewServer(oauthserver.Handler)
+	defer s.Close()
+
+	const iterations = 200
+
+	for i := 0; i < iterations; i++ {
+		i := i
```
apparentlymart:
```suggestion
	for i := range iterations {
```

In current versions of Go it's not necessary to shadow the loop induction variable like this, because each iteration has its own distinct `i`. We can also use the "range over integer" syntax that was added from https://github.com/golang/go/issues/61405.

(I think if you run `go fix ./internal/command` with Go 1.26 then it'll "modernize" this for you automatically, but if you do that then please only include changes to the code you've added here. There's other code in `package command` that _could_ be modernized but we're not ready to do that yet because of the policy in https://github.com/opentofu/opentofu/issues/3773.)

internal/command/login_test.go
```diff
@@ -369,3 +370,160 @@ func TestLogin(t *testing.T) {
...
+			abortCh := make(chan struct{}, 1)
+			c := &LoginCommand{
+				Meta: Meta{
+					WorkingDir:      workdir.NewDir("."),
+					View:            loginView,
+					BrowserLauncher: webbrowser.NewMockLauncher(ctx),
+					Services:        svcs,
+				},
+			}
+			c.ShutdownCh = abortCh
```
apparentlymart: Could this be set as part of the `Meta{...}` literal above, rather than separately afterwards? 🤔 

(This also applies to the similar code in the other test below.)

internal/command/login_test.go
```diff
@@ -369,3 +370,160 @@ func TestLogin(t *testing.T) {
...
+			defer testInputMap(t, map[string]string{
+				"approve": "yes",
+			})()
+
+			statusCh := make(chan int, 1)
+			go func() {
+				statusCh <- c.Run([]string{"example.com"})
+			}()
+
+			abortCh <- struct{}{}
```
apparentlymart:
```suggestion
			close(abortCh)
```

Although the code that interacts with `Meta.ShutdownCh` does currently accept sending a value to this channel as a "shutdown" signal, the documentation for `Meta.ShutdownCh` says that _closing_ this channel is the expected protocol.

(This also applies to the send to `abortCh` in the other test above.)

apparentlymart (commented):
Thanks again for working on this, @dongjune8931!

Overall this looks good to me. I have some comments inline that are just about some small details.

Also, just so you know what to expect: because I accepted the associated bug report unilaterally using our "fast path" process, I'll wait for another maintainer to approve before we merge this just because that's how we give the other maintainers an opportunity to disagree about whether this should've been fast path.

apparentlymart (approved):
Thanks for the updates!

This looks good to me, but as previously mentioned I'm going to wait to give other maintainers a chance to review before merging.

apparentlymart:
It seems that GitHub thinks there's a merge conflict with the `main` branch now, but it isn't describing the details of that merge conflict.

I have a feeling it's probably the changelog commit, even though it was reverted later. Might be best to squash all of your work here down into a single commit so that the steps of adding and then removing the changelog entry would be discarded, and then this will hopefully merge cleanly.

dongjune8931:
@apparentlymart
Done, squashed all commits into one. Thanks!

yottta (commented):
Thanks for the help.

Even though this works, it looks quite hacky.
I've been through this kind of issue several times and always the issue was the same: I didn't follow the separation of concerns around channels, meaning that "who opens the channel, owns the channel and it's responsible with closing it".
When starting to close channels everywhere, the handling tends to get pretty messy (as it's the case of this command).

Personally, I would recommend the following approach instead:
1. `main` goroutine is responsible with creating and closing. No other goroutine should call close on the channel.
2. the `main` goroutine is responsible with stopping the "channel producers". In this case it's quite easy:
    1. Because the `main` goroutine also controls the creation of the temporary http server, which is started in another goroutine, a simple `sync.WaitGroup` should be enough to signal when the "channel producers" returned. This will allow for the `main` goroutine to wait before closing the channel (guaranteeing no panic at all)
3. Once the producers returned, close the channel:
    1. If `c.ShutdownCh` received something, call `server.Close()` which will close any other possible writers to that channel, call `wg.Wait()` and then `close(codeCh)`
    2. If a code is received, do the same as the step above this one.

This way there is a clear separation on "who owns and reads the channel" and "who writes to the channel and when are allowed to do so".

I will restrain from approving this since now we have 2 different opinions here. If there will be another maintainer that will agree with the current approach, I have no problem on having this merged without my approval.

dongjune8931:
Hi @yottta, thank you for the detailed suggestion!

I was thinking through your approach (`server.Close()` → `wg.Wait()` → `close(codeCh)`), and I'm wondering if there could still be a potential data race.

The `WaitGroup` only tracks the `server.Serve` goroutine, not the individual HTTP handler goroutines spawned internally by `http.Server`. So after `server.Close()` + `wg.Wait()`, the `Serve` goroutine has returned — but an in-flight handler goroutine may still be running and attempting `select { case codeCh <- gotCode: }`. Calling `close(codeCh)` at that point could race with the handler's send.

Am I missing something, or could this be an issue with the suggested approach?

yottta:
> Hi @yottta, thank you for the detailed suggestion!
> 
> I was thinking through your approach (`server.Close()` → `wg.Wait()` → `close(codeCh)`), and I'm wondering if there could still be a potential data race.
> 
> The `WaitGroup` only tracks the `server.Serve` goroutine, not the individual HTTP handler goroutines spawned internally by `http.Server`. So after `server.Close()` + `wg.Wait()`, the `Serve` goroutine has returned — but an in-flight handler goroutine may still be running and attempting `select { case codeCh <- gotCode: }`. Calling `close(codeCh)` at that point could race with the handler's send.
> 
> Am I missing something, or could this be an issue with the suggested approach?

Good question! It might be a thing that I left unclear in my suggestion.

The `sync.WaitGroup` will be used wherever channel writes are performed, meaning that the handler that writes to the channel will have to do `wg.Add(1); defer wg.Done()` right at the top.

Once we call `server.Close()`, there will be no other goroutines spawned by the server, meaning that the handler will not be called anymore. 
```go
// Close immediately closes all active net.Listeners and any
// connections in state [StateNew], [StateActive], or [StateIdle]. For a
// graceful shutdown, use [Server.Shutdown].
```

Additionally, please check [`StateActive` docs](https://pkg.go.dev/net/http#ConnState). The way I understand that, is that by calling `server.Close()` also the ongoing requests gets terminated.

To be safe, we can also add in the same `wg` the `server.Serve` goroutine and that will ensure that the `wg.Wait()` will release only after the `server.Close()` returns, meaning that there will be no other ongoing calls running the producing handler.

Please let me know if there are still unclear things here.

dongjune8931:
Hi @yottta, thnaks for the clarification!

I checked the `StateActive` docs as you suggested. I undestand your point — `server.Close()` closes connections in `StateActive`, which means ongoing requests get their underlying TCP connection closed.

However, I think closing the TCP connection doesn't actualy stop the handler goroutine itself. The goroutine keeps running in Go's runtime until it returns — it would only notice the closed connection when it tries to write a response and gets an error. So tehre's still a window where:
1. `server.Close()` is called → TCP connections closed
2. `server.Serve` goroutine returns → `wg.Done()` → counter hits 0
3. A handler goroutine is still running but hasn't called `wg.Add(1)` yet
4. `wg.Wait()` returns early
5. `close(codeCh)` races with the handler's send

Since handler goroutines are spawned internaly by `http.Server`, we don't have the opportunity to call `wg.Add(1)` before they're spawned — which is usualy the safe pattern.

Given this, I can think of two alternatives:
1. Use `server.Shutdown()` instead of `server.Close()` — it waits for all active handler goroutines to finish before returning, so `close(codeCh)` after `wg.Wait()` would be completely safe.
2. Simply not call `close(codeCh)` at all — since the channel is buffered and the handler uses a non-blocking `select`, there's no deadlock risk and the GC will handle cleanup.

Which approch do you think makes more sense?

yottta:
> Hi @yottta, thnaks for the clarification!
> 
> I checked the `StateActive` docs as you suggested. I undestand your point — `server.Close()` closes connections in `StateActive`, which means ongoing requests get their underlying TCP connection closed.
> 
> However, I think closing the TCP connection doesn't actualy stop the handler goroutine itself. The goroutine keeps running in Go's runtime until it returns — it would only notice the closed connection when it tries to write a response and gets an error. So tehre's still a window where:
> 
> 1. `server.Close()` is called → TCP connections closed
> 2. `server.Serve` goroutine returns → `wg.Done()` → counter hits 0
> 3. A handler goroutine is still running but hasn't called `wg.Add(1)` yet
> 4. `wg.Wait()` returns early
> 5. `close(codeCh)` races with the handler's send
> 
> Since handler goroutines are spawned internaly by `http.Server`, we don't have the opportunity to call `wg.Add(1)` before they're spawned — which is usualy the safe pattern.
> 
> Given this, I can think of two alternatives:
> 
> 1. Use `server.Shutdown()` instead of `server.Close()` — it waits for all active handler goroutines to finish before returning, so `close(codeCh)` after `wg.Wait()` would be completely safe.
> 2. Simply not call `close(codeCh)` at all — since the channel is buffered and the handler uses a non-blocking `select`, there's no deadlock risk and the GC will handle cleanup.
> 
> Which approch do you think makes more sense?

Totally forgot about the `server.Shutdown()` 🙌 
I would say that the first option is the way to go.

dongjune8931:
Got it @yottta! I agree that `server.Shutdown()` is the cleaner approach  going with that. Thanks for the guidance!

dongjune8931:
@yottta Updated to use server.Shutdown() as suggested !!
I squashed into a single commit as well

internal/command/login.go
```diff
@@ -459,20 +471,14 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 			}
 		}),
 	}
+	wg.Add(1)
```
yottta: You can use [`wg.Go()`](https://pkg.go.dev/sync#WaitGroup.Go) here

internal/command/login_test.go
```diff
@@ -369,3 +370,159 @@ func TestLogin(t *testing.T) {
...
+				if r := recover(); r != nil {
+					t.Errorf("iteration %d: panic in LoginCommand.Run (would crash tofu login): %v", i, r)
+				}
+			}()
+
+			workDir := t.TempDir()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			loginView, _ := testView(t)
```
yottta:
```suggestion
			loginView, done := testView(t)
			defer done(t)
```
Recommended to cleanup the buffers created by `testView`.

internal/command/login_test.go
```diff
@@ -369,3 +370,159 @@ func TestLogin(t *testing.T) {
...
+				if r := recover(); r != nil {
+					t.Errorf("iteration %d: unexpected panic: %v", i, r)
+				}
+			}()
+
+			workDir := t.TempDir()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			loginView, _ := testView(t)
```
yottta:
```suggestion
			loginView, done := testView(t)
			defer done(t)
```

yottta (approved):
Just some small suggestions but now looks way better.
Thanks a lot for the work on this and for the great exchanges 🙌

internal/command/login.go
```diff
@@ -519,22 +522,34 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
...
-		// If we got no code at all then the server wasn't able to start
-		// up, so we'll just give up.
 		return nil, diags
+	case code = <-codeCh:
 	}
 
-	if err := server.Close(); err != nil {
-		// The server will close soon enough when our process exits anyway,
-		// so we won't fuss about it for right now.
+	if err := server.Shutdown(context.Background()); err != nil {
```
yottta:
```suggestion
	if err := server.Shutdown(ctx); err != nil {
```

internal/command/login.go
```diff
@@ -519,22 +522,34 @@ func (c *LoginCommand) interactiveGetTokenByCode(ctx context.Context, hostname s
 				"Current command was aborted by the calling code.",
 			),
 		)
-		code, ok = "", true
+		server.Shutdown(context.Background())
```
yottta:
```suggestion
    if err := server.Shutdown(ctx); err != nil {
		log.Printf("[WARN] login: callback server shutdown failed: %s", err)
	}
```

yottta (commented):
Just took another look and apparently, I missed something small.
EDIT: Apparently, the code consistency check caught the same thing 😄

internal/command/login.go
apparentlymart: The documentation for `server.Shutdown` says:

> If the provided context expires before the shutdown is complete, Shutdown returns the context's error, otherwise it returns any error returned from closing the [Server](https://pkg.go.dev/net/http#Server)'s underlying Listener(s).

It doesn't mention what the behavior is if the given context was already cancelled/expired before calling the function, but from reading the current implementation it seems like it currently closes the listener anyway and the cancel just stops it waiting for already-active connections to finish before returning.

So I think overall this seems fine... we might just see some spurious `WARN` messages in the log saying `callback server shutdown failed: context cancelled` when `tofu login` is interrupted by SIGINT, which is not a big deal.

dongjune8931:
Hi @apparentlymart @yottta, just checking in — is there anything else you'd like me to address before this gets merged? Happy to make any further changes!

apparentlymart:
Hi @dongjune8931,

Sorry for the delay. I was just working on some other things while you were discussing with @yottta and didn't notice the second approval until now. 

I'll merge this in a moment. Thanks for working on it!

Issue #3912, opened 2026-03-19
Phase out official releases for 32-bit CPU architectures (x86 and ARM)


apparentlymart:
In https://github.com/opentofu/opentofu/issues/3278 we agreed to phase out our official releases for the 32-bit CPU architectures across all operating systems.

This issue is currently a placeholder for making a more detailed plan and schedule for implementing that change. We expect to keep producing these builds throughout the v1.12 and v1.13 releases unless there's a surprising change in the ecosystem for any of these architectures in the meantime, but we've not planned any further than that so far.

This issue will hopefully become the parent issue for a series of tasks across the next few minor release milestones that lead to us eventually producing a minor release which includes official packages for only the 64-bit architectures.

apparentlymart:
I'm adding this to the v1.13.0 milestone for now just to remind us to discuss it while we're planning that release. However, I expect the result of that discussion to be that we remove this issue from the milestone and add one or more sub-issues to the milestone instead, since we're expecting to work toward this change gradually over consecutive release series.

apparentlymart:
Unfortunately we got distracted by various other things and didn't end up making a final plan for how to proceed here. Since we're getting close to the likely end of the v1.13 development period I'm making the following proposal for how to proceed:

1. Merge https://github.com/opentofu/opentofu/pull/4018 before the v1.13 feature freeze so that v1.13.0-beta1 and all other releases in that series will emit a warning from `tofu init` when running on one of the 32-bit CPU architectures.

    (Note that the PR makes this happen only for our "official" builds, so third-party builds such as Linux distribution packages would be unaffected and can continue compiling for whatever platforms they choose to support without any spurious warnings.)

2. Early in the v1.14 development period, update the nightly and main release processes to no longer produce builds for these platforms and revert the changes from https://github.com/opentofu/opentofu/pull/4018 since that warning will no longer be needed in later series.

    At the same time we do this we would also remove those platforms from our "Quick Checks" unit test runs for pull requests, so we'd then only be running these tests routinely on platforms that are directly supported by GitHub Actions and would no longer need to rely on QEMU emulation.

The overall effect of this is that v1.13 will proactively warn anyone still using these platforms that future series will no longer have official packages, and then v1.14.0-beta1 and onwards would not include official packages for these platforms.

[...]

PR #4018, opened 2026-04-09
Warn about forthcoming end of official support when using OpenTofu v1.13 on a 32-bit CPU architecture

apparentlymart:
This is a draft idea for one way we could potentially produce a more prominent announcement about ending official support for 32-bit CPU architectures, for https://github.com/opentofu/opentofu/issues/3912.

When we discussed this change of support with OpenTofu's technical steering committee they requested that there be at least one release that produces an in-band warning so that anyone currently relying on these builds is more likely to notice it.

However, we want to still make a best effort to help third-parties like Linux distribution maintainers produce their own builds for these CPU architectures if they want to, rather than actively _preventing_ OpenTofu from running on these platforms. It would be annoying if those third-party builds _also_ started producing these warnings, since it's up to those folks if and when they want to stop supporting any platform they currently support.

This PR therefore implements a compromise:

1. Introduces a new concept of "official builds", which is activated by an extra environment variable that's only set in the GitHub Actions workflow that we use to produce the official release packages.

    This is activated in the GitHub Action because it seems reasonable that a third-party distributor might try to adapt our `.goreleaser.yml` for their own use but might not notice the "official build" concept and so might accidentally mark their non-official builds as being official. It seems far less likely that a third-party distributor would reuse our GitHub Actions workflow.

2. _For official builds only_, the `tofu init` command in particular checks whether it's running in a `386` or `arm` build and generates a warning if so.

    This is included only in `tofu init` as a compromise because it's a command that developers are likely to run just once when initializing their working directory, and so it wouldn't be quite so annoying there as it would be to re-warn about it on every `tofu plan` or `tofu apply` run.

    There is some risk that someone won't run `tofu init` immediately after upgrading from an earlier release, but they probably will _eventually_ run it at some point during the life of the v1.13 series. Even if they don't, we'll be announcing this change of support in other ways too and so it's likely that most folks affected by this will see _some_ communication about it, even if not _this_ communication.

I'm opening this as a draft now just in preparation for a discussion about what we might do for https://github.com/opentofu/opentofu/issues/3912 during the v1.13 development period. The maintainers have not discussed this yet, and so we might decide to do something quite different in the end.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,6 +126,9 @@ jobs:
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_TOKEN }}
           RELEASE_FLAG_LATEST: ${{ inputs.latest }}
           RELEASE_FLAG_PRERELEASE: ${{ inputs.prerelease }}
+          # This causes version.IsOfficialBuild to return true when someone is
+          # running one of our official release builds.
+          BUILD_OPENTOFU_OFFICIAL: '1'
 
       - name: Remove GPG key
         if: always()
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -24,6 +24,12 @@ builds:
     ldflags:
       - "-s -w"
       - "-X 'github.com/opentofu/opentofu/version.dev=no'"
+      # BUILD_OPENTOFU_OFFICIAL is set by the OpenTofu project's main release
+      # process. It should typically _not_ be set for builds produced by
+      # third-parties such as Linux distribution maintainers, because in that
+      # case the resulting executable might generate irrelevant warnings that
+      # are intended only for users of the official release builds.
+      - "-X 'github.com/opentofu/opentofu/version.officialBuild={{if index .Env \"BUILD_OPENTOFU_OFFICIAL\" }}{{ .Env.BUILD_OPENTOFU_OFFICIAL }}{{end}}'"
 
     goos:
       - linux
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ UPGRADE NOTES:
 
     [Modern Windows versions now support OpenSSH](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse), and so we suggest that anyone currently relying on WinRM plan to migrate to using SSH instead.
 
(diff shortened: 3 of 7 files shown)


apparentlymart:
I can see that the unit test is failing on `linux_386` and `linux_arm` -- the first time in a while that either of those test runs has generated a useful signal, ironically! I don't have a `linux_386` or `linux_arm` system to run the tests on outside of GitHub Actions, so I'm going to need to rely on those to actions runs to debug this.

I'll figure out what's going wrong with that at some point, but hopefully it's still clear from the diff what general approach I'm suggesting here, even though I didn't get the implementation quite right.

**Edit:** Turns out there was one return path in `tofu init` that wasn't calling `view.Diagnostics` first, and that happened to be the one that my test case was exercising! :grinning: This is fixed now.

.github/workflows/release.yml
```diff
@@ -126,6 +126,9 @@ jobs:
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_TOKEN }}
           RELEASE_FLAG_LATEST: ${{ inputs.latest }}
           RELEASE_FLAG_PRERELEASE: ${{ inputs.prerelease }}
+          # This causes version.IsOfficialBuild to return true when someone is
+          # running one of our official release builds.
+          BUILD_OPENTOFU_OFFICIAL: '1'
```
diofeher, line 131: While I really like this idea, what prevents someone from just turning on this flag and saying they're an official OpenTofu build?

I know this is probably out of scope (since you're developing that for the 32-bit arch), but I can see malicious users might want to tamper with it, claiming the binary is an official build.

Unfortunately that adds another layer of complexity to manage our keys, etc :(

apparentlymart: I don't expect that anyone would do it intentionally because there would be no benefit of doing so: it would just mean if they make their own build for e.g. `linux_386` then it would generate the same warning our official builds for that platform would produce, which would be confusing but not beneficial.

I am not expecting that we'd ever use this "official build" flag to make any security-sensitive decision. Anyone who wants to make sure they are using an official build for a _security_ reason would continue to do that using our signed checksums.

apparentlymart:
I'm now marking this "Ready for review" as part of making a specific proposal for how to proceed with https://github.com/opentofu/opentofu/issues/3912.

lbordowitz (approved):

cam72cam (approved):

PR #4022, opened 2026-04-10
SECURITY.md: Summary of our upstream advisory policy

apparentlymart:
Our policy for handling advisories in upstream dependencies was previously documented only in the RFC where we adopted it, which is hard to find for anyone not already familiar with it.

We'd like for `SECURITY.md` to be the one place someone needs to look when they want to learn about our security policy, so this introduces a short summary of the main consequences of the policy from that RFC and then links to the full RFC in case a reader would like to learn the reasons behind those decisions.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -60,3 +60,29 @@ dealing with a public disclosure all timelines become ASAP (assuming the vulnera
 score >= 4). If the fix relies on another upstream project's disclosure timeline, that
 will adjust the process as well. We will work with the upstream project to fit their timeline and
 best protect our users.
+
+## Security advisories in upstream dependencies
+
+When a security problem is found in an library we depend on, we usually learn about it only once it
+has already been publicly announced.
+
+We use a scheduled job that periodically runs [`govulncheck`](https://go.dev/doc/tutorial/govulncheck)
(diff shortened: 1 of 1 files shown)


cam72cam (approved):

SECURITY.md
```diff
@@ -60,3 +60,29 @@ dealing with a public disclosure all timelines become ASAP (assuming the vulnera
 score >= 4). If the fix relies on another upstream project's disclosure timeline, that
 will adjust the process as well. We will work with the upstream project to fit their timeline and
 best protect our users.
+
+## Security advisories in upstream dependencies
+
+When a security problem is found in an library we depend on, we usually learn about it only once it
```
yottta, line 66:
```suggestion
When a security problem is found in a library we depend on, we usually learn about it only once it
```

apparentlymart: Thanks for catching this! I think this sentence originally had a more complicated noun phrase than "library" here and I simplified it during editing but forgot to change the article. :man_facepalming: 

Opening a new PR just to remove one character seems a little silly but we're unlikely to remember to correct this otherwise, so I'll do that in a moment. :grinning:

apparentlymart: Followup in https://github.com/opentofu/opentofu/pull/4036!

PR #4034, opened 2026-04-13
fix: replace panic with error diagnostics for invalid TF_STATE_PERSIST_INTERVAL

### LINKED ISSUE #4024: TF_STATE_PERSIST_INTERVAL with invalid value causes "OPENTOFU CRASH" instead of user-friendly error
## Community note

> [!TIP]

## OpenTofu Version

```shell
$ tofu version
OpenTofu v1.11.6
on linux_amd64
```

(Reproducible on current `main` as of v1.13 development period)

## OpenTofu Configuration Files

```hcl
# Any minimal configuration is sufficient to reproduce, e.g.:
terraform {
  required_providers {}
}

resource "null_resource" "example" {}
```

## Debug Output

Not required — the crash is immediately visible without `TF_LOG=trace`.

## Expected Behavior

When `TF_STATE_PERSIST_INTERVAL` is set to an invalid value (non-integer string, or an integer below the minimum of 20), `tofu apply` should print a clear, actionable error message and exit with a non-zero status code — the same way other configuration errors are reported.

Example of expected output:
```
╷
│ Error: Invalid value for TF_STATE_PERSIST_INTERVAL
│
│ The environment variable TF_STATE_PERSIST_INTERVAL must be an integer >= 20.
│ Got: "abc"
╵
```

## Actual Behavior

`tofu apply` crashes with an "OPENTOFU CRASH" message and a full goroutine stack trace, which is alarming and unhelpful for a simple user configuration mistake.

**Case 1 — non-integer value:**
```
TF_STATE_PERSIST_INTERVAL=abc tofu apply
```
```
OPENTOFU CRASH
...
panic: Can't parse value 'abc' of environment variable 'TF_STATE_PERSIST_INTERVAL'

goroutine 1 [running]:
runtime/debug.Stack(...)
...
```

**Case 2 — integer below minimum (20):**
```
TF_STATE_PERSIST_INTERVAL=5 tofu apply
```
```
OPENTOFU CRASH
...
panic: Can't use value lower than 20 for env variable TF_STATE_PERSIST_INTERVAL, got 5

goroutine 1 [running]:
...
```

## Steps to Reproduce

1. Create a minimal OpenTofu configuration (any `resource` block works).
2. Run `tofu init`.
3. Run one of the following:
   - `TF_STATE_PERSIST_INTERVAL=abc tofu apply`    — triggers panic on parse failure
   - `TF_STATE_PERSIST_INTERVAL=5 tofu apply`      — triggers panic on below-minimum check
   - `TF_STATE_PERSIST_INTERVAL=abc tofu destroy`  — same panic, `tofu destroy` also uses the same code path (`opApply`)
4. Observe the "OPENTOFU CRASH" output.

Note: `tofu plan` and `tofu refresh` are **not affected** — `getEnvAsInt` is only called in the apply code path.

## Root Cause (for reference)

`internal/backend/local/backend_apply.go`:

```go
// line 37-46
func getEnvAsInt(envName string, defaultValue int) int {
    if val, exists := os.LookupEnv(envName); exists {
        parsedVal, err := strconv.Atoi(val)
        if err == nil {
            return parsedVal
        }
        panic(fmt.Sprintf("Can't parse value '%s' of environment variable '%s'", val, envName))
    }
    return defaultValue
}

// line 115-119
if persistInterval < defaultPersistInterval {
    panic(fmt.Sprintf("Can't use value lower than %d for env variable %s, got %d",
        defaultPersistInterval, persistIntervalEnvironmentVariableName, persistInterval))
}
```

Both `panic()` calls are triggered by ordinary user input errors. Although a `panicHandler` wraps the goroutine and prevents a raw Go crash, it still emits the "OPENTOFU CRASH" message — appropriate for unexpected internal errors, but not for user mistakes.

The fix is to return a `tfdiags.Diagnostics` error from `opApply` instead of calling `panic()`.

## Additional Context

- The `panicHandler` in `internal/logging/panic.go` intentionally catches panics and formats them as "OPENTOFU CRASH" for unexpected bugs. It is not intended to be used as a substitute for proper input validation.
- `TF_STATE_PERSIST_INTERVAL` allows users to tune state persistence frequency. Any user who misconfigures this variable will see a crash-level message, which is poor UX.
- This is a "fast path" candidate: small, self-contained fix with clear expected behavior.
- I'd be happy to submit a PR to fix this if the issue is confirmed.

## References

- `internal/backend/local/backend_apply.go` lines 37–46, 115–119
- `internal/logging/panic.go` (`panicHandler`)

dongjune8931:
Resolves #4024

## Description

When `TF_STATE_PERSIST_INTERVAL` is set to an invalid value (a non-integer string or an integer below the minimum of 20), the previous implementation called `panic()` directly. Although a `panicHandler` wraps the goroutine and prevents a raw crash, it still emits the "OPENTOFU CRASH" message — which is intended for unexpected internal bugs, not user configuration errors.

This PR replaces both `panic()` calls with `tfdiags.Sourceless` error diagnostics, consistent with how other user-facing errors are handled throughout `opApply`.

**Before:**
```
$ TF_STATE_PERSIST_INTERVAL=abc tofu apply

Can't parse value 'abc' of environment variable 'TF_STATE_PERSIST_INTERVAL'
...
```

**After:**
```
$ TF_STATE_PERSIST_INTERVAL=abc tofu apply

╷
│ Error: Invalid environment variable value
│
│ The environment variable TF_STATE_PERSIST_INTERVAL must be a valid integer, got "abc".
╵
```

## Checklist

### Go checklist

### DIFF
--- a/internal/backend/local/backend_apply.go
+++ b/internal/backend/local/backend_apply.go
@@ -34,15 +34,21 @@ const (
 	persistIntervalEnvironmentVariableName = "TF_STATE_PERSIST_INTERVAL"
 )
 
-func getEnvAsInt(envName string, defaultValue int) int {
+func getEnvAsInt(envName string, defaultValue int) (int, tfdiags.Diagnostics) {
+	var diags tfdiags.Diagnostics
 	if val, exists := os.LookupEnv(envName); exists {
 		parsedVal, err := strconv.Atoi(val)
-		if err == nil {
-			return parsedVal
+		if err != nil {
+			diags = diags.Append(tfdiags.Sourceless(
+				tfdiags.Error,
+				"Invalid environment variable value",
+				fmt.Sprintf("The environment variable %s must be a valid integer, got %q.", envName, val),
+			))
+			return 0, diags
 		}
-		panic(fmt.Sprintf("Can't parse value '%s' of environment variable '%s'", val, envName))
+		return parsedVal, diags
 	}
-	return defaultValue
+	return defaultValue, diags
 }
 
 func (b *Local) opApply(
@@ -112,10 +118,20 @@ func (b *Local) opApply(
 	// stateHook uses schemas for when it periodically persists state to the
 	// persistent storage backend.
 	stateHook.Schemas = schemas
-	persistInterval := getEnvAsInt(persistIntervalEnvironmentVariableName, defaultPersistInterval)
+	persistInterval, intervalDiags := getEnvAsInt(persistIntervalEnvironmentVariableName, defaultPersistInterval)
+	diags = diags.Append(intervalDiags)
+	if intervalDiags.HasErrors() {
+		op.ReportResult(runningOp, diags)
+		return
+	}
 	if persistInterval < defaultPersistInterval {
-		panic(fmt.Sprintf("Can't use value lower than %d for env variable %s, got %d",
-			defaultPersistInterval, persistIntervalEnvironmentVariableName, persistInterval))
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Invalid environment variable value",
+			fmt.Sprintf("The environment variable %s must be at least %d, got %d.", persistIntervalEnvironmentVariableName, defaultPersistInterval, persistInterval),
+		))
+		op.ReportResult(runningOp, diags)
+		return
 	}
 	stateHook.PersistInterval = time.Duration(persistInterval) * time.Second
 
--- a/internal/backend/local/backend_apply.go
+++ b/internal/backend/local/backend_apply.go
@@ -42,7 +42,7 @@ func getEnvAsInt(envName string, defaultValue int) (int, tfdiags.Diagnostics) {
 			diags = diags.Append(tfdiags.Sourceless(
 				tfdiags.Error,
 				"Invalid environment variable value",
-				fmt.Sprintf("The environment variable %s must be a valid integer, got %q.", envName, val),
+				fmt.Sprintf("The environment variable %q is expected to be a valid integer but got %q which cannot be converted.", envName, val),
 			))
 			return 0, diags
 		}
@@ -128,7 +128,7 @@ func (b *Local) opApply(
 		diags = diags.Append(tfdiags.Sourceless(
 			tfdiags.Error,
 			"Invalid environment variable value",
-			fmt.Sprintf("The environment variable %s must be at least %d, got %d.", persistIntervalEnvironmentVariableName, defaultPersistInterval, persistInterval),
+			fmt.Sprintf("The value for the environment variable %q cannot be smaller than %d but got %d.", persistIntervalEnvironmentVariableName, defaultPersistInterval, persistInterval),
 		))
 		op.ReportResult(runningOp, diags)
 		return
--- a/internal/backend/local/backend_apply_test.go
+++ b/internal/backend/local/backend_apply_test.go
@@ -478,7 +478,6 @@ func TestLocal_applyInvalidPersistInterval(t *testing.T) {
 		TestLocalProvider(t, b, "test", applyFixtureSchema())
 
 		op, done := testOperationApply(t, "./testdata/apply")
-		defer done(t)
 
 		run, err := b.Operation(context.Background(), op)
 		if err != nil {
@@ -487,6 +486,9 @@ func TestLocal_applyInvalidPersistInterval(t *testing.T) {
 		<-run.Done()
 		if run.Result == backend.OperationSuccess {
 			t.Fatalf("expected operation to fail with invalid %s=abc", persistIntervalEnvironmentVariableName)
+		}
+		if got, want := done(t).Stderr(), "Invalid environment variable value"; !strings.Contains(got, want) {
+			t.Errorf("expected stderr to contain %q, got:\n%s", want, got)
 		}
 	})
 
@@ -497,7 +499,6 @@ func TestLocal_applyInvalidPersistInterval(t *testing.T) {
 		TestLocalProvider(t, b, "test", applyFixtureSchema())
 
 		op, done := testOperationApply(t, "./testdata/apply")
-		defer done(t)
 
 		run, err := b.Operation(context.Background(), op)
 		if err != nil {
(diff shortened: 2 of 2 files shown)


apparentlymart:
Hi @dongjune8931,

Please copy the following snippet from our pull request template into your PR description and make sure you are able to check all of the checkboxes.

```
## Checklist

<!-- Please check of ALL items in this list for all PRs: -->

### Go checklist

<!-- If your PR contains Go code, please make sure you check off all items on this list: --> 

```

Note that our process normally requires that an issue be assigned to a community contributor before they open a PR for it. I'm going to retroactively assign the issue to you now, but if you contribute again in future please wait to be assigned before opening a PR. Thanks!

diofeher:
@dongjune8931 Please have a read here of https://github.com/opentofu/opentofu/blob/main/contributing/DEVELOPING.md too

dongjune8931:
Hi @apparentlymart, I apologize for opening the PR without being assigned to the issue first. I wasn't aware of that process — I'll make sure to wait for assignment before opening a PR in the future.

I've also updated the PR description to follow the pull request template properly. Thanks for pointing that out!

---
Since I found this by analyzing the source code rather than encountering it in practice, should I skip the changelog entry? I recall receiving the same feedback on my previous PR

internal/backend/local/backend_apply_test.go
```diff
@@ -428,3 +428,85 @@ func TestApply_applyCanceledAutoApprove(t *testing.T) {
 	}
 
 }
+
+func TestGetEnvAsInt(t *testing.T) {
+	const testEnv = "TEST_GET_ENV_AS_INT"
+
+	t.Run("env not set returns default", func(t *testing.T) {
+		os.Unsetenv(testEnv)
```
yottta: Why is this `Unsetenv` needed? If the test does not work, then it's a problem somewhere else. Could you please check if there is another test where this env var is set using `os.Setenv` but never cleaned up?

internal/backend/local/backend_apply_test.go
```diff
@@ -428,3 +428,85 @@ func TestApply_applyCanceledAutoApprove(t *testing.T) {
...
+
+	t.Run("non-integer value returns error", func(t *testing.T) {
+		t.Setenv(testEnv, "abc")
+		_, diags := getEnvAsInt(testEnv, 20)
+		if !diags.HasErrors() {
+			t.Error("expected error but got none")
+		}
+	})
+
+	t.Run("float value returns error", func(t *testing.T) {
```
yottta: And if you have time, I would appreciate converting all these test cases to a [table driven testing](https://go.dev/wiki/TableDrivenTests).

internal/backend/local/backend_apply_test.go
```diff
@@ -428,3 +428,85 @@ func TestApply_applyCanceledAutoApprove(t *testing.T) {
...
+
+func TestLocal_applyInvalidPersistInterval(t *testing.T) {
+	t.Run("non-integer value causes error diagnostic", func(t *testing.T) {
+		t.Setenv(persistIntervalEnvironmentVariableName, "abc")
+
+		b := TestLocal(t)
+		TestLocalProvider(t, b, "test", applyFixtureSchema())
+
+		op, done := testOperationApply(t, "./testdata/apply")
+		defer done(t)
```
yottta: Could you use this to check that the expected error is created?

1. remove this line here
2. After `<-run.Done()`, you can add something like this:
   ```go
   output := done(t)
   stderr := output.Stderr()
   // check that stderr contains the expected error message
   ```

internal/backend/local/backend_apply_test.go
```diff
@@ -428,3 +428,85 @@ func TestApply_applyCanceledAutoApprove(t *testing.T) {
...
+
+		run, err := b.Operation(context.Background(), op)
+		if err != nil {
+			t.Fatalf("unexpected error starting operation: %v", err)
+		}
+		<-run.Done()
```
yottta, line 527: Same idea here, with asserting the error message.

yottta (commented):
Thanks a lot for helping with this.
It looks good, just left some comments on things that can be improved.

internal/backend/local/backend_apply_test.go
dongjune8931: Converted TestGetEnvAsInt to table-driven tests!

internal/backend/local/backend_apply_test.go
dongjune8931: Removed it — since t.Setenv automatically restores the original value after each test, it wasn't necessary.

internal/backend/local/backend_apply_test.go
dongjune8931: Added done(t).Stderr() assertion to verify the expected error message appears in the output.

internal/backend/local/backend_apply.go
```diff
@@ -34,15 +34,21 @@ const (
...
+	var diags tfdiags.Diagnostics
 	if val, exists := os.LookupEnv(envName); exists {
 		parsedVal, err := strconv.Atoi(val)
-		if err == nil {
-			return parsedVal
+		if err != nil {
+			diags = diags.Append(tfdiags.Sourceless(
+				tfdiags.Error,
+				"Invalid environment variable value",
+				fmt.Sprintf("The environment variable %s must be a valid integer, got %q.", envName, val),
```
yottta:
```suggestion
				fmt.Sprintf("The environment variable %q is expected to be a valid integer but got %q which cannot be converted.", envName, val),
```

internal/backend/local/backend_apply.go
```diff
@@ -112,10 +118,20 @@ func (b *Local) opApply(
...
+		op.ReportResult(runningOp, diags)
+		return
+	}
 	if persistInterval < defaultPersistInterval {
-		panic(fmt.Sprintf("Can't use value lower than %d for env variable %s, got %d",
-			defaultPersistInterval, persistIntervalEnvironmentVariableName, persistInterval))
+		diags = diags.Append(tfdiags.Sourceless(
+			tfdiags.Error,
+			"Invalid environment variable value",
+			fmt.Sprintf("The environment variable %s must be at least %d, got %d.", persistIntervalEnvironmentVariableName, defaultPersistInterval, persistInterval),
```
yottta:
```suggestion
			fmt.Sprintf("The value for the environment variable %q cannot be smaller than %d but got %d.", persistIntervalEnvironmentVariableName, defaultPersistInterval, persistInterval),
```

yottta (approved):
Left some suggestions since those sound better to me, but it might be more of a personal taste.
We could wait for other reviews before applying the suggestions.

dongjune8931:
Thank you for the review and the suggestions, @yottta! I'll wait for feedback from other maintainers before applying the change

dongjune8931:
> Left some suggestions since those sound better to me, but it might be more of a personal taste. We could wait for other reviews before applying the suggestions.

Updated as suggested — thanks for the feedback! @apparentlymart @yottta

yottta (approved):

PR #4037, opened 2026-04-14
azurerm: respect audience in request URL

lbordowitz:
Resolves #4007

This appears to be in good company, as this issue has been referenced in [other](https://github.com/Azure/login/issues/452) [repositories](https://github.com/hashicorp/terraform/issues/37353) as well as this one. I hope someone is able to help us test this fix. I can check that it works for the public cloud the same way it did before, but that won't do anything for the actual issue. I don't have access to a US Government cloud environment, and I've had to do a lot of guesswork to figure out how to support them appropriately.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/remote-state/azure/auth/oidc_auth.go
+++ b/internal/backend/remote-state/azure/auth/oidc_auth.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/opentofu/opentofu/internal/httpclient"
 	"github.com/opentofu/opentofu/internal/tfdiags"
@@ -53,9 +54,9 @@ func (cred *oidcAuth) Construct(ctx context.Context, config *Config) (azcore.Tok
 		// a fresh OIDC token as needed, enabling proper token refresh behavior.
 		func(ctx context.Context) (string, error) {
 			client := httpclient.New(ctx)
-			
+
 			if config.OIDCToken == "" && config.OIDCTokenFilePath == "" {
-				return getTokenFromRemote(client, config.OIDCAuthConfig)
+				return getTokenFromRemote(client, config.OIDCAuthConfig, config.CloudConfig)
 			}
 			return consolidateToken(config)
 		},
@@ -69,7 +70,7 @@ type TokenResponse struct {
 	Value string `json:"value"`
 }
 
-func getTokenFromRemote(client *http.Client, config OIDCAuthConfig) (string, error) {
+func getTokenFromRemote(client *http.Client, config OIDCAuthConfig, env cloud.Configuration) (string, error) {
 	// GET from the request URL, using the bearer token
 	req, err := http.NewRequest(http.MethodGet, config.OIDCRequestURL, nil)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/backend/remote-state/azure/auth/oidc_auth.go
```diff
@@ -105,6 +108,22 @@ func getTokenFromRemote(client *http.Client, config OIDCAuthConfig) (string, err
...
+func requestURLAudience(env cloud.Configuration) string {
+	switch env.ActiveDirectoryAuthorityHost {
+	case "https://login.chinacloudapi.cn/":
+		return "api://AzureADTokenExchangeChina"
+	case "https://login.microsoftonline.us/":
+		return "api://AzureADTokenExchangeUSGov"
+	default:
+		return "api://AzureADTokenExchange"
+	}
+}
```
apparentlymart, line 127: I was curious about where these `env.ActiveDirectoryAuthorityHost` values come from and whether it'd be possible for them to appear in other non-normalized forms, but from some digging I think I've found that there's a fixed set of possible values decided here:

https://github.com/opentofu/opentofu/blob/192b4c9d608f1bdf38076fc72d39a5e1f8496c97/internal/backend/remote-state/azure/auth/clients.go#L129-L146

Have I understood that correctly?

If so, perhaps it would be clearer to make these `case` statements refer to [the same strings from `github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud#pkg-variables) so that the relationship between these two is a little more obvious.

For example, instead of writing `"https://login.chinacloudapi.cn/"` literally here, we could write `cloud.AzureChina.ActiveDirectoryAuthorityHost` so that it refers to the same symbol we'd used to populate this value earlier.

I also worry that a future maintainer might not realize that these two codepaths both need to agree about which `ActiveDirectoryAuthorityHost` values are possible. Perhaps we should at least leave a comment in both this function and over in `CloudConfigForAddresses` that mentions that they both need to be updated together if the set of possible "clouds" changes in future.

(It's annoying that this upstream `cloud` package doesn't _also_ have the audience for each "cloud" so that we could centralize this all in one place, given that they already spelled out a one-to-one relationship between these in the documentation you linked, but of course that's outside of our control so 🤷‍♂️ .)

apparentlymart (approved):
Automatically selecting a suitable audience string instead of just letting the user reconfigure it is a nice addition!

I left a comment about it but I understand our implementation approach here is quite constrained by how the upstream SDK is designed.

**Edit:** I noticed this was a draft only after I already approved it :man_facepalming:, but FWIW what's here at the time I'm writing this makes sense to me.

internal/backend/remote-state/azure/auth/oidc_auth.go
lbordowitz: Interestingly enough, these URLs *did* change! (Kind of, a `/` character was added at the end in the most recent upgrade). However, they're never persisted, so we only need to be careful of that tricky codepath you mentioned. I'll add a developer sanity comment.

lbordowitz:
All good, Martin. It's set to draft pending completion of end-to-end tests. Since we're in beta, I figured I'd do the full battery of Azure tests to double-check the dependency upgrades as well.

diofeher (approved):
Looks good to me, but I think a CHANGELOG entry should be added.

lbordowitz:
I've tested OIDC, rebased, and added a changelog entry. You can give it another once-over if you'd like; I plan to merge this come Monday.

yottta (approved):

PR #4042, opened 2026-04-16
lang/funcs: cidrsubnet now allows newbits up to 128 for IPv6

### LINKED ISSUE #4040: The cidrsubnets function limits newbits to 32 even when using IPv6
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.6
on linux_amd64
```

There are no relevant configuration in this case.

### Debug Output

It is not relevant here.

### Expected Behavior

The `cidrsubnets` function should be able to handle bigger `newbits` values than 32 for IPv6 addresses.

This should work and give back `["2001:db8:102:304::/120"]`.
```
~$ tofu console 
> cidrsubnets("2001:db8:0102:0304::/64", 56)
╷
│ Error: Invalid function argument
│ 
│   on <console-input> line 1:
│   (source code not available)
│ 
│ Invalid value for "newbits" parameter: may not extend prefix by more than 32 bits.
╵
```

### Actual Behavior

You can't use the `cidrsubnets` function to create IPv6 ranges which are "more than 32 bits smaller" than the `prefix`.

My use case is that I want to create the first /120 of a /64 prefix. Currently this is the workaround I had to apply.
```
~$ tofu console 
> cidrsubnets(cidrsubnets("2001:db8:0102:0304::/64", 32)[0], 24)[0]
"2001:db8:102:304::/120
```

### Steps to Reproduce

1. tofu console <<< 'cidrsubnets("2001:db8:0102:0304::/64", 56)'

### Additional Context

I am not 100% sure that this is strictly a bug as this behavior is documented but the behavior itself looks buggy. Maybe this 32 limitation comes from an IPv4-only era where it made sense but is not intuitive when you work with IPv6 addresses.

If you think this is not a bug but a feature request please close this one and I will open a feature request ticket instead.

### References

_No response_

apparentlymart:
The original design for this function seems to have mistakenly thought that the upstream library used to implement the calculations could not accept newbits>32, perhaps due to that having been true for some of its other functions in an earlier release.

However, this restriction is not needed in the current version of the upstream library: it performs its calculations on a byte-by-byte basis treating the IP address as a byte-slice, rather than using the Go "int" type.

Closes https://github.com/opentofu/opentofu/issues/4040.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ UPGRADE NOTES:
 
 ENHANCEMENTS:
 
+- The `cidrsubnets` function now supports prefix extensions greater than 32 bits when the base CIDR block uses an IPv6 address. ([#4042](https://github.com/opentofu/opentofu/pull/4042))
 - The `local-exec` provisioner now automatically sets the `TRACEPARENT` environment variable in child processes when OpenTelemetry tracing is active, following the W3C Trace Context specification. ([#4014](https://github.com/opentofu/opentofu/issues/4014))
 
 BUG FIXES:
--- a/internal/lang/funcs/cidr.go
+++ b/internal/lang/funcs/cidr.go
@@ -163,13 +163,6 @@ var CidrSubnetsFunc = function.New(&function.Spec{
 			if length < 1 {
 				return cty.UnknownVal(cty.String), function.NewArgErrorf(i+1, "must extend prefix by at least one bit")
 			}
-			// For portability with 32-bit systems where the subnet number
-			// will be a 32-bit int, we only allow extension of 32 bits in
-			// one call even if we're running on a 64-bit machine.
-			// (Of course, this is significant only for IPv6.)
-			if length > 32 {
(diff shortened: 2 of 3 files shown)


diofeher (approved):

lbordowitz (approved):

yottta (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ UPGRADE NOTES:
 
 ENHANCEMENTS:
 
+- The `cidrsubnets` function now supports prefix extensions greater than 32 bits when the base CIDR block uses an IPv6 address. ([#4042](https://github.com/opentofu/opentofu/issues/4042))
```
```suggestion
- The `cidrsubnets` function now supports prefix extensions greater than 32 bits when the base CIDR block uses an IPv6 address. ([#4042](https://github.com/opentofu/opentofu/pull/4042))
```

Not truly needed since GH redirects to the right content type either way.

apparentlymart:
We did the process a little out of order here, but since the associated issue is now accepted and this hasn't changed significantly since folks already reviewed it, I'm going to merge this now.

PR #4048, opened 2026-04-22
rfc: Behavior Differences in the New OpenTofu Runtime

apparentlymart:
This is currently a very early draft of a document that I hope will eventually become an overview of the behavior differences we're intentionally accepting for the new runtime because we believe the new behavior more accurately models author intent and/or solves some challenges with the current implementation that have previously been discussed elsewhere.

However, I'm intending for now that we treat this as a "living document": most of what I've described here is not actually true yet because it's discussing features we've not implemented yet. I've written it in present tense only because that's the form we'd want this to be in once finalized and accepted.

Our main task here for now is to decide whether these behaviors are actually what we want to implement, and we might want to do some prototyping in order to answer that for some of these. For anything where we answer "yes", later drafts of this document should include specific details about how exactly someone migrating from old runtime to new should respond in each of these situations. If we cannot find a reasonable way for someone to respond when migrating then we should change the design to something less disruptive.

[Rendered version](https://github.com/opentofu/opentofu/blob/rfc-new-runtime-behavior-changes/rfc/20260421-new-runtime-differences.md)

### DIFF
--- a/rfc/20260421-new-runtime-differences.md
+++ b/rfc/20260421-new-runtime-differences.md
@@ -1,0 +1,324 @@
+# Behavior Differences in the New OpenTofu Runtime
+
+In [A new approach to configuration evaluation, planning, and applying](./20251001-eval-plan-apply-architecture.md)
+we agreed to try for a new way to implement OpenTofu's core functionality,
+learning from our experiences maintaining the current language runtime we
+inherited from our predecessor.
+
+During the OpenTofu v1.12 development period we implemented a "walking skeleton"
+experimental version of that design to learn whether it would actually work
+in practice. We successfully reimplemented the primary functionality of planning
+and applying changes to managed resource instances in that new runtime style,
+and although we didn't fully implement any other behaviors yet we have explored
+them enough to feel confident that they are achievable.
+
+However, if the new approach only led to us reimplementing the same behavior in
+a different way it would be hard to justify spending so much time and effort on
+it. Although we could perhaps justify it as making it easier to maintain and
+implement new functionality in future, we can't be sure that is true until we're
+actually doing ongoing maintenence and development on it.
+
+Aside from internal design improvements, the new design is intentionally
+arranged to allow us to perform more precise analysis of the configuration so
+that our interpretation and execution of it can more closely match the likely
+author intent. For example, it finds dependencies between resources using
+dynamic analysis rather than static analysis, so given an expression like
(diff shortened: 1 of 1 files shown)


apparentlymart:
https://github.com/hashicorp/terraform/issues/38724 shows a real-world example of the problem I described in [Data Resource Instance Behavior Changes](https://github.com/opentofu/opentofu/blob/d356958bb2b9491f177ba6ea662158bff535e1c4/rfc/20260421-new-runtime-differences.md#data-resource-instance-behavior-changes)[^1], where OpenTofu (and its predecessor) currently sometimes fail to detect when a data resource instance should not be read until the apply phase.

The problem in this case is that today's rule for whether to delay reading a data resource instance is based on whether the configuration is _known_, but that produces the wrong result here because we already know the final value of `azurerm_resource_group.main` will be `"rg-test"`. What we're missing is that testing whether the final value is known is the wrong rule: we should instead be representing whether that known value is already set in the remote system or if that value is the outcome of a change that hasn't been applied yet.

In the new behavior I've previously prototyped, the `name` attribute of `azurerm_resource_group.main` would have a special new cty mark representing that its value is expected to change during the apply phase, so that we can know there's a change pending even though we can already predict that the final value will be `"rg-test".

We could then adjust the rules for planning data resource instances so that their reads get delayed to the apply phase if any part of their configuration is marked as "expected to change during apply", regardless of whether the marked value is known or not.

Note though that this still has the potential gap I mentioned in the linked RFC text: if the author subsequently changes some argument of `azurerm_resource_group.main` without changing the `name` argument, and if the provider is able to apply that change without replacing the entire resource group object, then the same problem would occur: the data resource instance would appear to be readable during the planning phase because its _name_ is already known and is not expected to change, but yet the other attributes of that object that the data resource instance would be reading have not been changed yet and so it would read stale values. If we wanted to make _that_ work then we'd need to mark an entire resource instance as "expected to change during apply" if any part of it is changing, which would have more false positives but perhaps that's better than a false negative because at least it could give better feedback about what's going on.

(In my prototype the "expected to change during apply" mark also carried the identity of the resource instance the changing value belonged to, so that the plan UI could then explain exactly why a data resource instance read is being delayed, saying something like `(because azurerm_resource_group.main has pending changes)` just after the heading that says it must be read during the apply phase.)

[^1]: Note that this is a link to the current version of the RFC at the time I'm writing this comment, so the content at that URL won't update for subsequent revisions of the document. Use the link in the opening comment above for the latest version from the PR branch.

PR #4049, opened 2026-04-22
Warn when relying on non-default GODEBUG values

apparentlymart:
This change is primarily motivated by https://github.com/opentofu/opentofu/issues/3415, implementing something like what I proposed in https://github.com/opentofu/opentofu/issues/3415#issuecomment-3999975239, but I took a broader approach there that handles _all_ `GODEBUG` settings that might be used currently or in the future, rather than just the `winsymlink` setting.

---

"GODEBUG" is a mechanism that the Go runtime forces on OpenTofu that allows dynamic changes to various nuances of Go standard library behavior.

We cannot control when GODEBUG workarounds are added and removed from Go, so if someone chooses to rely on them then they are likely to get broken by a later release of OpenTofu that uses a newer version of Go. Therefore we'd like anyone relying on these to tell us so we can try to find a more sustainable solution to their problem early on while the workaround is still available to them. This is particularly important in light of the newer policy in https://github.com/golang/go/issues/76163 which calls for the Go runtime to panic immediately on startup (before any of OpenTofu's own code runs) once a previously-supported GODEBUG key has been removed, since by that point it would be too late for us to produce a user-friendly diagnostic.

We also occasionally force certain GODEBUG workarounds on by default in our official builds in anticipation of breakage caused by upstream Go changes. In that case we'd still like to know if anyone is relying on them so that we can remove those workarounds in a future release, but we'd prefer to get the reports about it in an issue or discussion we'd already created for that purpose.

This therefore introduces warning diagnostics any time something in the "init", "plan", or "apply" commands relies on non-default GODEBUG settings. For ones we enabled intentionally we provide a specific URL to report usage as part of the diagnostic message, but we also have a generic message for any GODEBUG settings the user chose to enable for themselves for reasons we probably don't know about yet but would like to learn more about.

This is included in only that subset of commands as a compromise because those primary workflow commands are the most likely to interact with Go library features that are affected by these workarounds -- they most commonly affect functionality related to networking, TLS, etc -- and additional diagnostics in these locations should not to cause breakage for anyone consuming the machine-readable output, but there are other commands where additional diagnostics are more likely to cause problems.

Unfortunately it isn't really possible to unit-test this because it relies directly on information reported by the Go metrics system and there's no facility to produce fake metrics for testing. However, Go 1.26 (which we're currently using) there is a GODEBUG setting `netedns0=0` which causes a warning any time OpenTofu makes a DNS request, so I've tested this manually by running `tofu init` with that workaround enabled in a configuration that depends on at least one provider (causing DNS lookups for the provider registry) and seen it generate the expected warning message referring to "netedns0". I also temporarily modified the code to add a report URL for that key to make sure the alternate form of the diagnostic message is working as expected, and temporarily forced some DNS lookups into plan and apply to verify that the warnings are also produced by those commands when relevant.

These new warnings replace our previous quieter messages about this in the TF_LOG=warn output, because those were only useful in situations where GODEBUG was causing a problem that led to someone opening a bug report, but we actually want to know of situations where GODEBUG was needed in order for OpenTofu to _succeed_ and so for that we need more prominent messages to let the operator know there's something they ought to report to us even though there wasn't an error.

---

Although this is for https://github.com/opentofu/opentofu/issues/3415, this will not close that issue. Instead, it's intended to give us more confidence to take action on that issue during the v1.14 development period.

- [ ] ~I have added tests for all relevant use cases of my code, and those tests are passing.~ (see above)

### DIFF
--- a/internal/command/apply.go
+++ b/internal/command/apply.go
@@ -122,10 +122,20 @@ func (c *ApplyCommand) Run(rawArgs []string) int {
 
 	// Run the operation
 	op, diags := c.RunOperation(ctx, be, opReq)
+
+	// Applying is one of the operations most likely to activate Go runtime
+	// behavior that's affected by GODEBUG, so we'll take this opportunity
+	// to warn about any reliance on non-default runtime behavior so that
+	// hopefully folks will report problems to us while their workaround is
+	// still available, instead of waiting until the workaround gets removed
+	// in a later Go releases.
+	diags = diags.Append(c.godebugUsageWarnings())
+
 	view.Diagnostics(diags)
 	if diags.HasErrors() {
 		return 1
 	}
+	diags = nil // avoid re-reporting any warnings below
 
 	if op.Result != backend.OperationSuccess {
 		return op.Result.ExitStatus()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ ENHANCEMENTS:
 
 - The `cidrsubnets` function now supports prefix extensions greater than 32 bits when the base CIDR block uses an IPv6 address. ([#4042](https://github.com/opentofu/opentofu/pull/4042))
 - The `local-exec` provisioner now automatically sets the `TRACEPARENT` environment variable in child processes when OpenTelemetry tracing is active, following the W3C Trace Context specification. ([#4014](https://github.com/opentofu/opentofu/issues/4014))
+- Some commands now produce warning diagnostics when execution relied on various workarounds in the Go runtime library ("GODEBUG" settings), encouraging the reader to report any situations where those workarounds are necessary so that we can identify a more sustainable solution, because these workarounds can be removed or change behavior in future versions of Go outside of our control and are not covered by OpenTofu's compatibility promises. ([#4049](https://github.com/opentofu/opentofu/pull/4049))
 
 BUG FIXES:
 
--- a/cmd/tofu/main.go
+++ b/cmd/tofu/main.go
@@ -339,12 +339,6 @@ func realMain() int {
 		return 1
 	}
 
-	// We might generate some additional log lines if OpenTofu relied on any
-	// non-default Go runtime behaviors enabled by GODEBUG settings, because
-	// they might be relevant when trying to reproduce certain problems for
-	// debugging or bug reporting purposes.
-	logGodebugUsage()
-
 	// if we are exiting with a non-zero code, check if it was caused by any
 	// plugins crashing
 	if exitCode != 0 {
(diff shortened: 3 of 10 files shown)


internal/command/apply.go
```diff
@@ -122,10 +122,20 @@ func (c *ApplyCommand) Run(rawArgs []string) int {
...
+	// hopefully folks will report problems to us while their workaround is
+	// still available, instead of waiting until the workaround gets removed
+	// in a later Go releases.
+	diags = diags.Append(c.godebugUsageWarnings())
+
 	view.Diagnostics(diags)
 	if diags.HasErrors() {
 		return 1
 	}
+	diags = nil // avoid re-reporting any warnings below
```
apparentlymart, line 138: This is fixing a bug that's independent of the main point of this PR: any warnings generated by running the operation would appear twice unless there was also an error, because there's another call to `view.Diagnostics(diags)` below. That situation was also relevant for any warnings added by `godebugUsageWarnings`.

apparentlymart:
Interestingly it seems like part of our test suite is relying on `winsymlink=0` when running on Windows. :grinning: 

I've run out of time for today but I'll investigate that more tomorrow.

apparentlymart:
Interestingly, the test that triggers the warning about `winsymlink` being non-default is able to pass without that workaround enabled, but some different tests begin failing:

```
--- FAIL: TestInit_fromModule_cwdDest (0.03s)
    init_test.go:120: err: GetFileAttributesEx C:\Users\RUNNER~1\AppData\Local\Temp\TestInit_fromModule_cwdDest945771748\001\hello.tf: The system cannot find the file specified.
--- FAIL: TestInit_fromModule_dstInSrc (0.04s)
    init_test.go:169: err: GetFileAttributesEx C:\Users\RUNNER~1\AppData\Local\Temp\TestInit_fromModule_dstInSrc4012406067\001\foo\issue518.tf: The system cannot find the file specified.
FAIL
FAIL	github.com/opentofu/opentofu/internal/command	128.880s
```

That's consistent with the reason why we ended up enabling `winsymlink=0` in the first place (discussed in https://github.com/opentofu/opentofu/pull/3289), but it doesn't help explain why `TestPlan_jsonGoldenReference` triggers the warning about `winsymlink` reliance but yet seems to pass just fine when that workaround isn't enabled. :thinking:

apparentlymart:
This is stalled because of the problem described in https://github.com/opentofu/opentofu/issues/3415#issuecomment-4314309507. If we were to add this PR now then anyone using absolute filesystem paths in their module source addresses would see the `winsymlink=0` warning whenever they run `tofu init` on a Windows system, and there's no point is us shipping a request to tell us about a problem we already know about.

PR #4051, opened 2026-04-23
`UIInput` must return when the first `SIGINT`/`SIGTERM` signal is received

### LINKED ISSUE #4017: Graceful failure from Control-C does not clean up S3 lockfile
### Community note

> [!TIP]

### OpenTofu Version

```shell
$ tofu version
OpenTofu v1.11.4
on darwin_arm64
+ provider registry.opentofu.org/tailscale/tailscale v0.28.0
```

```hcl
terraform {
  encryption {
    key_provider "aws_kms" "state_key" {
      kms_key_id = "arn:aws:kms:us-east-1:000:alias/opentofu-state-encryption"
      key_spec   = "AES_256"
      region     = "us-east-1"
      profile    = "core"

      assume_role = {
        role_arn     = "arn:aws:iam::000:role/OrganizationAccountAccessRole"
        session_name = "opentofu-state-encryption"
      }

      encrypted_metadata_alias = "state_key"
    }

    method "aes_gcm" "state" {
      keys = key_provider.aws_kms.state_key
    }

    state {
      enforced = true
      method   = method.aes_gcm.state
    }

    plan {
      enforced = true
      method   = method.aes_gcm.state
    }
  }

  backend "s3" {
    bucket       = "tofu-state-000"
    key          = "opentofu-core/tooling/tailscale/terraform.tfstate"
    region       = "us-east-1"
    profile      = "core"
    encrypt      = true
    kms_key_id   = "arn:aws:kms:us-east-1:000:alias/opentofu-state-encryption"
    use_lockfile = true

    assume_role = {
      role_arn     = "arn:aws:iam::000:role/OrganizationAccountAccessRole"
      session_name = "opentofu-state-backend"
    }
  }
}
```

### Debug Output

https://gist.github.com/spkane/a1b0f49cacc133bf1bd29c610046f611

### Expected Behavior

```console
❯ tofu apply
var.tailscale_oauth_client_id
  Tailscale OAuth client ID (create in admin console under Settings > OAuth clients)

  Enter a value:                  # HIT CONTROL_C AND THEN WAIT}

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...

❯ tofu apply
var.tailscale_oauth_client_id
  Tailscale OAuth client ID (create in admin console under Settings > OAuth clients)

  Enter a value: 
```

### Actual Behavior

```console
❯ tofu apply
var.tailscale_oauth_client_id
  Tailscale OAuth client ID (create in admin console under Settings > OAuth clients)

  Enter a value:

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...

❯ tofu apply
╷
│ Error: Error acquiring the state lock
│
│ Error message: operation error S3: PutObject, https response error StatusCode: 412,
│ RequestID: XA19RPMDM05BZZVP, HostID:
│ Ltw96ETc/l+smH=, api error
│ PreconditionFailed: At least one of the pre-conditions you specified did not hold
│ Lock Info:
│   ID:        91e631e1-6f92-...
│   Path:      tofu-state-000/opentofu-core/tooling/tailscale/terraform.tfstate
│   Operation: OperationTypeApply
│   Who:       me@MacBook-Pro.local
│   Version:   1.11.4
│   Created:   2026-04-08 21:15:23.923896 +0000 UTC
│   Info:
│
│
│ OpenTofu acquires a state lock to protect the state from being written
│ by multiple users at the same time. Please resolve the issue above and try
│ again. For most commands, you can disable locking with the "-lock=false"
│ flag, but this is not recommended.
```

### Steps to Reproduce

1. `tofu init`
2.`tofu apply` - Control-C while being prompted for a variable.
2.`tofu apply`

### Additional Context

_No response_

### References

_No response_

yottta:
When OpenTofu needs to prompt the user for variable with no provided values, it goes through the `UIInput`, which before this change, did not handle `SIGTERM`, but only `SIGINT`. 
This behavior was detached from the general way non-windows graceful signals are handled by OpenTofu.

This PR removes the specific signals handling from the `UIInput` and instead plugs the context meant for graceful shutdown into it.
Before these changes, the `UIInput` would have become unusable once `SIGINT` was returned but now it relies on the received `context.Context.Done()` channel to know if it can ask for values or not.
 
Using the procedure explained in #4017, I can confirm that both signals (`SIGTERM` and `SIGINT`) work as intended now.

### Sending `SIGINT` (`Ctrl+C`)
```shell
#> tofu apply

var.tailscale_oauth_client_id
  Enter a value: ^C

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...

╷
│ Error: No value for required variable
│
│   on main.tf line 10:
│   10: variable "tailscale_oauth_client_id" {
│
│ The root module input variable "tailscale_oauth_client_id" is not set, and has no default value. Use a -var or -var-file command line argument to provide a value for this variable.
╵
```

### Sending `SIGTERM` (`kill <pid>`)
```shell
#> tofu apply

var.tailscale_oauth_client_id
  Enter a value:

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...

╷
│ Error: No value for required variable
│
│   on main.tf line 10:
│   10: variable "tailscale_oauth_client_id" {
│
│ The root module input variable "tailscale_oauth_client_id" is not set, and has no default value. Use a -var or -var-file command line argument to provide a value for this variable.
╵
```

Resolves #4017
## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/backend.go
+++ b/internal/backend/backend.go
@@ -331,6 +331,10 @@ type Operation struct {
 	// for unmatched import targets and where any generated config should be
 	// written to.
 	GenerateConfigOut string
+
+	// StopCtx is used by some of the backends to be able to shut down gracefully when the caller
+	// asks for it.
+	StopCtx context.Context
 }
 
 // HasConfig returns true if and only if the operation has a ConfigDir value
--- a/internal/backend/backend.go
+++ b/internal/backend/backend.go
@@ -170,7 +170,13 @@ type Local interface {
 	// backend's implementations of this to understand what this actually
 	// does, because this operation has no well-defined contract aside from
 	// "whatever it already does".
-	LocalRun(context.Context, *Operation) (*LocalRun, statemgr.Full, tfdiags.Diagnostics)
+	//
+	// Even though both contexts contain the tracing information, there is a crucial difference between the two:
+	//  - The first one is non-cancellable meaning that it's safe to be used for thing like state unlocking and other
+	//    critical operations that need to run even when the process is asked to end gracefully.
+	//  - The second one is a cancellable context. This is the context that should be used to cancel operations
+	//    for a graceful shutdown.
+	LocalRun(context.Context, context.Context, *Operation) (*LocalRun, statemgr.Full, tfdiags.Diagnostics)
 }
 
 // LocalRun represents the assortment of objects that we can collect or
@@ -331,10 +337,6 @@ type Operation struct {
 	// for unmatched import targets and where any generated config should be
 	// written to.
 	GenerateConfigOut string
-
-	// StopCtx is used by some of the backends to be able to shut down gracefully when the caller
-	// asks for it.
-	StopCtx context.Context
 }
 
 // HasConfig returns true if and only if the operation has a ConfigDir value
--- a/internal/backend/remote/backend_context.go
+++ b/internal/backend/remote/backend_context.go
@@ -27,7 +27,8 @@ import (
 var _ backend.Local = (*Remote)(nil)
 
 // LocalRun implements backend.Local.
-func (b *Remote) LocalRun(ctx context.Context, _ context.Context, op *backend.Operation) (*backend.LocalRun, statemgr.Full, tfdiags.Diagnostics) {
+// Refer to the comments of backend.Local for more details about ctx vs stopCtx.
+func (b *Remote) LocalRun(ctx context.Context, stopCtx context.Context, op *backend.Operation) (*backend.LocalRun, statemgr.Full, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 	ret := &backend.LocalRun{
 		PlanOpts: &tofu.PlanOpts{
--- a/internal/backend/remote/backend_context.go
+++ b/internal/backend/remote/backend_context.go
@@ -27,7 +27,7 @@ import (
 var _ backend.Local = (*Remote)(nil)
 
 // LocalRun implements backend.Local.
-func (b *Remote) LocalRun(ctx context.Context, op *backend.Operation) (*backend.LocalRun, statemgr.Full, tfdiags.Diagnostics) {
+func (b *Remote) LocalRun(ctx context.Context, _ context.Context, op *backend.Operation) (*backend.LocalRun, statemgr.Full, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 	ret := &backend.LocalRun{
 		PlanOpts: &tofu.PlanOpts{
--- a/internal/backend/local/backend.go
+++ b/internal/backend/local/backend.go
@@ -330,9 +330,6 @@ func (b *Local) Operation(ctx context.Context, op *backend.Operation) (*backend.
 	runningOp.Cancel = cancel
 
 	op.StateLocker = op.StateLocker.WithContext(stopCtx)
-	if op.StopCtx == nil {
-		op.StopCtx = stopCtx
-	}
 
 	panicHandler := logging.PanicHandlerWithTraceFn()
 
--- a/internal/backend/local/backend.go
+++ b/internal/backend/local/backend.go
@@ -330,6 +330,9 @@ func (b *Local) Operation(ctx context.Context, op *backend.Operation) (*backend.
 	runningOp.Cancel = cancel
 
 	op.StateLocker = op.StateLocker.WithContext(stopCtx)
+	if op.StopCtx == nil {
+		op.StopCtx = stopCtx
+	}
 
 	panicHandler := logging.PanicHandlerWithTraceFn()
 
--- a/internal/backend/local/backend_apply.go
+++ b/internal/backend/local/backend_apply.go
@@ -88,7 +88,7 @@ func (b *Local) opApply(
 	op.Hooks = append(op.Hooks, stateHook)
 
 	// Get our context
-	lr, _, opState, contextDiags := b.localRun(ctx, op)
+	lr, _, opState, contextDiags := b.localRun(ctx, stopCtx, op)
 	diags = diags.Append(contextDiags)
 	if contextDiags.HasErrors() {
 		op.ReportResult(runningOp, diags)
--- a/internal/backend/local/backend_local.go
+++ b/internal/backend/local/backend_local.go
@@ -28,7 +28,7 @@ import (
 var _ backend.Local = (*Local)(nil)
 
 // backend.Local implementation.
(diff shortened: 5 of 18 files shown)


internal/backend/backend.go
```diff
@@ -331,6 +331,10 @@ type Operation struct {
 	// for unmatched import targets and where any generated config should be
 	// written to.
 	GenerateConfigOut string
+
+	// StopCtx is used by some of the backends to be able to shut down gracefully when the caller
+	// asks for it.
+	StopCtx context.Context
```
apparentlymart: I don't really like the idea of adding this here because the situation with contexts in the "backend" part of the system is _already_ very complicated and confusing, with at least three slighly-different contexts involved in every operation:

https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/backend/local/backend.go#L311-L330

I understand that this field is just another reference to the same `stopCtx` from the snippet above, but it's still more mental load of remembering how this context is different from the other two and therefore when it's appropriate to use it.

If adding this is the only reasonable way to fix this bug without making big risky changes then I suppose I could accept this additional complexity temporarily with the assumption that we'd follow up relatively soon with some refactoring to make these codepaths use contexts in a more normal way, but first I'd prefer it if we could find some way to avoid spreading this existing complexity into even more parts of the local backend. 😖

yottta: Maybe I should have this stated in the description, but let's do it here.
I don't like it either. Initially I propagated it through a new argument of the [`Local.LocalRun`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/backend/backend.go#L173), but having 2 contexts in the same method looks even weirder. Thinking about the maintenance cost, by having 2 contexts in a method signature might be "better" in terms of forcing the contributor to pay additional attention on "why there are 2 contexts and which one should I use?" instead of "hiding" that into the `Operation` type.

Another thing that I thought on doing is to pass into the [`Local.localRun`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/backend/local/backend_apply.go#L91) the `stopCtx`, but after a quick check, that would replace the currently passed (non cancellable) context with one that could disrupt the logic downstream. So that seems way to risky to do in the current setup.

I will dig more to find a better way, but feel free to let me know of the way(s) you see this 🤔

yottta: In 408964f78578148baeaa5d33816d977da545e169 I went back to the approach of changing the [`Local.LocalRun`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/backend/backend.go#L173) definition to accept a second `context.Context`. It's not clean, but it's not hidden either.

For the commands on which I created the `stopCtx` by using [`Meta.InterruptibleContext()`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/command/meta.go#L354), I can confirm that the usage of the `m.ShutdownCh` does not interfere with other flows down the line.
Saying this, because if `m.ShutdownCh` is used in the same flow more than once, it can lead to only one of the places to be cancelled while the other places, using the same channel, will receive no event.
The `m.ShutdownCh` is used only in:
* `Meta.InterruptibleContext` which is used only in some of the commands' `Run` method.
* `Meta.RunOperation` which is used only by the [`tofu apply`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/command/apply.go#L124), [`tofu plan`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/command/plan.go#L106) and [`tofu refresh`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/command/refresh.go#L103) commands.
* In 2 other commands: [`tofu test`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/command/test.go#L307) and [`tofu login`](https://github.com/opentofu/opentofu/blob/0b5321e870a58c4ac694bbbc0e455d02862b0638/internal/command/login.go#L514)

Between any of these there is no overlap so there is no issue in using the `Meta.InterruptibleContext()` in the commands that call directly `Local.LocalRun`.

WDYT?

apparentlymart: Passing two separate contexts _is_ weird, but it's also how other parts of `backend` already deal with the same messy situation so at least it's consistently weird! :grinning: 

In the past I've considered using one `context.Context` argument only for the "main" context that has all of the values (like tracing span) attached to it but change `stopCtx` arguments to just directly be the `<-chan struct{}` value returned by `stopCtx.Done`, so that it's clearer that we're using that extra thing _only_ for cancellation and not for anything else. But I always talk myself out of making that change because it seems like just a bunch of churn that doesn't improve the situation enough to be worth it.

In the long run I'd like to find a way to rework this so that we can pack all of the information we need into either the main `context.Context` argument _or_ other arguments that are of different types (not `context.Context`), but we've already got a lot of refactoring going on so I don't think it's a good time to make such a risky change to the backend API right now.

Overall I think what you've done in the latest updates seems like a good compromise for now, allowing us to fix the reported bug in a way that's consistent with our existing technical debt. :+1:

apparentlymart (approved):
Honestly it's always hard to follow through all of the different contexts in the backend code so I did my best to review this but I want to be explicit that I'm not 100% confident and would suggest someone else take a second look too.

Although this was to fix a bug, I'd recommend that we skip backporting this one and just save it for the v1.13 release because this change touches a lot of different parts of the code and the bug it's fixing does not seem to be a very common problem in practice given that this problem was here for many years but only recently did someone notice it.

internal/backend/remote/backend_context.go
```diff
@@ -27,7 +27,7 @@ import (
 var _ backend.Local = (*Remote)(nil)
 
 // LocalRun implements backend.Local.
-func (b *Remote) LocalRun(ctx context.Context, op *backend.Operation) (*backend.LocalRun, statemgr.Full, tfdiags.Diagnostics) {
+func (b *Remote) LocalRun(ctx context.Context, _ context.Context, op *backend.Operation) (*backend.LocalRun, statemgr.Full, tfdiags.Diagnostics) {
```
cam72cam: Even though it's unused, let's keep stopCtx here instead of `_` for documentation.

cam72cam (approved):
Given the complexity of ctx handling within OpenTofu, I think this is the cleanest surgical change that we could do.  It has a low likelihood of breaking anything else critical within the application and is clear to follow.

PR #4052, opened 2026-04-24
[RFC] Symbol Libraries

cam72cam:
[Rendered Markdown](https://github.com/opentofu/opentofu/blob/rfc_symbol_libraries/rfc/20260424-symbol-libraries.md)

## Checklist

### DIFF
--- a/rfc/20260424-symbol-libraries.md
+++ b/rfc/20260424-symbol-libraries.md
@@ -184,6 +184,36 @@ values {
 
 How do I have "unexported" symbols?
 
+The easiest way to do this without introduing any new syntax is to take an approach similar to go's `internal` pattern.
+
+```hcl
+# ./lib/contents.sym.hcl
+
+# Import the internal library
+symbols "internal" {
+  source = "."
+  file  = "./internal/types.sym.hcl"
+}
+
+
+# Re-export the custom type defined in the internal library
+typedef "custom" {
+  type = symbols::internal::types(custom)
+}
+
+```
+
+```hcl
+# ./lib/internal/types.sym.hcl
+
+typedef "custom" {
+  type = object({id = string, size = number})
+}
+
+typedef "other" {
+  type = list(symbols::types(custom))
+}
+```
 
 #### Symbol file usage
 
--- a/rfc/20260424-symbol-libraries.md
+++ b/rfc/20260424-symbol-libraries.md
@@ -1,0 +1,332 @@
+# Symbol Libraries
+
+Issue: https://github.com/opentofu/opentofu/issues/1962
+
+One of the goals of many engineers using OpenTofu is to be as DRY as possible. The current design of OpenTofu's language has some hard edges that prevent easy re-use of types, logic, and values.
+
+### Types
+
+One common problem is the inability to re-use custom types between constructs (currently just variables), both in the same module and between modules. This makes working with complex types harder, as well as encouraging copy-paste anywhere types are re-used.
+
+The only way to work around this limitation is with code-gen on-top of your OpenTofu setup.
+
+References:
+* https://github.com/opentofu/opentofu/issues/1962
+* https://github.com/opentofu/opentofu/issues/2704
+* https://github.com/hashicorp/terraform/issues/27365
+
+### Logic
+
+Another common problem is the inability to define and re-use logic between expressions, both within the same module and between modules.
+
+The current way to accomplish this is to create a module with only variables, locals, and outputs. This requires a whole new module to be pulled into your configuration for each call to the "module function" and is a lot of heavy lifting for what should be a "simple" feature.
+
+References:
+* https://github.com/opentofu/opentofu/issues/4050
+* https://github.com/opentofu/opentofu/issues/793
+* https://github.com/hashicorp/terraform/issues/27696
+
+### Values
+
+If you want to re-use constant values between modules, you either need to use code-gen, an orchestration tool, or create a module with only outputs. All of these are valid techniques, though it could be useful to have a native solution with less overhead.
+
+Of the three concepts presented in this RFC, this one is the least strong. It could also be implemented using the functions construct above and merely serves as syntactic sugar.
+
+## Proposed Solution
+
+We propose adding symbol libraries to enable easy and well defined re-use of types, logic, and values in OpenTofu. The ideas in this RFC were spawned out of @apparentlymart's comment in https://github.com/opentofu/opentofu/issues/1962#issuecomment-2398192022.
+
+A symbol file may contain constant expressions used to define types, logic, and values. Referencing these symbols should made possible in a consistent and simple fashion.
+
+A symbol file may also import the contents of another symbol library.
+
+### User Documentation
+
+
+#### Symbol File Contents
+
+```hcl
+
+#### Types #####
+
+typedef "simple_type" {
+  # https://opentofu.org/docs/language/expressions/types/
+  # Any primitive type may be used here
+  type = int
+}
+
+typedef "complex_type" {
+  type = object({
+    ncpus = int
+    # Types may reference other types within the same symbol library (or other imported libraries)
+    memory_size = symbols::types(simple_type)
+  })
+}
+
+typedef "defaults_type" {
+  type = object({
+    # Defaults (via optional()) will be stored alongside the type
+    complex = optional(symbols::types(complex_type), { ncpus = 1 })
+  })
+}
+
+#### Values #####
+
+# Nearly identical concept to "locals" in the tofu configs, but usable externally
+values {
+  simple = 10
+  custom_regex = "<some complex regex>"
+  # Most builtin functions are allowed, TBD if we allow "file", "plantimestamp" and other non-constant functions
+  upper_regex = upper(value.custom_regex)
+}
+
+
+#### Functions ####
+
+# Simple function definition
+function "add" {
+  description = "add two numbers together" # Optional
+  type        = number                     # Optional
+  parameter "a" {
+    description = "first number"
+    type = number
+  }
+  parameter "b" {
+    description = "second number"
+    type = number
+  }
+  return = param.a + param.b 
+}
+
+# Function with parameter validation
+# https://opentofu.org/docs/language/expressions/custom-conditions/#input-variable-validation
+function "divide" {
+  parameter "a" {
+    type = number
+  }
+  parameter "b" {
+    type = number
+    validation {
+      condition     = param.b != 0
+      error_message = "Divide by zero"
+    }
+  }
+  return = param.a / param.b 
+}
+
+# Function using variadic parameters and multiple expressions
+function "greeting" {
+  type = list(string)
+  parameter "prefix" {
+    type    = "string"
+    default = "Hello "
+  }
+  parameter "name" {
+    type    = string
+    variadic = true
+  }
+  locals {
+    messages = [for x in param.name: "${param.prefix} ${x}!"]
+  }
+  return = tolist(local.messages)
+}
+
+# Function that uses custom types and calls other functions, both builtin and user defined
+typedef "vec3" {
+  type = object({ x = number, y = number, z = number})
+}
+function "vec3_length" {
+  parameter "vec" {
+    type = symbols::types(vec3)
+  }
+  locals {
+    xx = param.x * param.x
+    yy = param.y * param.y
+    zz = param.z * param.z
+    squared_sum = symbols::add(symbols::add(local.xx, local.yy), local.zz)
+  }
+  return sqrt(local.squared_sum)
+}
+
+#### Libraries ####
+
+# Import syntax
+symbols "namespace" {
+  # The package in which the symbols live
+  source = "./lib" # Same conventions as registry modules, though without static evaluation
+
+  # One of the decisions we need to make in this RFC is if we allow importing all of the symbols in a given package, 
+  # or if a symbol block can only represent the contents of a single file
+  file = "contents.sym.hcl"
+}
+
+# Usage in types
+typedef "exported" {
+  type = symbols::namespace::types(type_name)
+}
+
+typedef "embedded" {
+  type = list(symbols::namespace::types(type_name))
+}
+
+# Usage in functions
+function "myfunc" {
+  return symbols::namespace::func_name()
+}
+
+# Usage in values
+
+values {
+  val = symbols.namespace.other_val
+}
+```
+
+
+
+How do I have "unexported" symbols?
+
+
+#### Symbol file usage
+
+```hcl
+# ./local-lib/exported.sym.hcl
+
+typedef "items" {
+    type = list(string)
+}
+
+function "non_empty" {
+  parameter "in" {
+    type = symbols::type(items) 
+  }
+  return alltrue([for x in param.in: length(x) != 0])
+}
+
+function "assert_non_empty" {
+  parameter "in" {
+    type = symbols::type(items)
+    validation {
+      condition = non_empty(param.in)
+      error_message = "One or more of the elements in ${param.in} is empty"
+    }
+  }
+  return param.in
+}
+
+values {
+    default_items = ["foo", "bar", "baz"]
+}
+```
+
+```hcl
+# main.tofu
+
+# Import ./local-lib/exported.sym.hcl as "lib"
+symbols "lib" {
+  source = "./local-lib" # Static evaluation not allowed in this source
+  file   = "exported.sym.hcl" 
+}
+
+# This variable pre-emptively 
+variable "my_items" {
+  type = symbols::lib::type(items)
+  #type = list(string) would also work here
+  validation {
+    condition = symbols::lib::non_empty(var.my_items)
+    error_message = "One or more elements is empty"
+  }
+  default = symbols.lib.default_items
+}
+
+variable "my_items_unchecked" {
+  type = symbols::lib::type(items)
+}
+
+locals {
+  # This will raise an error from within the function validation if var.my_items_unchecked contains empty elements
+  my_items_checked = symbols::lib::assert_non_empty(var.my_items_unchecked)
+}
+
+resource "provider_type" "ident" {
+  value = local.my_items_checked
+}
+```
+
+Notes:
+* Function recursion is not supported due to current limitations in HCL
+* A hackathon prototype that includes a partial implementation of these features is available in the [symbol_libraries_experiment branch](https://github.com/opentofu/opentofu/tree/symbol_libraries_experiment)
+
+### Technical Approach
+
+The main approach for this feature is to create a library that allows for "compilation" of symbol libraries and that library's integration into OpenTofu. Although this library may be used externally to OpenTofu some day, the initial plan is to keep it internal.
+
+#### Symbol Library Implementation
+
+Contracts:
+* Parse a symbol call
+  - Given a `hcl.Body`, validate the body and return a structure that represents the symbol call.
+  - This is primarily a helper for standardizing parsing `symbols "namespace" { ... }` and does not perform any complex logic or evaluation.
+  - Similar to the existing `configs.decodeThingBlock()` pattern
+* Parse a symbol file
+  - Given a `hcl.Body`, validate the body and return a structure that represents the symbol file.
+  - Similar to the existing `configs.Parser{}.loadConfigFileBody()` pattern
+* Compile a symbol library.
+  - Given:
+    - A set of symbol files
+    - A symbol loader function, which takes a symbol call and returns a symbol library
+    - A set of builtin functions
+  - Return a symbol library which contains:
+    - Types compiled from `typedef` blocks
+    - Functions compiled from `function` blocks
+    - Values compiled from `values` blocks
+
+Most of the proposed implementation follows existing patterns within the `configs` package and will not be covered here. The interesting differences however are in how HCL handles type lookups and in how we need to "compile" the library into a static set of functionality that requires no more evaluation.
+
+The simpler of the two changes is to our fork of the HCL language. We already use the [typeext extension](https://github.com/opentofu/hcl/tree/opentofu/ext/typeexpr) heavily in OpenTofu. It however does not support the addition of custom types. The `typeext` package is fairly simple, and extending it to support custom types is a fairly straightforward task. A possible implementation can be seen in the [patch for the hackathon experiment](https://gist.github.com/cam72cam/ac7d3900c7d5b96c8d87a0ef803302a2).
+
+The other possible change to HCL is to make changes to the [userfunc extension](https://github.com/opentofu/hcl/tree/opentofu/ext/userfunc). The syntax shown in the user examples above is inspired by, but ultimately more complex than the existing userfunc extension. We will need to decide which syntax to use, and when either it makes sense to keep that implementation within the symbol library implementation or to move it into HCL.
+
+The more complex piece of the library implementation is how to handle the compilation of the symbol files into a library. Ultimately, some form of evaluator will need to be built. It will need to understand the dependencies between the elements within the given symbol files and ensure that they are compiled in the correct order.
+
+The approach taken in the hackathon project was to use the new [workgraph](https://pkg.go.dev/github.com/apparentlymart/go-workgraph) library (also used by the new engine) to build the evaluator/scope constructs. Once the scope was built, we simply iterate through and request all of the known values (which performs the compilation). There are many ways in which to perform this compilation. Given the current limited scope of what is allowed / is possible within symbol files (although complex), it is an easily solved problem.
+
+#### OpenTofu Integration
+
+The biggest challenge to integrating with OpenTofu at this juncture is that we are re-building the tofu engine as part of ongoing work, as well as swapping out the static evaluation implementation with engine integration. Our hope is that the simplicity of the symbol library integration contract means that the integration will not be hampered by or delay the ongoing engine work.
+
+One of the critical touch-points is supporting symbol libraries in variable types. These need to be known prior to static evaluation, as that depends on the variable types. In practice, the current `configs/config_build.go` code can be juggled around to make the symbol library loading and static evaluation steps explicit and properly integrated. In theory, it will be easier to do this integration natively within the new engine.
+
+The only other component that needs to be integrated is how and where we install symbol libraries themselves. The hackathon prototype treats them as modules to be installed, but is that the right solution?
+
+### Open Questions
+
+* How does this impact `tofu show -config -json`?
+  - We have integrated that command into the backend of the registry-ui.
+  - Will we need to add a new `tofu init -symbols-only` mode to support this use case?
+  - Does this have security implementations / prevent us from exposing `file/templatefile`?
+* Tooling Integration
+  - Tofu-ls?
+  - Linting?
+  - Dependabot?
+* JSON Syntax support
+  - Do we want to support json syntax for the symbol files?
+  - Are there any features that explicitly can't be supported via the json syntax?
+* Types are currently aliases and the actual names do not matter
+  - Do users expect the same type definition under different names to be incompatible?
+* Does the `symbols::(namespace::)types(type_name) syntax make sense?
+  - `symbols(namespace.type_name)`?
+  - `symbols::types(namespace.type_name)`?
+  - `types(symbols.namespace.type_name)`?
+  - Unless we want to make more significant changes to HCL, function call expr syntax is the easiest option here.
+* How do we want to represent symbol language versioning?
+* As we consider module locking, do we also want to consider symbol locking?
+
+### Future Considerations
+
+* Functions outside of symbol libraries?
+  - What use cases are there to support functions in .tofu files themselves?
+  - Is that complexity worth it?
+* Integration with static eval
+  - Would we ever want to support static eval or symbol values within the `source` field of symbol libraries?
+* Recursion support
+  - HCL in it's current form makes this impossible. It would be neat to try to support it *someday*
--- a/rfc/20260424-symbol-libraries.md
+++ b/rfc/20260424-symbol-libraries.md
@@ -52,12 +52,12 @@ A symbol file may also import the contents of another symbol library.
 typedef "simple_type" {
   # https://opentofu.org/docs/language/expressions/types/
   # Any builtin type may be used here
-  type = int
+  type = number
 }
 
 typedef "complex_type" {
   type = object({
-    ncpus = int
+    ncpus = number
     # Types may reference other types within the same symbol library (or other imported libraries)
     memory_size = symbols::types(simple_type)
   })
@@ -66,7 +66,7 @@ typedef "complex_type" {
 typedef "defaults_type" {
   type = object({
     # Defaults (via optional()) will be stored alongside the type
-    complex = optional(symbols::types(complex_type), { ncpus = 1 })
+    complex = optional(symbols::types(complex_type), { ncpus = 1, memory_size = 1024 })
   })
 }
 
@@ -118,11 +118,10 @@ function "divide" {
 function "greeting" {
   type = list(string)
   parameter "prefix" {
-    type    = "string"
-    default = "Hello "
+    type    = string
   }
   parameter "name" {
-    type    = string
+    type     = string
     variadic = true
   }
   locals {
@@ -145,7 +144,7 @@ function "vec3_length" {
     zz = param.z * param.z
     squared_sum = symbols::add(symbols::add(local.xx, local.yy), local.zz)
   }
-  return sqrt(local.squared_sum)
+  return = sqrt(local.squared_sum)
 }
 
 #### Libraries ####
@@ -171,7 +170,7 @@ typedef "embedded" {
 
 # Usage in functions
 function "myfunc" {
-  return symbols::namespace::func_name()
+  return = symbols::namespace::func_name()
 }
 
 # Usage in values
@@ -197,20 +196,20 @@ typedef "items" {
 
 function "non_empty" {
   parameter "in" {
-    type = symbols::type(items) 
+    type = symbols::types(items) 
   }
-  return alltrue([for x in param.in: length(x) != 0])
+  return = alltrue([for x in param.in: length(x) != 0])
 }
 
 function "assert_non_empty" {
   parameter "in" {
-    type = symbols::type(items)
+    type = symbols::types(items)
     validation {
       condition = non_empty(param.in)
       error_message = "One or more of the elements in ${param.in} is empty"
     }
   }
-  return param.in
+  return = param.in
 }
 
 values {
@@ -229,7 +228,7 @@ symbols "lib" {
 
 # This variable pre-emptively 
 variable "my_items" {
-  type = symbols::lib::type(items)
+  type = symbols::lib::types(items)
   #type = list(string) would also work here
   validation {
     condition = symbols::lib::non_empty(var.my_items)
@@ -239,7 +238,7 @@ variable "my_items" {
 }
 
 variable "my_items_unchecked" {
-  type = symbols::lib::type(items)
+  type = symbols::lib::types(items)
 }
 
 locals {
--- a/rfc/20260424-symbol-libraries.md
+++ b/rfc/20260424-symbol-libraries.md
@@ -205,8 +205,8 @@ function "assert_non_empty" {
   parameter "in" {
     type = symbols::types(items)
     validation {
-      condition = non_empty(param.in)
-      error_message = "One or more of the elements in ${param.in} is empty"
+      condition = symbols::non_empty(param.in)
+      error_message = "One or more of the elements in ${jsonencode(param.in)} is empty"
     }
   }
   return = param.in
--- a/rfc/20260424-symbol-libraries.md
+++ b/rfc/20260424-symbol-libraries.md
@@ -2,7 +2,7 @@
 
 Issue: https://github.com/opentofu/opentofu/issues/1962
 
-One of the goals of many engineers using OpenTofu is to be as DRY as possible. The current design of OpenTofu's language has some hard edges that prevent easy re-use of types, logic, and values.
+Most engineers using OpenTofu strive to write configuration that is as DRY as possible. The current design of OpenTofu's language has some hard edges that prevent easy re-use of types, logic, and values.
 
 ### Types
 
@@ -51,7 +51,7 @@ A symbol file may also import the contents of another symbol library.
 
 typedef "simple_type" {
   # https://opentofu.org/docs/language/expressions/types/
-  # Any primitive type may be used here
+  # Any builtin type may be used here
   type = int
 }
 
@@ -282,9 +282,9 @@ Contracts:
 
 Most of the proposed implementation follows existing patterns within the `configs` package and will not be covered here. The interesting differences however are in how HCL handles type lookups and in how we need to "compile" the library into a static set of functionality that requires no more evaluation.
 
-The simpler of the two changes is to our fork of the HCL language. We already use the [typeext extension](https://github.com/opentofu/hcl/tree/opentofu/ext/typeexpr) heavily in OpenTofu. It however does not support the addition of custom types. The `typeext` package is fairly simple, and extending it to support custom types is a fairly straightforward task. A possible implementation can be seen in the [patch for the hackathon experiment](https://gist.github.com/cam72cam/ac7d3900c7d5b96c8d87a0ef803302a2).
+The simpler of the two changes is to our fork of the HCL language. We already use the [typeext extension](https://github.com/opentofu/hcl/tree/opentofu/ext/typeexpr) heavily in OpenTofu. It however does not support the addition of custom types. The `typeext` package is fairly simple, and extending it to support custom types is a straightforward task. A possible implementation can be seen in the [patch for the hackathon experiment](https://gist.github.com/cam72cam/ac7d3900c7d5b96c8d87a0ef803302a2).
 
-The other possible change to HCL is to make changes to the [userfunc extension](https://github.com/opentofu/hcl/tree/opentofu/ext/userfunc). The syntax shown in the user examples above is inspired by, but ultimately more complex than the existing userfunc extension. We will need to decide which syntax to use, and when either it makes sense to keep that implementation within the symbol library implementation or to move it into HCL.
+The other possible change to HCL is in altering the [userfunc extension](https://github.com/opentofu/hcl/tree/opentofu/ext/userfunc). The syntax shown in the user examples above is inspired by, but ultimately more complex than the existing userfunc extension. We will need to decide which syntax to use, and when either it makes sense to keep that implementation within the symbol library implementation or to move it into HCL.
 
 The more complex piece of the library implementation is how to handle the compilation of the symbol files into a library. Ultimately, some form of evaluator will need to be built. It will need to understand the dependencies between the elements within the given symbol files and ensure that they are compiled in the correct order.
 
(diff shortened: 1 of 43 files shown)


ilmax:
Maybe a silly question, but why using typeset and not just type for a user defined type?

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,333 @@
...
+function "greeting" {
+  type = list(string)
+  parameter "prefix" {
+    type    = string
+  }
+  parameter "name" {
+    type     = string
+    variadic = true
+  }
+  locals {
```
cam72cam, line 127: What do you think of `locals` here vs `body`?

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### Functions ####
+
+# Simple function definition
+function "add" {
+  description = "add two numbers together" # Optional
+  type        = number                     # Optional
+  parameter "a" {
+    description = "first number"
+    type = number
```
yottta, line 92: Is it possible to use a type from the library? In some cases, I see it useful to have a complex type defined and associated functions to work with it.

rfc/20260424-symbol-libraries.md
yottta: Just my 2 cents here: if the content can support more "natural" programming syntax, like "if" statements without direct assignment, it should be `body`.
If this defines variables whose values are computed by using regular HCL syntax, then I would go with `locals`.

rfc/20260424-symbol-libraries.md
yottta: Answered several blocks below. Great! 👏

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,333 @@
...
+# Usage in values
+
+values {
+  val = symbols.namespace.other_val
+}
+```
+
+
+
+How do I have "unexported" symbols?
```
yottta, line 185: Hm. Interesting question 🤔 
Maybe could be the other way around? Everything is unexported and export only what it's wanted?
How about a special block to control the visibility of symbols?
```hcl
library {
    exported = [symbols::myfuncs, "*"]
    unexported = [types::embedded]
}
```
With some rules:
* If `library` block is is missing, everything is exported (default rule)
* If both, `exported` and `unexported` mention the same symbol, show an error diagnostic.
* Both can support a special symbol "*" where everything is either exported or not.

This offers the flexibility to specify only a subset of the symbols as being exported or unexported, allowing users to choose their flavor of controlling the visibility.

This is just my opinion, and I would be more than happy to hear other arguments around this.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,333 @@
...
+  - `types(symbols.namespace.type_name)`?
+  - Unless we want to make more significant changes to HCL, function call expr syntax is the easiest option here.
+* How do we want to represent symbol language versioning?
+* As we consider module locking, do we also want to consider symbol locking?
+
+### Future Considerations
+
+* Functions outside of symbol libraries?
+  - What use cases are there to support functions in .tofu files themselves?
+  - Is that complexity worth it?
```
yottta: I would say to keep it contained in the symbols file. Easier to reason and fix issues.

yottta (commented):
This looks great.
Regarding the versioning of these, personally I wouldn't do more than it's already done for modules.
In other words, these files are designed to be able to live next to `.tofu`/`.tf` files, meaning that these can easily be part of modules.
Therefore, if somebody wants to create a versioned and distributable library, they can easily use the existing mechanism of installing and versioning modules.

rfc/20260424-symbol-libraries.md
cam72cam: I forgot to fill this out when writing the RFC, I have a bit of a simpler proposal: 39a8236

yottta: And would that mean that if the path of imported library contains "internal/" and is not a sibling of the file that imports that, we will return an error about it? 🤔

cam72cam: I figured it could be a convention to start, I'm not sure we need to be as defensive as go about this.

rfc/20260424-symbol-libraries.md
ImIOImI: In Terramate, there is existing art on how to handle internally scoped `local` blocks that I have been using and I rather like:

So this:
```
function "vec3_length" {
  locals {
    xx = param.x * param.x
    yy = param.y * param.y
    zz = param.z * param.z
    squared_sum = symbols::add(symbols::add(local.xx, local.yy), local.zz)
  }
  return = sqrt(local.squared_sum)
}
```

becomes:
```
function "vec3_length" {
  lets {
    xx = param.x * param.x
    yy = param.y * param.y
    zz = param.z * param.z
    squared_sum = symbols::add(symbols::add(let.xx, let.yy), let.zz)
  }
  return = sqrt(let.squared_sum)
}
```

[lets block](https://terramate.io/docs/cli/reference/blocks/lets)

I find that this helps me parse the difference between globally scoped variables and variables scoped within a block very easily.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,363 @@
...
+
+#### Libraries ####
+
+# Import syntax
+symbols "namespace" {
+  # The package in which the symbols live
+  source = "./lib" # Same conventions as registry modules, though without static evaluation
+
+  # One of the decisions we need to make in this RFC is if we allow importing all of the symbols in a given package, 
+  # or if a symbol block can only represent the contents of a single file
```
ImIOImI, line 158: I think that when the source is a folder, that it should automatically import any hcl symbols in that folder. 

For more fine tuning we can just build out the folder structure of `.lib` like `.lib\string-functions`

I don't see importing individual files to be particularly helpful. Furthermore, file imports are counter intuitive to the way that module imports work.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,363 @@
...
+  }
+  return = sqrt(local.squared_sum)
+}
+
+#### Libraries ####
+
+# Import syntax
+symbols "namespace" {
+  # The package in which the symbols live
+  source = "./lib" # Same conventions as registry modules, though without static evaluation
```
ImIOImI, line 155: What's the rationale for skipping static evaluation here? For people like myself that generate these source declarations from templates, this isn't a big deal... but if static evaluation is important enough to include in module sources, it seems important enough to include here as well.

cam72cam: There are a few things to consider here:

### Integration with tools like Dependabot and the language server

This is already a problem with static eval that we have not solved.

### Config loading complexity

Right now, we can break config loading into four(ish) phases.
1. Turning the config files in a directory into a collection (`configs.Module`) of mostly understood concepts
2. Refining that collection with information from symbol libraries
  - This populates things like variable types that might be used in static evaluation
3. Refining that collection with static evaluation
4. Loading child module calls, recursively

We will likely be making some significant changes to that workflow as we adopt the constructs of the new engine, and I suspect that it would make interleaving those steps much easier.

To keep the concepts simple for now, I'm pitching that at least the first iteration of this feature does not include static eval.

apparentlymart:
Just a random thing I thought about while considering this: :grinning: 

We should probably think about some sort of integrated testing mechanism for these symbol libraries. Tests for these should be much more straightforward than tests for full modules because by definition there are no side-effects, and folks who currently use `.tftest.hcl` files with complex mocks in them just so they can test complex expressions in their modules might prefer to factor those complex expressions out into a symbol library and test it separately without all that extra overhead.

Since functions in symbol libraries are effectively a basic functional programming language expressed in HCL's grammar, we could consider supporting both of the popular testing strategies for such languages:

- Example-based testing: the sort of testing that's more common in procedural languages, where you just provide a set of example inputs and the outputs that are expected for each of those inputs.
- Property-based testing: write a set of expressions representing general rules that are expected to hold for any arbitrary set of inputs and their associated result, and then have the test harness generate a variety of examples automatically to ensure that those rules hold regardless of the chosen inputs.

Each of those strategies is useful in some different (but potentially overlapping) situations, so I don't think of this as a choice between them so much as evaluating each of them separately to decide if it's worth including. They might both be.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,363 @@
...
+  # https://opentofu.org/docs/language/expressions/types/
+  # Any builtin type may be used here
+  type = number
+}
+
+typedef "complex_type" {
+  type = object({
+    ncpus = number
+    # Types may reference other types within the same symbol library (or other imported libraries)
+    memory_size = symbols::types(simple_type)
```
lbordowitz: References to `simple_type` are a dissimilar from how we reference resources, locals, or vars, where the labels are referenced using a special dot notation.

Also: I do not know if `types` is a reserved keyword or not, or whether that makes a difference here.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,363 @@
...
+
+How do I have "unexported" symbols?
+
+The easiest way to do this without introduing any new syntax is to take an approach similar to go's `internal` pattern.
+
+```hcl
+# ./lib/contents.sym.hcl
+
+# Import the internal library
+symbols "internal" {
```
lbordowitz, line 193: I'm not a fan of a special keyword to indicate something is `internal`. I think it would be better practice to force a user to explicitly export a typedef instead, and it better follows the `output` pattern in modules.

lbordowitz (commented):
Putting in my two cents. I'm a big fan of the overall thrust of this RFC, although there are a couple details I'd like more clarification on because they would affect design.

rfc/20260424-symbol-libraries.md
cam72cam: This is not a special keyword, I just chose it as a potential convention.  This follows the existing pattern of how modules work.

rfc/20260424-symbol-libraries.md
cam72cam: There are a few different ways we could do the type integration, with differing levels of impact on changes to HCL and other consumers.

The ones I considered:
```hcl
# Function based syntax requires less changes to HCL and overall simpler implementation IMO
type = symbols::<namespace::>::types(<type_name>) # What's in this RFC
type = symbols::types(<namespace.><type_name>) # Potentially closer to what you are expecting?
type = types(symbols.<namespace.><type_name>)
# All of the existing inline type definition code is type based.

# More "typical" reference style
type = symbols.<namespace.><type_name>
# I could see this causing confusion as this feels more like a "normal value" that could be passed via a variable instead of a "special type integration".
```

rfc/20260424-symbol-libraries.md
lbordowitz: Go modules, or OpenTofu modules?

cam72cam: OpenTofu modules.

Yantrio:
I think overall the concepts that are listed in this RFC are okay and the thoughts/points/discussions I wanted to raise have already been raised by others.

But, One thing that is nagging at me reading this is that I don't think we're anywhere near landed on a packaging/distribution/consumption story, and I think we should do this before we close this RFC as I feel this will inform some decisions we make.

For example, the use case of a provider author shipping this alongside their provider is a very different story to an average end-user wanting to template out some repeated things in their main.tf root module.

Because we haven't yet nailed down the use case, it's hard to make choices about what's there now, and also how we distrubute it.

Would you mind elaborating on the RFC a bit more about intended use-cases, and also discuss how you think authors of these symbol libraries (whoever they are) would distribute and/or consume them? 

For example, I would love to explore
- Is this part of modules?
- Do we set up a new primative in the registry for this
	- If so, what's our story around private registries and market fragmentation?
- Do we distribute a "compiled" item that contains all of this (an artifact of some sort) or do we distribute raw code (similar to providers vs modules).

rfc/20260424-symbol-libraries.md
cam72cam: I'm tempted to use `lets` here as it does not conflict with existing naming, which would potentially cause confusion.

rfc/20260424-symbol-libraries.md
cam72cam: Anyone have a preference on which approach we take here?

yottta: All looks good to me but for the sake of consistency, I would say to go with the `more "typical" reference style`.
```hcl
type = symbols.<namespace.><type_name>
```

apparentlymart:
I have yet another random observation that's not related to any specific part of the text. :confounded: 

When I was originally chattering about "symbol libraries" in different issues one of the reasons I kept returning to it is that it fills a gap I see in today's OpenTofu: there's no unit of reuse for _just plain code_ (values and expressions), separate from resource declarations.

I'd personally love it if what we design here is general enough to make sense to import into any context where OpenTofu evaluates expressions. That might include:

- Import into an OpenTofu module
- Import into a test scenario (`.tftest.hcl`) file.
- Import into a ["stack configuration file"](https://gist.github.com/apparentlymart/8b895867777a099727454e1ebce5e5b3#explicit-multi-instancing-of-configuration-stacks), if we ultimately decide to implement something like that. (I'm imagining these files as a more thorough successor to `.tfvars` files that can encapsulate everything needed to represent something like "environment" or "workspace" or whatever.)
- Import into an expanded version of `templatefile` that somehow allows the caller to let the template use functions and values from their symbol libraries.

Since a symbol library is just a bunch of values and pure functions, they could in principle be imported into all of these different contexts and shared between them. We could even eventually offer it as a separate library that other HCL-based software could use to represent reusable values and expressions... Terragrunt could potentially support it in their HCL-based language too, for example, and then authors could have a single symbol library shared across both Terragrunt and OpenTofu.

I realize that what I've discussed here is scope creep. I'm sharing it to discuss whether that scope creep is desirable or not, and if so what additional design constraints that might imply for how we think about source addresses and distribution methods for these.

apparentlymart:
@cam72cam is intending to make some more revisions to this before the next round of review, and so I'm going to mark this as draft just to get it out of our review queue.

sam-delap:
I don’t know that I see a ton of value for embedding this directly in OpenTofu vs. leveraging something like Pkl: https://github.com/apple/pkl to define structs/reusable types (and then to @cam72cam’s point, publish those to a registry, which pkl does via https or a package:// resolution syntax that’s compatible with standard release publishing).

HCL’s typing dependency on go-cty also makes this difficult to implement in an HCL-based syntax, as one of go-cty’s typing constraints is that it’s built around dynamic types for individual symbols: https://github.com/zclconf/go-cty/blob/main/docs/concepts.md#value-operations

(see also https://github.com/zclconf/go-cty/blob/main/docs/concepts.md#unknown-values-and-the-dynamic-pseudo-type)

Tracking that we’ve got an initial implementation here, and would love to learn more abt the community’s overall desire to embed something like this in the framework.

rfc/20260424-symbol-libraries.md
yottta: Navigated to this line to leave a similar feedback so this comment will just reiterate my previously reaction of agreement with the above comment.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+values {
+  val = symbols.namespace.other_val
+}
+```
+
+
+
+How do I have "unexported" symbols?
+
+The easiest way to do this without introduing any new syntax is to take an approach similar to go's `internal` pattern.
```
yottta: It's not clear how the `internal` will work. Could you clear this out a little bit more please?

rfc/20260424-symbol-libraries/1-org-monolith/modules/org/pipeline/pipeline.tofu
```diff
@@ -0,0 +1,39 @@
...
+
+variable "transformer" {
+  type = object({
+    id = string
+  })
+}
+
+# tfcoremock stands in for a cloud resource
+resource "tfcoremock_simple_resource" "pipeline" {
+  id = "${var.info.division}-${var.info.department}-${var.info.team}-${var.info.project}-pipeline"
```
yottta, line 26: Please use a different field here since `id` is read only. [Docs](https://search.opentofu.org/provider/hashicorp/tfcoremock/latest/docs/resources/simple_resource#read-only)

yottta: Same for all the places where `id` is used as argument and not as attribute.

rfc/20260424-symbol-libraries/2-public-module/post_symbols/public/consumer/consumer.tofu
```diff
@@ -0,0 +1,11 @@
+symbols "lib" {
+  source = "../producer"
+}
+
+variable "produced" {
+  type = symbols::lib::types(producer)
+}
+
+resource "tfcoremock_simple_resource" "consumed" {
+  string = jsonencode(var.produced)
```
yottta, line 10: If I understand this correctly, what comes from the `var.produced` is an object type that contains `id` and `data { text, other }` fields.
Based on the [`tfcoremock` docs](https://search.opentofu.org/provider/hashicorp/tfcoremock/latest/docs/resources/simple_resource), one field is read only and one does not even exist.

I would like to see these examples in line with real life. Otherwise, it might create confusion for any future reader.

rfc/20260424-symbol-libraries/3-provider-library/pre_symbols/project/mod/module.tofu
```diff
@@ -0,0 +1,9 @@
+variable "entity" {
+  type = object({
+    id = string
+  })
```
yottta, line 4: Did you omit on purpose the rest of the arguments of the `tfcoremock_simple_resource`?

rfc/20260424-symbol-libraries.md
lbordowitz: Would you be opposed to the syntax `<namespace>::types.<type_name>`? Wrapping it in parentheses makes it feel like a function call to some function called `types` rather than an indication of a reference. Of course, maybe this syntax is imperfect too, since, as you say, it gives the impression of a "normal value." Yet it is familiar, like how we reference such varied objects as modules, resources, and locals. Perhaps `types` is treated like a a special "map", and it should be `<namespace>::types[<typename>]`, but that opens up the scary portal to the "type" type.

Rather than bikeshedding this, could we have a review of literature on how types are referenced in other languages? I think the underlying goal is ergonomics, familiarity, and low friction to use; the exact syntax does not matter as much to me as these goals being achieved.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+}
+function "vec3_length" {
+  parameter "vec" {
+    type = symbols::types(vec3)
+  }
+  locals {
+    xx = param.x * param.x
+    yy = param.y * param.y
+    zz = param.z * param.z
+    squared_sum = symbols::add(symbols::add(local.xx, local.yy), local.zz)
```
lbordowitz, line 145: This could be `+`, right? The `symbols::add` is just to show off using functions?

rfc/20260424-symbol-libraries.md
lbordowitz: I think it's like this (and please correct me if I'm off-base, @cam72cam !): Consider a scenario where someone would want to have a type, function, or value and not export it. You import a symbol and don't re-export it, it's by default internal. Want it used by consumers of your module? Then you can choose to explicitly re-export it.

This is one of the things that tripped me up in the way this RFC was worded: **the folder does not need to be named internal**.  It's just a somewhat-similar pattern.

For example, if we were on a local filesystem, we could just symlink to the internal and it would be the same; that is, there's nothing special about its name nor any code needed for it. Alternatively, if the symbols library were loaded from the registry, only explicitly exported things are visible to the user.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+* JSON Syntax support
+  - Do we want to support json syntax for the symbol files?
+  - Are there any features that explicitly can't be supported via the json syntax?
+* Types are currently aliases and the actual names do not matter
+  - Do users expect the same type definition under different names to be incompatible?
+* Does the `symbols::(namespace::)types(type_name) syntax make sense?
+  - `symbols(namespace.type_name)`?
+  - `symbols::types(namespace.type_name)`?
+  - `types(symbols.namespace.type_name)`?
+  - Unless we want to make more significant changes to HCL, function call expr syntax is the easiest option here.
```
lbordowitz, line 361: That's unfortunate, because I find it somewhat confusing.

lbordowitz (commented):
So far, I'm most stuck on the way type definitions are done; values and functions feel fluent and straightforward, and I have no objections to the way they're depicted in this RFC.

rfc/20260424-symbol-libraries.md
cam72cam: That is correct @lbordowitz !

rfc/20260424-symbol-libraries/1-org-monolith/modules/org/pipeline/pipeline.tofu
cam72cam: The docs are wrong, which is shown as a conflict by your link already.

rfc/20260424-symbol-libraries/2-public-module/post_symbols/public/consumer/consumer.tofu
cam72cam: As for ID, The docs are wrong, which is shown as a conflict by your link already.

I did test all of these examples and they function just fine.  I think you need to re-read the example here.

rfc/20260424-symbol-libraries/3-provider-library/pre_symbols/project/mod/module.tofu
cam72cam: Yes, I defined an explicit type for the used fields. As more are needed, they can be exposed.

It is in service of showing off why https://github.com/opentofu/opentofu/pull/4052/changes/BASE..8187efae2c9e660870887a242ff359045f21ce7b#diff-b87e2cbb1785dc78f6db8fa48c27bafca350e0e0ca9021da37374a64f3420eb2R3 (needing to change the variable definition in multiple places) is a pain.

rfc/20260424-symbol-libraries.md
cam72cam: That is correct, just to show off.

rfc/20260424-symbol-libraries.md
cam72cam: If I were designing it from scratch, I might not prefer functions.  However, most of the existing type code *is built to use function syntax*. I designed this to follow the current conventions of the language.

```hcl
type = object({
  foo = optional(string)
  other = map(string)
})
```

Would `type = types(symbols.<namespace.><type_name>)` be a reasonable compromise? It keeps the existing functional convention, but uses a simpler selector syntax.

lbordowitz: The compromise you've provided feels like the closest we might get to the ideal I'm trying to communicate. I'm not opposed to it, but I do have a few nagging qualms.

In your example, `object`, `optional`, and `map` are all "generic" types taking parameters, whether that's the type to have a map/optional on, or the schema of the object. When we write `symbols::types(simple_type)`, we are not specifying a type that we can generically pass as a parameter for another concept, we are making a semantic reference. A function call does not feel like reference.

I'm tempted to ask for `symbols::types[simple_type]`, but that feels weird compared to a straight reference. I do want to be cognizant of the practical limitations [you talk about here](https://github.com/opentofu/opentofu/pull/4052/changes/BASE..8187efae2c9e660870887a242ff359045f21ce7b#diff-cbfc6b2339f208e199724a757d9ea9bb2630fbea6b984a0c7c2bf96f0b62df98R351). What kind of changes would need to be made to HCL to get this to work?

apparentlymart: Currently the syntax with `::` is a function-call-specific thing in HCL, and is honestly a bit of a hack due to the idea of function namespaces being retrofitted long after the `hcl.EvalContext` structure was established.

In theory I suppose a similar hack could be applied to the `EvalContext.Variables` field, where the calling application can optionally put symbol names containing colons as keys in that map and then HCL's parser and evaluator would treat it as a special case, but that seems like a pretty big divergence from the core behavior from upstream HCL since we'd need to modify both the parser and the evaluation logic.

(I do have the _very early beginnings_ of a fresh attempt at [ZCL](https://github.com/zclconf/go-zcl) -- the codebase that HCL 2 forked from -- in which I tried out a very different approach to evaluation that makes these colon-separated symbols a more first-class thing, and also makes _type expressions themselves_ a first-class thing, but it'd be very heavy lift to get that into a mature enough state where OpenTofu could use it, if we decided we wanted to do that.)

apparentlymart: I wonder if we should try for a version of this proposal that lacks support for custom types at first and focuses only on reusable functions and variables, since that's still a good amount of value to authors and doesn't seem to require any new syntax or quirky workarounds.

We could then take some more time to consider our options for how to support custom types in a more ergonomic way, rather than rushing in to a workaround that we may come to regret. I'd rather take a little more time to consider some different approaches, but we needn't delay other aspects of this proposal for that since the other features of this new mini-language are pretty independent.

rfc/20260424-symbol-libraries.md
apparentlymart: I'm not opposed to using a different noun, but I'd prefer it to not be specifically "lets" because that's a pretty confusing use of English grammar that seems likely to cause learning roadblocks. ("let" is a verb, not a noun.)

One additional option we have in this new language is to let these function-local symbols appear without any prefix at all. The main OpenTofu module language requires `local` in front because the module language prioritizes referring to resources and so everything that isn't a resource needs to be in a separate namespace, but for this new language we have no such constraint and so we could let the locally-defined symbols live in the top-level namespace:

```hcl
function "vec3_length" {
   NAME_TBD {
    xx = param.x * param.x
    yy = param.y * param.y
    zz = param.z * param.z
    squared_sum = symbols::add(symbols::add(xx, yy), zz)
  }
  return = sqrt(squared_sum)
}
```

Sadly we would still need to name the block they are declared in (`NAME_TBD` above) as long as we want that to be the thing that's nested, but we can at least avoid the extra prefix noise on every reference. Honestly it would be nice to just flatten the whole thing so that there's no nesting at all, but of course that would require us to reserve some name to represent the return value. That seems a little awkward but maybe not so bad after you get used to it? 🤷‍♂️ 

```hcl
function "vec3_length" {
  parameter "x" {}
  parameter "y" {}
  parameter "z" {}

  # Now a nested block for the entire set of definitions,
  # so that we can interpret it using hcl.Body.JustAttributes
  # without causing conflict with the parameter decls.
  body {
    xx = x * x
    yy = y * y
    zz = z * z
    squared_sum = symbols::add(symbols::add(xx, yy), zz)

    # "result" is now essentially just a regular local symbol,
    # except that the function evaluation would take the
    # value of this one as the final function result.
    result = sqrt(squared_sum)
  }
}
```

I'd consider also dropping the prefix for any symbols declared at the top-level outside of functions, and making this symbol libraries language have lexical scoping where a prefixless reference just resolves to whatever declaration is "closest", as is familiar from many modern general-purpose programming languages.

(To be honest, every time I look at this I find myself considering implementing a new grammar that's more suited to the problem we're trying to solve here. HCL wasn't really designed for this sort of thing and we're really pushing the boundaries of its ergonomics here. But designing and implementing an entirely new grammar and associated parser is a much heavier lift of course, and not something we should take on lightly.)

rfc/20260424-symbol-libraries.md
Yantrio: I think I agree with @apparentlymart here. I worry that we need to solidify typing and type-conversion/coercion before shipping this just to ensure that we do it correctly. I wonder though: is there a route to getting this RFC in for functions and values and then extending it with types? I dont see any reason we can't do that.

rfc/20260424-symbol-libraries.md
Yantrio: I just want to go on record that im not a huge fan of the golang approach and having some sort of internal package that doesnt get exported. I would rather have an explicit way to say what is exported or not per item (Similar to typescript functions and the `export` keyword or `defaults` and not through what I would say is mis-using scopes)

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+
+We also need to determine how and where we install symbol library packages. The hackathon prototype treats them as modules to be installed, but is that the right solution?
+
+The injection of the symbol library data into the OpenTofu engine/evaluator is a trivial concern, if the prototype is anything to go by.
+
+### Open Questions
+
+* How does this impact `tofu show -config -json`?
+  - We have integrated that command into the backend of the registry-ui.
+  - Will we need to add a new `tofu init -symbols-only` mode to support this use case?
```
Yantrio, line 338: I suspect we should have something similar to `tofu show -module=/path` but we need to somehow initialize which does complicate things :(

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+  - Does this have security implementations / prevent us from exposing `file/templatefile`?
+* Tooling Integration
+  - Tofu-ls?
```
Yantrio: we should add tofu-ls support for sure.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+  - Linting?
+  - Dependabot?
+* JSON Syntax support
+  - Do we want to support json syntax for the symbol files?
+  - Are there any features that explicitly can't be supported via the json syntax?
+* Types are currently aliases and the actual names do not matter
+  - Do users expect the same type definition under different names to be incompatible?
```
Yantrio, line 354: My opinion here is that we should allow for type-coercion and this means that similar to your 3-provider-library example, someone can ship a set of types that you can coerce your value into.

Yantrio: To be more concrete, I'd love to give an example of a wrapper tool that would like to expect outputs of modules to be in a specific format. The author of that tool (CLI/TUI/TACOS etc) should be able to ship a symbol library, and if people type their outputs in that format, the tool can read those items.

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,366 @@
...
+* Does the `symbols::(namespace::)types(type_name) syntax make sense?
+  - `symbols(namespace.type_name)`?
+  - `symbols::types(namespace.type_name)`?
+  - `types(symbols.namespace.type_name)`?
+  - Unless we want to make more significant changes to HCL, function call expr syntax is the easiest option here.
+* How do we want to represent symbol language versioning?
```
Yantrio, line 366: And can someone bring in multiple versions of the same symbol lib? How is that referenced in the syntax of `symbols::(namespace::)types...`

Yantrio (commented):
Overall I like the approach. I have some hesitation about shipping it without a nice discussion on type conversion and coercion. 

One of the things I wanted to go on record saying is that I really like the idea of keeping symbol libraries in their own filename (not inside .tf or .tofu files) because this creates a clear boundary to stop users being confused. I dont think we should allow end users to reference resources, datasources, etc within these functions or variables as that creates a lot of complexity that I think we dont need.

rfc/20260424-symbol-libraries.md
cam72cam: Same way that we do with modules.  The consumer chooses the namespace, as is shown above.

tsarna:
At @cam72cam 's suggestion, I'm mentioning [functy](https://github.com/tsarna/functy) here, which may be helpful for some of the things you're trying to do.

`functy` is a small but full-featured imperative language with conditionals, loops, switch/case, try/catch/finally, etc . It has a mostly Go-inspired syntax and, importantly, **uses _exactly_ HCL's expression syntax and the same underlying type system (`cty`)**. It was designed specifically to feel "native" to HCL users and to integrate seamlessly with HCL/cty. `functy` functions compile to `cty.Function` objects that get registered in an `hcl.EvalContext` like any other function. 
Basically, it was built for exactly the case where the [HCL userfunc extension] (https://github.com/hashicorp/hcl/tree/main/ext/userfunc) isn't enough, and you wish HCL had "real" language for defining functions, where you want control flow, temporary variables, error handling, etc.

`functy` code does not embed directly in HCL, but instead lives in parallel `.cty` files in the same directory.

Also, functy has a lot of the tooling and creature comforts you'd want: a CLI with `run`, `eval`, `check`, `test`, `fmt`, `symbols`, a very basic interactive `repl`, a VS Code editor extension, and I'm planning an LSP for it.

It seems like the goal of this project is to allow defining and reusing variables, types, and functions in OpenTofu, and I think `functy` can help to some extent with all three.

## Functions

For functions I think it's the clear winner, because that's _exactly_ the gap it's trying to fill. I tried building an imperative language in HCL first, and it _sucks_. Trying to work within HCL syntax, you end up with something that has awful ergonomics: verbose, ugly, and with weird syntax rules that are hard to remember (at least for me) when thinking with a programming mindset.

Using sibling files allows a friendly syntax that feels familiar and natural to _both_ programmers and people who live in HCL. It has all the benefits of a "real" language, but without that jarring mental shift back and forth when using a "foreign" language like Starlark/JS/Lua/whatever.

Also, because everything is natively `cty`, there's no weird gotchas caused by trying to map values back and forth to a foreign type system, which will never quite match 100%.

## Variables

The `functy` parser _optionally_ allows top-level variable and/or constant declarations, with types and initializers:

```functy
namespace foo::bar

const pi : number = 3.1415927
var answer : number = 42
```

It's 100% up to the host program to decide if they're allowed, and to give them semantics. `functy` just collects the definitions and lets the host decide what to do with them.

(In the system I built `functy` for, constants map directly to values in the EvalContext, while variables are another way of spelling `var "name" { ... }` blocks in the HCL config. This is totally up to the host program though.)

An important HCL limitation here is that it only allows `::` in function calls, there is no namespacing for variables. This is where the `symbol("foo::bar")` thing in the current proposal comes in, I think. `functy` can't do anything about that. It does however track in which namespace top-level vars/consts are declared, so again the host can do whatever it wants with that, including using it to back a `symbol()` lookup function.

So, while it doesn't solve this problem, it does offer an alternate and friendlier way to spell these things than HCL blocks, and allows them to be defined close to the functions to which they relate.

## Types

`functy` use the familiar HCL type definition syntax but with support for capsule types, and has a typedef feature:

```functy
type Point = object({ x = number, y = number })

func midpoint(a: Point, b: Point) -> Point {
    return { x = (a.x + b.x) / 2, y = (a.y + b.y) / 2 }
}
```

~~Types are currently *not* namespaced in functy, but that could be changed if desirable. You'd still need some way of dealing with the fact that `::` wouldn't be allowed in type expressions in HCL, like a `type()` function, or some kind of re-export mechanism.~~

tsarna:
A more specific proposal:

 * https://github.com/tsarna/functy/blob/main/OPENTOFU-SYMBOLS.md mirrors the current RFC, but with `.sym.hcl` files replaced with functy `.cty` files. The `.tofu` side is nearly unchanged from the original, so this really only affects the _supplier_ side of symbols, they'd be _consumed_ the same way as in the original RFC.

 * https://github.com/tsarna/functy/blob/main/symbols/symbols.go is a prototype of a wrapper that provides everything I think OpenTofu would need to implement the above (modulo a few planned improvements to functy around namespace handling). It would just need to be wired in:
     * Parse `symbols` blocks in `.tofu` files
     * Hand them to the wrapper
     * Add the results from the wrapper into OpenTofu's `hcl.EvalContext` (`.Functions` and `.Variables["symbols"]`)
     * Hook in the type resolution using the provided `.Type(label, name)` lookup method.
 
 Compared to the `.sym.hcl` approach, this offers:
 
  * improved syntax
  * _significantly_ more expressive function capabilities (branches, loops, error handling, ...)

tsarna:
I have a draft implementation at https://github.com/tsarna/opentofu/tree/tsarna-functy-symbols. It was more involved than I expected, but I have the basics going. Per repo policy, there is no LLM-generated code, though I did use Claude Code for research and (prose-only) review feedback on my changes.

What works today:
 * `symbols` blocks with a local source (`source = "./blah"`)
 * `symbols::symbolsblocklabel::functionname(...)` function calls
 * `symbols.symbolsblocklabel.constname` value references
 * Fairly comprehensive tests.
 
 What doesn't work yet:
  * The new runtime. I'll look at this next.
  * Remote `source = ...`
  * Type references. The direction for type references seems unsettled, but a Module.SymbolLibrary.TypeLookup(symbolsblocklabel, typename) -> cty.Type function is provided, ready to hook in.
  
Feedback welcome!

Small example:

main.tofu:
```opentofu 
symbols "mylib" {
   source = "./mylib"
}

output "doubled_answer" {
  # 84
  value = symbols::mylib::double(symbols.mylib.answer)
}

# Internally, Module.SymbolLibrary.TypeLookup("mylib", "Point") -> an appropriate cty.Type.
# TBD how to reference that from .tofu files.
```

mylib/lib.cty:
```functy
const answer = 42

type Point = object({ x = number, y = number })

func double(n: number) -> number {
    return n * 2
}
```

apparentlymart:
@tsarna feel free to ignore the new language runtime for now if you like. It's far from feature-complete anyway, so I think it's better to focus on figuring out both what user experience we want and how this would integrate into the current runtime for now, and then whatever we decide to move forward with we can figure out how to do something equivalent in the new runtime afterwards.

I've not looked closely at what you shared yet since my attention is personally elsewhere right now and just wanted to quickly let you know it's okay to ignore the new runtime while you're prototyping, but I know @cam72cam is planning another round of work on this RFC soon and so we'll have an opportunity to look more closely at what you've implemented then.

rfc/20260424-symbol-libraries.md
cam72cam: My one concern with dropping the prefix is what we decide to do for name conflicts between the functions within a symbol library's scope and the core functions.

```
function tolist { }

values {
  my_val = tolist(something)
}
```

I think I'd be ok with this as long as we clearly document that the symbol library takes precedence in this scenario.

rfc/20260424-symbol-libraries.md
cam72cam: Documented better in d831ebd88e

rfc/20260424-symbol-libraries.md
cam72cam: Documented in d831ebd88e

rfc/20260424-symbol-libraries.md
cam72cam: IMO types are a very important part of this feature and I'd like to ship v1 with them if at all possible.

I've gone through most of the related OpenTofu and Terraform issues related to types in rfc/20260424-symbol-libraries/type_research.md with @apparentlymart 's help.

The ask here seems pretty clear to me.

Are there any major objections, outside of the reference conventions?

rfc/20260424-symbol-libraries.md
cam72cam: Resolved in 6e155ea86a

rfc/20260424-symbol-libraries.md
cam72cam: Added notes in 6e155ea86a

rfc/20260424-symbol-libraries.md
cam72cam: 6e155ea86a

rfc/20260424-symbol-libraries.md
cam72cam: Detailed response in 6e155ea86a. The examples in the RFC folder cover that use case as well.

Yantrio (approved):

rfc/20260424-symbol-libraries.md
cam72cam: Some additional context is that the convert RFC (https://github.com/opentofu/opentofu/pull/4440) essentially requires us to use a functional-only form without parameters.

cam72cam: I've updated the rfc to follow the structure
```hcl
type = symbols::<namespace::><type>()
// For Example
type = symbols::custom_type()
type = symbols::mylib::custom_type()
```

tsarna: I have changes to hcl's ext/typeexpr to implement this `x::y::z()` style in terms of a "Resolver" interface. OpenTofu already has a fork of hcl so this seemed like the path of least resistance. It's also generic enough that it might be acceptable upstream.

I don't think it would be significantly harder to implement `x.y.z` or `types(x.y.z)`, which to me read more naturally, but I went with what's in the RFC.

I haven't worked on the tofu side yet.

apparentlymart:
Hi @sam-delap and @tsarna,

Sorry we've not responded more directly to your alternative language proposals so far. The discussion on this RFC has already been quite large, and so I think we should discuss alternative proposals in separate issues instead of trying to discuss everything here... GitHub's issues/PRs UI doesn't scale well for larger discussions, unfortunately.

If you'd like to discuss those specific other languages as alternative ways to write symbol libraries, please open new issues where we can discuss each of them and mention this one as part of the issue description. In the long run we'd want to go through the RFC process but I'm suggesting issues first just because I think that some initial discussion would help figure out what details an RFC about each of these ought to cover, since we've not previously attempted to include languages other than HCL in OpenTofu.

Thanks!

rfc/20260424-symbol-libraries.md
```diff
@@ -0,0 +1,386 @@
...
+
+#### OpenTofu Integration
+
+The biggest challenge to integrating with OpenTofu at this juncture is that we are re-building the tofu engine as part of ongoing work, as well as swapping out the static evaluation implementation with engine integration. Our hope is that the simplicity of the symbol library integration contract means that the integration will not be hampered by or delay the ongoing engine work.
+
+One of the critical touch-points is supporting symbol libraries in variable types. These need to be known prior to static evaluation, as that depends on the variable types. In practice, the current `configs/config_build.go` code can be juggled around to make the symbol library loading and static evaluation steps explicit and properly integrated. In theory, it will be easier to do this integration natively within the new engine.
+
+We also need to determine how and where we install symbol library packages. The hackathon prototype treats them as modules to be installed, but is that the right solution?
```
yottta: Do you plan replacing this question with an actual statement? AFAIR, was decided to leave these be distributed as modules. Correct me if I am wrong.

yottta (approved):

rfc/20260424-symbol-libraries.md
cam72cam: After some additional discussion between @apparentlymart and myself, our thoughts are to leave this as local for now and to not drop prefixes.

This is for a few reasons:
* Keeps the reference syntax identical between normal module code and symbol libraries
* Dropping the prefix would add lots of complication detecting potential name conflicts
* Potential conflicts with the builtin type system names (optional, object, etc...)

We can revisit this in subsequent language versions.

I'm still open to an alternative to "local" as the name, though "lets" is a bit too overloaded.

rfc/20260424-symbol-libraries.md
cam72cam: Summarizing a discussion I had with @apparentlymart:

Either single file source or directory sources could be argued.  There is a worry that users will expect symbol files in the same directory as a module to be automatically loaded.  If this is a big enough concern, we should make the error quite clear and prescriptive if we detect someone trying to use non-namespaced symbol references within a module.

rfc/20260424-symbol-libraries.md
cam72cam: In summary, we see the advantage of allowing static eval in these source fields.  However due to the technical challenges above, we are holding off on considering how to implement this until after the new engine is complete.

rfc/20260424-symbol-libraries.md
cam72cam: 683f176509

lbordowitz (approved):

PR #4055, opened 2026-04-27
Unify the state related flags in one struct

yottta:
OpenTofu has several flags that control how the state is handled:
* `-state`: indicates the input file containing the state
* `-state-out`: indicates the output file where the newly updated state should be written
* `-lock`: enables or disables the locking of the state
* `-lock-timeout`: specifies the timeout duration when the lock of the state is triggered
* `-backup`: when specified, it backups the previous state into it, before writing the new one

Before this PR, all these flags were mostly registered via `extendedFlagSet`, but that implementation lacked granularity. But most of the commands don't use all of those flags, but only a subset, and some commands use a subset with different default values. 
Even more, the `Meta` structure contained individual attributes to store and pass these values to the concerned logic in the system.
Another thing, some commands, even if they didn't register the `-lock` flag (which provides a default value as `true`), still wanted to have the lock enabled but without a way for the user to control its value. Before this, that `Meta` attributes was configured specifically by the commands that required that.

As can be seen in the list of inconveniences listed above, the handling of these was pretty much scattered and hard to follow. 
Now, all of these are unified under one struct, with a constructor that ensures that lock is enabled and methods to granularly register flags as are needed.

Some things that needs to be known before jumping into the review:
* Any command that passed a non-nil value in the `extendedFlagSet` should register all the flags from the `arguments.State`.
* Some commands register the flags in 2 steps since some flags are registered with their general default values but some are registered with a different default value.
* Previously, I added in the `arguments.Backend` also the flags for `-lock`/`-lock-timeout`, but since now I managed to unify everything, I removed those too.
* Check the `NOTE:` from `commands/state.mv.go` since that's the part that gets a little bit out of the way of doing things around. Maybe you have other suggestions ❓ 

Part of #3595

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/cli.go
+++ b/internal/backend/cli.go
@@ -55,7 +55,7 @@ type CLIOpts struct {
 	// StateArgs carries the state related configurations. Refer to the
 	// arguments.State for more details.
 	// In this particular case, an empty arguments.State#BackupPath will result
-	// in skipping the backup entirely and the new state will just ovewrite the old
+	// in skipping the backup entirely and the new state will just overwrite the old
 	// one.
 	StateArgs arguments.State
 
--- a/internal/backend/cli.go
+++ b/internal/backend/cli.go
@@ -6,6 +6,7 @@
 package backend
 
 import (
+	"github.com/opentofu/opentofu/internal/command/arguments"
 	"github.com/opentofu/opentofu/internal/command/views"
 	"github.com/opentofu/opentofu/internal/tofu"
 )
@@ -51,16 +52,12 @@ type CLIOpts struct {
 	// of their actions.
 	View views.BackendRemote
 
-	// StatePath is the local path where state is read from.
-	//
-	// StateOutPath is the local path where the state will be written.
-	// If this is empty, it will default to StatePath.
-	//
-	// StateBackupPath is the local path where a backup file will be written.
-	// If this is empty, no backup will be taken.
-	StatePath       string
-	StateOutPath    string
-	StateBackupPath string
+	// StateArgs carries the state related configurations. Refer to the
+	// arguments.State for more details.
+	// In this particular case, an empty arguments.State#BackupPath will result
+	// in skipping the backup entirely and the new state will just ovewrite the old
+	// one.
+	StateArgs arguments.State
 
 	// ContextOpts are the base context options to set when initializing a
 	// OpenTofu context. Many of these will be overridden or merged by
--- a/internal/command/arguments/backend_test.go
+++ b/internal/command/arguments/backend_test.go
@@ -8,7 +8,6 @@ package arguments
 import (
 	"flag"
 	"testing"
-	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/opentofu/opentofu/internal/tfdiags"
@@ -53,65 +52,6 @@ func TestBackend_AddIgnoreRemoteVersionFlag(t *testing.T) {
 
 			if got := backend.IgnoreRemoteVersion; got != tc.want {
 				t.Errorf("IgnoreRemoteVersion = %v, want %v", got, tc.want)
-			}
-		})
-	}
-}
-
-func TestBackend_AddStateFlags(t *testing.T) {
-	testCases := map[string]struct {
-		args            []string
-		wantLock        bool
-		wantLockTimeout time.Duration
-	}{
-		"default values": {
-			args:            nil,
-			wantLock:        true,
-			wantLockTimeout: 0,
-		},
-		"lock set to false": {
-			args:            []string{"-lock=false"},
-			wantLock:        false,
-			wantLockTimeout: 0,
-		},
-		"lock set to true explicitly": {
-			args:            []string{"-lock=true"},
-			wantLock:        true,
-			wantLockTimeout: 0,
-		},
-		"lock-timeout set": {
-			args:            []string{"-lock-timeout=10s"},
-			wantLock:        true,
-			wantLockTimeout: 10 * time.Second,
-		},
-		"lock-timeout set in minutes": {
-			args:            []string{"-lock-timeout=5m"},
-			wantLock:        true,
-			wantLockTimeout: 5 * time.Minute,
-		},
-		"both flags set": {
-			args:            []string{"-lock=false", "-lock-timeout=30s"},
-			wantLock:        false,
-			wantLockTimeout: 30 * time.Second,
-		},
-	}
-
-	for name, tc := range testCases {
-		t.Run(name, func(t *testing.T) {
-			backend := &Backend{}
-			fs := flag.NewFlagSet("test", flag.ContinueOnError)
-			backend.AddStateFlags(fs)
-
-			if err := fs.Parse(tc.args); err != nil {
-				t.Fatalf("unexpected error parsing flags: %v", err)
-			}
-
-			if got := backend.StateLock; got != tc.wantLock {
-				t.Errorf("StateLock = %v, want %v", got, tc.wantLock)
-			}
-
-			if got := backend.StateLockTimeout; got != tc.wantLockTimeout {
-				t.Errorf("StateLockTimeout = %v, want %v", got, tc.wantLockTimeout)
 			}
 		})
 	}
@@ -350,8 +290,6 @@ func TestBackend_AllFlags(t *testing.T) {
 			args: nil,
 			want: Backend{
 				IgnoreRemoteVersion: false,
-				StateLock:           true,
-				StateLockTimeout:    0,
 				ForceInitCopy:       false,
 				Reconfigure:         false,
 				MigrateState:        false,
@@ -360,16 +298,12 @@ func TestBackend_AllFlags(t *testing.T) {
 		"all flags set": {
 			args: []string{
 				"-ignore-remote-version",
-				"-lock=false",
-				"-lock-timeout=1m",
 				"-force-copy",
 				"-reconfigure",
 				"-migrate-state",
 			},
 			want: Backend{
 				IgnoreRemoteVersion: true,
-				StateLock:           false,
-				StateLockTimeout:    time.Minute,
 				ForceInitCopy:       true,
 				Reconfigure:         true,
 				MigrateState:        true,
@@ -378,13 +312,10 @@ func TestBackend_AllFlags(t *testing.T) {
 		"mixed flags": {
 			args: []string{
 				"-ignore-remote-version=true",
-				"-lock-timeout=30s",
 				"-migrate-state",
 			},
 			want: Backend{
 				IgnoreRemoteVersion: true,
-				StateLock:           true,
-				StateLockTimeout:    30 * time.Second,
 				ForceInitCopy:       false,
 				Reconfigure:         false,
 				MigrateState:        true,
@@ -397,7 +328,6 @@ func TestBackend_AllFlags(t *testing.T) {
 			backend := &Backend{}
 			fs := flag.NewFlagSet("test", flag.ContinueOnError)
 			backend.AddIgnoreRemoteVersionFlag(fs)
-			backend.AddStateFlags(fs)
 			backend.AddMigrationFlags(fs)
 
 			if err := fs.Parse(tc.args); err != nil {
--- a/internal/command/arguments/extended.go
+++ b/internal/command/arguments/extended.go
@@ -39,6 +39,7 @@ type State struct {
 
 	// StatePath specifies a non-default location for the state file. The
 	// default value is blank, which is interpreted as "terraform.tfstate".
+	// Represents the local path where state is read from.
 	StatePath string
 
 	// StateOutPath specifies a different path to write the final state file.
@@ -48,8 +49,8 @@ type State struct {
 
 	// BackupPath specifies the path where a backup copy of the state file will
 	// be stored before the new state is written. The default value is blank,
-	// which is interpreted as StateOutPath +
-	// ".backup".
+	// which is interpreted as StateOutPath + ".backup" or, in some cases, the backup
+	// is skipped altogether.
 	BackupPath string
 }
 
--- a/internal/command/arguments/extended.go
+++ b/internal/command/arguments/extended.go
@@ -54,16 +54,6 @@ type State struct {
 	BackupPath string
 }
 
-// NewStateFlags can be used to create the [State] instance with [State.Lock] initialised as "true", instead of creating
-// a new object inline.
-// It's not needed to be used if the lock related flags are registered, since that provides the same default value
-// for the lock flag.
-func NewStateFlags() *State {
-	return &State{
-		Lock: true,
-	}
-}
-
 // AddFlags is the sole logic of registering the state related flags in OpenTofu.
 func (s *State) AddFlags(f *flag.FlagSet, lockFlags bool, stateInPath bool, stateOutPath bool, backupPath bool) {
 	if lockFlags {
--- a/internal/command/arguments/extended.go
+++ b/internal/command/arguments/extended.go
@@ -54,8 +54,10 @@ type State struct {
 	BackupPath string
 }
 
-// NewStateFlags must be used when creating a new [State] instance, instead of creating a new object inline.
-// This is because it provides the default value for [State.Lock] even if the flags are not registered.
+// NewStateFlags can be used to create the [State] instance with [State.Lock] initialised as "true", instead of creating
+// a new object inline.
+// It's not needed to be used if the lock related flags are registered, since that provides the same default value
+// for the lock flag.
 func NewStateFlags() *State {
 	return &State{
 		Lock: true,
--- a/internal/command/arguments/extended.go
+++ b/internal/command/arguments/extended.go
@@ -74,8 +74,13 @@ func (s *State) AddFlags(f *flag.FlagSet, lockFlags bool, stateInPath bool, stat
 		f.StringVar(&s.StateOutPath, "state-out", "", "state-path")
 	}
 	if backupPath {
-		f.StringVar(&s.BackupPath, "backup", "", "backup-path")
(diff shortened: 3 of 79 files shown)


internal/command/meta_backend_test.go
```diff
@@ -2126,11 +2126,11 @@ func testMetaBackend(t *testing.T) *Meta {
 	var m Meta
 	view, _ := testView(t)
 	m.View = view
+	m.stateArgs = arguments.State{}
```
diofeher:
```suggestion
	m.stateArgs = arguments.State{Lock: true}
```

Missing from https://github.com/opentofu/opentofu/pull/4055/changes#diff-2ea40f9edb3d161001e053f3b0da8b678141b6638cb2afbfcd0d1f2f252d4febL2133

internal/command/arguments/extended.go
```diff
@@ -48,11 +49,38 @@ type State struct {
...
 	// be stored before the new state is written. The default value is blank,
-	// which is interpreted as StateOutPath +
-	// ".backup".
+	// which is interpreted as StateOutPath + ".backup" or, in some cases, the backup
+	// is skipped altogether.
 	BackupPath string
 }
 
+// AddFlags is the sole logic of registering the state related flags in OpenTofu.
+func (s *State) AddFlags(f *flag.FlagSet, lockFlags bool, stateInPath bool, stateOutPath bool, backupPath bool) {
```
diofeher: The thing I don't like about using parameters like that is that it makes the caller code harder to read. It would be easier to have:

```
# Parameter version (previously mentioned)
console.State.AddFlags(cmdFlags, true, false, false, false)

# Struct version
console.State.AddFlags(cmdFlags, StateArgs: {
  Lock: true
})
```

I'm not super strong about that, though.

internal/command/console.go
```diff
@@ -50,22 +50,13 @@ func (c *ConsoleCommand) Run(rawArgs []string) int {
...
 	// FIXME: the -input flag value is needed to initialize the backend and the
 	// operation, but there is no clear path to pass this value down, so we
 	// continue to mutate the Meta object state for now.
 	c.Meta.input = args.ViewOptions.InputEnabled
 
-	// TODO meta-refactor: when the stateLock and stateLockTimeout are extracted to be configured separately, remove
-	// these and use a common way to configure this
-	// The stateLock=true is here this way because this command used before meta.extendedFlagSet which did the same
-	// and left for the command to configure flags for this if needed.
-	c.Meta.stateLock = true
```
diofeher, line 67: I don't fully understand the comment (even though I reviewed that before, sorry). But code-wise, since we set stateLock to true for this command, shouldn't we have Lock: true as the default above?

internal/backend/cli.go
```diff
@@ -51,16 +52,12 @@ type CLIOpts struct {
...
-	//
-	// StateBackupPath is the local path where a backup file will be written.
-	// If this is empty, no backup will be taken.
-	StatePath       string
-	StateOutPath    string
-	StateBackupPath string
+	// StateArgs carries the state related configurations. Refer to the
+	// arguments.State for more details.
+	// In this particular case, an empty arguments.State#BackupPath will result
+	// in skipping the backup entirely and the new state will just ovewrite the old
```
diofeher:
```suggestion
	// in skipping the backup entirely and the new state will just overwrite the old
```

internal/command/arguments/extended.go
yottta: Happy that you opened the conversation. I went through multiple iterations of this, and ended up with this one just because it forces you to think clearly on what is needed and what not while gives a clear understanding what it offers and what out of its offerings its needed in each specific context.

Having an API similar with the one you suggested, bit me in the past, and this might be just "a me issue" 🤷 

Another approach that I still think that would make sense, is to remove this method altogether while providing only specific methods for each type of flag it supports. Something like:
```go
func (s *State) StateInFlag(f *flag.FlagSet)                                   { /*...*/ }
func (s *State) StateInFlagWithDefault(f *flag.FlagSet, def string)            { /*...*/ }
func (s *State) StateOutFlag(f *flag.FlagSet)                                  { /*...*/ }
func (s *State) StateOutFlagWithDefault(f *flag.FlagSet, def string)           { /*...*/ }
func (s *State) LockFlag(f *flag.FlagSet)                                      { /*...*/ }
func (s *State) LockTimeoutFlag(f *flag.FlagSet)                               { /*...*/ }
func (s *State) LockTimeoutFlagWithDefault(f *flag.FlagSet, def time.Duration) { /*...*/ }
func (s *State) BackupFlag(f *flag.FlagSet)                                    { /*...*/ }
func (s *State) BackupFlagWithDefault(f *flag.FlagSet, def string)             { /*...*/ }
```
But that would end up in a more of a verbose usage:
```go
	apply.State.LockFlag(cmdFlags)
	apply.State.LockTimeoutFlag(cmdFlags)
	apply.State.StateInFlag(cmdFlags)
	apply.State.StateOutFlag(cmdFlags)
	apply.State.BackupFlag(cmdFlags)
```

I agree that it's more readable this way, but in the same time, it's not a code path that is (or would be) extensively used.

WDYT?

internal/command/console.go
yottta: That's one thing that I wanted to tackle with these changes. Apparently, didn't make it clear enough in the PR description. Sorry for that.

The idea is that not the `command` package should set that specifically, but the `arguments` package should be responsible for this part.
And the way this PR changes things, it makes that default value for this to be somehow a side-effect.
What I try to say with this, is that the `arguments` implementation for each command, has 2 ways to provide the value for this particular attributes:
* Register the `-lock` flag that is configured directly [with a `true` default value](https://github.com/opentofu/opentofu/blob/978526b3e002f7df11ec72b41e2301a12082e033/internal/command/arguments/extended.go#L60C30-L60C34)
* If the command does not offer a way to control the locking behavior but it still relies on having the state locked, initialise the `arguments.State` with `Lock: true`

This way, the `Run` function in the `command` package, will just pass those values down the line.
At this point, we still do this by setting those values into the `Meta` structure (ie: `c.stateArgs = *args.State`), but in a future iteration, once we have the backend loading and state loading extracted in their independent components, this `arguments.State` (most probably together with others, like `arguments.Backend`) will be an argument of that backend/state component API.
I envision its signature to be looking something like this:
```go
	backendOpts := BackendOpts{
		Config:      backendConfig,
		View:        view.Backend(),
		StateArgs:   args.State,
		BackendArgs: args.Backend,
	}
	b, backendDiags := backend.Load(ctx, &backendOpts, enc.State())
```

---

And more specifically about that comment:
1. Looking back, this command used `Meta.extendedFlagSet` which had this `stateLock` [configured specifically](https://github.com/opentofu/opentofu/blob/v1.11/internal/command/meta.go#L659-L661)
2. When we did the refactor (for this particular command), we moved that flag to the `arguments` package, and configured that [manually here to not break things](https://github.com/opentofu/opentofu/pull/3764/changes#diff-5c5b5ae504fabf75611915b88827dd51226a04b3c4111fc5c53995af2e1bb308L35) (even though, arguably, it wasn't needed since the `-lock` flag with its `true` default value was already registered) 

This change wants to remove that tech debt.


IDK if this makes things clearer or more confusing, so please let me know about that.

internal/backend/cli.go
yottta: Fixed in 5ce4d25e60f1e73536603a36d5b4f880bb36730e.

internal/command/meta_backend_test.go
yottta: Fixed in 5ce4d25e60f1e73536603a36d5b4f880bb36730e

internal/command/arguments/extended.go
cam72cam: Another option is to do typed bools:

```go
type lockFlagOption bool

const lockFlagIncluded = lockFlagOption(true)
const lockFlagExcluded = lockFlagOption(false)

console.State.AddFlags(cmdFlags, logFlagIncluded, stateInPathExcluded, stateOutPathExcluded, backupPathExcluded)
```

A fourth option could be a bit mask

```go
type stateFlagOption int
const (
  lockFlagOption stateFlagOption = 1 << iota
  stateInPathOption
  stateOutPathOption
  backupPathOption
)

console.State.AddFlags(cmdFlags, lockFlagOption | backupPathOption)
```

(disregard poor naming choices)

yottta: Thanks for the suggetsions. I love the bit mask approach. Let me give it a shot! I missed that option.

yottta: Changed in c05d2a62be4260f4dfbe005ac0ca29be4f432eec

internal/command/arguments/login.go
```diff
@@ -29,9 +30,12 @@ func ParseLogin(args []string) (*Login, func(), tfdiags.Diagnostics) {
 		// Even though the command does not use the -var/-var-file content, we will keep this for the moment
 		// just to keep backwards compatibility for users (in case any of them are using these flags with this command)
 		Vars: &Vars{},
+		// State is only initialised and no flags are registered since the login command needs to lock the
+		// state by default, with no user input on that.
+		State: &State{Lock: true},
```
lbordowitz, line 35: Why does state need to be locked during login?

internal/command/state_push.go
```diff
@@ -55,11 +55,11 @@ func (c *StatePushCommand) Run(rawArgs []string) int {
 	}
 	// TODO meta-refactor: remove these assignments once we have a clear way to propagate these to the logic
 	//  that uses them
-	c.Meta.variableArgs = args.Vars.All()
-	c.stateLock = args.Backend.StateLock
-	c.stateLockTimeout = args.Backend.StateLockTimeout
 	c.ignoreRemoteVersion = args.Backend.IgnoreRemoteVersion
 
+	c.Meta.variableArgs = args.Vars.All()
+	c.stateArgs = *args.State
```
lbordowitz: I only noticed this here, but we're mixing usages of whether we refer to the embedded `Meta` struct or not. It is style consistency issue, though I'd just categorize this as a minor nit and also non-blocking.

internal/command/state_mv.go
```diff
@@ -146,18 +154,18 @@ func (c *StateMvCommand) Run(rawArgs []string) int {
 	stateToMgr := stateFromMgr
 	stateTo := stateFrom
 
-	if args.StateOutPath != "" {
-		c.statePath = args.StateOutPath
-		c.backupPath = args.BackupPathOut
+	if args.State.StateOutPath != "" {
+		c.stateArgs.StatePath = args.State.StateOutPath
+		c.stateArgs.BackupPath = args.BackupPathOut
```
lbordowitz, line 159: I'm not arguing you did this refactor incorrectly, I'm arguing about the logic of this block, so this is out-of-scope for this PR...

What even is the logic of `BackupPathOut` vs `BackupPath`? It just seems like a conflicting way of setting the backup path. I mean, I get `StatePath` vs `StateOutPath` because one talks about the state input and the other the state output... but there's no "backup input" like... what?

internal/command/arguments/init_test.go
```diff
@@ -218,65 +230,6 @@ func TestParseInit_backendCloudSynchronization(t *testing.T) {
 	}
 }
 
-func TestParseInit_backendFlags(t *testing.T) {
-	testCases := map[string]struct {
-		args        []string
-		wantBackend Backend
-	}{
-		"ignore-remote-version": {
```
lbordowitz, line 226: I understand removing the two state-based flag tests. But is removing the others really necessary? Consider restoring this test, albeit without the state-based flags cases.

internal/command/arguments/extended_test.go
```diff
@@ -0,0 +1,294 @@
...
+				t.Errorf("unexpected error (-want,+got)\n%s", diff)
+			}
+			if diff := cmp.Diff(tc.want, s, cmpOpts); diff != "" {
+				t.Errorf("unexpected result (-want,+got)\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestStateFlagsRegistering(t *testing.T) {
```
lbordowitz, line 192: I have mixed feelings about this test. On the one hand, it's useful to make sure that, when registering all flags, we ensure that when such a flag is set but unregistered it returns an error. On the other, this test kinda enforces an order in which the flags ought to be checked. If we ever change that order, this test would need to be changed, and that feels weird. There's nothing that necessarily needs to be changed, but I felt it shouldn't have gone unsaid.

internal/command/arguments/backend_test.go
```diff
@@ -58,65 +57,6 @@ func TestBackend_AddIgnoreRemoteVersionFlag(t *testing.T) {
...
-			wantLock:        true,
-			wantLockTimeout: 0,
-		},
-		"lock-timeout set": {
-			args:            []string{"-lock-timeout=10s"},
-			wantLock:        true,
-			wantLockTimeout: 10 * time.Second,
-		},
-		"lock-timeout set in minutes": {
-			args:            []string{"-lock-timeout=5m"},
```
lbordowitz, line 88: It doesn't appear that any tests currently check the lock-timeout flag with minutes; please add such a test, probably in the `extended_test.go`.

lbordowitz (commented):
A couple of testing and style nits, but overall an excellent refactor.

internal/command/arguments/login.go
yottta: Really good question.
I didn't put too much thinking into these details. I just wanted to keep the behavior as close to what it was before starting the whole CLI rework story.

---

But since you asked, I did some git archeological work on the history of the [`command/login.go`](https://github.com/opentofu/opentofu/commits/ffe056bacb9f2cf403fa3b6b894c5fbe1fa850a7/command/login.go?browsing_rename_history=true&new_path=internal/command/login.go&original_branch=1d3548a0e76cd207e235b60e40bffbd3984ebd6d).
To provide a little bit of context, the `Meta` struct had this `extendedFlagSet` method (removed recently by https://github.com/opentofu/opentofu/pull/3958). That method had this [default setup of locking the state](https://github.com/opentofu/opentofu/blob/0b346e589a27c296d650e39b5ab23a2080be66c6/command/meta.go#L391).
In the `command/login.go` history I found [this particular commit](https://github.com/opentofu/opentofu/commit/0b346e589a27c296d650e39b5ab23a2080be66c6#diff-0d69e07fda36f3d91e7f4fcb4f7f05d291ac363dce927afbec461777fc067e0eR19-R42) that changed the usage of `defaultFlagSet` to `extendedFlagSet`. From the looks of that commit, this was done purely for the `-input` argument and the state locking was just enabled without any side-effect.
I agree, that by looking at the history of the command and its internal logic, this seems truly unnecessary. Though, I will like not to do it as part of this changeset. I will take a note on this conversation and will revisit later.

If you disagree, please let me know.

internal/command/state_push.go
yottta: Changed in 39c055bbe8319642b30c18db7af06bc901c4dc50. Did it mostly for writes, not for reads.

internal/command/arguments/login.go
apparentlymart: Indeed, I don't know any reason why the login command should need to hold a state lock.

In particular: this command must be runnable before the backend is initialized, because it prepares credentials that `tofu init` might use during its work.

However, if the login command was handling these options before your change then I agree we could keep it for now and then fix this in a separate PR, so that we separate refactoring from behavior changes.

internal/command/state_mv.go
yottta: Isn't it? I banged my head against the walls with this too.
The way I managed to understand it, is that the state loading logic is mangled through different layers and the particular pain point here is that it relies on those state attributes from the `Meta` structure.

_In my attempt explain this clearly, I will describe this in terms of the flags and not on the `stateArgs` attributes._
So this command operates on 2 state files: the source and the target.
Source:
* Uses `-state` as state input
* Uses `-backup` as the backup for the state given in `-state`

Target:
* Uses `-state-out` as the state input
* Uses `-backup-out` as the backup for the state given in `-state-out`

So this particular block you highlighted can be read like:
> if the user specified a target state for the resources to be moved to, load that state too. To load that we need to initialise first the `Meta.stateArgs` attributes that are used by the state loading logic


Hopefully, this makes sense.

internal/command/arguments/init_test.go
yottta: Removed the whole method because the test cases overlapped with the other testing methods in [this file](https://github.com/opentofu/opentofu/blob/39c055bbe8319642b30c18db7af06bc901c4dc50/internal/command/arguments/backend_test.go).
_Still, I would like to rework that file fully one day, because it has some tests that overlap in checks_

If you still think that I should reinstate those, please let me know.

internal/command/arguments/backend_test.go
yottta: Integrated the idea in the changes added in f88e68e75bf2b6a48a7d5767294a0f2d22846467

internal/command/arguments/extended_test.go
yottta: Yeah, thanks. Didn't like it either 😓 
Fixed in f88e68e75bf2b6a48a7d5767294a0f2d22846467

internal/command/arguments/login.go
lbordowitz: Agreed. Put a pin in it, it's out-of-scope for this refactor. But let's not forget; please make an issue or something so it's being tracked.

yottta: I always have my notes for it. It doesn't look like an issue is necessary for such a small thing 🤷 
I might create an issue later, if we will identify more things like this.

internal/command/console.go
diofeher: Sorry, I missed this comment, but this makes sense. Thanks for explaining it so clearly!

diofeher (approved):

cam72cam (approved):

PR #4062, opened 2026-04-28
resources: Unit tests for ManagedResourceType.PlanChanges

apparentlymart:
While working through the "walking skeleton" phase of developing the new runtime we intentionally kept the test coverage pretty light so that it'd be easier to make substantial changes to the design if we learned new information as we went along.

Since we're now switching to a more incremental mode of development, it'll be helpful to have a little more test coverage. This is a retroactive addition of a unit test suite for the function that encapsulates the interactions with providers for planning changes to managed resource instances, since (FIXMEs and TODOs notwithstanding) this seems pretty unlikely to be refactored significantly in the near future.

This establishes a very simple fake provider client implementation that's convenient to use in the tests here. Along with just being much simpler than the one in `package tofu` (no call tracking, etc), including this here also avoids the need for `package resources` to depend on `package tofu`, which would therefore make it impossible for the old runtime to call into here. As we add new code to support other resource-related provider operations later the additional tests can hopefully share this fake provider client implementation.

This is for https://github.com/opentofu/opentofu/issues/3414, in preparation for ongoing development.

### DIFF
--- a/internal/resources/resources_test.go
+++ b/internal/resources/resources_test.go
@@ -1,0 +1,209 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resources
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/opentofu/opentofu/internal/providers"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// fakeProviderClient is an implementation of [providers.Interface] that just
+// delegates to various function pointers in its own fields, so that we can more
+// easily unit test the methods in this package that make calls to these
+// methods and then react to what they return.
+//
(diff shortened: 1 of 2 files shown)


cam72cam (approved):

internal/resources/resources_test.go
```diff
@@ -0,0 +1,209 @@
...
+		return providers.PlanResourceChangeResponse{
+			PlannedState: req.ProposedNewState,
+		}
+	}
+	return f.planResourceChange(ctx, req)
+}
+
+// ApplyResourceChange implements [providers.Interface].
+func (f *fakeProviderClient) ApplyResourceChange(ctx context.Context, req providers.ApplyResourceChangeRequest) providers.ApplyResourceChangeResponse {
+	if f.planResourceChange == nil {
```
diofeher: Shouldn't it be `applyResourceChange`?

apparentlymart: Yes, it should. There aren't any tests using this one yet but I'll fix it now anyway because it'll likely cause confusion later otherwise. 👍🏻

PR #4063, opened 2026-04-28
OCI Registries: Use separate auth tokens for each repository, rather than just one auth token per registry

### LINKED ISSUE #3316: OCI Registries: Use separate auth tokens for each repository, rather than just one auth token per registry
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.10.6
```

```hcl
module "eks" {
  source = "oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0"
  cluster = "example-eks"
}
```

### Debug Output

OpenTofu debug logs -

```
2025-09-29T17:00:25.576Z [TRACE] ModuleInstaller: eks address "oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0" will be handled by go-getter
Downloading oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0 for eks...
2025-09-29T17:00:25.576Z [TRACE] getmodules: fetching "oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0" to ".terraform/modules/eks"
2025-09-29T17:00:25.577Z [TRACE] OCI credentials discovery reading /home/vscode/.config/containers/auth.json
2025-09-29T17:00:25.577Z [TRACE] OCI credentials discovery reading /home/vscode/.docker/config.json
2025-09-29T17:00:25.578Z [TRACE] OCI credentials discovery reading /home/vscode/.dockercfg
2025-09-29T17:00:25.578Z [DEBUG] OCI registry client will consider credentials from /home/vscode/.docker/config.json
2025-09-29T17:00:25.579Z [TRACE] HTTP client GET request to https://harbor.example.com/v2/
2025-09-29T17:00:25.866Z [DEBUG] Executing docker-style credentials helper "dev-containers-c9219a27-feac-4e4e-8422-538330db1e63" for https://harbor.example.com
2025-09-29T17:00:26.327Z [ERROR] Docker-style credential helper "dev-containers-c9219a27-feac-4e4e-8422-538330db1e63" failed for https://harbor.example.com: exit status 1
2025-09-29T17:00:26.328Z [TRACE] HTTP client GET request to https://harbor.example.com/service/token?service=harbor-registry
2025-09-29T17:00:26.358Z [TRACE] modsdir: writing modules manifest to .terraform/modules/modules.json
╷
│ Error: Failed to download module
│ 
│   on main.tf line 1:
│    1: module "eks" {
│ 
│ Could not download module "eks" (main.tf:1) source code from "oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0": error downloading
│ 'oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0': configuring client for harbor.example.com/project/tfmod-aws-eks: failed to contact OCI
│ registry at "harbor.example.com": GET "https://harbor.example.com/v2/": GET
│ "https://harbor.example.com/service/token?service=harbor-registry": response status code 401: Unauthorized
╵
```

From Harbor logs
```
"GET /service/token?service=harbor-registry HTTP/1.1" 401 0 "-" "OpenTofu/1.10.6" 0.005 0.005 .
```

### Expected Behavior

Module is downloaded correctly as long as - 
1. harbor.example.com is accessible from the machine running opentofu, and
2. `project/tfmod-aws-eks` is a public repository which allows anonymous pulls 

### Actual Behavior

```
Could not download module "eks" (main.tf:1) source code from "oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0": error downloading 'oci://harbor.example.com/project/tfmod-aws-eks?tag=v1.0': configuring client for harbor.example.com/project/tfmod-aws-eks: failed to contact OCI registry at "harbor.example.com": GET "https://harbor.example.com/v2/": GET "https://harbor.example.com/service/token?service=harbor-registry": response status code 401: Unauthorized
```

### Steps to Reproduce

`tofu init`

### Additional Context

We have a private self-hosted Harbor registry running in our environment. This registry instance hosts multiple _public_ projects (public simply means projects that allow anonymous pulls, not that it is internet facing). 

All repositories within these internal _public_ projects are accessible anonymously. This setup works fine when pulling images. However, we get this 401 unauthorized error when trying to use a public project/repo to pull OpenTofu modules.

The following call to fetch the token succeeds, note the additional query parameters specifying `scope=repository/project/module`. This is absent in the OpenTofu call, and likely the cause of 401 errors.

```
curl "https://harbor.example.com/service/token?scope=repository%3Aproject%3Atfmod-aws-eks&service=harbor-registry" 
#
"GET /service/token?scope=repository% 3Aproject%3Atfmod-aws-eks&service=harbor-registry HTTP/1.1" 200 805 "-" "curl/8.5.0" 0.007 0.007 .
```

### References

_No response_

leinad87:
Resolves #3316 

It adds a test to make sure OpenTofu can use separate credentials for diferent repositories in the same registry. Also, removes extra `oras.Ping` request that returns 401 even when registry allos anonymous pulls.

## Checklist

### Go checklist

 

### DIFF
--- a/cmd/tofu/oci_distribution.go
+++ b/cmd/tofu/oci_distribution.go
@@ -105,11 +105,6 @@ func getOCIRepositoryStore(ctx context.Context, registryDomain, repositoryName s
 		return nil, err // This is only for registryDomain validation errors, and we should've caught those much earlier than here
 	}
 	reg.Client = client
-	err = reg.Ping(ctx) // tests whether the given domain refers to a valid OCI repository and will accept the credentials
-	if err != nil {
-		tracing.SetSpanError(span, err)
-		return nil, fmt.Errorf("failed to contact OCI registry at %q: %w", registryDomain, err)
-	}
 	repo, err := reg.Repository(ctx, repositoryName)
 	if err != nil {
 		tracing.SetSpanError(span, err)
--- a/cmd/tofu/oci_distribution.go
+++ b/cmd/tofu/oci_distribution.go
@@ -105,6 +105,14 @@ func getOCIRepositoryStore(ctx context.Context, registryDomain, repositoryName s
 		return nil, err // This is only for registryDomain validation errors, and we should've caught those much earlier than here
 	}
 	reg.Client = client
+	// NOTE: We intentionally do not call reg.Ping(ctx) here.
+	// The Ping method targets the GET /v2/ endpoint, which some registries
+	// respond to with a 401 challenge, even for repositories with an
+	// anonymous pull configuration, which is interpreted as an error
+	//
+	// Since ORAS handles authentication transparently on the first real
+	// API call, the Ping is both redundant and problematic for this type of
+	// registry.
 	repo, err := reg.Repository(ctx, repositoryName)
 	if err != nil {
 		tracing.SetSpanError(span, err)
--- a/cmd/tofu/oci_distribution_test.go
+++ b/cmd/tofu/oci_distribution_test.go
@@ -59,8 +59,6 @@ func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
 // Test that getOCIRepositoryORASClient uses the correct credentials for each repository
 // when a registry has multiple repositories with different credentials.
 func TestGetOCIRepositoryORASClient_PerRepositoryCredentials(t *testing.T) {
-	ctx := t.Context()
-
 	// Configure two repositories with different credentials for same registry
 	credsPolicy := ociauthconfig.NewCredentialsConfigs([]ociauthconfig.CredentialsConfig{
 		&testPerRepoCredentialsConfig{
@@ -73,24 +71,31 @@ func TestGetOCIRepositoryORASClient_PerRepositoryCredentials(t *testing.T) {
 	})
 
 	for _, id := range [...]string{"a", "b"} {
-		client, err := getOCIRepositoryORASClient(ctx, "registry.example.com", "repo-"+id, credsPolicy)
-		if err != nil {
-			t.Fatalf("creating client for repo-%s: %s", id, err)
-		}
+		t.Run(id, func(t *testing.T) {
+			ctx := t.Context()
+			client, err := getOCIRepositoryORASClient(ctx, "registry.example.com", "repo-"+id, credsPolicy)
 
-		credentials, err := client.Credential(ctx, "registry.example.com")
-		if err != nil {
-			t.Fatalf("resolving credential for repo-%s: %s", id, err)
-		}
+			if err != nil {
+				t.Fatalf("failed creating credential: %s", err)
+			}
 
-		if credentials.Username != "user-"+id {
-			t.Errorf("repo-%s: wrong username: got %s, want %s", id, credentials.Username, "user-"+id)
-		}
-		if credentials.Password != "password-"+id {
-			t.Errorf("repo-%s: wrong password: got %s, want %s", id, credentials.Password, "password-"+id)
-		}
+			credentials, err := client.Credential(ctx, "registry.example.com")
+			if err != nil {
+				t.Fatalf("failed resolving credential: %s", err)
+			}
+
+			want, got := "user-"+id, credentials.Username
+			if want != got {
+				t.Fatalf("username mismatch: %s != %s", got, want)
+			}
+
+			want, got = "password-"+id, credentials.Password
+			if want != got {
+				t.Fatalf("password mismatch: %s != %s", got, want)
+			}
+
+		})
 	}
-
 }
 
 // testPerRepoCredentialsConfig is a test-only implementation of
--- a/cmd/tofu/oci_distribution_test.go
+++ b/cmd/tofu/oci_distribution_test.go
@@ -13,6 +13,8 @@ import (
 	"testing"
 
 	"github.com/opentofu/opentofu/internal/command/cliconfig/ociauthconfig"
+	"github.com/opentofu/opentofu/internal/command/e2etest/fakeocireg"
+	orasRemote "oras.land/oras-go/v2/registry/remote"
 )
 
 func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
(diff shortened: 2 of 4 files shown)


cmd/tofu/oci_distribution_test.go
```diff
@@ -51,3 +55,65 @@ func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
...
+			t.Fatalf("creating client for repo-%s: %s", id, err)
+		}
+
+		credentials, err := client.Credential(ctx, "registry.example.com")
+		if err != nil {
+			t.Fatalf("resolving credential for repo-%s: %s", id, err)
+		}
+
+		if credentials.Username != "user-"+id {
+			t.Errorf("repo-a: wrong username: got %q, want %q", credentials.Username, "user-"+id)
```
diofeher: repo-a is hardcoded here

cmd/tofu/oci_distribution_test.go
```diff
@@ -51,3 +55,65 @@ func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
...
+		}
+		if credentials.Password != "password-"+id {
+			t.Errorf("repo-a: wrong password: got %q, want %q", credentials.Password, "password-"+id)
```
diofeher: repo-a is hardcoded here

cmd/tofu/oci_distribution.go
```diff
@@ -105,11 +105,6 @@ func getOCIRepositoryStore(ctx context.Context, registryDomain, repositoryName s
 		return nil, err // This is only for registryDomain validation errors, and we should've caught those much earlier than here
 	}
 	reg.Client = client
-	err = reg.Ping(ctx) // tests whether the given domain refers to a valid OCI repository and will accept the credentials
-	if err != nil {
-		tracing.SetSpanError(span, err)
-		return nil, fmt.Errorf("failed to contact OCI registry at %q: %w", registryDomain, err)
-	}
```
diofeher, line 112: Do you think it makes sense to turn that removal into a comment explaining why there's no need to check if it's a valid domain and call `reg.Repository` below?

I'm not super strong about this, but just thought that may give future readers some context

cmd/tofu/oci_distribution_test.go
```diff
@@ -51,3 +55,65 @@ func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
 		t.Errorf("wrong error\ngot: %s\nwant substring: %s", gotErr, wantErr)
 	}
 }
+
+// Test that getOCIRepositoryORASClient uses the correct credentials for each repository
+// when a registry has multiple repositories with different credentials.
+func TestGetOCIRepositoryORASClient_PerRepositoryCredentials(t *testing.T) {
+	ctx := t.Context()
+
+	// Configure two repositories with different credentials for same repository
```
diofeher:
```suggestion
	// Configure two repositories with different credentials for same registry
```

diofeher (changes requested):

cmd/tofu/oci_distribution.go
leinad87: yes, it make sense. I've just added some info at that point.

cmd/tofu/oci_distribution_test.go
```diff
@@ -51,3 +55,65 @@ func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
...
+			domain: "registry.example.com",
+			repos: map[string]ociauthconfig.Credentials{
+				"repo-a": ociauthconfig.NewBasicAuthCredentials("user-a", "password-a"),
+				"repo-b": ociauthconfig.NewBasicAuthCredentials("user-b", "password-b"),
+			},
+		},
+	})
+
+	for _, id := range [...]string{"a", "b"} {
+		client, err := getOCIRepositoryORASClient(ctx, "registry.example.com", "repo-"+id, credsPolicy)
```
yottta: The only thing that I would like to note, is that instead of having the id of the repository used everywhere in the error messages, I find it easier to follow if you spawn independent sub tests for each case.
In this particular situation, it would look something like:
```go
    for _, id := range [...]string{"a", "b"} {
        // ...
		t.Run(id, func(t *testing.T) {
			t.Fatalf("failed to resolve credential: %s", err)
		})
        // ...
	}
```

Not mandatory, but more of a suggestion.

leinad87: You are totally right! I'm not very fluent in go, I've just learn about subtests. Let me change it

cmd/tofu/oci_distribution.go
```diff
@@ -105,11 +105,14 @@ func getOCIRepositoryStore(ctx context.Context, registryDomain, repositoryName s
 		return nil, err // This is only for registryDomain validation errors, and we should've caught those much earlier than here
 	}
 	reg.Client = client
-	err = reg.Ping(ctx) // tests whether the given domain refers to a valid OCI repository and will accept the credentials
```
yottta, line 108: For the removal of this, can you add a new test that confirms the previous issue and which confirms also the current behavior?
The added test doesn't have much to do with the changes here.

leinad87: Do you have any idea to test `getOCIRepositoryStore`? May be mocking a http server that simulates a container registry is too ambitious.

apparentlymart: We do have a "mock" OCI Distribution implementation in `internal/command/e2etest/fakeocireg`, but it's focused on the needs of some of the tests in `internal/command/e2etest` rather than being for general use across all of our OCI-registry-related functionality.

`e2etest` _is_ where we usually test the glue code and wiring from `package main` though, so maybe we could cover this with a new end-to-end test in `internal/command/e2etest/provider_oci_mirrors_test.go`. We would unfortunately need to teach `fakeocireg` to be able to implement the authentication-related parts of the protocol though, which is non-trivial. I don't have a better idea... this part of the system is just generally hard to test. :confounded:

leinad87: We can add to `fakeocireg` the `GET /v2` request that returns 401 to simulate ACR, if someone reintroduce the `Ping` call.

The underlying question is: why do we need Ping here at all? Its only purpose is to fail fast on setup problems (e.g. unreachable registry, misconfigured auth). But in practice it produces false negatives against registries that respond to `GET /v2/` with a `401` challenge even when anonymous pulls are allowed — exactly the case that motivated removing it. Since ORAS already performs authentication transparently on the first real manifest/blob request, `Ping` is both redundant and actively harmful for these registries, so the early-failure benefit isn't worth the cost. From ORAS documentation:

> Ping checks whether or not the registry implement Docker Registry API V2 or OCI Distribution Specification. Ping can be used to check authentication when an auth client is configured.

apparentlymart: If I'm understanding the code correctly, it seems like the "ping" request is currently handled in a very minimal way by the `fakeocireg` implementation here:

https://github.com/opentofu/opentofu/blob/9e746cc2d159933d665cb8680819fccb99195531/internal/command/e2etest/fakeocireg/handler.go#L45-L50

Perhaps we could compromise by changing that to just immediately return status 500 with a comment above explaining that, although this endpoint exists in the OCI Distribution protocol, OpenTofu must not use it because it would cause an incorrectly-scoped auth token to be created.

That way we'd get a failing test if someone mistakenly reintroduces this request in future, but we wouldn't need to make an overly-complicated mock that covers a bunch of functionality that OpenTofu would never _really_ use in practice anyway.

I'd like to know if @yottta would consider that to be a good enough compromise before you actually implement it, though.

yottta: Yep, that sound good enough. Just want to be sure that we have this tested and documented properly for whoever will land on it in the future.

cmd/tofu/oci_distribution_test.go
```diff
@@ -51,3 +57,116 @@ func TestOCICredentialsLookupEnv_DockerCredHelper(t *testing.T) {
...
+}
+
+// TestGetOciRepository verifies that getOCIRepositoryStore can successfully
+// connect to an OCI registry that returns a 401 Unauthorized response for
+// the `GET /v2/` endpoint, which is a behavior seen with some registries.
+//
+// This test ensures that OpenTofu does not use the problematic `oras.Ping()`
+// function, which would fail against such registries. Instead, authentication
+// is handled transparently by ORAS on the first real API call.
+func TestGetOciRepository(t *testing.T) {
```
yottta:
```suggestion
func TestGetOciRepositoryNoPing(t *testing.T) {
```

yottta (approved):
Just a small nitpick, but other than that, looks good.

diofeher (approved):

apparentlymart:
Thanks, @leinad87!

leinad87:
> Thanks, @leinad87!

Thanks to the team for the support. This is my first contribution to an open-source project.

PR #4064, opened 2026-04-28
Save errored.tfstate during graph walk panic

### LINKED ISSUE #4027: Discussion on state file handling during panic
[original title "Filesystem statemgr does not persist statefile on panic"]

While chasing down #4019, I discovered that after https://github.com/opentofu/opentofu/pull/579/ we don't save terraform.tfstate during a panic for the local filesystem backend.

cam72cam:
Resolves #4027

## Checklist

### Go checklist

 

### DIFF
--- a/cmd/tofu/testing.go
+++ b/cmd/tofu/testing.go
@@ -1,0 +1,29 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package main
+
+// e2eTestingFeatures can be set to any non-empty string using Go linker
+// arguments in order to enable the use of e2e test features for a
+// particular OpenTofu build:
+//
+//	go install -ldflags="-X 'main.e2eTestingFeatures=yes'" ./cmd/tofu
+//
+// By default this variable is initialized as empty, in which case
+// experimental features are not available.
+//
+// The OpenTofu release process should arrange for this variable to be
+// set for alpha releases and development snapshots, but _not_ for
+// betas, release candidates, or final releases.
+//
+// (NOTE: Some experimental features predate the rule that experiments
+// are available only for alpha/dev builds, and so intentionally do not
+// make use of this setting to avoid retracting a previously-documented
+// open experiment.)
+var e2eTestingFeatures string
+
+func e2eTestingFeaturesEnabled() bool {
+	return e2eTestingFeatures != ""
+}
--- a/cmd/tofu/testing.go
+++ b/cmd/tofu/testing.go
@@ -12,16 +12,7 @@ package main
 //	go install -ldflags="-X 'main.e2eTestingFeatures=yes'" ./cmd/tofu
 //
 // By default this variable is initialized as empty, in which case
-// experimental features are not available.
-//
-// The OpenTofu release process should arrange for this variable to be
-// set for alpha releases and development snapshots, but _not_ for
-// betas, release candidates, or final releases.
-//
-// (NOTE: Some experimental features predate the rule that experiments
-// are available only for alpha/dev builds, and so intentionally do not
-// make use of this setting to avoid retracting a previously-documented
-// open experiment.)
+// e2e testing features are not available.
 var e2eTestingFeatures string
 
 func e2eTestingFeaturesEnabled() bool {
--- a/internal/command/apply_e2etesting.go
+++ b/internal/command/apply_e2etesting.go
@@ -1,0 +1,30 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package command
+
+import (
+	"os"
+
+	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/plans"
+	"github.com/opentofu/opentofu/internal/states"
+	"github.com/opentofu/opentofu/internal/tofu"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// e2eTestingApplyHooks simulates particularly nasty
+// scenarios within OpenTofu's apply engine, such
+// as panics due to programming errors
+type e2eTestingApplyHook struct {
+	tofu.NilHook
+}
+
+func (e *e2eTestingApplyHook) PreApply(addr addrs.AbsResourceInstance, gen states.Generation, action plans.Action, priorState, plannedNewState cty.Value) (tofu.HookAction, error) {
+	if resourceString := os.Getenv("TOFU_E2E_APPLY_RESOURCE_PANIC"); resourceString == addr.String() {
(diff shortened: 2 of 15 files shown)


cam72cam:
I'm still looking at how I might be able to add an E2E test for this.

internal/command/apply_e2etesting.go
```diff
@@ -0,0 +1,30 @@
...
+
+// e2eTestingApplyHooks simulates particularly nasty
+// scenarios within OpenTofu's apply engine, such
+// as panics due to programming errors
+type e2eTestingApplyHook struct {
+	tofu.NilHook
+}
+
+func (e *e2eTestingApplyHook) PreApply(addr addrs.AbsResourceInstance, gen states.Generation, action plans.Action, priorState, plannedNewState cty.Value) (tofu.HookAction, error) {
+	if resourceString := os.Getenv("TOFU_E2E_APPLY_RESOURCE_PANIC"); resourceString == addr.String() {
```
yottta, line 26: _About this and the wiring in `command/apply.go`_

I would change the env var name and purpose to opt-in different features in the future, something like `TOFU_E2E_OPERATION_EDITOR`.

After this, I would move the handling in a separate, singular place, where all these “e2e operation editors” will be stored and have everything encapsulated into a function that receives only the `Operation` pointer, checks the env var and based on that applies the editors:

```go
func applyE2EOpEditors(sysCfg system.Config, opReq *backend.Operation) {
	if !sysCfg.E2ETestingFeaturesEnabled {
		return
	}
	feats := strings.Split(os.Getenv("TOFU_E2E_OPERATION_EDITOR"), ",")
	for _, feat := range feats {
		switch {
		case strings.HasPrefix(feat, "apply:panichook:"):
			res := strings.TrimPrefix(feat, "apply:panichook:")
			opReq.Hooks = append(opReq.Hooks, &e2eTestingApplyHook{res: res})
		default:
			panic(fmt.Errorf("invalid feature provided: %s", feat))
		}
	}
}
```

cmd/tofu/testing.go
```diff
@@ -0,0 +1,29 @@
...
+// experimental features are not available.
+//
+// The OpenTofu release process should arrange for this variable to be
+// set for alpha releases and development snapshots, but _not_ for
+// betas, release candidates, or final releases.
+//
+// (NOTE: Some experimental features predate the rule that experiments
+// are available only for alpha/dev builds, and so intentionally do not
+// make use of this setting to avoid retracting a previously-documented
+// open experiment.)
```
apparentlymart: This part of the comment seems relevant only to the "experimental features enabled" flag that I assume you cribbed this comment from. 😀

internal/e2e/e2e.go
```diff
@@ -253,6 +253,7 @@ func GoBuild(pkgPath, tmpPrefix string) string {
 	args := []string{
 		"go",
 		"build",
+		`-ldflags=-X 'main.e2eTestingFeatures=yes'`,
```
apparentlymart: Having this implemented down at this level seems unfortunate since it means we're making it even harder for us to run the e2etests against real release binaries in future, should we choose to do that again. (It's what the e2etest suite was originally made for.)

We don't have any specific plans to use the e2etests as part of the release process right now and so I can accept this if the alternatives are all significantly more complicated, but I do have a possible compromise I'd like to offer for consideration:

- Change this `GoBuild` function so that it has a new `buildArgs ...string` variadic argument whose items get appended to the `go build` command line. Most callers would not provide any additional arguments, thereby retaining the previous behavior.
- In the new tests _specifically about this feature_, call `GoBuild` inline with the additional `ldflags` argument instead of using the main executable that is built in `TestMain`, and pass _that_ executable to the `NewBinary` function instead.

    (Note that there are already several tests that call `GoBuild` inline for one reason or another, so this is a relatively well-trodden path although admittedly not previously for the `tofu` executable in particular.)
- Make the tests that rely on a specially-configured build of `tofu` check `canRunGoBuild` and skip if it isn't set:

    https://github.com/opentofu/opentofu/blob/2334579470dcc3eaae36b6711f9cb635b78ade37/internal/command/e2etest/main_test.go#L19-L29

    ...which will then mean that, as with the various other tests that already run `e2e.GoBuild` inline, these tests would be skipped if the e2etest suite is run against a pre-built real release built instead of an executable built dynamically by `TestMain`.

    (The reason this exists is that in the "test a real release build" mode the suite is designed to run in an environment where the Go toolchain isn't available at all, with the assumption that the toolchain was run in an earlier step that actually built the `tofu` release executables that are now being tested.)

This approach would make it easier for us to reinstate e2etest as part of the release process later if we choose to, and I think achieves that with only a moderate amount of additional complexity that'd be self-contained only inside the tests related to this feature.

cmd/tofu/testing.go
cam72cam: Yep, I got part way through keeping the relevant portions and must have gotten distracted. I'll do another pass.

internal/command/apply_e2etesting.go
cam72cam: I like keeping this simple and explicit for now, we can introduce this complexity later if needed.

internal/e2e/e2e.go
cam72cam: Done in [61c04be](https://github.com/opentofu/opentofu/pull/4064/commits/61c04bedd411ae282d1582d7e9fb8ee26663ae8a)

cmd/tofu/testing.go
cam72cam: Fixed in [61c04be](https://github.com/opentofu/opentofu/pull/4064/commits/61c04bedd411ae282d1582d7e9fb8ee26663ae8a)

yottta (approved):

internal/command/e2etest/testdata/apply-panic/main.tf
```diff
@@ -0,0 +1,17 @@
+// the provider-plugin tests uses the -plugin-cache flag so terraform pulls the
+// test binaries instead of reaching out to the registry.
```
apparentlymart:
```suggestion
// the provider-plugin tests uses the -plugin-dir flag so OpenTofu pulls the
// test binaries instead of reaching out to the registry.
```

internal/command/e2etest/main_test.go
```diff
@@ -50,6 +50,7 @@ func setup() func() {
 		return func() {}
 	}
 
+	// Build with e2e testing features turned on
```
apparentlymart: I don't think this comment is true anymore, right? This is the normal build and then `TestApplyPanic` makes its own build with these features turned on.

apparentlymart (approved):

internal/command/e2etest/testdata/apply-panic/main.tf
cam72cam: [898b604](https://github.com/opentofu/opentofu/pull/4064/commits/898b604c2a06eb621fbaf9501638d328031d3511)

internal/command/e2etest/main_test.go
cam72cam: [898b604](https://github.com/opentofu/opentofu/pull/4064/commits/898b604c2a06eb621fbaf9501638d328031d3511)

Issue #3951, opened 2026-03-26
Adopt Unicode 17


apparentlymart:
We expect that the OpenTofu v1.13.x series will be built using Go 1.27, and Go 1.27 is adopting Unicode 17 so we'll need to deal with some of the interactions described in [How OpenTofu Uses Unicode](https://github.com/opentofu/opentofu/blob/866b067c0daf18165ced10879992a1d49d3de8df/docs/unicode.md).

This is mainly about updating a few of our dependencies to versions that use Unicode 17, once upstream support is available:

- We indirectly depend on [go-textseg](https://github.com/apparentlymart/go-textseg), which has some initial support for Unicode 16 but [it's currently incomplete](https://github.com/apparentlymart/go-textseg/issues/6), and does not yet support Unicode 17.

    (The grapheme cluster part of the text segmentation annex has not been significantly changed between Unicode 16 and 17, so once the support for the Unicode 16 algorithms is complete adopting 17 should just be a matter of regenerating based on the updated character tables.)
- `github.com/zclconf/go-cty` depends on `go-textseg` and uses it for some of the string-manipulation functions that OpenTofu exposes in its language. At the time of writing there is not yet a cty release that uses Unicode 17.
-  `github.com/hashicorp/hcl` uses the Unicode definition of "identifier" based on the `ID_Start` and `ID_Continue` properties in the Unicode character database. The set of characters with those properties may have changed in Unicode 17.

    HCL also depends on `go-textseg` and uses it for things such as counting how many "columns" into a line a source range is. At the time of writing there is not yet an upstream HCL release that uses Unicode 17.

    (The OpenTofu project has its own fork of HCL anyway, so we could incorporate these updates locally if we need to.)

(Note that Go got a little "stuck" on Unicode 15 for a while and so is effectively skipping over Unicode 16. That is echoed in the upstream dependencies mentioned here, since they also typically follow what Go is supporting in the standard library.)

princebirring:
I'd like to work on this issue. I'll submit a PR soon.

apparentlymart:
Hi @princebirring! Thanks for starting this. 

If you'd be willing, I think it would be better to try to submit a change to upstream HCL (the HashiCorp repository) first, since we try to keep our fork as close to upstream as possible.

Of course it's up to them whether they'd accept a change like this, but hopefully they would be since other software that uses HCL would also benefit.

apparentlymart:
Some progress on this:

[...]

PR #4067, opened 2026-04-28
Add support for instances in test resource overrides

### LINKED ISSUE #3264: Enhanced testing support for resources with for_each and count

lbordowitz:
Resolves #3264 

This change introduces an `OverrideTrie` concept which is used to either identify individual resource instances or a coarse ranges of resource instances relevant to a particular set of value overrides. It also introduces a novel use of the splat expression: you can "select" against all instances of a targeted resource and have that instance key be respected. This might be relevant to imports and moves; I've kept OverrideTrie generic in case it's useful elsewhere, and I'm open to renaming it should the need arise.

A couple of notes that may make this easier to review:

- I've split this PR into a couple of commits to make it easier to review.
  - The actual override logic is mostly implemented in `node_resource_abstract_instance.go`. This is where the magic happens.
- This uses an [unmerged version of HCL](https://github.com/opentofu/hcl/compare/opentofu...f-splat-as-traversal). It has `AbsTraversalPatternForExpr`, which enables using wildcard expression in resource and module indexes. Big, big thanks to Martin for implementing this; I might know where to begin, but certainly not how it ought to work.
- A good chunk of the override trie code is written specifically to test for a backwards-compatibility feature. Currently, overrides can be like `module.a.resource.element` and that overrides `module.a["x"].resource.element["iron"]` and `module.a.["y"].resource.element["copper"], etc. However, we don't want to mix syntax: if someone's using a wildcard, they should know better, and they ought to specify every range using a wildcard rather than relying on the "old" (i.e. current) way of managing all resources.

Additionally, there are a couple `TODO`s still in there. Those are relevant, and they are open questions I'm hoping to hash out through the review process. Another question: should I engineer the `OverrideTrie` to be safe for concurrent usage? I'm not totally sure what the execution model looks like when it enters the `getProvider` function of `node_resource_abstract_instance.go`. I'm hoping to learn a little more through this PR!

## Checklist

### Go checklist

 

### DIFF
--- a/internal/addrs/override_trie.go
+++ b/internal/addrs/override_trie.go
@@ -59,7 +59,7 @@ func NewOverrideTrie[T any]() *OverrideTrie[T] {
 // containing the value for the address.
 //
 // The "NoKey" key is treated specially and made equivalent to WildCard. This is
-// to provide backwards compatibility; before this was implemented, a non-instanced
+// to provide backwards compatibility; before this was implemented, an unkeyed
 // resource address was used to refer to every instance associated with the address.
 //
 // val is expected to be non-nil; it might complicate overrides if no value
@@ -72,29 +72,29 @@ func (ot *OverrideTrie[T]) Set(addr *AbsResourceInstance, val T, src *hcl.Range)
 	for i, mod := range addr.Module {
 		next, usesNoKey := ot.subSet(current, mod.InstanceKey)
 		if usesNoKey {
-			ot.SetNoKeyEvidence(i, addr, src)
+			ot.setNoKeyEvidence(i, addr, src)
 		}
-		ot.TrackModernAddressing(mod.InstanceKey)
+		ot.trackModernAddressing(mod.InstanceKey)
 		current = next
 	}
 	last, usesNoKey := ot.subSet(current, addr.Resource.Key)
 	if usesNoKey {
-		ot.SetNoKeyEvidence(len(addr.Module), addr, src)
+		ot.setNoKeyEvidence(len(addr.Module), addr, src)
 	}
-	ot.TrackModernAddressing(addr.Resource.Key)
+	ot.trackModernAddressing(addr.Resource.Key)
 	last.value = new(val)
 }
 
-func (ot *OverrideTrie[T]) TrackModernAddressing(key InstanceKey) {
+func (ot *OverrideTrie[T]) trackModernAddressing(key InstanceKey) {
 	_, usesWildcard := key.(WildcardKey)
 	ot.usesModernAddresses = ot.usesModernAddresses || usesWildcard
 }
 
-// SetNoKeyEvidence provides evidence that the NoKey instance key was used in a
+// setNoKeyEvidence provides evidence that the NoKey instance key was used in a
 // particular resource override. This is later used when getting a key; if
 // this trie uses modern address syntax, but no key is used when a key is
 // called for, this is how we obtain that evidence.
-func (ot *OverrideTrie[T]) SetNoKeyEvidence(i int, addr *AbsResourceInstance, src *hcl.Range) {
+func (ot *OverrideTrie[T]) setNoKeyEvidence(i int, addr *AbsResourceInstance, src *hcl.Range) {
 	if ot.noKeyEvidenceMap == nil {
 		ot.noKeyEvidenceMap = make([][]*hcl.Range, len(addr.Module)+1)
 		for i := range len(addr.Module) + 1 {
@@ -118,8 +118,8 @@ func (ot *OverrideTrie[T]) subSet(current *OverrideTrie[T], key InstanceKey) (*O
 	}
 	next, ok := current.trie[key]
 	if !ok {
-		current.trie[key] = NewOverrideTrie[T]()
-		next = current.trie[key]
+		next = NewOverrideTrie[T]()
+		current.trie[key] = next
 		next.root = ot
 	}
 	return next, usesNoKey
@@ -199,8 +199,8 @@ func (ot *OverrideTrie[T]) checkKey(i int, key InstanceKey, addrString string) t
 		for _, noKeyRange := range ot.noKeyEvidenceMap[i] {
 			diags = diags.Append(&hcl.Diagnostic{
 				Severity: hcl.DiagError,
-				Summary:  "Please switch overrides to wildcard syntax (i.e. \"[*]\") to refer to for-each resources",
-				Detail:   fmt.Sprintf("When trying to override %s", addrString),
+				Summary:  "Invalid resource override usage",
+				Detail:   fmt.Sprintf("The mixed usage of un-keyed and wildcard references in an override is not allowed, but detected in %q. Switch overrides to wildcard syntax (i.e. \"[*]\" to refer to for-each or count resources.)", addrString),
 				Subject:  noKeyRange,
 			})
 		}
--- a/internal/addrs/override_trie.go
+++ b/internal/addrs/override_trie.go
@@ -1,0 +1,196 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+package addrs
+
+import (
+	"fmt"
+
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// OverrideTrie provides a step-wise method to obtain values
+// for overridden resource. Each instance of the OverrideTrie
+// represents one "hop" in the resource address, traversing
+// through modules.
+//
+// A resource without a key, which is expected to have a key, may use
+// either the Wildcard key "[*]" or no key at all, but not both. See Set
+// and Get for more specific information on how that is handled.
+type OverrideTrie[T any] struct {
+	trie  map[InstanceKey]*OverrideTrie[T]
+	value *T
+
+	// usesModernAddresses is used in the root trie to track
+	// whether a wildcard literal was ever used for the
+	// index of an override
+	usesModernAddresses bool
+
+	// noKeyEvidenceMap, for each step, provides a list of address ranges used in overrides
+	// where, at that step of the process, the instance key was NoKey. This is used
+	// for error handling to provide user feedback on which addresses to fix.
+	noKeyEvidenceMap [][]*AbsResourceInstance
+
+	root *OverrideTrie[T]
+}
+
+// NewOverrideTrie creates a new trie for mapping override values to addresses.
+//
+// A default value is provided in this constructor.
+// This is used when the address is not found in the trie, that is,
+// it is not set as an override.
+func NewOverrideTrie[T any]() *OverrideTrie[T] {
+	return &OverrideTrie[T]{
+		trie:  make(map[InstanceKey]*OverrideTrie[T]),
+		value: nil,
+	}
+}
+
+// Set takes an address and value and loads it into the OverrideTrie. Each
+// instance key, for a module or resource, creates a trie, with the "leaf" trie
+// containing the value for the address.
+//
+// The "NoKey" key is treated specially and made equivalent to WildCard. This is
+// to provide backwards compatibility; before this was implemented, a non-instanced
+// resource address was used to refer to every instance associated with the address.
+//
+// val is expected to be non-nil; it might complicate overrides if no value
+// is provided but the resource is still considered "overridden"
+func (ot *OverrideTrie[T]) Set(addr *AbsResourceInstance, val T) {
+	current := ot
+	for i, mod := range addr.Module {
+		next, usesNoKey := ot.subSet(current, mod.InstanceKey)
+		if usesNoKey {
+			ot.SetNoKeyEvidence(i, addr)
+		}
+		ot.TrackModernAddressing(mod.InstanceKey)
+		current = next
+	}
+	last, usesNoKey := ot.subSet(current, addr.Resource.Key)
+	if usesNoKey {
+		ot.SetNoKeyEvidence(len(addr.Module), addr)
+	}
+	ot.TrackModernAddressing(addr.Resource.Key)
+	last.value = new(val)
+}
+
+func (ot *OverrideTrie[T]) TrackModernAddressing(key InstanceKey) {
+	_, usesWildcard := key.(WildcardKey)
+	ot.usesModernAddresses = ot.usesModernAddresses || usesWildcard
+}
+
+// SetNoKeyEvidence provides evidence that the NoKey instance key was used in a
+// particular resource override. This is later used when getting a key; if
+// this trie uses modern address syntax, but no key is used when a key is
+// called for, this is how we obtain that evidence.
+func (ot *OverrideTrie[T]) SetNoKeyEvidence(i int, addr *AbsResourceInstance) {
+	if ot.noKeyEvidenceMap == nil {
+		ot.noKeyEvidenceMap = make([][]*AbsResourceInstance, len(addr.Module)+1)
+		for i := range len(addr.Module) + 1 {
+			ot.noKeyEvidenceMap[i] = make([]*AbsResourceInstance, 0)
+		}
+	}
+	ot.noKeyEvidenceMap[i] = append(ot.noKeyEvidenceMap[i], addr)
+}
+
+// subSet prepares one step in the module or resource chain
+// as a trie of instance keys. It also provides evidence that a NoKey
+// was used in one step of the override; if, during Get, a key is used,
+// and the OverrideTrie is using "Modern Addresses Ranges" (i.e.
+// override addresses with wildcards), we can return an error and call
+// out the particular address.
+func (ot *OverrideTrie[T]) subSet(current *OverrideTrie[T], key InstanceKey) (*OverrideTrie[T], bool) {
+	usesNoKey := false
+	if key == NoKey {
+		key = WildcardKey{UnknownKeyType}
+		usesNoKey = true
+	}
+	next, ok := current.trie[key]
+	if !ok {
+		current.trie[key] = NewOverrideTrie[T]()
+		next = current.trie[key]
+		next.root = ot
+	}
+	return next, usesNoKey
+}
+
+// Get returns the value in the OverrideTrie associated with the address. If part of the
+// address is not found, but a WildCard address is set in the trie, that sub-trie is
+// recursively searched against the query.
+//
+// If the address could not be found in the OverrideTrie as an override, Get returns nil.
+//
+// A wildcard instance key anywhere in the provided address will produce an error; it
+// make sense to use this to obtain a single value when referencing a wildcard. Additionally,
+// if the override addresses had no key in a module or resource where a key was expected,
+// this method will also produce an error for that.
+func (ot *OverrideTrie[T]) Get(addr *AbsResourceInstance) (*T, tfdiags.Diagnostics) {
+	n := len(addr.Module) + 1
+	keyList := make([]InstanceKey, n)
+	for i, mod := range addr.Module {
+		keyList[i] = mod.InstanceKey
+	}
+	keyList[n-1] = addr.Resource.Key
+
+	value, diags := ot.recursiveGet(0, n, keyList, addr.String())
+	return value, diags
+}
+func (ot *OverrideTrie[T]) recursiveGet(i, n int, keyList []InstanceKey, addrString string) (*T, tfdiags.Diagnostics) {
+	if ot == nil {
+		return nil, nil
+	}
+	if i == n {
+		return ot.value, nil
+	}
+	keyCheckDiags := ot.checkKey(i, keyList[i], addrString)
+	if keyCheckDiags.HasErrors() {
+		return nil, keyCheckDiags
+	}
+	value, diags := ot.trie[keyList[i]].recursiveGet(i+1, n, keyList, addrString)
+	if diags.HasErrors() {
+		return nil, diags
+	}
+	if value != nil {
+		return value, diags
+	}
+
+	// Could not find against concrete instance key, try on Wildcard key
+	return ot.trie[WildcardKey{UnknownKeyType}].recursiveGet(i+1, n, keyList, addrString)
+}
+
+func (ot *OverrideTrie[T]) checkKey(i int, key InstanceKey, addrString string) tfdiags.Diagnostics {
+	if ot.root != nil && ot.root != ot {
+		return ot.root.checkKey(i, key, addrString)
+	}
+	if _, usesWildcard := key.(WildcardKey); usesWildcard {
+		return tfdiags.Diagnostics{
+			tfdiags.Sourceless(
+				tfdiags.Error,
+				"Wildcard key not expected in when retrieving override values",
+				fmt.Sprintf("In the provided resource address \"%s\", a wildcard accessor is used for the instance key. A specific key should always be used, where applicable.", addrString),
+			),
+		}
+	}
+	if ot.noKeyEvidenceMap == nil || !ot.usesModernAddresses {
+		return nil
+	}
+	// check if NoKey is being used in a place it shouldn't
+	// i.e. this key isn't NoKey, but the override was NoKey
+	// at this step
+	var diags tfdiags.Diagnostics
+	if key != NoKey && len(ot.noKeyEvidenceMap[i]) > 0 {
+		for _, noKeyAddr := range ot.noKeyEvidenceMap[i] {
+			// TODO this results in a crazy amount of diagnostics...
+			// Like, for every instance of every instance of every instance, and every override therein,
+			// has an error output example. How do I avoid this? Hash on AbsResource or something?
+			diags = diags.Append(tfdiags.Sourceless(
+				tfdiags.Error,
+				// TODO I'm open to a less-verbose version of this summary
+				"The override address cannot contain unkeyed for-each resources if it is also using the wildcard syntax (i.e. \"[*]\"). Please switch entirely to wildcard syntax for test overrides.",
+				fmt.Sprintf("Using %s to override %s", noKeyAddr.String(), addrString),
+			))
+		}
+	}
+	return diags
+}
--- a/internal/addrs/override_trie.go
+++ b/internal/addrs/override_trie.go
@@ -7,6 +7,7 @@ package addrs
 import (
 	"fmt"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/tfdiags"
 )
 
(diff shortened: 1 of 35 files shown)


cam72cam:
> should I engineer the OverrideTrie to be safe for concurrent usage?

At this point all of the "writes" in the configs package are handled within a single go-routine and are completed before any of your "reads" happen in other areas of OpenTofu.

The only other "write" is contained within a function that does not spawn any routines for processing.

For now, I think leave it as is. If you are concerned, add a comment about future concurrent usage in the docstring.

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,202 @@
...
+
+	root *OverrideTrie[T]
+}
+
+// NewOverrideTrie creates a new trie for mapping override values to addresses.
+//
+// A default value is provided in this constructor.
+// This is used when the address is not found in the trie, that is,
+// it is not set as an override.
+func NewOverrideTrie[T any](defaultVal T) *OverrideTrie[T] {
```
cam72cam: I'm in favor of getting rid of the defaultValue and assuming consumers will use the boolean returned in Get() below.

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3218,19 +3218,48 @@ func (n *NodeAbstractResourceInstance) getProvider(ctx context.Context, evalCtx
 			}
 		}
 
+		trie := addrs.NewOverrideTrie(overrideValues)
 		// Overridden by the provider (overrides mocks)
 		for _, res := range n.ResolvedProvider.OverrideResources {
-			if res.TargetParsed.Equal(n.Addr.ConfigResource()) && res.Mode == n.Addr.Resource.Resource.Mode {
-				overrideValues = res.Values
-				break
+			if res.TargetParsed.AffectedAbsResource().Equal(n.Addr.AffectedAbsResource()) && res.Mode == n.Addr.AffectedAbsResource().Resource.Mode {
```
cam72cam, line 3224: I'm 92% sure that we don't need the mode check here as it's included in the AffectedAbsResource.

cam72cam: Though now that I look at it, it's not a new check in this PR.

internal/addrs/override_trie.go
lbordowitz: I think I must've hallucinated default values, because they're just not anywhere in the code.

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,196 @@
...
+		for _, noKeyAddr := range ot.noKeyEvidenceMap[i] {
+			// TODO this results in a crazy amount of diagnostics...
+			// Like, for every instance of every instance of every instance, and every override therein,
+			// has an error output example. How do I avoid this? Hash on AbsResource or something?
+			diags = diags.Append(tfdiags.Sourceless(
+				tfdiags.Error,
+				// TODO I'm open to a less-verbose version of this summary
+				"The override address cannot contain unkeyed for-each resources if it is also using the wildcard syntax (i.e. \"[*]\"). Please switch entirely to wildcard syntax for test overrides.",
+				fmt.Sprintf("Using %s to override %s", noKeyAddr.String(), addrString),
+			))
```
apparentlymart: I don't have an immediate great answer for you, but for similar situations in the past the solution has involved moving the diagnostics handling upstream somewhere, by e.g. having this function return an `error` (or some other machine-understandable signal) capturing the context needed to produce the diagnostic and then making some caller aggregate them and produce only one diagnostic per distinct problem.

I notice this diagnostic also doesn't include source location information about the expression that caused the problem, and maybe having the diagnostics produced upstream in a more central place would also mean the diagnostics can be generated in a place that knows the source location for each override? 🤔 

(I must admit I'm not sure exactly what "central place" would make sense here. I'm sorry this is not a more complete suggestion.)

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"testing"
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+)
+
+// parseAbsResourceRangeStr is copied from ParseAbsResourceInstanceStr (more or less),
+// but with functions edited to refer to ranges instead.
+// TODO should this be outside of the test file?
```
apparentlymart: We do have an existing pattern of having both `ParseSomething(hcl.Traversal)` and `ParseSomethingStr(string)`, but I'd agree it's debatable whether we should proactively add `Str` version of this one in the exported API until we know we actually need it outside of tests.

But I think I _would_ choose to place this unexported version of the function in the non-test file alongside the `ParseAbsResourceRange` function because if I were considering adding a `ParseAbsResourceRangeStr` function for some new reason in future I doubt I would think to look in here to see if an implementation were already written, but I'd be much more likely to find it if it were right next to the `ParseAbsResourceRange` function, where I'd presumably try to write the exported function anyway.

(This is an opinion very loosely held.)

apparentlymart (commented):
This looks good to me! The new data structure in `addrs` is a nice way to solve this.

It seems like we need to figure out what we're going to do with that HCL change: try to submit it upstream, or just immediately include it in our fork?

I would actually suggest we do both concurrently: put it in our fork for now but also open an PR upstream to see if they are interested. If they are willing to accept something like it but want some different API details then we can always update our callers to use the final API later, since nothing outside of the OpenTofu codebases should be directly depending on our fork.

internal/addrs/override_trie.go
lbordowitz: I have a couple of potential solutions:

- Add a "fuse" to the trie. When the fuse is "tripped", don't run any more queries, just return nil with no errors and let the one error speak for itself.
- Since the problem occurs for the override addresses, perhaps we could obtain source locations during `Set`. Those happen in `config`, and I can definitely get the source range of the target from there.

These are aimed at the resource situation. Neither of these helps for the case of overrides in providers; I'll think on that a bit more.

lbordowitz:
Note: the newly introduced `refuse` property in the `OverrideTrie` does introduce a race condition. If multiple instances of a resource are doing a `Get`, it could happen that one of them triggers `ot.refuse = true` while at the same time another is checking `ot.refuse` for whether or not to skip this `Get`.

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+// Set takes an address and value and loads it into the OverrideTrie. Each
+// instance key, for a module or resource, creates a trie, with the "leaf" trie
+// containing the value for the address.
+//
+// The "NoKey" key is treated specially and made equivalent to WildCard. This is
+// to provide backwards compatibility; before this was implemented, a non-instanced
+// resource address was used to refer to every instance associated with the address.
```
yottta:
```suggestion
// to provide backwards compatibility; before this was implemented, an unkeyed
// resource address was used to refer to every instance associated with the address.
```
Sounds better to my ear, but hey, I might be wrong so push back if you disagree.

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,209 @@
...
+	usesNoKey := false
+	if key == NoKey {
+		key = WildcardKey{UnknownKeyType}
+		usesNoKey = true
+	}
+	next, ok := current.trie[key]
+	if !ok {
+		current.trie[key] = NewOverrideTrie[T]()
+		next = current.trie[key]
+		next.root = ot
```
yottta:
```suggestion
		next = NewOverrideTrie[T]()
		current.trie[key] = next
		next.root = ot
```

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,209 @@
...
+	// Note: the Swap at the end prevents this error from being returned
+	// more than once per resource. We rely on logic shortcutting for it
+	// to not be executed when the first two conditions are not met.
+	var diags tfdiags.Diagnostics
+	if key != NoKey && len(ot.noKeyEvidenceMap[i]) > 0 && !ot.refuse.Swap(true) {
+		for _, noKeyRange := range ot.noKeyEvidenceMap[i] {
+			diags = diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Please switch overrides to wildcard syntax (i.e. \"[*]\") to refer to for-each resources",
+				Detail:   fmt.Sprintf("When trying to override %s", addrString),
```
yottta:
```suggestion
				Summary:  "Invalid resource override usage",
				Detail:   fmt.Sprintf("The mixed usage of un-keyed and wildcard references in an override is not allowed but detected in %q. Switch overrides to wildcard syntax (i.e. \"[*]\") to refer to for-each resources.", addrString),
```

I am not sure about this, but I tend to like shorter and more clear summaries, while the details contain more information.
Here's the [general guidance](https://github.com/opentofu/opentofu/blob/main/docs/diagnostics.md#diagnostic-description-writing-style) on the agreed writing style of the diagnostics.

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	last, usesNoKey := ot.subSet(current, addr.Resource.Key)
+	if usesNoKey {
+		ot.SetNoKeyEvidence(len(addr.Module), addr, src)
+	}
+	ot.TrackModernAddressing(addr.Resource.Key)
+	last.value = new(val)
+}
+
+func (ot *OverrideTrie[T]) TrackModernAddressing(key InstanceKey) {
```
yottta: This looks to be part of the internal api of the trie. Do we want to export it?

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,209 @@
...
+	_, usesWildcard := key.(WildcardKey)
+	ot.usesModernAddresses = ot.usesModernAddresses || usesWildcard
+}
+
+// SetNoKeyEvidence provides evidence that the NoKey instance key was used in a
+// particular resource override. This is later used when getting a key; if
+// this trie uses modern address syntax, but no key is used when a key is
+// called for, this is how we obtain that evidence.
+func (ot *OverrideTrie[T]) SetNoKeyEvidence(i int, addr *AbsResourceInstance, src *hcl.Range) {
```
yottta: This looks to be part of the internal api of the trie. Do we want to export it?

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,127 @@
...
+// SPDX-License-Identifier: MPL-2.0
+package addrs
+
+import (
+	"testing"
+)
+
+// shorthand function to obtain known good strings.
+// Please please please do not use this outside of tests!!!
+func getAbsResourceRangeOrPanic(str string) AbsResourceInstance {
```
yottta:
```suggestion
func mustAbsResourceRange(str string) AbsResourceInstance {
```
This is a quite common pattern in Go. [Here is one of its usage in the stdlib](https://pkg.go.dev/text/template#Must).

Even more, there is a quite extensive use of this pattern in the `addrs` package, so you could export this in case it's needed to be used later in other packages. `grep -ri "func must" internal/addrs/*.go`

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,127 @@
...
+	out, diags := parseAbsResourceRangeStr(str)
+	if diags.HasErrors() {
+		panic(diags)
+	}
+	return out
+}
+
+type override struct {
+	Address *AbsResourceInstance
+	Values  string
```
yottta:
```suggestion
	Value  string
```

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,127 @@
...
+		{
+			TestName: "error on wildcard",
+			Overrides: []override{
+				{
+					Address: new(getAbsResourceRangeOrPanic(`module.vps["apac"].tofu_network.spiderweb`)),
+					Values:  "australia",
+				},
+			},
+			Query:     new(getAbsResourceRangeOrPanic(`module.vps[*].tofu_network.spiderweb`)),
+			WantError: true,
```
yottta: I (and future I) would appreciate a more specific error here. Is way more easier to build a mental model of the changes once you can see input/output/expected error in one block and not to jump from this `true` value to the actual code to figure out why an error is expected here.

Furthermore, if something is broken later in the trie implementation, it might still return an error, but not the expected one, resulting in a change of final UX. I found several issues like this in this project.

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,127 @@
...
+				trie.Set(override.Address, override.Values, nil)
+			}
+
+			got, diags := trie.Get(test.Query)
+			if diags.HasErrors() {
+				if !test.WantError {
+					// unexpectedly encountered an error
+					t.Errorf("got an error from trie override retrieval: %s", diags.Err().Error())
+				}
+				return
```
yottta:
```suggestion
                 if got != nil { ... }
				return
```

internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl
```diff
@@ -0,0 +1,45 @@
...
+
+override_resource {
+    target = test_resource.cats["pet_a"]
+    values = {
+        id = "Lassie"
+    }
+}
+
+override_resource {
+    target = test_resource.cats
```
lbordowitz, line 13: Should this be allowed, or should this be an error? Currently, I only have "modern syntax" in the `OverrideTrie` checking for wildcard keys, but maybe it should check for any keys. After all, once any keys are being used in an override, that implies usage of instances in test resource override targets.

Simultaneously, I want to ensure that people who have solely unkeyed overrides have their tests continue to work the way they expect. But that's not the case here, is it...

internal/addrs/override_trie.go
```diff
@@ -0,0 +1,209 @@
...
+// setNoKeyEvidence provides evidence that the NoKey instance key was used in a
+// particular resource override. This is later used when getting a key; if
+// this trie uses modern address syntax, but no key is used when a key is
+// called for, this is how we obtain that evidence.
+func (ot *OverrideTrie[T]) setNoKeyEvidence(i int, addr *AbsResourceInstance, src *hcl.Range) {
+	if ot.noKeyEvidenceMap == nil {
+		ot.noKeyEvidenceMap = make([][]*hcl.Range, len(addr.Module)+1)
+		for i := range len(addr.Module) + 1 {
+			ot.noKeyEvidenceMap[i] = make([]*hcl.Range, 0)
+		}
```
yottta:
```suggestion
		for j := range len(addr.Module) + 1 {
			ot.noKeyEvidenceMap[j] = make([]*hcl.Range, 0)
		}
```

I know that this shadows the method argument, but would be clearer to have a different name.
It's more of a nitpick, not really needed to be changed.

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,133 @@
...
+// SPDX-License-Identifier: MPL-2.0
+package addrs
+
+import (
+	"strings"
+	"testing"
+)
+
+// shorthand function to obtain known good strings.
+// Please please please do not use this outside of tests!!!
```
yottta:
```suggestion
// mustAbsResourceRange parses the given string into a AbsResourceInstance.
// It panics, if the given string is not parsable.
```

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,133 @@
...
+			TestName: "use default",
+			Overrides: []override{
+				{
+					Address: new(mustAbsResourceRange(`module.vps["apac"].tofu_network.spiderweb`)),
+					Value:   "australia",
+				},
+			},
+			Query:       new(mustAbsResourceRange(`module.vps["us-central1"].tofu_network.spiderweb`)),
+			WantMissing: true,
+			Want:        new("somewhere"),
```
yottta: Why this `Want` value? 🤔 Confused here.

yottta: Related to https://github.com/opentofu/opentofu/pull/4067/changes#r3187482988

internal/addrs/override_trie_test.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+
+			if test.WantMissing && got != nil {
+				t.Error("expected to get nothing, but found something")
+			}
+			if !test.WantMissing && got == nil {
+				t.Error("expected something, but didn't find anything")
+			} else if !test.WantMissing && *test.Want != *got {
+				t.Errorf("wrong result: expected %s, got %s\n", *test.Want, *got)
+			}
```
yottta, line 123:
```suggestion
			got, diags := trie.Get(test.Query)
			var gotErr string
			if diags.HasErrors() {
				gotErr = diags.Err().Error()
			}
			if !strings.Contains(gotErr, test.ErrorSubstr) {
				t.Errorf("unexpected content of the returned error\ngot:\n\t%s\nshould contain:\n\t%s", gotErr, test.ErrorSubstr)
			}

			if diff := cmp.Diff(test.Want, got); diff != "" {
				t.Errorf("unexpected returned value (-want,+got):\n%s", diff)
			}```

Comment (yottta) [reply internal/addrs/override_trie_test.go:123]: And remove the `WantMissing` altogether.
This caught my eyes because of my previous comment, where a test case was configured with a `Want` value that is not even checked, creating some confusion.
This new version always checks the err and the returned value with no complex cases.

File: internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl (hunk 14 above)
Comment (yottta) [reply internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl:13]: Maybe I am not familiar enough with the current inner workings of this, but this looks good to me. I read it something like this:
* test_resource.cats has 2 keyed overrides (for `pet_c` and `pet_a`)
* test_resource.cats has 1 un-keyed override, or wildcard in the new system (that applies to the missing keyed override, `pet_b`) (quite similar with the test in `internal/command/testdata/test/override_instance_default_instance_wildcard/main_provider.tftest.hcl`, but that uses the modern system)

That sounds good to me, like if there are N elements, a user can override 2 out of those, and for the rest of N-2, the general override should take place, like a `default` case in a `switch` statement.


> But that's not the case here, is it...

Could you share more about this concern? I lack heavily on this part of the system.

File: internal/command/test_test.go [hunk 19]
```diff
@@ -1654,3 +1654,139 @@ func TestTest_DeprecatedOutputs(t *testing.T) {
 		t.Fatalf("expected status code 0 but got %d: %s", code, output.All())
 	}
 }
+
+func TestTest_InstanceOverride(t *testing.T) {
```
yottta, line 1658: For the tests that have failed tests, can you add the `expectedErr` value please?

internal/tofu/node_resource_abstract_instance.go
```diff
@@ -3218,19 +3218,36 @@ func (n *NodeAbstractResourceInstance) getProvider(ctx context.Context, evalCtx
...
 		}
+
+		overrideValuesP, providerOverrideDiags := trie.Get(&n.Addr)
+		if overrideValuesP != nil {
+			overrideValues = *overrideValuesP
+		}
+
+		if providerOverrideDiags.HasErrors() {
+			return nil, providers.ProviderSchema{}, providerOverrideDiags.Err()
+		}
```
yottta, line 3236: I see that the `break` statement has been removed. Do we really want to do it? Once we have an override value, does it make any sense to check the others? 🤔

internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl
lbordowitz: Right, that's exactly how the overrides work. You have the wildcard or the "no-key" wildcard, and that determines the general case.

My question was more around whether a "no-key" wildcard should be allowed here, or if we should enforce usage of the "true" splat-based wildcard. I'm supporting the "no-key" wildcard as a backwards compatibility feature: before I implemented splat in this PR, you could not use it for targeting. But if someone is using *any* keys in their overrides, that means they ought to use splat wildcards instead of no-key wildcards... right? Or not? If they ought to: this test case is invalid, and should error-out, saying "use splat on this override". If it's fine, this test case remains valid.

The question centers around the [Modern Addressing](https://github.com/opentofu/opentofu/pull/4067/changes#diff-08beffed7b00c976b2d01ee388590a855c7e3bd2aa803fb4b91a26834d910c7dR88-R92) check in the override trie. Is using *any* key considered sufficiently modern? If so, then I need to revise the logic there and update some tests.

internal/addrs/override_trie_test.go
lbordowitz: That was leftover from when the `OverrideTrie.Get` returned a default value. It is no longer necessary and I've removed it.

internal/command/test_test.go
lbordowitz: I don't think I can. They're not really "errors" because there's no actual "error" happening in those tests. It's an expected logical issue, and I wanted to use it highlight that "this usage is intended to not work like that" or "the override will not propagate here".

internal/tofu/node_resource_abstract_instance.go
lbordowitz: We're not checking other override values. We're loading up the trie with all the overrides set in the provider, and then looking up the override for this node.

I like to use the metaphor "chewing your food." The root-level overrides are [associated address-wise during the config phase](https://github.com/opentofu/opentofu/pull/4067/changes/BASE..0c236f348e2771c310d9daa55c373b8392ddd63f#diff-249a4f321bf10846d86f2674032e699eaf5d0fe834588aa46284df4a68cb6696R1105), and then the trie is used for overrides; our food is chewed for us before we check overrides on the resource (below).

Provider overrides, on the other hand, are unprocessed. We have to check each one that might be relevant to the abs resource (*not* abs resource instance!) and set up the trie accordingly each time. Unlike before, we can't just break when we find the first override, because there may be multiple overrides relevant to the resource, depending on the instance. We have to chew our food here.

internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl
yottta: The only reason that crosses my mind and makes me say "yes, error if keyed are used with non-keyed overrides", is to keep the used features in the same line.
I am trying to think of the future, where if we allow both, how polluted the configurations would be with mixed syntaxes?

Furthermore, allowing both of these, does it add more friction and maintenance cost for the project? 🤔 
Because, as I see it right now, I am not sure that it adds anything significant.

And do any of the maintainers see any downsides (feature(s) evolution wise) on keeping both ways of overrides working together? Because I see none, but others might have different points of views.

---

_My initial response and a question for future us_

In that case, let me throw a new twist: would be possible to show a warning? 😅 
Do we want to deprecate the un-keyed overrides in the future and remove those fully?
If so, I would allow both ways of overrides to work and show a warning when detected both syntaxes, with the intent of removing un-keyed overrides in a future version.

internal/command/test_test.go
yottta: Ok, that makes sense.

internal/tofu/node_resource_abstract_instance.go
yottta: Great explanation! Thanks a lot. Totally on board!

internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl
lbordowitz: A warning might be the way to go: warn for mixed unkeyed + keyed, error if unkeyed + wildcard. Honestly, I'm not completely stoked with the way I'm detecting keys/"modern syntax" in the override trie as it is, and I welcome any suggestions.

cam72cam:
@lbordowitz let's try to get the opentofu/hcl PR created and in before this is merged :+1:

internal/command/testdata/test/override_instance_default_instance_mixed/main_provider.tftest.hcl
yottta: Just to make clear my take here: I am for a warning if both, new and old ways of addressing, are detected.
Later, if we decide that we want to remove fully the old ways of addressing, we can add a new warning for one or two development cycles and force users to migrate before removing it altogether.

_random thought: this could also be a linting concern_ 🤔

yottta (approved):

apparentlymart:
I submitted the HCL change upstream, but in the meantime I also prepared a PR to add it to our own HCL fork so that we don't need to wait for upstream: https://github.com/opentofu/hcl/pull/6.

apparentlymart:
The HCL change for "traversal patterns" is now available on our fork.

I'm expecting to pull some other changes into there for https://github.com/opentofu/opentofu/issues/3951, but I don't think we need to wait until that is completed before updating this PR's branch to refer to the current tip of our HCL fork so we can merge it.

apparentlymart (approved):

cam72cam (approved):

PR #4071, opened 2026-04-30
RFC: Installation and Caching of Remote Dependencies

apparentlymart:
This is a proposal for adopting a "cache by default"s strategy for remote dependency packages, and for adding module package tracking to the dependency lock file in support of that.

This is currently only a draft focused on a possible set of user-experience decisions. I'm opening it mainly to start a discussion about the open questions at the end of the document, so we can hopefully find consensus about what the goals are before designing the implementation details.

[Rendered Version](https://github.com/opentofu/opentofu/blob/rfc-remote-dependency-installation-and-caching/rfc/20260428-remote-dependency-install-cache.md)

### DIFF
--- a/rfc/20260428-remote-dependency-install-cache.md
+++ b/rfc/20260428-remote-dependency-install-cache.md
@@ -1,0 +1,951 @@
+# Installation and Caching of Remote Dependencies
+
+This document proposes a new approach to OpenTofu's handling of remote module
+packages and provider packages.
+
+This is a broad proposal that potentially interacts with (but does not
+necessarily completely solve) at least the following issues:
+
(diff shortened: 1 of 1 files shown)


Rezarys:
While looking at this RFC, I put together a small standalone tool that does local module caching for `terraform init`/`tofu init` (I'm the author): https://github.com/Rezarys/tfmodcache

It reads the declared modules, mirrors them to disk keyed by source and exact version, then serves subsequent `init` calls from that mirror instead of the network (git sources, registry sources with pinned versions, and archive sources). It confirms the approach in this RFC is workable outside of core: the tricky parts are exact-version pinning and rewriting the local `modules.json` manifest so `init` trusts the cache.

Built with AI assistance, reviewed and tested by me. Sharing in case the manifest format or lookup logic is a useful reference; it is not under active development.

PR #4072, opened 2026-04-30
Extract global configurations in one struct

yottta:
Part of #3595

This is a small PR that extracts pre-command static-like configuration into a separate struct with the the intent of using that later to decouple logic more easily.

## Checklist

### Go checklist

 


internal/command/views/refresh_test.go
```diff
@@ -24,7 +24,7 @@ func TestRefreshHuman_operation(t *testing.T) {
 	v := NewRefresh(arguments.ViewOptions{ViewType: arguments.ViewHuman}, NewView(streams).SetRunningInAutomation(true)).Operation()
 	if hv, ok := v.(*OperationHuman); !ok {
 		t.Fatalf("unexpected return type %t", v)
-	} else if hv.inAutomation != true {
+	} else if hv.view.runningInAutomation != true { // redundant but it replaces a previously more relevant check
```
diofeher: Since this check is redundant, could it be removed?

internal/command/arguments/system.go
diofeher, line 1: There's only one issue for me with placing these arguments here. From what I was looking at in the docs, none of these arguments are passed through the CLI, but rather included in blocks or in `.tofurc`. So it's kind of strange to put them in arguments, which are mostly used for CLI args.

yottta: > There's only one issue for me with placing these arguments here. From what I was looking at in the docs, none of these arguments are passed through the CLI, but rather included in blocks or in `.tofurc`. So it's kind of strange to put them in arguments, which are mostly used for CLI args.

Can you check 1d7c716abf065fdd4734312543f0ebb1325ab73a to see if it's better?

diofeher: I liked that!

diofeher (approved):

apparentlymart (approved):
I must admit that I was initially skeptical about the new `package system`, thinking it seemed like overkill for just one struct type, but after thinking about it more I changed my mind:

- We already have `package workdir` for encapsulating the various information we track on a per-working-directory basis, so it seems reasonable to have `package system` for representing similar information at a "global" scope.
- You already mentioned that you intend for this package to grow to include more of the logic from `package main` later, which aligns with the evolution of `package workdir` which gradually absorbed various functionality that was previously just inline in `package command`.

    (As we _do_ continue to move more logic into `package system`, I'd like for us to keep the specific decisions about how we access environment variables and what specific directories we search for different things inside `package main`, because that package represents our final concrete definition of how everything is "wired together" in a real CLI build vs. in unit tests or in other possible `package main`s we might add in future for use-cases like a "code playground" compiled to WebAssembly. But we can discuss that tradeoff more in later PRs, of course.)
- The code in `package cliconfig` and `package clistate` has some unclear conceptual overlap with `package system` and `package workdir` (respectively), but I expect we can refine the separation of concerns between those in later work rather than worry a lot about it now.

internal/command/views/operation_test.go
```diff
@@ -465,7 +465,7 @@ func TestOperation_planNextStep(t *testing.T) {
 // clearer.
 func TestOperation_planNextStepInAutomation(t *testing.T) {
 	streams, done := terminal.StreamsForTesting(t)
-	v := NewOperation(arguments.ViewHuman, true, NewView(streams))
+	v := NewOperation(arguments.ViewHuman, NewView(streams).SetRunningInAutomation(true))
```
lbordowitz, line 468: Why does this have `SetRunningInAutomation`, but none of the other tests with removed `true` args have that?

lbordowitz (approved):
General question: what is the scope of the `SystemCfg`? How do we prevent it from getting out of hand the way `Meta` did?

That question shouldn't block this PR, but it's something on my mind.

internal/command/views/operation_test.go
yottta: Good catch. Fixed in f121778fa30783e45363d3bef08804bb7eb52f1b

yottta:
> General question: what is the scope of the `SystemCfg`? How do we prevent it from getting out of hand the way `Meta` did?
> 
> That question shouldn't block this PR, but it's something on my mind.

As @apparentlymart already summarised it, if you look at the values that are used to initialise this new `system.Config`, those are mostly due to some overlapping with the `cliconfig` package.
I am not yet 100% that this struct will survive when the whole CLI rework will be done, but for now is more of a shim to remove as much as possible from the `Meta` structure.
For the other fields in this new struct (`RunningInAutomation` and `AllowExperimentalFeatures`) I intend to have these configured in a future [`rootCmd`](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#create-rootcmd) and _maybe_ propagated through `context.WithValue` or another struct, like `rootCmdCfg` that will be used to configure the rest of the commands.

The `Meta` struct as it's today, will not exist anymore, but maybe will be there something small that will be read-only. The main thing that really bothers me (and others) is the amount of playing around with the values on the `Meta` struct attributes instead of passing the values to the (non-existent yet) components to do their job and return the result.
This adds a lot of cognitive pressure when trying to understand a flow and suddenly it doesn't work because some flow configured, silently, another field in `Meta` that changed the behavior.

And when talking about "altering" the values of a struct, right now it's possible because all those fields (especially from the `arguments` package) are exported. So technically, there is no way to prevent it to get out of hands today.
But, one final blow that I want to do once we get close to finish this work, is to unexport everything and provide some sort of getters or some kind of builders for the components that those are are intended for.
This way, it will create clear boundaries of what is allowed and what not.

This is not clear cut right now, but all this work I do with all these small and maybe apparently useless PRs is to slowly reduce from the spaghetti code we have today and to remove the amount of dependencies between different packages and instead streamline the flow of configuration and the call-stack through the packages.

Hopefully this makes things clearer. If not, feel free to disagree. I would love to dig more into this 😄

PR #4073, opened 2026-04-30
Replace backend individual attributes from `Meta` with `arguments.Backend`

yottta:
Part of #3595

Removing even more individual attributes from the `Meta` structure. This time, removed the backend related configuration and replaced those with the `arguments.Backend`.
From here on, there are no other clear cut refactoring that can be done so will continue with the [`Private: do not set these` part of the `Meta` struct](https://github.com/opentofu/opentofu/blob/ce973d0b297941d48cafeb156cde377c9d3296c4/internal/command/meta.go#L161-L195).

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/import.go
+++ b/internal/command/import.go
@@ -56,7 +56,18 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 		}
 		return cli.RunResultHelp
 	}
-	c.configureBackendFlags(args)
+	c.Meta.variableArgs = args.Vars.All()
+	c.stateArgs = *args.State
+	c.backendArgs = *args.Backend
+
+	// TODO meta-refactor: remove this only when there is clear path of passing these from the "arguments" package to
+	// the place where these needs to be used
+	c.Meta.parallelism = args.Parallelism
+
+	// FIXME: the -input flag value is needed to initialize the backend and the
+	// operation, but there is no clear path to pass this value down, so we
+	// continue to mutate the Meta object state for now.
+	c.Meta.input = args.ViewOptions.InputEnabled
 
 	// Parse the provided resource address.
 	traversalSrc := []byte(args.ResourceAddress)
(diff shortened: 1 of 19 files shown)


internal/command/meta_backend.go
```diff
@@ -1544,9 +1544,9 @@ func (m *Meta) assertSupportedCloudInitOptions(mode cloud.ConfigChangeMode) tfdi
 				))
 			}
 		}
-		if m.migrateState {
+		if m.backendArgs.MigrateState {
 			name := "-migrate-state"
-			if m.forceInitCopy {
+			if m.backendArgs.ForceInitCopy {
 				// -force copy implies -migrate-state in "tofu init",
 				// so m.migrateState is forced to true in this case even if
```
diofeher:
```suggestion
				// so m.backendArgs.migrateState is forced to true in this case even if
```

diofeher (approved):

internal/command/meta_backend.go
yottta: Fixed in ac7666861a56568c69ae3d6f0f3620f77c4b04c8

internal/command/import.go
```diff
@@ -56,7 +56,18 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 		}
 		return cli.RunResultHelp
 	}
-	c.configureBackendFlags(args)
+	c.Meta.variableArgs = args.Vars.All()
+	c.stateArgs = *args.State
+	c.backendArgs = *args.Backend
```
lbordowitz, line 61: Are `State` and `Backend` guaranteed to be non-nil?

lbordowitz (approved):
This appears to be functionally equivalent to what was there before; a good refactor. Looks nice!

internal/command/import.go
yottta: If they are not, it's a bug in the implementation which will result in a panic in at least one test.
These fields are only initialised in the `arguments` package.

Issue #4076, opened 2026-05-03
`tofu init -backend=false` still reads and attempts to decrypt local state when encryption is configured


mburumaxwell:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
OpenTofu v1.11.6
on darwin_arm64
```

### OpenTofu Configuration Files

```hcl
# providers.tf
terraform {
  required_version = ">= 1.11"

  encryption {
    key_provider "pbkdf2" "main" { passphrase = var.passphrase }
    method "aes_gcm" "main" { keys = key_provider.pbkdf2.main }
    state { method = method.aes_gcm.main }
  }
}

variable "passphrase" {
  sensitive = true
}

# main.tf — minimal resource to produce a state file on first apply
resource "terraform_data" "example" {}
```

### Debug Output

```txt
2026-05-03T18:24:28.086+0100 [TRACE] OpenTelemetry: OTEL_TRACES_EXPORTER not set, OTel tracing is not enabled
2026-05-03T18:24:28.086+0100 [INFO]  OpenTofu version: 1.11.6
2026-05-03T18:24:28.086+0100 [DEBUG] using github.com/hashicorp/go-tfe v1.95.0
2026-05-03T18:24:28.086+0100 [DEBUG] using github.com/opentofu/hcl/v2 v2.20.2-0.20251021132045-587d123c2828
2026-05-03T18:24:28.086+0100 [DEBUG] using github.com/zclconf/go-cty v1.17.0
2026-05-03T18:24:28.086+0100 [INFO]  Go runtime version: go1.26.2
2026-05-03T18:24:28.086+0100 [INFO]  CLI args: []string{"tofu", "init", "-backend=false"}
2026-05-03T18:24:28.086+0100 [TRACE] Stdout is a terminal of width 146
2026-05-03T18:24:28.086+0100 [TRACE] Stderr is a terminal of width 146
2026-05-03T18:24:28.086+0100 [TRACE] Stdin is a terminal
2026-05-03T18:24:28.086+0100 [DEBUG] Attempting to open CLI config file: /Users/me/.tofurc
2026-05-03T18:24:28.086+0100 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
[...]

PR #4077, opened 2026-05-04
fix: skip local encrypted state read when init runs with -backend=false (#4076)

mumberrymountain:
fix issue #4076

**Problem**

init with `-backend=false` could still fail in a directory that already contains an encrypted local state file, surfacing as an error while refreshing state.

**Cause**

init.go
```go
default:
	// load the previously-stored backend config
	back, backDiags = c.Meta.backendFromState(ctx, enc.State())
}
```

With `-backend=false`, init still fell into the default branch and call backendFromState, which reconstructs a backend from previously stored working-directory metadata and wires it with the encryption handle passed in for this run. 

A non-nil backend then led init to open the workspace state manager and refresh state, which attempts to read the on-disk state—including an encrypted terraform.tfstate—and fails in that situation.

**Fix**

```go
switch {
	case args.BackendFlagSet && !args.FlagBackend:
		// The user explicitly passed -backend=false,
		// so we must neither initialize a new backend nor load any
		// previously-initialized one. Loading the previously-initialized
		// backend would otherwise try to read (and, if encryption is
		// configured, decrypt) the local state file, defeating the whole
		// purpose of -backend=false and breaking workflows where the
		// encryption key is intentionally unavailable.
		back = nil
	case args.FlagCloud && rootModEarly.CloudConfig != nil:
		back, backendOutput, backDiags = c.initCloud(ctx, rootModEarly, args.FlagConfigExtra, enc, view.Backend())
	case args.FlagBackend:
		back, backendOutput, backDiags = c.initBackend(ctx, rootModEarly, args.FlagConfigExtra, enc, view.Backend())
	default:
		// load the previously-stored backend config
		back, backDiags = c.Meta.backendFromState(ctx, enc.State())
}
```

When `-backend=false` is explicitly set, do not take that path. skip loading a previously initialized backend so init does not refresh workspace state for provider hints in this mode. Add a regression test for this issue.

Resolves # 4076

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -207,6 +207,15 @@ To initialize the configuration already in this working directory, omit the
 	var backendOutput bool
 
 	switch {
+	case args.BackendFlagSet && !args.FlagBackend:
+		// The user explicitly passed -backend=false,
+		// so we must neither initialize a new backend nor load any
+		// previously-initialized one. Loading the previously-initialized
+		// backend would otherwise try to read (and, if encryption is
+		// configured, decrypt) the local state file, defeating the whole
+		// purpose of -backend=false and breaking workflows where the
+		// encryption key is intentionally unavailable.
+		back = nil
 	case args.FlagCloud && rootModEarly.CloudConfig != nil:
 		back, backendOutput, backDiags = c.initCloud(ctx, rootModEarly, args.FlagConfigExtra, enc, view.Backend())
 	case args.FlagBackend:
--- a/internal/command/init.go
+++ b/internal/command/init.go
@@ -207,7 +207,7 @@ To initialize the configuration already in this working directory, omit the
 	var backendOutput bool
 
 	switch {
-	case args.BackendFlagSet && !args.FlagBackend:
+	case !args.FlagBackend && args.BackendFlagSet:
 		// The user explicitly passed -backend=false,
 		// so we must neither initialize a new backend nor load any
 		// previously-initialized one. Loading the previously-initialized
--- a/internal/command/init_test.go
+++ b/internal/command/init_test.go
@@ -3262,6 +3262,44 @@ func TestInit_skipEncryptionBackendFalse(t *testing.T) {
 	})
 }
 
+t.Run("init succeeds with -backend=false even when an encrypted state file is already present", func(t *testing.T) {
+	td := t.TempDir()
+	testCopyDir(t, testFixturePath("init-encryption-with-state"), td)
+	t.Chdir(td)
+
+	overrides := metaOverridesForProvider(testProvider())
+	view, done := testView(t)
+	providerSource, closeCallback := newMockProviderSource(t, map[string][]string{
+		"hashicorp/aws": {"5.0", "5.8"},
+	})
+	defer closeCallback()
+	m := Meta{
+		WorkingDir:       workdir.NewDir("."),
+		testingOverrides: overrides,
+		View:             view,
+		ProviderSource:   providerSource,
+	}
+
+	c := &InitCommand{
+		Meta: m,
+	}
+
+	args := []string{
+		"-backend=false",
+	}
+	code := c.Run(args)
+	output := done(t)
+	if code != 0 {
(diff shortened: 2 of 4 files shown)


internal/command/init_test.go
```diff
@@ -3260,6 +3260,44 @@ func TestInit_skipEncryptionBackendFalse(t *testing.T) {
...
+		output := done(t)
+		if code != 0 {
+			t.Fatalf("init should run successfully with -backend=false even with an encrypted state file present\nexit code: %d\nstderr:\n%s\nstdout:\n%s", code, output.Stderr(), output.Stdout())
+		}
+		if strings.Contains(output.Stderr(), "Error refreshing state") {
+			t.Fatalf("init must not attempt to read/refresh the local state when -backend=false is set\nstderr:\n%s", output.Stderr())
+		}
+		if strings.Contains(output.Stderr(), "can not be read without an encryption configuration") {
+			t.Fatalf("init must not attempt to decrypt the local state when -backend=false is set\nstderr:\n%s", output.Stderr())
+		}
```
yottta:
```suggestion
		if stderr := output.Stderr(); stderr != "" {
			t.Errorf("expected to have no errors but got:\n%s", stderr)
		}
```

Since the command is not meant to fail anymore, we shouldn't leave the error messages non-existence checks since these could create more confusion than making things clearer.
If you really want to, leave a comment on the sub-test to make it clear that was added during fixing this specific issue and was added to check that the init does not load any backend which expects no error messages.

internal/command/init_test.go
```diff
@@ -3260,6 +3260,44 @@ func TestInit_skipEncryptionBackendFalse(t *testing.T) {
 			t.Fatalf("generated error should contain the string \"Error: Unable to fetch encryption key data\"\ninstead got : %s\n", output.Stderr())
 		}
 	})
+
+	t.Run("init succeeds with -backend=false even when an encrypted state file is already present", func(t *testing.T) {
```
yottta, line 3266: I would add this sub test in its own test function since `TestInit_skipEncryptionBackendFalse` has a little bit of a different purpose IMO. It checks that it skips encryption, not the backend per se.
Or you can rename the function to cover a broader spectrum.
Maybe `TestInit_skipEncryptionAndBackendOnCliBackendFalse` 🤷

yottta (approved):
Mostly nitpicks, but other than that, this looks really good!
Thanks for the help 🙌

internal/command/init_test.go
mumberrymountain: Agree. Applied your suggestion, replacing the two specific error string checks with a single `stderr == ""` assertion.

internal/command/init_test.go
mumberrymountain: Yeah, I just moved the sub-test into its own function `TestInit_backendFalse_skipsBackendFromStateOnPreviouslyInitializedDir` to better reflect what's being tested

mumberrymountain:
Hey, @yottta thanks for the review! I applied your feedback into commit.

internal/command/init.go
```diff
@@ -207,6 +207,15 @@ To initialize the configuration already in this working directory, omit the
 	var backendOutput bool
 
 	switch {
+	case args.BackendFlagSet && !args.FlagBackend:
```
diofeher:
```suggestion
	case !args.FlagBackend && args.BackendFlagSet:
```

Really small nit: our other switch cases here start the checks with "args.Flag*" as the first member of the expression. I think swapping the two members of this expression would facilitate reading here.

mumberrymountain: Good Point, I applied your feedback into commit.

diofeher: Thanks!

diofeher (approved):

mburumaxwell:
Hi @diofeher, when does this get released?

yottta:
> Hi @diofeher, when does this get released?

This is going to be released in v1.13. There is no an exact date for it but should be around September.

PR #4078, opened 2026-05-04
website: Updates for the v1.12 release

apparentlymart:
As well as the usual mechanical replacements of any mention of the current release version number, this includes some other broader changes:

- The page about upgrading from earlier versions now just refers generally to "earlier versions", because continuing the tradition of listing every previous version of OpenTofu explicitly would make the page title and content very unwieldy now that there are four earlier releases.
- The "What's New" page is now just a short overview of the new features followed by a link to the longer-form content in the announcement blog post, since that makes the page easier to scan and keeps a single main place for the full announcement content.
- The sidebar is slightly reordered so that upgrading from previous OpenTofu versions is now sorted before migrating from Terraform. The page title is also now "Migrating from Terraform" rather than "Migrating to OpenTofu" so that it's clearer who is the audience for that set of pages.
- And finally, just some general copyediting to freshen some of this per-release content that has been with us for a long time, and to focus some of the pages a little more tightly.

---

### DIFF
--- a/website/docs/intro/upgrading.mdx
+++ b/website/docs/intro/upgrading.mdx
@@ -1,19 +1,19 @@
 ---
-sidebar_position: 5
-sidebar_label: Upgrading from OpenTofu 1.8.x/1.9.x/1.10.x
+sidebar_position: 4
+sidebar_label: Upgrading from earlier versions of OpenTofu
 description: |-
-    Learn how to upgrade OpenTofu from version 1.8.x/1.9.x/1.10.x to 1.11.0.
+    Learn how to upgrade OpenTofu to 1.12.
 ---
 
-# Upgrading from OpenTofu 1.8.x/1.9.x/1.10.x
+# Upgrading from earlier versions of OpenTofu
 
-OpenTofu 1.11.x is mostly compatible with previous OpenTofu versions. This migration guide will take you through the process of upgrading OpenTofu to version 1.11.0.
+OpenTofu 1.12.x is largely compatible with previous OpenTofu versions. This migration guide will take you through the process of upgrading to OpenTofu v1.12.
 
 ## Step 0: Prepare a disaster recovery plan
 
-Although OpenTofu 1.11 is mostly compatible with previous versions, you should take the necessary precautions to prevent accidents. Make sure you have an up to date and *tested* disaster recovery plan.
+Although OpenTofu v1.12 is mostly compatible with previous versions, you should take the necessary precautions to prevent accidents. Make sure you have an up-to-date and *tested* disaster recovery plan.
 
-## Step 1: Apply all changes with OpenTofu 1.7.x/1.8.x/1.9.x
+## Step 1: Apply all changes with your current OpenTofu version
 
 Before proceeding, make sure that you apply all changes with `tofu apply`. Running `tofu plan` should result in no planned changes. While you can switch to OpenTofu with pending changes, it is not recommended.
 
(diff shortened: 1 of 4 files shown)


website/docs/intro/upgrading.mdx
```diff
@@ -28,50 +28,44 @@ OpenTofu has compared your real infrastructure against your
...
+If any of the following steps fail, we recommend that you stop following these instructions and refer to [Rolling back and reporting issues](#rolling-back-and-reporting-issues) below.
 
 :::
 
 Now you are ready to migrate. Run `tofu init` in the directory where your code resides. OpenTofu will download
 any providers and modules referenced in your configuration from the OpenTofu registry.
 
-:::note
-
-If you are using the S3 backend - You will need to run `tofu init -reconfigure` to reinitialize the backend.
```
yottta, line 66: I wouldn't remove this note altogether but I would rework it to be a little bit more generic "using some of the existing backends might require to run `tofu init -reconfigure` to reinitialize the backend, action that will be indicated by a failure of `tofu init` in case it's needed".

website/docs/intro/whats-new.mdx
```diff
@@ -1,55 +1,28 @@
...
 
-# What's new in OpenTofu 1.11?
+# What's new in OpenTofu 1.12?
 
-## New features
+Here are some highlights from the new features in this release:
 
-### Ephemeral Resources / Write-Only Attributes
+- **Dynamic `prevent_destroy`**: The `prevent_destroy` argument in a resource's lifecycle block can now refer to other symbols within the same module, such as input variables.
+- **Improved provider checksum handling**: `tofu init` now automatically includes a full set of checksums for all platforms using both `zh:` and `h1:` hashes, reducing the need for manually running `tofu providers lock`.
```
yottta, line 13: I would add a warning here to catch the reader's attention and set the right expectations regarding the lock file, which in most cases, will suffer changes.
I know that this is also mentioned in the blog post, but since it's a feature that will result in changes of the user's lockfile, I think that's a good thing to make people aware of this everywhere.

yottta (approved):

website/docs/intro/upgrading.mdx
apparentlymart: I don't know of a reason why this would be needed for any backend in v1.12.

I believe this was here because in some earlier release there was a change to the schema of the S3 backend that made OpenTofu unsure about whether migration was needed or not, and so this instruction was here to deal with that problem. But I don't think we've made any changes like that for v1.12. Did I miss something from work in a different backend?

I'm hesitant to just make a general statement that it's always okay to "reconfigure" here, because the error message this option bypasses is shown when the user needs to make a decision that might cause new problems if decided incorrectly, and so just saying that it's always okay to reconfigure would undermine the purpose of that error message.

(The fact that changing the S3 backend schema caused this to trigger was very unfortunate in the first place, and so I hope it's we make similar schema changes in future we'll find a way to avoid it and make OpenTofu make the safe decision automatically itself.)

website/docs/intro/whats-new.mdx
apparentlymart: That's a good idea. 

I think I will handle it by changing the heading for "deprecations" to instead be a more general "upgrade notes" (matching the terminology from the changelog) and then add a short section about checksums alongside the other two sections about WinRM and 32-bit CPU architectures.

website/docs/intro/upgrading.mdx
```diff
@@ -86,34 +80,18 @@ configuration and found no differences, so no changes are needed.
 
 ## Step 6: Test out a small change
 
-Before you begin using OpenTofu for larger changes, test out `tofu apply` with a smaller, non-critical
-change.
+Before you begin using OpenTofu for larger changes, we recomment testing out `tofu apply` with a smaller, non-critical change.
```
diofeher:
```suggestion
Before you begin using OpenTofu for larger changes, we recommend testing out `tofu apply` with a smaller, non-critical change.
```

website/docs/intro/whats-new.mdx
diofeher: `reducing the need for manually running `tofu providers lock`.`

Should we mention the cases where we still want to run this command?

diofeher (approved):

website/docs/intro/whats-new.mdx
apparentlymart: The longer text in the blog post (linked after this list) talks more about when `tofu providers lock` is needed or not needed. I would prefer to keep the text on this page brief and rely on the blog post for the longer content.

(I only mentioned `tofu providers lock` as part of the summary here because being forced to run that command has been a major way folks have described being annoyed by the current situation, and so I'm hoping that mentioning it makes it clearer to those folks that this is talking about a solution to their problem.)

PR #4081, opened 2026-05-07
Fix panic during closing of the ephemeral resources

yottta:
Resolves _No issue_

Discovered this during another test, and coincidentally, the closing of the ephemeral resources took exactly the amount of time that is [configured as timeout for the closing operation](https://github.com/opentofu/opentofu/blob/9e746cc2d159933d665cb8680819fccb99195531/internal/shared/ephemeral_resource.go#L225).
When that exact condition is met, the [writing of the diagnostics](https://github.com/opentofu/opentofu/blob/9e746cc2d159933d665cb8680819fccb99195531/internal/shared/ephemeral_resource.go#L215) for the `CloseEphemeralResource` fails because the close function returned and the [defer statements have been executed, closing the diagnostics channel](https://github.com/opentofu/opentofu/blob/9e746cc2d159933d665cb8680819fccb99195531/internal/shared/ephemeral_resource.go#L221-L223).

To avoid this kind of issues, as a general rule of thumb, in Go, the writer is responsible with the channel closing, and not the reader, so now the goroutine that writes to the channel it also closes it.
The fix includes also a failover, to avoid a deadlock in case the channel is full and there is no other reader of it.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/shared/ephemeral_resource.go
+++ b/internal/shared/ephemeral_resource.go
@@ -147,6 +147,10 @@ func OpenEphemeralResourceInstance(
 	closeCh := make(chan context.Context, 1)
 	diagsCh := make(chan tfdiags.Diagnostics, 1)
 	go func() {
+		// The writer is responsible with closing the channel, not the receiver.
+		defer func() {
+			close(diagsCh)
+		}()
 		var diags tfdiags.Diagnostics
 		renewAt := openResp.RenewAt
 		privateData := openResp.Private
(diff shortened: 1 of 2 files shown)


lbordowitz:
I think you can fix the race conditions if you swap out `OpenEphemeralResourceCalled` and `CloseEphemeralResourceCalled` for atomic booleans.

lbordowitz (approved):

PR #4082, opened 2026-05-07
getproviders: Don't clobber mismatching cache entries

apparentlymart:
Prevously OpenTofu's provider installer would try to install a package even if there was already a directory there which doesn't match the package contents. That's effective in making us more likely to end up with a working provider cache directory, but risks clobbering a package directory that the operator intentionally modified for some reason.

We'll now require that if an existing directory (or symlink to one) is present at the place where we'd need to put our cache entry then its contents must already match what we're trying to install, thereby making this a no-op. If the existing contents don't match then we'll fail with an error to let the operator decide whether they need to keep something from their modified directory before deleting it.

In earlier versions of OpenTofu, silently replacing an existing directory was actually sometimes done intentionally to ensure that the cache would definitely match the dependency lock file, but we no longer need to do that because as of OpenTofu v1.9 the provider installer now exits early (without downloading anything at all) if a matching package is already present, so we never end up trying to replace a package that was already present on disk unless it's the case we're now trying to catch as an error here.

The handling of this is in `PackageLocalArchive` because the two network-based location types (HTTP archive and OCI blob archive) work by first fetching the archive to a temporary local location and then asking the local archive location to finish the installation.

This is covered by e2etests rather than a unit test because successfully hitting this error requires both the `providercache` and `getproviders` packages to cooperate to let execution reach this late step without any earlier code doing an early exit due to the directory already being present. The e2etest runs through that entire codepath to make sure we reach the error message we're expecting to reach.

---

### DIFF
--- a/internal/getproviders/package_location_local_archive.go
+++ b/internal/getproviders/package_location_local_archive.go
@@ -69,22 +69,44 @@ func (p PackageLocalArchive) InstallProviderPackage(ctx context.Context, meta Pa
 	filename := meta.Location.String()
 	span.SetAttributes(traceattrs.FilePath(filename))
 
-	// NOTE: Packages are immutable, but we may want to skip overwriting the existing
-	// files in due to specific scenarios defined below.
-
-	if _, err := os.Stat(targetDir); err == nil {
-		// If the package might already be installed, we should try to skip overwriting the contents.
-		// When run with TF_PLUGIN_CACHE_DIR or similar, a given provider might already be executing
-		// and therefore locking the provider binary in the target directory (preventing the overwrite below)
-		//
-		// This does incur the overhead of two additional hash computations and could be
-		// skipped with smarter checks around re-use scenarios in the future.
+	// If there is already a package at the location we would've been installing
+	// to then that's okay if the content already matches what we would've
+	// installed, but we reject it otherwise so the operator can investigate.
+	if info, err := os.Lstat(targetDir); err == nil {
+		log.Printf("[TRACE] There's already a directory entry at %s, so we'll check if it matches our expectations", targetDir)
 
 		targetHash, targetErr := PackageHashV1(PackageLocalDir(targetDir))
-		fileHash, fileErr := PackageHashV1(meta.Location)
-
-		if targetHash == fileHash && fileErr == nil && targetErr == nil {
-			// Package is properly installed, bad or missing lock file will be caught elsewhere
+		// If the existing entry is an empty directory then we permit that just
+		// because sometimes a caller might want to create the target directory
+		// themselves before installing into it, such as if the target directory
+		// is a temporary directory created with [os.MkdirTemp]. Only a direct
+		// empty directory is allowed here, not a symlink to an empty directory.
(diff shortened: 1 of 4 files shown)


apparentlymart:
This is going to be backported, so there's intentionally no changelog entry here and instead we'll add it to the earlier series changelog as part of the subsequent backport PR.

internal/getproviders/package_location_local_archive.go
```diff
@@ -69,22 +69,42 @@ func (p PackageLocalArchive) InstallProviderPackage(ctx context.Context, meta Pa
...
-		// If the package might already be installed, we should try to skip overwriting the contents.
-		// When run with TF_PLUGIN_CACHE_DIR or similar, a given provider might already be executing
-		// and therefore locking the provider binary in the target directory (preventing the overwrite below)
-		//
-		// This does incur the overhead of two additional hash computations and could be
-		// skipped with smarter checks around re-use scenarios in the future.
+	// If there is already a package at the location we would've been installing
+	// to then that's okay if the content already matches what we would've
+	// installed, but we reject it otherwise so the operator can investigate.
+	if info, err := os.Lstat(targetDir); err == nil {
```
lbordowitz, line 75: Phew, that is a load-bearing Lstat!

Making sure I understand correctly: we use Lstat so that we get the opaque link's information. That symlink will never be a directory, i.e. `info.IsDir()` is always false for symlinks, regardless of what it points to. Is that correct?

lbordowitz (approved):
Great tests, looks good!

internal/getproviders/package_location_local_archive.go
apparentlymart: Yeah, that's exactly right. If we use just `Stat` here then the check below would allow a symlink referring to a directory in addition to allowing a directory directly.

(The test case that uses a symlink to an empty directory is covering that difference. Switching thus to `Stat` makes that test fail, so we should be able to avoid accidentally breaking this again later.)

apparentlymart: I realized that the comment here wasn't actually talking about the "symlink to empty directory" situation -- that was previously mentioned only in the e2etest, which is far away from here -- so I added some more words to the comment about that.

While I was doing that I got caught up trying to read the double-negative conditionals in the `if` statement below and so I decided it would be clearer to write it as first deciding a positive "is an empty directory" flag and _then_ negate it for the `if` statement, so hopefully that also helps to make the control flow a little easier to follow.

yottta (approved):
Looks great! 🙌 
And thanks for the test in `hash_test.go`. That calmed down my cross OS compatibility concerns for that emptyDirHash. 😅

apparentlymart:
Backport PRs, for future reference:

- [v1.12](https://github.com/opentofu/opentofu/pull/4087) (merged before v1.12.0, so included throughout that series)
- [v1.11](https://github.com/opentofu/opentofu/pull/4088) (from v1.11.7 onwards)
- [v1.10](https://github.com/opentofu/opentofu/pull/4089) (from v1.10.10 onwards)

PR #4086, opened 2026-05-08
backend/s3: Close response body in getLockInfoFromS3

SebTardif:
## What

`getLockInfoFromS3` calls `s3Client.GetObject` but never closes `resp.Body`,
leaking an HTTP connection on every call.

## How it leaks

```go
// internal/backend/remote-state/s3/client.go, line 481
resp, err := c.s3Client.GetObject(ctx, getParams, ...)
if err != nil {
    // ... error handling returns, no body to close (correct)
    return nil, err
}

// Body is read here but never closed
lockInfo := &statemgr.LockInfo{}
err = json.NewDecoder(resp.Body).Decode(lockInfo)
```

The AWS SDK v2 requires callers to close `GetObject` response bodies. Each
unclosed body holds an HTTP connection from the transport pool. This function
is called during `s3Lock` (when another lock exists) and `s3Unlock` (to verify
lock ownership), so the leak occurs on every contested lock operation and every
unlock.

In long-running automation with frequent state operations, leaked connections
accumulate and can exhaust the connection pool, causing subsequent S3 requests
to stall waiting for an available connection.

The sibling method `Get` in the same file already handles this correctly:

```go
// line 178 - correct pattern
defer output.Body.Close()
```

## Fix

Add `defer resp.Body.Close()` after the error check, matching the pattern
used by `Get`.

Introduced in #2521.

## Checklist

### Go checklist

- [x] I have run existing tests to ensure my code does not break anything.

### DIFF
--- a/internal/backend/remote-state/s3/client.go
+++ b/internal/backend/remote-state/s3/client.go
@@ -487,6 +487,7 @@ func (c *RemoteClient) getLockInfoFromS3(ctx context.Context) (*statemgr.LockInf
 
 		return nil, err
 	}
+	defer resp.Body.Close()
 
 	lockInfo := &statemgr.LockInfo{}
 	err = json.NewDecoder(resp.Body).Decode(lockInfo)


yottta (approved):
Such a slip up from me 🤦 . Thanks a lot for this! 👏

diofeher (approved):

cam72cam (approved):

PR #4110, opened 2026-05-15
tofu: Don't panic if provider startup fails during validate

apparentlymart:
If a `ProviderManager` returns error diagnostics when asked to start up a provider then the providers.Interface it returns is likely to be uninitialized, so we mustn't try to use it to validate the provider configuration in that case.

(I'm fixing this because somebody reported seeing a panic caused by this, and the panic is obscuring whatever the real error is. Once this is fixed it'll presumably reveal another problem by returning an error diagnostic.)

### DIFF
--- a/internal/tofu/node_provider.go
+++ b/internal/tofu/node_provider.go
@@ -103,9 +103,11 @@ func (n *NodeApplyableProvider) Execute(ctx context.Context, evalCtx EvalContext
 	if op == walkValidate {
 		log.Printf("[TRACE] NodeApplyableProvider: validating configuration for %s", n.Addr)
 		instance, newDiags := evalCtx.Providers().NewProvider(ctx, n.Addr.Provider)
-		n.instances[addrs.NoKey] = instance
-		for key, data := range instanceData {
-			diags = diags.Append(n.ValidateProvider(ctx, evalCtx, instance, key, data))
+		if !newDiags.HasErrors() { // We can't validate if we didn't successfully start the provider plugin
+			n.instances[addrs.NoKey] = instance
+			for key, data := range instanceData {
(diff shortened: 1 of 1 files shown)


apparentlymart:
I've not added a changelog entry here yet because this is just an interim change to unmask whatever underlying error message exists here so we can learn what's _actually_ causing the provider to fail to launch. I'm expecting that once we have this in v1.12.1 we can ask the original requester to test again and hopefully we'll then find out what the root bug is and can mention _that_ in the changelog when we fix it.

yottta (approved):
By challenging the stacktrace with the [latest commit from the v1.12 branch](https://github.com/opentofu/opentofu/blob/bdcbf091aab886499ed7718d7f558b428fffce39/internal/tofu/node_provider.go#L183), the change makes sense.
```shell
runtime error: invalid memory address or nil pointer dereference
...
panic({0x30c6ba0?, 0x63e7a10?})
	runtime/panic.go:860 +0x12c
github.com/opentofu/opentofu/internal/tofu.(*NodeApplyableProvider).ValidateProvider(0x13090aeef5b0, {0x3b5dc58, 0x13090890d230}, {0x3b8f4c0, 0x130911e034a0}, {0x0, 0x0}, {0x0, 0x0}, {{{{0x0, ...}}, ...}, ...})
	github.com/opentofu/opentofu/internal/tofu/node_provider.go:183 +0xb34
github.com/opentofu/opentofu/internal/tofu.(*NodeApplyableProvider).Execute(0x13090aeef5b0, {0x3b5dc58, 0x13090890d230}, {0x3b8f4c0, 0x130911e034a0}, 0x4)
```

diofeher (approved):

PR #4112, opened 2026-05-18
Prepare config loader to be used independently, without relying on the Meta struct

yottta:
Part of #3595.

After several attempts on how to remove completely the config loader concern from the Meta structure I had to settle with the compromise presented in this PR.

* Since the previous config loader was created by using static data rather than dynamic, there is no obvious reason to have this created by the Meta struct.
* But to keep the logic as close as possible to the existing behavior, we want to defer the return of the loader initialisation errors to the moments when its member methods are invoked.

Considering the points above, I chose to create a new `lazyLoader` that can be **created** before creating the commands structs (because the information needed is already at hand) and initialised later when it's actually used.
The `lazyLoader` implementation will return its initialisation error whenever one of its methods that return errors is invoked. (❓ should we cache the error once and do not attempt to initialise again every time a non-returning-error-method is called?)

The reason for the compromise is due to the way current tests are configured:
* The commands are exported structs that are initialised during the tests with the bare minimul dependencies for that particular use case. The config loader is loaded and initialised by the command logic when it needs it (via the `Meta` struct).
* The config loader cannot be created in the `Run` methods because it needs to be accessible for all of the commands' methods, since some commands autocompletion rely on that too.

Therefore, for the time being (until we will have time to rework the tests with a custom config loader), the `Meta.configLoader` will be kept.
Later, I intend to create constructor-like functions for each command and those functions will have as their arguments clear dependencies that each command have.
Once we have those functions, we can migrate all the tests to that and the initialisation of the config loader (and other components) could happen there, making it a common code path for both, the normal flow and testing flow.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/backend/backend.go
+++ b/internal/backend/backend.go
@@ -273,7 +273,7 @@ type Operation struct {
 
 	// ConfigLoader is a configuration loader that can be used to load
 	// configuration from ConfigDir.
-	ConfigLoader *configload.Loader
+	ConfigLoader configload.Loader
 
 	// DependencyLocks represents the locked dependencies associated with
 	// the configuration directory given in ConfigDir.
--- a/internal/command/apply.go
+++ b/internal/command/apply.go
@@ -13,6 +13,7 @@ import (
 	"github.com/opentofu/opentofu/internal/backend"
 	"github.com/opentofu/opentofu/internal/command/arguments"
 	"github.com/opentofu/opentofu/internal/command/views"
+	"github.com/opentofu/opentofu/internal/configs/configload"
 	"github.com/opentofu/opentofu/internal/encryption"
 	"github.com/opentofu/opentofu/internal/plans/planfile"
 	"github.com/opentofu/opentofu/internal/tfdiags"
@@ -276,7 +277,7 @@ func (c *ApplyCommand) OperationRequest(
 	opReq.View = view.Operation()
 
 	var err error
-	opReq.ConfigLoader, err = c.initConfigLoader()
+	opReq.ConfigLoader, err = configload.Initialize(c.configLoader())
 	if err != nil {
 		diags = diags.Append(fmt.Errorf("Failed to initialize config loader: %w", err))
 		return nil, diags
--- a/internal/command/apply.go
+++ b/internal/command/apply.go
@@ -277,7 +277,7 @@ func (c *ApplyCommand) OperationRequest(
 	opReq.View = view.Operation()
 
 	var err error
-	opReq.ConfigLoader, err = configload.Initialize(c.configLoader())
+	opReq.ConfigLoader, err = configload.Initialise(c.configLoader())
 	if err != nil {
 		diags = diags.Append(fmt.Errorf("Failed to initialize config loader: %w", err))
 		return nil, diags
--- a/internal/command/import.go
+++ b/internal/command/import.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/mitchellh/cli"
+	"github.com/opentofu/opentofu/internal/configs/configload"
 	"github.com/opentofu/opentofu/internal/tracing"
 
 	"github.com/opentofu/opentofu/internal/addrs"
@@ -76,7 +77,7 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 	if travDiags.HasErrors() {
 		// NOTE: The call to registerSynthConfigSource works well with the view.Diagnostics too since the view is
 		// configured in [Meta.initConfigLoader] with a callback to get the sources when it prints the diagnostics.
-		c.registerSynthConfigSource("<import-address>", traversalSrc) // so we can include a source snippet
+		c.configLoader().ForceFileSource("<import-address>", traversalSrc) // so we can include a source snippet
 		view.Diagnostics(diags)
 		view.InvalidAddressReference()
 		return 1
@@ -86,7 +87,7 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 	if addrDiags.HasErrors() {
 		// NOTE: The call to registerSynthConfigSource works well with the view.Diagnostics too since the view is
 		// configured in [Meta.initConfigLoader] with a callback to get the sources when it prints the diagnostics.
-		c.registerSynthConfigSource("<import-address>", traversalSrc) // so we can include a source snippet
+		c.configLoader().ForceFileSource("<import-address>", traversalSrc) // so we can include a source snippet
 		view.Diagnostics(diags)
 		view.InvalidAddressReference()
 		return 1
@@ -111,7 +112,7 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 		return 1
 	}
 
-	if !c.dirIsConfigPath(args.ConfigPath) {
+	if !c.configLoader().IsConfigDir(args.ConfigPath) {
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "No OpenTofu configuration files",
@@ -216,7 +217,7 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 	// Build the operation
 	opReq := c.Operation(ctx, b, view.Backend(), enc)
 	opReq.ConfigDir = args.ConfigPath
-	opReq.ConfigLoader, err = c.initConfigLoader()
+	opReq.ConfigLoader, err = configload.Initialize(c.configLoader())
 	if err != nil {
 		diags = diags.Append(tfdiags.Sourceless(
 			tfdiags.Error,
--- a/internal/command/import.go
+++ b/internal/command/import.go
@@ -75,8 +75,8 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 	traversal, travDiags := hclsyntax.ParseTraversalAbs(traversalSrc, "<import-address>", hcl.Pos{Line: 1, Column: 1})
 	diags = diags.Append(travDiags)
 	if travDiags.HasErrors() {
-		// NOTE: The call to registerSynthConfigSource works well with the view.Diagnostics too since the view is
-		// configured in [Meta.initConfigLoader] with a callback to get the sources when it prints the diagnostics.
+		// NOTE: The call to Loader.ForceFileSource works well with the view.Diagnostics too since the view is
+		// configured in [Meta.configLoader] with a callback to get the sources when it prints the diagnostics.
 		c.configLoader().ForceFileSource("<import-address>", traversalSrc) // so we can include a source snippet
 		view.Diagnostics(diags)
 		view.InvalidAddressReference()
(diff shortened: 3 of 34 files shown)


internal/backend/backend.go
```diff
@@ -267,7 +267,7 @@ type Operation struct {
 
 	// ConfigLoader is a configuration loader that can be used to load
 	// configuration from ConfigDir.
-	ConfigLoader *configload.Loader
```
yottta, line 276: `configload.Loader` is now an interface.

internal/command/apply.go
```diff
@@ -273,7 +274,7 @@ func (c *ApplyCommand) OperationRequest(
 	opReq.View = view.Operation()
 
 	var err error
-	opReq.ConfigLoader, err = c.initConfigLoader()
+	opReq.ConfigLoader, err = configload.Initialize(c.configLoader())
```
yottta: Before going into graph walks and goroutine based flows, we want to be sure that the config loader is initialised properly.
We don't want to mess around with the same diagnostic returned multiple times by the parallel node execution.

internal/command/meta.go
```diff
@@ -497,12 +504,6 @@ func isAutoVarFile(path string) bool {
 func (m *Meta) checkRequiredVersion(ctx context.Context) tfdiags.Diagnostics {
 	var diags tfdiags.Diagnostics
 
-	loader, err := m.initConfigLoader()
```
yottta, line 518: The error handling was deferred for later (several lines below). Multiple places like this can be seen in this PR.

internal/command/meta_config.go
```diff
@@ -179,40 +161,17 @@ func (m *Meta) loadSingleModuleWithTests(ctx context.Context, dir string, testDi
...
 
-// dirIsConfigPath checks if the given path is a directory that contains at
-// least one OpenTofu configuration file (.tf or .tf.json), returning true
-// if so.
-//
-// In the unlikely event that the underlying config loader cannot be initialized,
-// this function optimistically returns true, assuming that the caller will
-// then do some other operation that requires the config loader and get an
-// error at that point.
-func (m *Meta) dirIsConfigPath(dir string) bool {
```
yottta, line 207: This kind of default logic has been moved into `lazy.go`. For whoever will review it. Please double check that these deleted methods defaults are reflected in that file.

internal/configs/configload/lazy.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+// lazyLoader implements Loader. When created, it does not initialise the
+// actual implementation of the loader but instead defers that to the first
+// call to its methods.
+//
+// Each Loader method that returns no error will return a sane default that the system
+// is known to work with in case the underlying loader initialisation will fail during
+// those methods' calls.
+// If the initialisation fails during the call on these methods, the underlying loader will
```
yottta, line 25: The same question as in the PR description: should we cache the error once and do not attempt to initialise again every time a non-returning-error-method is called?

internal/configs/configload/loader.go
```diff
@@ -134,28 +166,12 @@ func (l *Loader) ImportSources(sources map[string][]byte) {
...
-// AllowLanguageExperiments specifies whether subsequent LoadConfig (and
-// similar) calls will allow opting in to experimental language features.
-//
-// If this method is never called for a particular loader, the default behavior
-// is to disallow language experiments.
-//
-// Main code should set this only for alpha or development builds. Test code
-// is responsible for deciding for itself whether and how to call this
-// method.
-func (l *Loader) AllowLanguageExperiments(allowed bool) {
```
yottta, line 161: Moved to `configload.Config` instead.

internal/configs/configload/loader.go
```diff
@@ -177,3 +181,43 @@ func (l *loader) ImportSourcesFromSnapshot(snap *Snapshot) {
...
+	if l.lastLoadedRoot == nil {
+		log.Printf("[ERROR] Unable to determine if module source is remote due to missing config load")
+		return nil
+	}
+
+	current := l.lastLoadedRoot
+	for _, part := range path {
+		child, childOk := current.Children[part]
+		if !childOk {
+			log.Printf("[ERROR] Unable to determine if module source is remote due to missing child")
```
apparentlymart: I don't understand how this function relates to "determine if the module source is remote"... was this just copied from the function above? 🤔

internal/command/import.go
```diff
@@ -86,7 +87,7 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 	if addrDiags.HasErrors() {
 		// NOTE: The call to registerSynthConfigSource works well with the view.Diagnostics too since the view is
 		// configured in [Meta.initConfigLoader] with a callback to get the sources when it prints the diagnostics.
-		c.registerSynthConfigSource("<import-address>", traversalSrc) // so we can include a source snippet
+		c.configLoader().ForceFileSource("<import-address>", traversalSrc) // so we can include a source snippet
```
apparentlymart: The comment above is still talking about `registerSyncConfigSource`, rather than `configLoader().ForceFileSource`.

internal/command/import.go
```diff
@@ -76,7 +77,7 @@ func (c *ImportCommand) Run(rawArgs []string) int {
 	if travDiags.HasErrors() {
 		// NOTE: The call to registerSynthConfigSource works well with the view.Diagnostics too since the view is
 		// configured in [Meta.initConfigLoader] with a callback to get the sources when it prints the diagnostics.
-		c.registerSynthConfigSource("<import-address>", traversalSrc) // so we can include a source snippet
+		c.configLoader().ForceFileSource("<import-address>", traversalSrc) // so we can include a source snippet
```
apparentlymart: The comment above is still talking about `registerSyncConfigSource`, rather than `configLoader().ForceFileSource`.

internal/configs/configload/lazy.go
apparentlymart: From what I read here it seems like caching the error would primarily be a performance thing (avoiding `NewLoader` repeatedly) rather than a correctness thing, right? From the caller's perspective the only difference would be how long a function takes to return and not what value it returns.

If that's true then I think what you have here right now is fine. We don't typically try to heavily optimize error paths.

I suppose it's possible that `NewLoader` could begin succeeding after it previously failed if another process is concurrently modifying files on disk, but that's also not a situation we typically try to handle robustly: if files are being modified on disk while a `tofu` process is working with them then various kinds of inconsistencies could happen already.

apparentlymart (commented):
The separation of concerns between `command.Meta`, `package configload`, and `package configs` has always been a little unclear, so thanks for starting to tidy that up!

internal/configs/configload/lazy.go
yottta: > From what I read here it seems like caching the error would primarily be a performance thing (avoiding NewLoader repeatedly) rather than a correctness thing, right?

Exactly. I was thinking about situations where the configuration is quite large. Though, checking again the logic performed during the initialisation, the error that could be returned comes from reading the modules manifest. Even though a configuration uses thousands of modules (which I don't think is a common encountering), the reading of such a file would still take an irrelevant amount of time.

> I suppose it's possible that `NewLoader` could begin succeeding after it previously failed if another process is concurrently modifying files on disk, but that's also not a situation we typically try to handle robustly: if files are being modified on disk while a `tofu` process is working with them then various kinds of inconsistencies could happen already.

Exactly this was my thinking of caching this error: if it failed once, the process should fail whenever an-error-returning-loader-method is called and the non-nil error is already loaded. Would make no sense to try loading again the configuration (even if it would succeed in a later call), since that could have already caused some divergent loaded information in separate parts of the system.

But I agree that all these reasons refer to situations with a really low probability of occurrence, so it feels more like a premature optimisation more than anything.

internal/configs/configload/loader.go
yottta: Both of these methods (`ModuleSourceAddrs` and `IsRemoteModuleSource`) were added in https://github.com/opentofu/opentofu/pull/3996/changes#diff-b56d379b07c52805ec95fdf7f8727c53967257ccf8d3d7f50ed2dbb4b0e89447R168.

Good catch. Fixed in 5f9533e40908223bccf4727e53336adf7c2109c9

internal/command/import.go
yottta: Fixed in e4c2ca9b294b2fdb8602c3c53d3b4a847f852dfb

internal/command/import.go
yottta: Fixed in e4c2ca9b294b2fdb8602c3c53d3b4a847f852dfb

internal/configs/configload/lazy_test.go
```diff
@@ -0,0 +1,607 @@
...
+	})
+	t.Run("fails when manifest is malformed", func(t *testing.T) {
+		d := t.TempDir()
+		writeConfigFile(t, filepath.Join(d, "main.tf"), []byte(""))
+		writeManifestFile(t, d, []byte("invalid content"))
+
+		ll := NewLazy(&Config{ModulesDir: d})
+		il, err := Initialise(ll)
+		if err == nil {
+			t.Fatalf("expected but got nothing: %s", err)
```
diofeher:
```suggestion
			t.Fatalf("expected but got nil")
```

Just to simplify as err is nil here

yottta: Fixed in 18ce789b9d0c33d52130a7afe0be64521d6caba7

apparentlymart (approved):

internal/configs/configload/lazy.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil
+	}
+	return l.Sources()
+}
+
+// LoadConfig implements Loader
+func (c *lazyLoader) LoadConfig(ctx context.Context, rootDir string, call configs.StaticModuleCall) (*configs.Config, hcl.Diagnostics) {
+	l, err := c.init()
+	if err != nil {
+		return nil, hcl.Diagnostics{
```
lbordowitz: Wrap this diagnostic in a function, and re-use that everywhere rather than repeating the same diag strings.

internal/command/meta_config.go
```diff
@@ -409,50 +362,39 @@ func (m *Meta) inputForSchema(given cty.Value, schema *configschema.Block, view
...
-// If the loader cannot be created for some reason then an error is returned
-// and no loader is created. Subsequent calls will presumably see the same
-// error. Loader initialization errors will tend to prevent any further use
-// of most OpenTofu features, so callers should report any error and safely
-// terminate.
-func (m *Meta) initConfigLoader() (*configload.Loader, error) {
-	if m.configLoader == nil {
-		loader, err := configload.NewLoader(&configload.Config{
-			ModulesDir: m.WorkingDir.ModulesDir(),
+// The configload.Loader that is initialised it's a lazy one, meaning that the
```
lbordowitz: "The configload.Loader is lazy-initialized, meaning that the"....

Side-Note: you are using the British spelling of initialise; since you've been consistent with that, just keep with it. I'll use "initialize" in some of my comments simply because that's what I'm used to, and my spell-checker complains otherwise.

lbordowitz (approved):
Looks great. The defaults of the deleted functions appear to be correctly captured here. I have a couple of nits, but nothing blocking.

internal/command/meta_config.go
yottta: Funny! 😄  I am aware of this issue I have for a long time and I still try to fix it, but I rarely think concisely to this.
My formal education was on British but after that, most of the words I learnt (initially wrote "learned" 🤣) were from American based content and now I have a soup of words in my head.
Thanks for pointing it out, this will help with the general awareness.

Fixed in fabbd1bce6cdc7f11c349a86803ad562395ffdaf

internal/configs/configload/lazy.go
yottta: Fixed in fabbd1bce6cdc7f11c349a86803ad562395ffdaf

PR #4120, opened 2026-05-20
removed dependency from legacy state files

KrishnaSindhur:
After this [PR](https://github.com/opentofu/opentofu/pull/4119), which is not useful because the whole `legacy/tofu` package is on track for removal.
The change will help to reduce the dependency of `legacy/tofu`

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/unlock_test.go
+++ b/internal/command/unlock_test.go
@@ -13,8 +13,8 @@ import (
 	"github.com/opentofu/opentofu/internal/backend/remote-state/inmem"
 	"github.com/opentofu/opentofu/internal/command/arguments"
 	"github.com/opentofu/opentofu/internal/command/workdir"
-
-	legacy "github.com/opentofu/opentofu/internal/legacy/tofu"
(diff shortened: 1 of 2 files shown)


yottta (approved):

internal/command/unlock_test.go
```diff
@@ -23,18 +23,18 @@ func TestUnlock(t *testing.T) {
...
 		f, err := os.Create(statePath)
 		if err != nil {
 			t.Fatalf("err: %s", err)
 		}
-		err = legacy.WriteState(legacy.NewState(), f)
-		f.Close()
-		if err != nil {
+		sf := statefile.New(states.NewState(), "test-lineage", 1)
+		if err := statefile.WriteForTest(sf, f); err != nil {
+			f.Close()
```
yottta: You can replace these two `f.Close()` statements with one `defer f.Close()` before this `if` block.

lbordowitz (approved):

PR #4084, opened 2026-05-08
communicator/ssh: Remove private key and certificate material from error messages

SebTardif:
## What

`signCertWithPrivateKey()` in the SSH communicator formats private key
content, certificate content, and parsed key/signer structs into error
messages using `%q`. These error strings propagate to terminal output,
CI/CD logs, and cloud platform UIs, exposing credential material to
anyone with log access.

## How it leaks

The four `fmt.Errorf` calls in `signCertWithPrivateKey` include sensitive
values via `%q`:

```go
// Line 403 - full PEM-encoded private key
return nil, fmt.Errorf("failed to parse private key %q: %w", pk, err)

// Line 412 - full SSH certificate content
return nil, fmt.Errorf("failed to parse certificate %q: %w", certificate, err)

// Line 433 - parsed raw private key struct (*rsa.PrivateKey, *ecdsa.PrivateKey, etc.)
// Go's %q on a struct prints all fields including private exponents and primes
return nil, fmt.Errorf("failed to create signer from raw private key %q: %w", rawPk, err)

// Line 438 - SSH signer derived from the private key
return nil, fmt.Errorf("failed to create cert signer %q: %w", usigner, err)
```

When a user provides a malformed key or certificate in their `connection`
block, the parse error propagates up through the provisioner to:
- Terminal output (visible in scrollback and screen recordings)
- CI/CD logs (GitHub Actions, Jenkins, GitLab CI, stored indefinitely)
- Cloud platform UIs (Terraform Cloud, Spacelift, env0, Scalr)

## Fix

Remove the sensitive data from all four format strings. The wrapped error
(`%w`) already contains the actionable diagnostic from `golang.org/x/crypto/ssh`
(e.g., "ssh: no key found", "ssh: invalid openssh private key format").
The key material itself adds no debugging value since the user already
knows which key they configured.

The sibling function `readPrivateKey()` was already safe (it only formats
`err`, not the key content).

## Related

Same fix on the Terraform side:
- [hashicorp/terraform#38543](https://github.com/hashicorp/terraform/pull/38543) - Redact private key material from SSH error messages

Related redaction work:
- [hashicorp/terraform#30067](https://github.com/hashicorp/terraform/pull/30067) - Redact sensitive values from function errors
- [moby/moby#33884](https://github.com/moby/moby/pull/33884) - Redact secret data on "secret create"

Introduced in #414.

## Checklist

[...]

PR #4121, opened 2026-05-21
Fix file descriptor leaks in provisioners and commands

SebTardif:
## What this PR does

Fixes four file descriptor leaks across provisioners and commands:

1. **`local-exec/resource_provisioner.go`**: `os.Pipe()` creates pr (read) and pw (write). Only pw.Close() was called; pr leaked one OS FD per local-exec invocation. Added `defer pr.Close()`.

2. **`command/login.go`**: `defer resp.Body.Close()` placed after `io.ReadAll`; early return on ReadAll error leaked the HTTP body. Moved defer immediately after the response error check.

3. **`command/workspace_new.go`**: `os.Open(statePath)` had no close; file handle leaked for the duration of the function. Added `defer f.Close()`.

4. **`provisioners/file/resource_provisioner.go`**: `os.CreateTemp` in `getSrc()` never closed the file handle. Added explicit close before returns.

These are the same class of bugs fixed in hashicorp/terraform PR #38585.

## AI Disclosure

Developed with AI assistance (Grok by xAI) in a human-in-the-loop workflow. All code reviewed and verified by the human author.

Signed-off-by: Sebastien Tardif <sebtardif@ncf.ca>

### DIFF
--- a/internal/builtin/provisioners/file/resource_provisioner.go
+++ b/internal/builtin/provisioners/file/resource_provisioner.go
@@ -162,9 +162,11 @@ func getSrc(v cty.Value) (string, bool, error) {
 		}
 
 		if _, err = file.WriteString(content.AsString()); err != nil {
+			file.Close()
 			return "", true, err
 		}
 
+		file.Close()
 		return file.Name(), true, nil
 
 	case !src.IsNull():
--- a/internal/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/internal/builtin/provisioners/local-exec/resource_provisioner.go
@@ -167,6 +167,7 @@ func (p *provisioner) ProvisionResource(ctx context.Context, req provisioners.Pr
 		))
 		return resp
 	}
+	defer pr.Close()
 
 	var cmdEnv []string
 	cmdEnv = os.Environ()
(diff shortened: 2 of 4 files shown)


Yantrio:
Hey @SebTardif, thanks again for the contribution, but I'm going to have to close this one.

Two things:

1. **No accepted issue.** I [flagged this on your last PR](https://github.com/opentofu/opentofu/pull/4084#issuecomment-4407429841). We request that you have an `accepted` issue before work starts, so we can discuss scope and decide whether it's something we want to take. I let #4084 through because it was tiny and self-contained, but the rule is there for a reason and we shouldn't keep waving it through. If you think these FD leaks are worth fixing, please open an issue first so we can triage it and discuss it together.

2. **AI-assisted submissions.** Your disclosure says this was developed with Grok. Our [DEVELOPING.md](https://github.com/opentofu/opentofu/blob/main/contributing/DEVELOPING.md) mentions this this and we are currently working on making this more explicit.
We ask contributors to avoid LLM-based coding assistants, because the training data almost certainly includes BSL-licensed Terraform code and we have no way to verify what's original vs. regurgitated.

You also removed the pull request template where you would declare that you wrote this without using AI assistance.

We appreciate that you disclosed it up front, but for both reasons above I'm going to close this.

PR #4128, opened 2026-05-22
2453 copywrite tool inaccurate

### LINKED ISSUE #2453: Copywrite (tool) is inaccurate
Our current copyright headers look like this:

```
// Copyright (c) The OpenTofu Authors
```

However, in certain situations [copywrite](https://github.com/hashicorp/copywrite), the tool we use to update copyright headers, fails to update old-style copyright headers like this:

```
// Copyright (c) HashiCorp, Inc.
```

The reason for this, as @ollevche found, is likely the [fuzzy logic here](https://github.com/hashicorp/copywrite/blob/9d021bf61a094a5eac6ae3084ceed2dda4700a73~1/addlicense/main.go#L449).

> [!NOTE]
> **Help wanted:** If you have a suggestion for a good replacement, please post in the comments.

josephedward:
Resolves #2453

## Replace `copywrite` with `nwa` (check) + `reuse` (fix) for correct dual-copyright enforcement

The hashicorp copyright uses fuzzy matching logic to detect existing copyright headers. This causes it to silently skip files that still carry the old-style `// Copyright (c) HashiCorp, Inc.` header. It does not update them, and it does not fail the CI check, leaving non-compliant headers undetected. This PR replaces copywrite with nwa (https://github.com/B1NARY-GR0UP/nwa)  for copyright header validation, and reuse (https://reuse.software/) for adding/updating headers.
  
### nwa
  
nwa validates headers against an exact template (.nwa-header.txt) rather than matching. Any file that doesn't carry the precise dual-header block:

```
// Copyright (c) The OpenTofu Authors
```
will fail CI.
  
### reuse
  
  reuse (via the fsfe/reuse Docker image) is used in scripts/add-copyright-headers.sh to add or update headers. It can **annotate --merge-copyrights** to handle both cases in a single pass. Files with only the old HashiCorp-only header,  the OpenTofu line is prepended. Files with no header at all, the full dual header is inserted. This replaces the previous copywrite headers invocation which would silently fail on the older headers. 

### Changes

- `scripts/add-copyright-headers.sh` -  Replaced `go tool copywrite` with `docker run fsfe/reuse annotate --merge-copyrights` 
- `.github/workflows/checks.yml` - Replaced `copywrite headers --plan` with `nwa check --rawtmpl` 
- `.nwa-header.txt` - Exact header template used by `nwa check` 
- `.reuse/templates/opentofu.jinja2` - Jinja2 template used by `reuse annotate` 
- `go.mod` / `go.sum` - Swapped `github.com/hashicorp/copywrite` for `github.com/B1NARY-GR0UP/nwa` in `tool {}` block 
- `internal/configs/removed_test.go` - Fixed the one file currently carrying only the old HashiCorp-only header 

### Testing locally

```bash
# Fix headers (requires Docker)
./scripts/add-copyright-headers.sh

# Check headers (requires Go toolchain only)
go tool github.com/B1NARY-GR0UP/nwa check \
  --rawtmpl .nwa-header.txt \
  --skip "**/*.pb.go" \
  --skip "**/*_string.go" \
  --skip "**/mock*.go" \
  "**/*.go"
```
  
  Checklist
  
  - [x] I have read the contribution guide (https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md).
  
  Go checklist
  

yottta (changes requested):
Thanks a lot for the submission.

Though, please revisit and rebase the PR correctly because it seems that it was created from an older commit. It contains unrelated changes added in recent PRs.

josephedward:
> Thanks a lot for the submission.
> 
> Though, please revisit and rebase the PR correctly because it seems that it was created from an older commit. It contains unrelated changes added in recent PRs.

Sorry about that, accident. Fixed.

yottta:
Hi @josephedward,

I took a look at this when you submitted it but sidetracked with more urgent matters.
Thanks for the submitted changes. Looking more in-depth at this approach, I am not really keen on having this accepted.
The amount of dependencies that are pulled into that can create security risks is a little bit overwhelming, especially for a job like this.

I was checking also the https://github.com/google/addlicense tool that feels way more lightweight and more suitable for this.

These are just my thoughts on this. I will have a conversation with the maintainers about this and will be back with a more joint feedback.

Thanks a lot for the work here and that you started this conversation that was kind of stale.

josephedward:
> I was checking also the https://github.com/google/addlicense tool that feels way more lightweight and more suitable for this.

Sure, @yottta, the concern about dependencies makes sense. From doing a bit of research `addlicense` is upstream of `copywrite`, they share the same root. So `removed_test.go` would be skipped by `addlicense` for the same reason it's skipped by `copywrite` - it only checks for the presence of a license line, not who it belongs to. So you could keep `copywrite` since `copywrite.hcl` is already configured - use it for `add-copyright-headers.sh` (it works fine for adding), but then could add a grep guard to flag files that mention “HashiCorp" but not “The OpenTofu Authors”: 

```
- name: "Copyright headers"
  run: |
    # Guard: catch files with only the old HashiCorp-only header
    bad_files=""
    while IFS= read -r f; do
      if ! grep -q "Copyright (c) The OpenTofu Authors" "$f"; then
        bad_files="$bad_files\n  $f"
      fi
    done < <(grep -rl --include="*.go" "Copyright (c) HashiCorp, Inc\." . || true)

    if [[ -n "$bad_files" ]]; then
      echo >&2 "ERROR: the following files have only the old HashiCorp copyright header:"
      printf >&2 '%b\n' "$bad_files"
      echo >&2 "Run scripts/add-copyright-headers.sh to fix."
      exit 1
    fi

    go tool github.com/hashicorp/copywrite headers --plan
    if [[ $? != 0 ]]; then
      echo >&2 "ERROR: some files are missing required copyright headers. Run scripts/add-copyright-headers.sh locally and then commit the updated files."
      exit 1
    fi
```

The PR would be: 
- Fix internal/configs/removed_test.go
- Add the grep guard to the existing CI step

This way, no new tools, no Python, no dep changes. I considered submitting something like that but long bash scripts in CI make me nervous, but for a repo with one broken file and a pattern that's unlikely to recur often, it's probably ok. And it looks like I could run [shellcheck](https://www.shellcheck.net/), which I think is available on the runners, and/or maybe inline the grep guard as a go script. Want me to try that?

yottta:
> > I was checking also the https://github.com/google/addlicense tool that feels way more lightweight and more suitable for this.
> 
> Sure, @yottta, the concern about dependencies makes sense. From doing a bit of research `addlicense` is upstream of `copywrite`, they share the same root. So `removed_test.go` would be skipped by `addlicense` for the same reason it's skipped by `copywrite` - it only checks for the presence of a license line, not who it belongs to. So you could keep `copywrite` since `copywrite.hcl` is already configured - use it for `add-copyright-headers.sh` (it works fine for adding), but then could add a grep guard to flag files that mention “HashiCorp" but not “The OpenTofu Authors”:
> 
> ```
> - name: "Copyright headers"
>   run: |
>     # Guard: catch files with only the old HashiCorp-only header
>     bad_files=""
>     while IFS= read -r f; do
>       if ! grep -q "Copyright (c) The OpenTofu Authors" "$f"; then
>         bad_files="$bad_files\n  $f"
>       fi
>     done < <(grep -rl --include="*.go" "Copyright (c) HashiCorp, Inc\." . || true)
> 
>     if [[ -n "$bad_files" ]]; then
>       echo >&2 "ERROR: the following files have only the old HashiCorp copyright header:"
>       printf >&2 '%b\n' "$bad_files"
>       echo >&2 "Run scripts/add-copyright-headers.sh to fix."
>       exit 1
>     fi
> 
>     go tool github.com/hashicorp/copywrite headers --plan
>     if [[ $? != 0 ]]; then
>       echo >&2 "ERROR: some files are missing required copyright headers. Run scripts/add-copyright-headers.sh locally and then commit the updated files."
>       exit 1
>     fi
> ```
> 
> The PR would be:
> 
> * Fix internal/configs/removed_test.go
> * Add the grep guard to the existing CI step
> 
> This way, no new tools, no Python, no dep changes. I considered submitting something like that but long bash scripts in CI make me nervous, but for a repo with one broken file and a pattern that's unlikely to recur often, it's probably ok. And it looks like I could run [shellcheck](https://www.shellcheck.net/), which I think is available on the runners, and/or maybe inline the grep guard as a go script. Want me to try that?

If you are willing to give it a try, that would save us a lot of time. Thanks a lot for your willingness to help with this! 🙌

josephedward:
Ok, so after feedback:

- Removed the nwa, reuse approaches 
- Added scripts/check-copyright-headers.sh:
   - finds *.go and *.proto files
   - skips generated files
   - checks that the first lines exactly match the dual header for OpenTofu + HashiCorp
   - fails with a list of offending files if not
- Updated `.github/workflows/checks.yml` to:
   - run shellcheck on the two shell scripts
   - call `bash scripts/check-copyright-headers.sh` instead of `go tool github.com/hashicorp/copywrite headers —plan`


The end result is CI catches files that are missing the OpenTofu header or still carry only the old HashiCorp‑only header, and it can be reproduced locally (`bash scripts/check-copyright-headers.sh`).No new Go or Docker dependencies are introduced; everything is implemented as a script.

josephedward:
Closing for #4233, which implements the org-level shared workflow approach. Happy to help roll out to the rest of the org's repos.

Issue #4131, opened 2026-05-23
Destroy-time provisioners cannot access Ephemeral values, preventing zero-state secret cleanup patterns


mulder999:
### Community note

> [!TIP]
> 👋 Hi there, OpenTofu community! The OpenTofu team prioritizes issues based on upvotes. Please make sure to upvote this issue and describe how it affects you in detail in the comments to show your support.

### OpenTofu Version

```shell
OpenTofu v1.11.5
on linux_amd64
```

### OpenTofu Configuration Files

```hcl
variable "api_secret" {
  type      = string
  ephemeral = true
}

resource "terraform_data" "custom_record" {
  triggers_replace = {
    id = "12345"
  }

  # This works perfectly on create
  provisioner "local-exec" {
    command = "curl -X POST -d '{\"key\":\"${var.api_secret}\"}' https://api.example.com/create"
  }

  # This fails compilation due to referencing var.api_secret outside of 'self'
  provisioner "local-exec" {
    when    = destroy
    command = "curl -X POST -d '{\"key\":\"${var.api_secret}\"}' https://api.example.com/delete"
  }
}
```

### Debug Output

not relevant for a conceptual flaw

### Expected Behavior

The purpose of ephemeral values is to keep secrets out of the state file on disk while keeping them available in system memory during runtime execution. Because a destroy plan is a live execution phase that has access to current module input variables, destroy-time provisioners should have a mechanism to read active ephemeral context inputs.

### Actual Behavior

1. If you reference an ephemeral variable or local containing an ephemeral value directly in the destroy command, the engine throws: Error: Invalid reference from destroy provisioner because destroy-time provisioners can only reference attributes via self.

2. If you try to map the ephemeral variable into triggers_replace so it can be read via self.triggers_replace.my_key, the engine throws: Error: Ephemeral value used in non-ephemeral context because triggers_replace persists to the state file on disk.

This leaves the engine in a catch-22 state: It is impossible to use ephemeral credentials to authenticate a deletion hook.

### Steps to Reproduce

Conceptual flaw

yottta:
Thanks for the submission @mulder999,

Looked into this a little bit and yes, this is doable. 
But, even though the macro reason is clear on why destroy time provisioners are not allowed to reference the non-self blocks, there are other micro reasons that elude me right now. Maintainers will discuss this soon and we will let you know about the decision on this one.

---

[...]

PR #4132, opened 2026-05-25
[WIP] Allow variables in destroy provisioners

yottta:
These changes provides a POC on how we could allow the variables to be accessible in the evaluation context of the destroy time provisioners.

Even though the currently provided changes do make this work, the reason for why this limitation has been introduced initially still needs to be understood in-depth and ensure that there are no unwanted side effects and behaviors introduced by such changes.

The initial limitation has been introduced in https://github.com/opentofu/opentofu/commit/8547603ff5645edcdcaf2f49fd8a6d18e4cda3b3

NOTE: Even though the current changes work for a particular use case (the configuration shared in #4131), we have to dig deeper to be sure that we cover also edge cases.

Possible solution for #4131 

Resolves # 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/configs/provisioner.go
+++ b/internal/configs/provisioner.go
@@ -163,7 +163,7 @@ func onlySelfRefs(body hcl.Body) hcl.Diagnostics {
 		for _, v := range attr.Expr.Variables() {
 			valid := false
 			switch v.RootName() {
-			case "self", "path", "terraform", "tofu":
+			case "self", "path", "terraform", "tofu", "var":
 				valid = true
 			case "count":
 				// count must use "index"
--- a/internal/lang/eval.go
+++ b/internal/lang/eval.go
@@ -119,6 +119,7 @@ func (s *Scope) EvalSelfBlock(ctx context.Context, body hcl.Body, self cty.Value
 
 	terraformAttrs := map[string]cty.Value{}
 	pathAttrs := map[string]cty.Value{}
+	varAttrs := map[string]cty.Value{}
 
 	// We could always load the static values for Path and Terraform values,
 	// but we want to parse the references so that we can get source ranges for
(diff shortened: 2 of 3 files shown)


internal/configs/provisioner.go
```diff
@@ -163,7 +163,7 @@ func onlySelfRefs(body hcl.Body) hcl.Diagnostics {
 		for _, v := range attr.Expr.Variables() {
 			valid := false
 			switch v.RootName() {
-			case "self", "path", "terraform", "tofu":
+			case "self", "path", "terraform", "tofu", "var":
```
yottta, line 166: The reference needs to be allowed from the config layer

internal/lang/eval.go
```diff
@@ -119,6 +119,7 @@ func (s *Scope) EvalSelfBlock(ctx context.Context, body hcl.Body, self cty.Value
 
 	terraformAttrs := map[string]cty.Value{}
 	pathAttrs := map[string]cty.Value{}
+	varAttrs := map[string]cty.Value{}
```
yottta, line 122: In the destroy time provisioner context, we want to inject also the module variables.
All the changes in this particular file are about this.

internal/tofu/node_resource_abstract.go
```diff
@@ -270,6 +270,23 @@ func (n *NodeAbstractResource) DestroyReferences() []*addrs.Reference {
 				refs, _ := lang.ReferencesInExpr(addrs.ParseRef, c.Managed.PreventDestroy)
 				result = append(result, refs...)
 			}
+
+			for _, p := range c.Managed.Provisioners {
```
yottta, line 274: In order to create the required edge between the resource and variable nodes, when the destroy time references are requested, the resource needs to check also the references inside the destroy time provisioner block.

yottta:
Closing since this was more of an experiment and in fact would break other parts of the system by allowing to create cyclic dependencies between resources.

PR #4133, opened 2026-05-25
Store replace triggered by schema in node

### LINKED ISSUE #4129: OpenTofu 1.12: `replace_triggered_by` validates attribute traversal against wrong resource schema
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.12.0
on darwin_arm64
```

```hcl
terraform {
  required_version = "1.12.0"
  required_providers {
    random = {
      source  = "hashicorp/random"
      version = "~> 3.6"
    }
  }
}

resource "random_integer" "dependency" {
  min = 1
  max = 100
}

resource "terraform_data" "dependee" {
  lifecycle {
    replace_triggered_by = [
      random_integer.dependency.result
    ]
  }
}
```

### Debug Output

https://gist.github.com/zhimsel/9efcd7d519eeaddcf2fc21935f5ba810

### Expected Behavior

No errors on `tofu validate` or `tofu plan`

### Actual Behavior

```console
$ tofu validate
│
│   on main.tf line 23, in resource "terraform_data" "dependee":
│   23:       random_integer.dependency.result
│
│ This object has no argument, nested block, or exported attribute named "result".
```

_Note: the same happens for `tofu plan` as well._

### Steps to Reproduce

1. Save the configuration above as `main.tf`
1. `tofu init`
1. `tofu validate` (alternatively: `tofu plan` or `tofu apply`)

### Additional Context

This is a regression from 1.11, where the same configuration validates and plans successfully. Introduced in #3967.

From what I can tell, the issue is that the resource validation is validating the attributes of `replace_triggered_by` targets against the _caller's_ schema, not the _targets_.

Specifically: `schemaForType` is set to the schema of the resource containing the `lifecycle` block (`terraform_data`, in the case of my example above): https://github.com/opentofu/opentofu/blob/v1.12.0/internal/tofu/node_resource_validate.go#L288

However, `ref.Remaining` (the attribute part of the `replace_triggered_by` traversal) is _validated against `schemaForType`_: https://github.com/opentofu/opentofu/blob/v1.12.0/internal/tofu/node_resource_validate.go#L369

Instead, it should look up the schema for `ref.Subject`'s resource type instead of reusing `schemaForType`.

> [!NOTE]
> The bug is masked when both resources are the same type (e.g. `terraform_data` -> `terraform_data`), because the wrong schema _happens to_ contain the referenced attribute. Simple, same-type repros won't surface the issue.

### References

- #3967 
- #3970

yottta:
When the `replace_triggered_by` references another resource than the one that is configured with that `lifecycle` argument, we need to use the targeted resource schema to validate the referenced attribute.

~⚠️ This is WIP because the implementation provided here in the `AttachSchemaTransformer` will impact the performance of large configurations, since before it was O(N) and now is O(N^2). That's because there is no way in this particular transformer to get only the nodes that are referenced by the node configured with the `replace_triggered_by`.~

~Before trying to improve the logic of this one, will try another approach in another PR.~
~ℹ️ Marked as WIP as we want to decide between this approach or the one provided in #4134.~

Resolves #4129

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/node_resource_abstract.go
+++ b/internal/tofu/node_resource_abstract.go
@@ -119,18 +119,19 @@ type NodeAbstractResource struct {
 }
 
 var (
-	_ GraphNodeReferenceable             = (*NodeAbstractResource)(nil)
-	_ GraphNodeReferencer                = (*NodeAbstractResource)(nil)
-	_ GraphNodeProviderConsumer          = (*NodeAbstractResource)(nil)
-	_ GraphNodeProvisionerConsumer       = (*NodeAbstractResource)(nil)
-	_ GraphNodeConfigResource            = (*NodeAbstractResource)(nil)
-	_ GraphNodeAttachResourceConfig      = (*NodeAbstractResource)(nil)
-	_ GraphNodeAttachResourceSchema      = (*NodeAbstractResource)(nil)
-	_ GraphNodeAttachProvisionerSchema   = (*NodeAbstractResource)(nil)
-	_ GraphNodeAttachProviderMetaConfigs = (*NodeAbstractResource)(nil)
-	_ GraphNodeTargetable                = (*NodeAbstractResource)(nil)
-	_ graphNodeAttachResourceDependsOn   = (*NodeAbstractResource)(nil)
-	_ dag.GraphNodeDotter                = (*NodeAbstractResource)(nil)
+	_ GraphNodeReferenceable                  = (*NodeAbstractResource)(nil)
+	_ GraphNodeReferencer                     = (*NodeAbstractResource)(nil)
+	_ GraphNodeProviderConsumer               = (*NodeAbstractResource)(nil)
+	_ GraphNodeProvisionerConsumer            = (*NodeAbstractResource)(nil)
+	_ GraphNodeConfigResource                 = (*NodeAbstractResource)(nil)
+	_ GraphNodeAttachResourceConfig           = (*NodeAbstractResource)(nil)
+	_ GraphNodeAttachResourceSchema           = (*NodeAbstractResource)(nil)
+	_ GraphNodeAttachReplaceTriggeredBySchema = (*NodeAbstractResource)(nil)
+	_ GraphNodeAttachProvisionerSchema        = (*NodeAbstractResource)(nil)
+	_ GraphNodeAttachProviderMetaConfigs      = (*NodeAbstractResource)(nil)
+	_ GraphNodeTargetable                     = (*NodeAbstractResource)(nil)
+	_ graphNodeAttachResourceDependsOn        = (*NodeAbstractResource)(nil)
+	_ dag.GraphNodeDotter                     = (*NodeAbstractResource)(nil)
 )
 
 // NewNodeAbstractResource creates an abstract resource graph node for
--- a/internal/tofu/node_resource_abstract.go
+++ b/internal/tofu/node_resource_abstract.go
@@ -85,7 +85,7 @@ type NodeAbstractResource struct {
 
 	ProvisionerSchemas map[string]*configschema.Block
 
-	ReplaceTriggeredBySchemas map[string]*configschema.Block
+	ReplaceTriggeredBySchemas map[addrs.Resource]*configschema.Block
 
 	// Set from GraphNodeTargetable
 	Targets []addrs.Targetable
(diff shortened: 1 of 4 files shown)


internal/tofu/node_resource_abstract.go
```diff
@@ -470,6 +472,29 @@ func (n *NodeAbstractResource) AttachProvisionerSchema(name string, schema *conf
 	n.ProvisionerSchemas[name] = schema
 }
 
+// GraphNodeAttachReplaceTriggeredBySchema
```
cam72cam, line 476: TODO: add type assertion to assertions block above

internal/tofu/transform_attach_schema.go
```diff
@@ -114,5 +130,29 @@ func (t *AttachSchemaTransformer) Transform(ctx context.Context, g *Graph) error
...
+					refAddr = rs
+				case addrs.ResourceInstance:
+					refAddr = rs.Resource
+				default:
+					continue
+				}
+				schema, ok := refMap[refAddr]
+				if ok {
+					log.Printf("[TRACE] AttachSchemaTransformer: attaching replace_triggered_by %q config schema to %s", refAddr, dag.VertexName(v))
+					tv.AttachReplaceTriggeredBySchema(*ref, schema)
```
apparentlymart: It's safe to assume that all instances of a particular resource must always have the same schema, because all instances of a resource must have the same resource type.

Therefore to get more compile-time checking and make the relationships easier to understand I suggest changing the signature of `AttachReplaceTriggeredBySchema` so that the first argument is an `addrs.Resource` rather than a general `addrs.Reference`.

Then the map inside the graph node can be of type `map[addrs.Resource]*configschema.Block` instead of `map[string]*configschema.Block`, making it explicit exactly what kind of address we're expecting to use as the key there.

(There have historically been a number of bugs where we used a string representation of an address as a key and then misunderstood what kind of address it was in later work, especially because the string representations of `addrs.Resource` and `addrs.ResourceInstance` are equal for a single-instance resource and that's the situation we are most likely to use in unit tests.)

internal/tofu/node_resource_validate.go
```diff
@@ -364,9 +364,18 @@ func (n *NodeValidatableResource) validateResource(ctx context.Context, evalCtx
 					continue
 				}
 
+				// If there are schemas registered for the replace_triggered_by references, try to use one of those because
+				// the resource referenced can be different than the current one.
+				refSchema := schemaForType.Block
```
apparentlymart: I don't think it's correct to default to `schemaForType.Block` if there's no element present in `n.ReplaceTriggeredBySchemas`, because there's no reason to believe that the target of the reference is more likely to be of the same type as the current resource.

We should probably treat that situation as a "this is a bug in OpenTofu' error instead, because if the schema transformer is written correctly then it should not be possible for the schema to be missing by the time we get here and it would be better for us to learn about it via a bug report if it _does_ happen.

internal/tofu/transform_attach_schema.go
cam72cam: Addressed in [bd7083b](https://github.com/opentofu/opentofu/pull/4133/commits/bd7083bcb8d06ef68f614a7297e855dd1aa6be59)

internal/tofu/node_resource_validate.go
cam72cam: Addressed in [bd7083b](https://github.com/opentofu/opentofu/pull/4133/commits/bd7083bcb8d06ef68f614a7297e855dd1aa6be59)

internal/tofu/node_resource_validate.go
```diff
@@ -360,15 +360,32 @@ func (n *NodeValidatableResource) validateResource(ctx context.Context, evalCtx
...
-				// it only checks the resource type and its primary attributes.
-				remainingDiags := schemaForType.Block.StaticValidateTraversal(ref.Remaining)
-				if remainingDiags.HasErrors() {
-					diags = diags.Append(remainingDiags)
+				refSchema, ok := n.ReplaceTriggeredBySchemas[refAddr]
+				if !ok {
+					diags = diags.Append(&hcl.Diagnostic{
+						Severity: hcl.DiagError,
+						Summary:  "Resource referenced in replace_triggered_by not declared",
+						Detail:   fmt.Sprintf("Resource %s references %s in replace_triggered_by, but it has not been declared.", n.Addr, refAddr),
```
apparentlymart:
```suggestion
						Summary:  "Undeclared resource in replace_triggered_by",
						Detail:   fmt.Sprintf("Resource %s refers to %s in replace_triggered_by, but it is not declared in this module.", n.Addr, refAddr),
```

This started as just my usual annoying tweak to avoid ambiguity in whether "references" is the plural noun or the verb, since we very often use the word "references" as a plural noun elsewhere in OpenTofu, but while I was here anyway I also did some other fiddling.

In particular, "not declared in this module" is something mentioned in some other similar error messages because newcomers don't tend to understand immediately that each module has its own namespace.

apparentlymart: It's interesting that we'd also get to this error message if there were a bug in the schema transformer that made it fail to register the schema. Since that's a "should never happen" case I think that's fine but maybe we could leave a comment just above to note that possibility for a future maintainer:

```go
// (We could also get here if [AttachSchemaTransformer] has a bug that
// makes it fail to register one of the needed schemas.)
```

apparentlymart (approved):
Some minor things inline but overall this seems like a good compromise!

internal/tofu/transform_attach_schema.go
```diff
@@ -132,21 +133,22 @@ func (t *AttachSchemaTransformer) Transform(ctx context.Context, g *Graph) error
...
 		if tv, ok := v.(GraphNodeAttachReplaceTriggeredBySchema); ok {
+			module := tv.ResourceAddr().Module
 			refs := tv.ReplaceTriggeredBy()
 			for _, ref := range refs {
-				var refAddr addrs.Resource
+				var refAddr addrs.ConfigResource
 				switch rs := ref.Subject.(type) {
 				case addrs.Resource:
-					refAddr = rs
+					refAddr = rs.InModule(module)
```
yottta, line 142: Thanks for the `InModule` change! 🙌 
This removes from my doubts with the previous approach!

PR #4134, opened 2026-05-25
[WIP] New transformer that stores replace_triggered_by schema in the configured node

### LINKED ISSUE #4129: OpenTofu 1.12: `replace_triggered_by` validates attribute traversal against wrong resource schema

yottta:
When the replace_triggered_by references another resource than the one that is configured with that lifecycle argument, we need to use the targeted resource schema to validate the referenced attribute.

The newly introduce transformer (ReplaceTriggeredBySchemaTransformer), needs to be executed after the `ReferenceTransformer` since it relies on the existence of references between the node configured with the `replace_triggered_by` and the node that it references. Without having these references already in place, we would have to iterate over all the nodes in the graph to find the one that it points to.

ℹ️ Marked as WIP as we want to decide between this approach or the one provided in #4133.

Resolves #4129

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/transform_replace_triggered_by_schema.go
+++ b/internal/tofu/transform_replace_triggered_by_schema.go
@@ -1,0 +1,94 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package tofu
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/configs/configschema"
+	"github.com/opentofu/opentofu/internal/dag"
+)
+
+// GraphNodeAttachReplaceTriggeredBySchema is an interface implemented by node types
+// that need one or more provider schemas attached for the replace_triggered_by.
+type GraphNodeAttachReplaceTriggeredBySchema interface {
+	ReplaceTriggeredBy() []*addrs.Reference
+
+	// AttachReplaceTriggeredBySchema is called during transform for each resource type
+	// type returned from ReplaceTriggeredBy, providing the configuration schema
+	// for each referenced provider in turn. The implementer should save these for
+	// later use in evaluating replace_triggered_by references.
+	AttachReplaceTriggeredBySchema(ref addrs.Reference, schema *configschema.Block)
+}
+
+// ReplaceTriggeredBySchemaTransformer checks the GraphNodeAttachReplaceTriggeredBySchema nodes and if any
+// reports references from the `replace_triggered_by` expressions, this tries to identify the schema
+// of the referenced resource type and attach it to the node containing the reference.
+// This way, the node configured with `replace_triggered_by` can validate the expression against the
+// actual targeted resource schema.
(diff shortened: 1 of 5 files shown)


internal/tofu/transform_replace_triggered_by_schema.go
```diff
@@ -0,0 +1,81 @@
...
+				if !ok {
+					continue
+				}
+				addr := crn.ResourceAddr().Resource
+				pcn, ok := d.(GraphNodeProviderConsumer)
+				if !ok {
+					log.Printf("[TRACE] ReplaceTriggeredBySchemaTransformer: skip processing %s since it's not GraphNodeProviderConsumer", d)
+					continue
+				}
+				deps[addr] = pcn
```
yottta, line 61: Isn't this a bug? What if there is the same resource type and the same resource name in 2 different modules (root and child module)?
I might be wrong here, but this might be an issue 🤔

cam72cam:
Closing in favor of https://github.com/opentofu/opentofu/pull/4133

yottta:
> Closing in favor of #4133

Just to provide additional context for the future: in a discussion maintainers had, we decided to go with the option where we change an existing transformer, since adding a new one (what was this PR all about) can introduce also unwanted side effects.

PR #4153, opened 2026-05-26
website: Update the documentation on "import" blocks

apparentlymart:
During the v1.12 development period we added support for the new "identity" argument but forgot to update the docs to mention it.

While adding that I also took the opportunity to revise some of the other text on this page, since much of it was using the term "resource" inconsistently to refer both to the objects in the remote system and to the declarations in the OpenTofu configuration. This now consistently refers to the objects in the remote system as "objects" and uses "resource" only to refer to OpenTofu's concept with that name.

This uses explicit [heading IDs](https://docusaurus.io/docs/next/markdown-features/toc#heading-ids) to preserve the anchor names previously associated with these headings even though the heading text is now changing.

If merged, this will also need to be backported to the `v1.12` branch in a subsequent PR.

---

### DIFF
--- a/website/docs/language/import/index.mdx
+++ b/website/docs/language/import/index.mdx
@@ -9,11 +9,11 @@ description: >-
 While we do not expect to make backwards-incompatible changes to syntax, the `-generate-config-out` flag and how OpenTofu processes imports during the plan stage and generates configuration may change in future releases.
 :::
 
-Use the `import` block to import existing infrastructure resources into OpenTofu, bringing them under OpenTofu's management. Unlike the `tofu import` command, configuration-driven import using `import` blocks is predictable, works with CICD pipelines, and lets you preview an import operation before modifying state.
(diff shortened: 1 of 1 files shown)


yottta (approved):

website/docs/language/import/index.mdx
```diff
@@ -34,44 +36,62 @@ resource "aws_instance" "example" {
...
-- `for_each` (optional) - Import several resources by iterating over a map or a set. See [Importing multiple resources](#importing-multiple-resources) below.
+- Either `identity` or `id` - To [specify which remote object to import](#import-id).
+- `to` - The resource instance address that OpenTofu will use to track this object in [OpenTofu state](../state/index.mdx).
+- `provider` (optional) - The provider instance to use for importing, as with [The Resource provider Meta-Argument](../../language/meta-arguments/resource-provider.mdx).
+- `for_each` (optional) - Import several objects as multiple instances of the same resource by iterating over a map or a set. See [Importing multiple resources](#importing-multiple-resources) below.
 
 If you do not set the `provider` argument, OpenTofu attempts to import from the default provider.
 
-### Import ID
+### Specifying what to import {#import-id}
```
yottta, line 46: Nice! TIL. Didn't know that the anchors can be specified independently from the heading. Cool!

PR #4156, opened 2026-05-27
Add a mention about the 'resource' vs 'object'

yottta:
Reviewing https://github.com/opentofu/opentofu/pull/4153, thought that would be a good idea to have this added to the glossary we started.
If you disagree with the way I wrote it, please let me know.
We can improve my wording or we can just close the PR if you find it useless.

## Checklist

### Go checklist

 

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -76,3 +76,10 @@ Reference: [link](https://opentofu.org/docs/language/expressions/)
 
 ### Variable (HCL)
 Anything that's available to refer to in the current evaluation context.
+
+## General documentation
+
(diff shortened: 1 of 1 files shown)


diofeher (approved):
I like that! But I think it would be better suited next to https://github.com/opentofu/opentofu/blob/cba3ae6608ffc5f2920d5d5ce7a03db0209b4791/docs/glossary.md#resourceresource-instanceresource-type

so we could remove "general documentation" as it feels not matching with a glossary

apparentlymart (approved):

PR #4158, opened 2026-05-27
Rework support for providers and ephemeral resources in the new engine

### LINKED ISSUE #4059: Implement provider instances and ephemeral resource instances in the new runtime
This is a placeholder issue for the world to first design how the main ephemeral objects (provider instances and ephemeral resource instances) will be handled in the new runtime, and then to implement what we design.

This is intentionally underspecified right now because we've not settled on a single specific path.

As part of this we should make decisions on what we want to do about [Ephemeral Resource Instance and Provider Instance behavior changes](https://github.com/opentofu/opentofu/blob/rfc-new-runtime-behavior-changes/rfc/20260421-new-runtime-differences.md#ephemeral-resource-instance-and-provider-instance-behavior-changes). At the time of writing that section calls for quite a different approach where these ephemeral objects would be resolved completely dynamically when needed based on dynamic reachability, but we've not yet actually settled on whether that results in a better user experience or whether it's too different to the current behavior.

cam72cam:
In the [Walking Skeleton](https://github.com/opentofu/opentofu/issues/3570) phase of the new engine, we tried to make the execution graph generated by plan and followed by apply as rigid as possible. This meant that providers and ephemeral resources were deeply interwoven in the graph, adding a lot of complexity along the way.

After a *long* discussion between myself (@cam72cam), @apparentlymart, and other maintainers, we decided to try removing provider and ephemeral resource information from the graph entirely.
We reached this conclusion for the following reasons:
* As we improve hcl and tofu's handling of more complex references, certain dynamic actions will become possible. A simple example is that a conditional on tofu.applying (ephemeral value) may change the resource dependency graph in drastic ways.
* Interactions with -target could become quite tricky, for similar reasons to the complex references above.
* At a fundamental level, "ephemerality" means that no expectation of similarity should be expected between plan and apply.

At this point, we drew a line in the sand for what we would support and what we would not:
* Providers and ephemeral resources *may* be used in a different order / different set between plan and apply.
  - This is technically possible, though I personally would try to avoid it.
* Managed and data resources and their dependency graph *may not* be different between plan and apply.
  - Being able to produce a clear error when this occurs is paramount to accepting this compromise and is critical to end users.

The general approach taken by this PR is:
* Remove most/all existing provider and ephemeral resource wiring in the plan and apply engines
* Re-add provider wiring, taking a dynamic approach
* Add ephemeral resource wiring, taking a dynamic approach and introducing a new type of mark for handling the scenario where a plan resource graph does not work during apply.

I have a local set of changes where I was able to run all of the existing ephemeral tests in the tofu package. Not all of the tests passed, but I traced all of the failures to existing known missing functionality in the new engine.

Resolves #4059

## Checklist

### Go checklist

 


internal/engine/planning/execgraph_resource.go
```diff
@@ -107,24 +99,11 @@ func (b *execGraphBuilder) AddResourceInstanceObjectSubgraphs(
...
-		if !ok {
-			var addProviderConfigDep func(execgraph.AnyResultRef)
-			providerClientRef, addProviderConfigDep, addProviderCloseDep = b.ProviderInstanceSubgraph(providerInstAddr)
-			providerClientRefs.Put(providerInstAddr, providerClientRef)
-			addProviderConfigDeps.Put(providerInstAddr, addProviderConfigDep)
-			addProviderCloseDeps.Put(providerInstAddr, addProviderCloseDep)
-		} else {
-			addProviderCloseDep = addProviderCloseDeps.Get(providerInstAddr)
-		}
+		//providerInstAddr := obj.ProviderInst
```
apparentlymart:
```suggestion
```

😀

apparentlymart (approved):

internal/engine/applying/operations_resource.go
```diff
@@ -12,11 +12,66 @@ import (
...
+				fmt.Sprintf("Invalid %s action", idType),
+				fmt.Sprintf(resourceDependencyMissingDetail, mark.Target, idName, "but was not present in the plan"),
+			))
+		case execgraph.ResourceInstanceDependencyMissingCauseNotExecuted:
+			diags = diags.Append(tfdiags.Sourceless(
+				tfdiags.Error,
+				fmt.Sprintf("Invalid %s action", idType),
+				fmt.Sprintf(resourceDependencyMissingDetail, mark.Target, idName, "but was not yet present in the planned execution order"),
+			))
+
```
yottta: A `default` branch with a `panic` call would help for the time being to be sure that new types will not remain unhandled.

internal/engine/internal/common/providers.go
```diff
@@ -0,0 +1,110 @@
...
+		pi.active.Put(addr, &grapheval.Once[providers.Configured]{})
+	}
+	pi.activeMu.Unlock()
+
+	once := pi.active.Get(addr)
+	return once.Do(ctx, func(ctx context.Context) (ret providers.Configured, diags tfdiags.Diagnostics) {
+
+		closeCh := make(chan struct{})
+		closer := func(ctx context.Context) tfdiags.Diagnostics {
+			println("CLOSING PROVIDER " + addr.String())
```
yottta: Log statement instead?

yottta (approved):
Reviewing this realised that I am really out of the loop with this one. 😢  So left only a couple of nit picks.

The general understanding, from the code, PR description and code comments is that this PR removes the relationships from the execution graph between ephemeral resources (or providers) and any other type of entity that goes into the execgraph.
**But**, we still execute the ephemeral resources, during plan, based on the references other blocks have to them, if any do actually reference those, right?

One thing that I want to get a better understanding on is the following:
> A simple example is that a conditional on tofu.applying (ephemeral value) may change the resource dependency graph in drastic ways

You mean that if an ephemeral block is marked as `enabled = tofu.applying`, the ephemeral resource would not be added to the execution graph during the plan phase but apply phase would still somehow have to pick it up? 🤔 
_sorry if it's a dumb question_

cam72cam:
@yottta that is correct. Any ephemerals and providers needed during plan / apply are only opened when used.

The issue is if targeting is in effect and the ephemerals pull in additional resource dependencies. We will improve the error handling and tune this further when we get to implementing targeting in the new engine.

internal/engine/applying/operations_resource.go
cam72cam: Fixed in 7e0c43497d

internal/engine/internal/common/providers.go
cam72cam: Fixed in 7e0c43497d

internal/engine/planning/execgraph_resource.go
cam72cam: Fixed in 7e0c43497d

PR #4159, opened 2026-05-27
Lots of small fixes and tweaks to get closer to running the existing tests against the new engine

cam72cam:
## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/applying/providers.go
+++ b/internal/engine/applying/providers.go
@@ -7,7 +7,9 @@ package applying
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/engine/internal/common"
 	"github.com/opentofu/opentofu/internal/providers"
@@ -25,6 +27,14 @@ func newProviderInstances(execOperations *execOperations) *common.ProviderInstan
 		configVal, diags := oracle.ProviderInstanceConfig(ctx, addr)
 		if diags.HasErrors() {
 			return nil, diags
+		}
+
+		if !configVal.IsWhollyKnown() {
+			return nil, tfdiags.Diagnostics{}.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid provider configuration",
+				Detail:   fmt.Sprintf("The configuration for %s depends on values that cannot be determined until apply.", addr),
+			})
 		}
 
 		// If _this_ call fails then unfortunately we'll end up duplicating
--- a/internal/engine/applying/providers.go
+++ b/internal/engine/applying/providers.go
@@ -7,9 +7,7 @@ package applying
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/addrs"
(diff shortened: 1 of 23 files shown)


internal/tofu/context_temp_runtime.go
```diff
@@ -229,6 +244,14 @@ func (n *newRuntimeModules) ModuleConfig(ctx context.Context, source addrs.Modul
 	case addrs.ModuleSourceLocal:
 		sourceDir = filepath.Clean(filepath.FromSlash(string(source)))
 	default:
+		// Hack in support for file:// paths to allow existing tests to run
```
yottta: Maybe a TODO then? To be easier to identify.

internal/engine/planning/providers.go
```diff
@@ -44,6 +46,14 @@ func newProviderInstances(planCtx *planContext, oracle *eval.PlanningOracle) *co
 		}
 		configVal := config.ConfigVal
 
+		if !configVal.IsWhollyKnown() {
+			return nil, tfdiags.Diagnostics{}.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid provider configuration",
+				Detail:   fmt.Sprintf("The configuration for %s depends on values that cannot be determined until apply.", addr),
+			})
+		}
```
apparentlymart: I don't think this is a correct fix because providers expect to be sent partially-unknown configuration and will reject it themselves if they can't handle it.

If you did this because something downstream was failing then I think we probably need to fix that downstream thing to be more tolerant of unknown values, but I'm not sure what to suggest without knowing what motivated you to add this in the first place. 😀 

(In a world where "deferring" is supported providers would  also have the option of telling us that the configuration isn't known enough so we can successfully generate a partial plan, but we don't have proper handling of _that_ situation just yet.)

internal/engine/applying/providers.go
```diff
@@ -27,6 +29,14 @@ func newProviderInstances(execOperations *execOperations) *common.ProviderInstan
 			return nil, diags
 		}
 
+		if !configVal.IsWhollyKnown() {
+			return nil, tfdiags.Diagnostics{}.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid provider configuration",
+				Detail:   fmt.Sprintf("The configuration for %s depends on values that cannot be determined until apply.", addr),
+			})
+		}
```
apparentlymart: Unlike the other comment I left just above this one, it _is_ invalid for there to be unknown values here during the apply phase.

But in that case it's a "this is a bug in OpenTofu" situation because nothing that a provider configuration could refer to should contain any unknown values during the apply phase. So failing here is correct, but the error message used here is not correct. 😀

apparentlymart (commented):
Thanks for working on this! I expect it was tedious...

Overall this looks good to me. Just some small notes inline about the handling of unknown provider configurations, but otherwise I think this is good to go.

internal/engine/planning/providers.go
cam72cam: Ah reading the old code, It's only needed for the import path

cam72cam: Reverted the commit that added this change.

internal/engine/applying/providers.go
cam72cam: Reverted the commit that added this change.

apparentlymart (approved):

PR #4174, opened 2026-06-01
Better support for implicit providers in new engine

cam72cam:
This is not an ideal solution, but it patches the existing workaround.

Follow up to #4059

## Checklist

### Go checklist

 


internal/lang/eval/internal/tofu2024/compile.go
```diff
@@ -102,7 +103,49 @@ func CompileModuleInstance(
...
+
+			// Already populated by this "transformer"
+			if found, ok := providerConfigNodes[pc]; ok {
+				return found
+			}
+
+			temp := compileModuleInstanceProviderConfigs(ctx,
+				nil,
+				nil,
+				// This is a fragile hack
```
yottta, line 131:
```suggestion
				// TODO: This is a fragile hack
```

yottta (approved):

internal/lang/eval/internal/tofu2024/compile.go
diofeher, line 131: In addition to what Andrei said, maybe add the context on why here?

cam72cam:
Closing as I've found a better approach.

PR #4208, opened 2026-06-02
Clean up some file descriptor leaks

### LINKED ISSUE #4122: File descriptor leaks in local-exec, file provisioner, and two commands
## Summary

Four file descriptor leaks across provisioners and commands. Each leaks one OS file descriptor per invocation, which can accumulate during long `tofu apply` runs with many resources.

### 1. `internal/builtin/provisioners/local-exec/resource_provisioner.go`

`os.Pipe()` creates two file descriptors (pr and pw). Only `pw.Close()` is called; the read end `pr` is never closed.

```go
pr, pw, err := os.Pipe()
// ... pw is closed, but pr leaks
```

Fix: add `defer pr.Close()` after the pipe is created.

### 2. `internal/command/login.go`

`defer resp.Body.Close()` is placed after `io.ReadAll`. If `ReadAll` returns an error, the early return skips the deferred close, leaking the HTTP response body.

Fix: move `defer resp.Body.Close()` immediately after the nil-error check on the response.

### 3. `internal/command/workspace_new.go`

`os.Open(statePath)` opens a file handle for state import but never closes it. The handle leaks for the duration of the command.

Fix: add `defer f.Close()` after the open.

### 4. `internal/builtin/provisioners/file/resource_provisioner.go`

`os.CreateTemp` in `getSrc()` creates a temporary file but the file handle is never closed before returning. Only the path is used afterward.

Fix: close the file handle before returning from `getSrc()`.

## Impact

Each leak is one FD per invocation. For local-exec and file provisioners, this means one leaked FD per resource using that provisioner in a single `tofu apply`. In large configurations with hundreds of resources, this can approach the OS file descriptor limit.

## Reference

The same class of bugs was fixed in hashicorp/terraform in _[Terraform PR link removed by @apparentlymart]_ (merged May 14, 2026).

A proposed fix for all four issues is available at [`SebTardif/opentofu@fix-provisioner-fd-leaks`](https://github.com/SebTardif/opentofu/tree/fix-provisioner-fd-leaks) ([diff](https://github.com/opentofu/opentofu/compare/main...SebTardif:opentofu:fix-provisioner-fd-leaks)). The change is +8/-5 lines across 4 files.

Yantrio:
Resolves #4122

## Checklist

### Go checklist

 

### DIFF
--- a/internal/builtin/provisioners/file/resource_provisioner.go
+++ b/internal/builtin/provisioners/file/resource_provisioner.go
@@ -162,7 +162,7 @@ func getSrc(v cty.Value) (string, bool, error) {
 		}
 
 		if _, err = file.WriteString(content.AsString()); err != nil {
-			file.Close()
+			_ = file.Close()
 			return "", true, err
 		}
 
--- a/internal/builtin/provisioners/file/resource_provisioner.go
+++ b/internal/builtin/provisioners/file/resource_provisioner.go
@@ -162,6 +162,12 @@ func getSrc(v cty.Value) (string, bool, error) {
 		}
 
 		if _, err = file.WriteString(content.AsString()); err != nil {
+			file.Close()
+			return "", true, err
+		}
+
+		err = file.Close()
+		if err != nil {
 			return "", true, err
 		}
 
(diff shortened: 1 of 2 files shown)


internal/builtin/provisioners/file/resource_provisioner.go
```diff
@@ -162,6 +162,12 @@ func getSrc(v cty.Value) (string, bool, error) {
 		}
 
 		if _, err = file.WriteString(content.AsString()); err != nil {
+			file.Close()
```
yottta:
```suggestion
			_ = file.Close()
```

Just to avoid linting issues.

yottta (approved):

Issue #4065, opened 2026-04-28
Implement `depends_on` for the new language runtime


apparentlymart:
The new language runtime (https://github.com/opentofu/opentofu/issues/3414) doesn't currently have support for declaring explicit dependencies with `depends_on`. The new dynamic analysis approach means we need to think a little differently about how this feature is implemented, though.

The most direct analogy to the current feature would be to evaluate the value that each `depends_on` traversal refers to, search throughout that value for any marks related to dependencies, and to transfer those marks to the value of the object that `depends_on` is associated with.

We are also considering a more flexible approach where `depends_on` elements are allowed to be arbitrary expressions rather than just references, and then we'd copy only the dynamically-detected dependencies of whatever the expression evaluates to. For example, it would be possible to refer to a specific attribute of a resource instance and so depend only on whatever the expression for that attribute depends on, rather than selecting all of the dependencies of the overall resource instance configuration. More on that in [`depends_on` behavior changes](https://github.com/opentofu/opentofu/blob/rfc-new-runtime-behavior-changes/rfc/20260421-new-runtime-differences.md#depends_on-behavior-changes).

Whatever specific details we choose, handling this for resource instances and provider instances in particular seems relatively straightforward: the special treatment of `depends_on` can probably be mostly encapsulated inside the `configgraph.ResourceInstance.Value` implementation.

Handling this for `depends_on` inside a `module` block is trickier because that sort of dependency is treated as making everything declared beneath the affected module call have the additional dependencies added to them, and so that's a cross-cutting concern that we'll need to figure out how to handle.

cam72cam:
I pushed some of my basic ideas to `f-engine/depends_on`. This implements the relatively easy side of resource depends_on. I did find one circular loop that makes the graph marshalling panic and is explicitly skipped separately than SkipExperimental().

Note: the "legacy" workaround I implemented for `module.name` does not wait for absolutely everything in the module.  It just waits for the resources. I've not fully considered the ramifications of this change.

I figured I'd write down my current thoughts on how module depends_on could work.

Option 1:
* We could add a `DependsOn []OnceValuer` through the module compilation chain, appended to at each child level
* This could then be passed to the resource instance compiler as an additional argument
* This approach would not force variables, locals, and outputs to wait for depends_on like they do now.  I assume this is probably fine?

Option 2:
* Add the depends on logic to the InstanceSelector in compileModuleInstance
* This is a simpler approach, but does not enforce the dependency in the execgraph or the state
[...]

PR #4215, opened 2026-06-03
Basic support for state dependencies

cam72cam:
This adds support for `addrs.ConfigResource`s being saved in state via the new engine.  It also introduces a way for the existing state to store `addrs.AbsResourceInstance`s calculated by the new engine in the existing state.

Depends On #4210

## Checklist

### Go checklist

 


cam72cam:
Found some fun as a followup to
https://github.com/opentofu/opentofu/pull/1647
and
https://github.com/opentofu/opentofu/pull/3922

Resolving that before continuing on here.

internal/lang/eval/internal/configgraph/resource_instance.go
```diff
@@ -191,7 +202,8 @@ func (ri *ResourceInstance) ResourceInstanceDependencies(ctx context.Context) it
 	// We ignore diagnostics here because callers should always perform a
 	// CheckAll tree walk, including a visit to this resource instance object,
 	// before trusting anything else that any configgraph nodes report.
-	resultVal := diagsHandledElsewhere(ri.Value(ctx))
+	// TODO consider if this should be ConfigValue vs Value based on the usages of this function
+	resultVal := diagsHandledElsewhere(ri.ConfigValue(ctx))
```
apparentlymart, line 271: FWIW I can't really justify this being `Value` vs. `ConfigValue`, because implicit dependencies come from the configuration and new ones can't possibly be introduced during the plan and apply process because there's no way to serialize them in the provider wire protocol anyway... in practice any marks on `Value` are added synthetically by copying them from the original configuration value.

I'm pretty sure that I just made an error when I wrote this, because resource instances are the only case where we have a difference between "value" and "config value" in this way.

cam72cam: Removed the comment in b28ec5495d

internal/engine/internal/execgraph/compiler.go
```diff
@@ -164,8 +164,9 @@ func (c *compiler) Compile() (*CompiledGraph, tfdiags.Diagnostics) {
 		ref := elem.Value
 		execFunc := c.compileResultRef(ref)
 		c.compiledGraph.resourceInstanceValues.Put(instAddr, func(ctx context.Context) cty.Value {
-			rawResult, ok, _ := execFunc(ctx)
+			rawResult, ok, diags := execFunc(ctx)
 			if !ok {
+				println(diags.Err().Error())
```
apparentlymart: If we're going to keep this then it probably wants to be `log.Println`, right?

internal/lang/eval/internal/configgraph/resource_instance.go
```diff
@@ -35,6 +36,9 @@ type ResourceInstance struct {
 	// is the provider to use when asking for config validation, etc.
 	Provider addrs.Provider
 
+	// Used to ensure marks
+	RepetitionData instances.RepetitionData
```
apparentlymart: I'm not 100% sure about this yet, but I think generalizing this to just be a `cty.ValueMarks` representing "various additional marks that must be applied to the value" is one answer to the module-dependencies part of https://github.com/opentofu/opentofu/issues/4065, since it could give the `tofu2024` package a way to sneak in some dependency marks from the module call without adding a bunch of complexity to `configgraph`, and just generally make these details each be a "surface-language-related" concern that could vary by language edition, instead of a core runtime concern.

I don't mean to say that you should change this immediately now -- this is a fine way to implement it for this PR -- but just noting that I'm considering changing this again in my next set of work.

internal/states/instance_object.go
```diff
@@ -47,6 +48,13 @@ type ResourceInstanceObject struct {
 	// altogether, or is now deposed.
 	Dependencies []addrs.ConfigResource
 
+	// DependsOn is a set of absolute address instances to other resources this
+	// instance depended on when it was applied. This is used to construct
+	// the dependency relationships for an object whose configuration is no
+	// longer available, such as if it has been removed from configuration
+	// altogether, or is now deposed.
+	DependsOn []addrs.AbsResourceInstance
+
```
apparentlymart, line 57: Let's also extend the comment of the `Dependencies` field to note that the new runtime uses it only when `DependsOn` is not populated, and extend _this_ comment with a note that it's completely ignored by the current language runtime, just to aid maintainers who are trying to understand what's going on here in the meantime while we have both uses at once. 😀

apparentlymart (approved):

internal/engine/internal/execgraph/compiler.go
cam72cam: Ah, this was some temporary logging I need to remove!

PR #4219, opened 2026-06-03
Refine the validations around ephemeral marks before writing to the in-memory state and the in-file state

### LINKED ISSUE #3959: Ephemeral Variables not usable for config of data terraform_remote_state
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.5
```

```hcl
variable "state_access_token" {
  type        = string
  description = "Access token for the project holding IaC for the layer below"
  ephemeral   = true # breaks usage in config below
  sensitive   = true
}

data "terraform_remote_state" "upstream" {
  backend = "http"

  config = {
    address  = "<...>/${var.environment}"
    username = "gitlab-ci-token"
    password = var.state_access_token
  }
}
```

### Debug Output

╷
│ Error: Ephemeral value used in non-ephemeral context
│ 
│   with data.terraform_remote_state.upstream,
│   on upstream.tf line 17, in data "terraform_remote_state" "upstream":
│   17:   config = {
│   18:     address  = "${var.upstream_state_url}/${var.environment}"
│   19:     username = "gitlab-ci-token"
│   20:     password = var.state_access_token
│   21:   }
│ 
│ Attribute ".config" is referencing an ephemeral value but ephemeral values can be referenced only by other ephemeral attributes or by write-only ones.

### Expected Behavior

Ephemeral Values should be usable for configuration of the terraform_remote_state data source.

### Actual Behavior

se error above

### Steps to Reproduce

1. `tofu init`
2. `tofu plan`

### Additional Context

_No response_

### References

_No response_

yottta:
Ensure that there are no ephemeral values written into the state file for any resource.
Do not allow ephemeral values to be stored into the temporary state for anything else than ephemeral resources.

This change is in response to the changes proposed in #3960.
Testing that, realised that we miss some guardrails when it comes to what `configschema.Attribute` is allowed or not to be marked as `WriteOnly` and what `configschema.Schema` can be marked or not as `Ephemeral`.

The general idea is that the schema for types that are meant to be written to state must not be allowed with such configuration: 
* The only `configschema.Schema` that is allowed to be marked as `Ephemeral` is the one used for ephemeral resource types and other ephemeral types (like provider) that by design are not persisted into the state or plan.
* The only `configschema.Attribute` that is allowed to be marked as `WriteOnly` is the one used for attributes inside *managed resources*, as instructed by the provider. This is really important, because the [provider write-only marked attributes value nullification is handled directly by the `terraform-plugin-framework`](https://github.com/hashicorp/terraform-plugin-framework/blob/main/internal/fwserver/write_only_nullification.go#L22). If it would be to allow data sources write-only attributes (which is suggested in #3960), that would mean that we have to implement in OpenTofu core this nullification strategy which defends the purpose of the protocol and the current design.

Therefore, as agreed with the maintainers, we intend to reject this enhancement in favor of a future [Backends as plugins](https://github.com/opentofu/opentofu/issues/382) availability. The reason for that, was loosely hinted in the above points: provider configuration is ephemeral by design. Hence, if the state will be handled through the providers protocol, then configuration that is needed to access it will go through the `provider` configuration block, which being ephemeral, will accept ephemeral values (as it already does).

Resolves #3959

## Checklist

### Go checklist

 

### DIFF
--- a/internal/addrs/provider_config.go
+++ b/internal/addrs/provider_config.go
@@ -258,6 +258,15 @@ func ParseAbsProviderConfigStr(str string) (AbsProviderConfig, tfdiags.Diagnosti
 	diags = diags.Append(addrDiags)
 	return addr, diags
 }
+
+func MustParseAbsProviderConfigStr(str string) AbsProviderConfig {
+	ret, diags := ParseAbsProviderConfigStr(str)
+	if diags.HasErrors() {
+		panic(fmt.Errorf("could not parse %q as AbsProviderConfig: %s", str, diags))
+	}
+	return ret
+}
+
 func ParseAbsProviderConfigInstanceStr(str string) (AbsProviderConfig, InstanceKey, tfdiags.Diagnostics) {
 	var diags tfdiags.Diagnostics
 	traversal, parseDiags := hclsyntax.ParseTraversalAbs([]byte(str), "", hcl.Pos{Line: 1, Column: 1})
--- a/internal/states/statefile/version4.go
+++ b/internal/states/statefile/version4.go
@@ -620,7 +620,7 @@ func appendInstanceObjectStateV4(rs *states.Resource, is *states.ResourceInstanc
 				tfdiags.Sourceless(
 					tfdiags.Error,
 					"Ephemeral detected in state writing",
-					fmt.Sprintf("%q has an ephemeral value in %q. This is an OpenTofu error. Please report it", rs.Addr.String(), tfdiags.FormatCtyPath(mark.Path)),
+					fmt.Sprintf("%q has an ephemeral value in %q. This is a bug in OpenTofu. Please report it", rs.Addr.String(), tfdiags.FormatCtyPath(mark.Path)),
 				),
 			}
 		}
--- a/internal/states/statefile/version4.go
+++ b/internal/states/statefile/version4.go
@@ -607,6 +607,19 @@ func appendInstanceObjectStateV4(rs *states.Resource, is *states.ResourceInstanc
 		providerInstance = rs.ProviderConfig.InstanceString(is.ProviderKey)
 	}
 
+	for _, mark := range obj.TransientPathValueMarks {
+		_, ok := mark.Marks[marks.Ephemeral]
+		if ok {
+			return nil, tfdiags.Diagnostics{
+				tfdiags.Sourceless(
+					tfdiags.Error,
+					"Ephemeral detected in state writing",
+					fmt.Sprintf("%q has an ephemeral value in %q. This is an OpenTofu error. Please report it", rs.Addr.String(), tfdiags.FormatCtyPath(mark.Path)),
+				),
+			}
+		}
+	}
+
 	// Extract paths from path value marks
 	var paths []cty.Path
 	for _, vm := range obj.AttrSensitivePaths {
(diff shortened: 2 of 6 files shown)


internal/states/statefile/version4.go
```diff
@@ -582,6 +582,25 @@ func appendInstanceObjectStateV4(rs *states.Resource, is *states.ResourceInstanc
...
+	// Before this change, the only values that could have been marked as ephemeral and written into state were already
+	// nullified and unmarked (see usage of Block.RemoveEphemeralFromWriteOnly)
+	for _, mark := range obj.TransientPathValueMarks {
+		_, ok := mark.Marks[marks.Ephemeral]
+		if ok {
+			return nil, tfdiags.Diagnostics{
+				tfdiags.Sourceless(
+					tfdiags.Error,
+					"Ephemeral detected in state writing",
+					fmt.Sprintf("%q has an ephemeral value in %q. This is an OpenTofu error. Please report it", rs.Addr.String(), tfdiags.FormatCtyPath(mark.Path)),
```
apparentlymart:
```suggestion
					fmt.Sprintf("%q has an ephemeral value in %q. This is a bug in OpenTofu. Please report it", rs.Addr.String(), tfdiags.FormatCtyPath(mark.Path)),
```

(Just to make this match how we typically say this in other diagnostics.)

apparentlymart (approved):

internal/states/statefile/version4.go
yottta: Fixed in f1d5c54600e5e68042f6a639fa7111fc760884d7

internal/tofu/node_resource_abstract_instance_test.go
```diff
@@ -190,6 +191,45 @@ aws_instance.foo:
...
+	}
+	evalCtx.installProvider(addrs.NewDefaultProvider("aws"), mockProvider)
+
+	err := node.writeResourceInstanceState(t.Context(), evalCtx, obj, workingState)
+	if err == nil {
+		t.Fatalf("expected error but got nothing")
+	}
+	expectedErr := `non ephemeral resource ("data.aws_region.foo") found to be written with ephemeral values`
+	if diff := cmp.Diff(expectedErr, err.Error()); diff != "" {
+		t.Fatalf("expected a particular error bug got the wrong one (-want,+got):\n%s", diff)
```
diofeher:
```suggestion
		t.Fatalf("expected a particular error but got the wrong one (-want,+got):\n%s", diff)
```

internal/addrs/provider_config.go
```diff
@@ -258,6 +258,15 @@ func ParseAbsProviderConfigStr(str string) (AbsProviderConfig, tfdiags.Diagnosti
 	diags = diags.Append(addrDiags)
 	return addr, diags
 }
+
+func MustParseAbsProviderConfigStr(str string) AbsProviderConfig {
```
diofeher, line 262: This is only used in tests, should we put it there?

yottta: Yes, we have other functions like this, for example https://github.com/opentofu/opentofu/blob/20d9ff08e783eb596f44b05b85c80a3dc81e31d9/internal/addrs/provider.go#L201

internal/tofu/node_resource_abstract_instance_test.go
yottta: Good catch. Fixed in bed651985be998d67bcbdd38d7244da5ecfa3d09

cam72cam (approved):

PR #4224, opened 2026-06-03
Add UpgradeResourceState provider call to the new Engine

lbordowitz:
Adding `UpgradeResourceState` to the new engine. This is an adaptation of code in [upgrade_resource_state.go](https://github.com/opentofu/opentofu/blob/main/internal/tofu/upgrade_resource_state.go#L52); I've even retained many of its comments, I can't tell whether they're still relevant or not so I've erred on the side of more info. It's worth mentioning that I've omitted a nontrivial amount of code that accommodates FlatAttrs. Anybody who is on the upgrade/migration path will not realistically have those in their state file when using the new engine.

A small note: I don't think there's any existing test to check that an upgrade call does not happen when the schema version in the state file is greater than the provider's resource schema version. Is that something I ought to add and double-check about? Is there anything else I should check about? While on the subject of tests, more than half of the tests which were skipped waiting for the "state upgrade" feature also require another feature, like plan-destroy mode or handling taint. I've added feature markers to them.

The function is getting a bit long in the LOC, maybe it could stand to be broken down into some private functions.

## Checklist

### Go checklist

 


apparentlymart:
I intend to look at this in more detail soon, but just wanted to quickly answer the question about comparing schema version numbers:

Despite the fact that there is a version number tracked in the schema of each resource type, OpenTofu asks for an upgrade unconditionally even when the version number hasn't changed.

This is because in practice the upgrade call runs two separate pieces of code that are both needed:

1. Built-in fixups done by the plugin SDK or framework itself, such as silently dropping any attributes that are no longer in the schema, automatically without any involvement from the provider developer.
2. Resource-type-specific upgrade logic that deals with more substantial changes that the SDK/Framework's built-in logic can't handle.

Although the original intention of the protocol was that the version number would increase each time _any_ upgrade logic is required, in practice provider developers don't tend to think about the first category of fixups that get done for them automatically, and so they generally don't realize they need to increase the version number. After running into various situations where there ought to have been a version number increase but there wasn't and that caused quirky behavior, our predecessor began just running the upgrade operation unconditionally as you've seen.

The schema version number now exists mainly for the benefit of external software that's trying to work with data from providers, such as systems that consume `tofu show -json -plan=...` output to enforce policy rules. In case where providers have made significant changes to the schema of a resource type, policy logic might in theory need to behave differently depending on the current schema version. But OpenTofu itself doesn't really use it for anything directly itself.

internal/tofu/context_temp_runtime_test.go
```diff
@@ -73,13 +74,60 @@ const (
 	ExperimentalBugExecGraph ExperimentalFlag = "Bug in generated Exec Graph"
 )
 
+var EnabledFlags []ExperimentalFlag = []ExperimentalFlag{
```
lbordowitz: I could probably change this to a map.

internal/engine/planning/plan_managed.go
```diff
@@ -122,7 +124,89 @@ func (p *planGlue) planDesiredManagedResourceInstance(
...
+		}
+
+		upgradedPrevState := &states.ResourceInstanceObjectFullSrc{
+			Value: states.ValueJSONWithMetadata{
+				ValueJSON:      src,
+				SensitivePaths: prevRoundState.Value.SensitivePaths,
+			},
+			Private:              prevRoundState.Private,
+			Status:               prevRoundState.Status,
+			ProviderInstanceAddr: *inst.ProviderInstance,
```
lbordowitz: This line is currently causing `TestContext2Apply_issue5254` to fail. Setting it to `prevRoundState.ProviderInstanceAddr` makes it pass. Will take a closer look at this next week.

lbordowitz: Or not?

internal/engine/planning/plan_managed.go
```diff
@@ -122,7 +124,89 @@ func (p *planGlue) planDesiredManagedResourceInstance(
 	var prevRoundPrivate []byte
 	prevRoundState := p.planCtx.prevRoundState.SyncWrapper().ResourceInstanceObjectFull(inst.Addr.CurrentObject())
 	if prevRoundState != nil {
-		obj, err := states.DecodeResourceInstanceObjectFull(prevRoundState, schema.Block.ImpliedType())
+		// While we know prevRoundState is non-nil, let's upgrade state, too.
+		// Let's do a schema version comparison before upgrade
```
apparentlymart, line 126: I notice in rebase you got the `FIXME` comment I added just above for implementing "upgrade", which you've now completed here so I suggest removing that comment. 😀 

```go
// FIXME: Need to "upgrade" the previous round state before we try to decode it.
```

internal/engine/planning/plan_managed.go
apparentlymart: I must admit I don't remember the full details of this myself but I _think_ what's going on here is that in the current runtime the provider instance address for a resource instance is only synchronized with the configuration during "refresh", so the "upgraded state" would still reflect the old one and then the "refreshed state" would reflect whatever the configuration now says.

That explanation seems to agree with what you shared in your comment above, at least. 😀

internal/tofu/context_refresh_test.go
```diff
@@ -1583,6 +1583,9 @@ aws_instance.bar:
 
 func TestContext2Refresh_schemaUpgradeFlatmap(t *testing.T) {
 	SkipExperimental(t, ExperimentalFeatureUpgradeState)
+	if experimentalRuntimeEnabled() {
+		t.Skip("Flatmap attributes are a relic of version 0.11, and will not be accepted in the new engine.")
+	}
```
apparentlymart: I wonder if we should make this still go through the `SkipExperimental` mechanism, even though in this case we're not actually intending to eventually "fix" it, just because that'll give us just a single symbol to grep for at a later point where we want to review what we've done and decide how (and whether) to adapt each of these tests to live in another part of the system.

Perhaps we can add another category of `ExpermentalFlag` for "not expected to be fixed" ones, so that we can see clearly in the table in `context_temp_runtime_test.go` what we've decided to leave non-working in the new implementation. Could prefix it `ExperimentalObsolete...`, and have this one be `ExperimentalObsoleteFlatmap` with a comment next to it explaining that we're intentionally dropping support for flatmap?

apparentlymart (commented):
This looks good!

Since there's still the question about which provider instance address we're supposed to record in the "upgraded state" I'm not approving this yet, but the general principle of it seems good to me aside from that detail (and some other little nits I left inline).

apparentlymart (approved):

cam72cam (approved):

PR #4231, opened 2026-06-04
add experimental engine build and test to makefile and github actions

lbordowitz:
I'd like to add test runs for the experimental engine to pull requests

## Checklist

### Go checklist

 

### DIFF
--- a/.github/workflows/experimental-engine.yml
+++ b/.github/workflows/experimental-engine.yml
@@ -1,0 +1,44 @@
+
+name: Experimental Engine
+
+on:
+  pull_request:
+    paths:
+      - 'internal/engine/**'
+      - 'internal/lang/**'
+      - 'internal/resources/**'
+
+permissions:
+  contents: read
+
+jobs:
+  exp-eng-tests:
+    name: Unit tests in the experimental engine for ${{ matrix.goos }}_${{ matrix.goarch }}
+    runs-on: ${{ matrix.runson }}
+    env:
+      TF_APPEND_USER_AGENT: Integration-Test
+      GOOS: ${{ matrix.goos }}
+      GOARCH: ${{ matrix.goarch }}
+    strategy:
+      matrix:
+        include:
+          - { runson: ubuntu-latest, goos: linux, goarch: "amd64" }
+      fail-fast: false
+    steps:
+      # 👇🏾 GH actions supports only "AMD64 arch", so we are using this action
+      # for testing on non amd64 envs like 386, arm64 etc...
+      - name: "Set up QEMU"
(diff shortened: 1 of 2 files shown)


.github/workflows/experimental-engine.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: Unit tests in the experimental engine for ${{ matrix.goos }}_${{ matrix.goarch }}
+    runs-on: ${{ matrix.runson }}
+    env:
+      TF_APPEND_USER_AGENT: Integration-Test
+      GOOS: ${{ matrix.goos }}
+      GOARCH: ${{ matrix.goarch }}
+    strategy:
+      matrix:
+        include:
+          - { runson: ubuntu-24.04-arm, goos: linux, goarch: "arm64" }
```
apparentlymart: For the kind of work we're doing right now I'd suggest that we reduce this matrix down to just `linux_amd64`, since running the tests twice isn't likely to give us enough additional information to justify spending the compute on it.

Once we get closer to things mostly working it would be interesting to grow this to include at least one other platform, but right now the basics aren't working properly on even one platform, so I think we can give GitHub Actions a break. 😀

.github/workflows/experimental-engine.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install Go toolchain
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: 'go.mod'
+
+      - name: "Unit tests"
+        run: |
+          TOFU_X_EXPERIMENTAL_RUNTIME=1 go test -ldflags "-X main.experimentsAllowed=yes" ./...
```
apparentlymart: Only the code in `package tofu` is sensitive to this `TOFU_X_EXPERIMENTAL_RUNTIME` environment variable -- the check for it is an unexported function in that package -- so to keep things light for now I'd suggest we focus on running just `./internal/tofu` here and let the main unit test run cover everything else, rather than redundantly re-running all of the other packages.

As with my other comment above it's possible that we'll grow this later as we get further along with the work, but I'd prefer to keep this light for now so we'll get faster feedback and spend less additional compute time on PRs that contain changes related to the new runtime.

apparentlymart (approved):
I like this idea of this!

I'd prefer to focus it a little more tightly for now though, so that we're only adding one additional job to the pull request checks when relevant files are affected, and that job will focus only on testing the codepath that are unique to the experimental mode.

yottta (approved):

cam72cam (approved):

PR #4233, opened 2026-06-05
Migrate from hashicorp/copywrite to a custom bash copyright check

yottta:
Part #2453 (we don't want the issue closed yet. We will close that once we have all the repositories from the OpenTofu organisation migrated to the new way of checking copyright headers)

This PR integrates a new [workflow template](https://github.com/opentofu/scripts/pull/52) that uses a custom written bash script that checks the copyright headers.

It also fixes the identified errors.

For more details on why we ended up with this tool, please read [this comment](https://github.com/opentofu/opentofu/issues/2453#issuecomment-4633453719).

Only for this specific repository, included also a make target to make it easier for the contributors to run it locally.

## Checklist

### Go checklist

 


diofeher:
I would be more relieved if this NWA tool used immutable releases...

yottta:
> I would be more relieved if this NWA tool used immutable releases...

As suggested and also agreed with the maintainers, changed this PR to use a custom written script for this job.

.github/workflows/checks.yml
```diff
@@ -186,13 +186,13 @@ jobs:
...
-        run: |
-          go tool github.com/hashicorp/copywrite headers --plan
-          if [[ $? != 0 ]]; then
-            echo >&2 "ERROR: some files are missing required copyright headers. Run `scripts/add-copyright-headers.sh` locally and then commit the updated files."
-            exit 1
-          fi
+  copyright:
+    permissions:
+      contents: read
+    uses: opentofu/scripts/.github/workflows/copyright.yml@copyright-workflow-bash
```
yottta: Do not resolve this comment until https://github.com/opentofu/scripts/pull/52 is merged. Once that is done, change this line to point to the `main` branch of the [scripts repository](https://github.com/opentofu/scripts)

diofeher: Loved the idea of a custom bash script :D

diofeher (approved):
I'm impressed by how many files were missing this header, and we didn't notice before. Thanks!

lbordowitz (approved):

Makefile
```diff
@@ -73,6 +63,12 @@ license-check:
 	rm -rf vendor/
 	git diff --exit-code
 
+# This runs the copyright checks against the files in this repository.
+# This is the same script that is used also in the .github/workflows/checks.yml. If this or that configuration is ever changed,
+# be sure that the changes are reflected in both places.
+copyright:
+	curl -fsSL https://raw.githubusercontent.com/opentofu/scripts/copyright-workflow-bash/sh/copyright_check.sh | bash -s -- "" "*.go *.proto" "*/.git* */vendor/* */node_modules/* */.workdir* */website/*"
```
yottta: Do not resolve this comment until https://github.com/opentofu/scripts/pull/52 is merged. Once that is done, change this line to point to the `main` branch of the [scripts repository](https://github.com/opentofu/scripts)

go.mod
```diff
@@ -246,21 +221,15 @@ require (
...
-	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/ryanuber/go-glob v1.0.0 // indirect
-	github.com/samber/lo v1.37.0 // indirect
 	github.com/shopspring/decimal v1.3.1 // indirect
 	github.com/spf13/cast v1.5.0 // indirect
-	github.com/spf13/cobra v1.6.1 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/spiffe/go-spiffe/v2 v2.6.0 // indirect
 	github.com/stretchr/testify v1.11.1 // indirect
-	github.com/thanhpk/randstr v1.0.4 // indirect
```
cam72cam, line 258: I love seeing all of these deps removed from go.mod 👍

cam72cam (commented):
Approved, though I'm not 100% sure about the changes to the old protofiles.  Not a big deal IMO

apparentlymart:
I do think we should be careful about modifying how the original copyright holder described their copyright, since the license calls for us to preserve those notices. If some of the code we forked from didn't include a starting year for the copyright then we should probably keep it that way.

Hopefully we could achieve that just by excluding these older protobuf files from being checked at all, since they are intentionally just frozen snapshots of files from upstream anyway and so we have no reason to be modifying them in subsequent PRs and thus little risk of accidentally removing the copyright comments from them.

yottta:
> I do think we should be careful about modifying how the original copyright holder described their copyright, since the license calls for us to preserve those notices. If some of the code we forked from didn't include a starting year for the copyright then we should probably keep it that way.
> 
> Hopefully we could achieve that just by excluding these older protobuf files from being checked at all, since they are intentionally just frozen snapshots of files from upstream anyway and so we have no reason to be modifying them in subsequent PRs and thus little risk of accidentally removing the copyright comments from them.

Fixed in 46c76efc2b47477c0b249431b6fea3f0cc05d0ab by removing the changes from the provider protocol files and excluding those from the scanning.
Please let me know if this looks ok.

cam72cam (approved):

Makefile
yottta: Done in c90139177be99dab5e322d61973c0404693ee677

.github/workflows/checks.yml
yottta: Done in c90139177be99dab5e322d61973c0404693ee677

Issue #1593, opened 2024-05-02
Pretty-printed JSON representation of a state increases file size by ~20-25%


alexott:
### OpenTofu Version

```shell
1.8
```

### Use Cases

We have a few use cases when the state size is very big, and we were looking into optimizing performance. What I noticed is that by default, we're using pretty-printed JSON when storing the state, and the resulting file size is much bigger than if we use compact representation, and this make read/write of state slower, not taking into account increased storage and egress costs when using cloud backends.

For example, here is comparison of the state with ~70k objects - original state (`mkdirs.tfstate`) is converted into compact representation by running `jq -c .` on it.

```
84096 -rw-rw-r--@ 1 alex  staff  42996911 Apr 27 10:53 mkdirs-compact.tfstate
114112 -rw-r--r--@ 1 alex  staff  57680266 Apr 27 10:50 mkdirs.tfstate
```

### Attempted Solutions

.

### Proposal

Enable compact representation by default. I can make PR to support it if it's ok

Yantrio:
There was a discussion around something like this here: https://github.com/opentofu/opentofu/pull/397

I still think that this should be an opt-in feature as per my comment here: https://github.com/opentofu/opentofu/pull/397#issuecomment-1719321845

PR #4239, opened 2026-06-08
RFC: Speed up tofu show <planfile> by embedding schemas into the planfile

Yantrio:
RFC to propose introducing speeding up `tofu show -json <planfile>` by embedding the schemas into the planfile. 

This feels like it should be a quick win to save a lot of time for people running `tofu` in a loop (TACOS, agents etc).

[Rendered view](https://github.com/opentofu/opentofu/blob/96b6c674f27172c3586edf9a112561feb2ef45ee/rfc/20260605-speed-up-tofu-show-planfile.md)

## Checklist

### DIFF
--- a/rfc/20260605-speed-up-tofu-show-planfile.md
+++ b/rfc/20260605-speed-up-tofu-show-planfile.md
@@ -1,0 +1,87 @@
+# Speed up tofu show -json <planfile> by embedding provider schemas in the planfile
+
+Today, `tofu show -json <planfile>` has become one of the primary mechanisms for other applications to interface with a planfile and with the rise of agentic development, we at OpenTofu should aim to imrpove the time which this command takes to run.
+
+I believe that this command is currently being run a lot in the following situations
+
+- TACOS automating tofu runs by splitting the plan and the apply phase
+- Generic CI/CD users who execute tofu
+- Cost estimation tooling (Infracost/etc)
+- Tooling around policy engines (OPA/Conftest/Sentinel-style tools)
+- Increasingly, AI/LLM based agents
+
+For all of these, `tofu show -json <planfile>` is on the hotpath, called once or many times per execution of `tofu plan`.
+
+This slow execution on the hotpath hits end users in different ways, but I would like to focus on 2 primary ways with this RFC:
+
+- **Feedback loop latency** - This is the story for agentic development, allowing an agent to run a tight loop of `tofu plan` -> read the json -> reason, infer, fix, edit -> plan again`. The factor that matters here is what I will call "Time to actionable output". Launching even a single provider with a large schema taxes every cycle of this loop. Over time this adds up.
+- **Compute and time at scale** - Teams that are running `tofu` at a large scale run plans hundreds or even thousands of times a day, and this adds up to a very non-negligable recurring compute cost over time.
+
+The problem today is that rendering a saved planfile as json requires the schema of the resources from the provider and OpenTofu obtains those schemas in an expensive way, by launching every referenced provider plugin and asking each one for it's schema over gRPC. It does this even though the plan was built with those exact schemas moments earlier in the plan execution. So every consumer re-derives from scratch a value that is fully determined by the plan. This issue is what this RFC plans to address.
+
+This RFC proposed storing a **minimal subset of provider schemas** that are required to render the planfile as json inside the plan file itsself, so that `tofu show` and other tooling built on it can decode and display the plan without requiring to launch providers.
+
+Beyond this being just a speed improvement, this turns the plan file in to a self-describing and portable artifact. Which means that the consumer of the planfile for analytics may not always need to be the same machine that created it.
+
+## Background - Why are schemas needed?
+
+A saved plan stores resource attributes as an opaque msgpack value blob. To turn this blobs back into a typed JSON OpenTofu needs to know the `cty.Type` for each resource's block, plus per-resource metadata such as the schema version and sensitivity marks.
+
+To get this `cty.Type` and metadata, OpenTofu needs to spawn the provider that manages the resource and ask for the schema over gRPC. This means spawning a seperate process, and in the case of large providers, transferring around megabytes of schema information.
+
+Because of this, **it is slow and it gets slower as you use more providers**. Each distinct provider in the plan means one more plugin to launch, one more gRPC call, one more teardown. Whilst this work is paralellized across pproviders, the cost of launching these providers is expensive.
+
+As mentioned above too, this **requires every provider to be installed locally**. If a provider is not present then OpenTofu is unable to determine the schema. There is no graceful degredation, its a hard fail. So if a plan file is copied from a CI runner to a reviewer's laptop, or an agent sandbox, this cannot display the planfile as json at all unless the environment is populated with providers correctly through `tofu init`.
+
+## Proposed Solution
+
+OpenTofu should persist the provider schemas that are needed by the `show` command **inside the planfile**, written at plan-creation time from the schemas already in memory, and have `show` prefer using those over launching providers.
+
+The plan file today is already an archive (zip), I propose that we introduce one new, optional entry alongside the existing entries. This area of the codebase is extensible and already ignores unknown entries in the zip.
+
+Crucially, and this is possibly an extra bit of work ontop of this, we only store what is needed to render the plan, NOT the full schema of every provider. This ensures that the planfile does not baloon in size too much, especially for users of providers with a large amount of resources.
+
+### Technical approach
+
+There are 3 parts that we need to address here to introduce this feature into OpenTofu:
+
+- What to store
+- How to store
+- How to read
+
+#### 1. What to store
+
+During rendering the json, `jsonplan`, `jsonstate` and `jsonconfig` all access parts of the schema that is stored in memory.
+
+From this we can derive the minimal state of what resource schmeas need storing in the planfile. For each provider referenced by the plan, we can store a condensed version of the ProviderSchema that only contains
+
+- The **provider configuration** schema
+- **managed resource type schemas and resource identity schemas** for resource types that are referenced in the config snapshot or in the state
+- **data source schemas** that are referenced similarly to resources.
+
+We can store this information in a format that is easily comprehendable by the json marshalling logic to reduce the compute overhead of reading this entry in the archive. If the trimming logic and the render logic ever seem to drift (For example, we are missing a resource identity) then the fallback path of the `show` command should be the existing functionality, by questining the provider.
+
+> ![NOTE]
+> Provisioner Schemas today are built into the opentofu binary and are not part of plugins. For this reason we do not need to store these alongside provider schemas.
+
+#### How to store
+
+Provider schemas already have a protobuf representation as defined by the `GetProviderSchema` response in `tfplugin{5|6}`, and the codebase has existing logic to handle conversion in both directions. Reusing this protobuf representation simplifies the implementation of this featureand avoids inventing a new format that could drift over time.
+
+The new proposed zip entry should be a new protobuf message, a map of provider address to the trimmed per-provider schema.
+
+To store this, we should pass the ProviderSchemas through to the logic that writes the planfile (where `planfile.Create` is called should be a good entrypoint). The trimming of the schemas can happen in here and we can re-use the same mechanism of writing to the archive that the other entries in the archive use.
+
+#### How to read
+
+... TODO
+
+### Open Questions
+
+- Should we have this as a default enabled feature? - It's nearly free at write time and there's lots of benefits for this, however we should be aware that some people want to keep their planfiles smaller.
+- Encryption - Schemas are not sensitive information, but they live inside the encrypted blob like everything else in the planfile, we should check that that's okay.
+- Should we do the same for human readable `tofu show`, or only `-json` ? I think it's a freebie but requires some investigation.
+
+### Ideas for the future
+
+- It would be nice to have a global persistent provider schema cache on disk instead of embedding it into the provider. This could be done alongside existing providder caching.
--- a/rfc/20260605-speed-up-tofu-show-planfile.md
+++ b/rfc/20260605-speed-up-tofu-show-planfile.md
@@ -1,6 +1,6 @@
 # Speed up tofu show -json <planfile> by embedding provider schemas in the planfile
 
-Today, `tofu show -json <planfile>` has become one of the primary mechanisms for other applications to interface with a planfile and with the rise of agentic development, we at OpenTofu should aim to imrpove the time which this command takes to run.
+Today, `tofu show -json <planfile>` has become one of the primary mechanisms for other applications to interface with a planfile and with the rise of agentic development, we at OpenTofu should aim to improve the time which this command takes to run.
 
 I believe that this command is currently being run a lot in the following situations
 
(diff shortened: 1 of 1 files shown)


Yantrio:
Note: I'm aware there's a `TODO` here, I need to do some more research into how this area of the codebase works.

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
+# Speed up tofu show -json <planfile> by embedding provider schemas in the planfile
+
+Today, `tofu show -json <planfile>` has become one of the primary mechanisms for other applications to interface with a planfile and with the rise of agentic development, we at OpenTofu should aim to imrpove the time which this command takes to run.
```
cam72cam: I had to look up `imrpove` in the dictionary

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+Beyond this being just a speed improvement, this turns the plan file in to a self-describing and portable artifact. Which means that the consumer of the planfile for analytics may not always need to be the same machine that created it.
+
+## Background - Why are schemas needed?
+
+A saved plan stores resource attributes as an opaque msgpack value blob. To turn this blobs back into a typed JSON OpenTofu needs to know the `cty.Type` for each resource's block, plus per-resource metadata such as the schema version and sensitivity marks.
+
+To get this `cty.Type` and metadata, OpenTofu needs to spawn the provider that manages the resource and ask for the schema over gRPC. This means spawning a seperate process, and in the case of large providers, transferring around megabytes of schema information.
+
+Because of this, **it is slow and it gets slower as you use more providers**. Each distinct provider in the plan means one more plugin to launch, one more gRPC call, one more teardown. Whilst this work is paralellized across pproviders, the cost of launching these providers is expensive.
```
cam72cam: `pproviders`

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+As mentioned above too, this **requires every provider to be installed locally**. If a provider is not present then OpenTofu is unable to determine the schema. There is no graceful degredation, its a hard fail. So if a plan file is copied from a CI runner to a reviewer's laptop, or an agent sandbox, this cannot display the planfile as json at all unless the environment is populated with providers correctly through `tofu init`.
+
+## Proposed Solution
+
+OpenTofu should persist the provider schemas that are needed by the `show` command **inside the planfile**, written at plan-creation time from the schemas already in memory, and have `show` prefer using those over launching providers.
```
cam72cam, line 38: Why not `**inside the statefile**` (also stored in the planfile)? I wonder if it would be a more comprehensive solution, instead of just solving this for planfile?

Yantrio: I feel like it is much less risky to implement this as a separate entry in the archive for backwards compatability reasons. 

I'm happy to discuss the option of storing elsewhere if you think it could work better.

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+- **Compute and time at scale** - Teams that are running `tofu` at a large scale run plans hundreds or even thousands of times a day, and this adds up to a very non-negligable recurring compute cost over time.
+
+The problem today is that rendering a saved planfile as json requires the schema of the resources from the provider and OpenTofu obtains those schemas in an expensive way, by launching every referenced provider plugin and asking each one for it's schema over gRPC. It does this even though the plan was built with those exact schemas moments earlier in the plan execution. So every consumer re-derives from scratch a value that is fully determined by the plan. This issue is what this RFC plans to address.
+
+This RFC proposed storing a **minimal subset of provider schemas** that are required to render the planfile as json inside the plan file itsself, so that `tofu show` and other tooling built on it can decode and display the plan without requiring to launch providers.
```
diofeher:
```suggestion
This RFC proposed storing a **minimal subset of provider schemas** that are required to render the planfile as json inside the plan file itself, so that `tofu show` and other tooling built on it can decode and display the plan without requiring to launch providers.
```

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+Beyond this being just a speed improvement, this turns the plan file in to a self-describing and portable artifact. Which means that the consumer of the planfile for analytics may not always need to be the same machine that created it.
```
diofeher: I like that!

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+To get this `cty.Type` and metadata, OpenTofu needs to spawn the provider that manages the resource and ask for the schema over gRPC. This means spawning a seperate process, and in the case of large providers, transferring around megabytes of schema information.
+
+Because of this, **it is slow and it gets slower as you use more providers**. Each distinct provider in the plan means one more plugin to launch, one more gRPC call, one more teardown. Whilst this work is paralellized across pproviders, the cost of launching these providers is expensive.
+
+As mentioned above too, this **requires every provider to be installed locally**. If a provider is not present then OpenTofu is unable to determine the schema. There is no graceful degredation, its a hard fail. So if a plan file is copied from a CI runner to a reviewer's laptop, or an agent sandbox, this cannot display the planfile as json at all unless the environment is populated with providers correctly through `tofu init`.
```
diofeher:
```suggestion
As mentioned above too, this **requires every provider to be installed locally**. If a provider is not present then OpenTofu is unable to determine the schema. There is no graceful degradation, it's a hard fail. So if a plan file is copied from a CI runner to a reviewer's laptop, or an agent sandbox, this cannot display the planfile as json at all unless the environment is populated with providers correctly through `tofu init`.
```

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+- What to store
+- How to store
+- How to read
+
+#### 1. What to store
+
+During rendering the json, `jsonplan`, `jsonstate` and `jsonconfig` all access parts of the schema that is stored in memory.
+
+From this we can derive the minimal state of what resource schmeas need storing in the planfile. For each provider referenced by the plan, we can store a condensed version of the ProviderSchema that only contains
```
diofeher:
```suggestion
From this we can derive the minimal state of what resource schemas need storing in the planfile. For each provider referenced by the plan, we can store a condensed version of the ProviderSchema that only contains
```

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Proposed Solution
+
+OpenTofu should persist the provider schemas that are needed by the `show` command **inside the planfile**, written at plan-creation time from the schemas already in memory, and have `show` prefer using those over launching providers.
+
+The plan file today is already an archive (zip), I propose that we introduce one new, optional entry alongside the existing entries. This area of the codebase is extensible and already ignores unknown entries in the zip.
+
+Crucially, and this is possibly an extra bit of work ontop of this, we only store what is needed to render the plan, NOT the full schema of every provider. This ensures that the planfile does not baloon in size too much, especially for users of providers with a large amount of resources.
```
diofeher: Are there other downsides to this proposal besides adding size to the plan file?

diofeher (approved):
Other than small nits, looks good to me!

rfc/20260605-speed-up-tofu-show-planfile.md
apparentlymart: I would suggest that we don't add anything new to the state snapshot format because state snapshots tend to be subject to more constraints on storage size.

Some of the state storage implementations write into storage that has maximum size limits, and even for those that don't state snapshots are typically stored indefinitely and so additional bloat inside them adds up.

A little bloat of the plan file format is easier to justify because there's not really any useful reason to retain a plan file after you've used it with `tofu apply PLANFILE` -- it becomes immediately invalidated as soon as the state snapshot it was created from is not current anymore. At that point, if you already ran `tofu show -plan=PLANFILE` and saved its output then you've already done the only remaining useful thing you could do with that plan file.

rfc/20260605-speed-up-tofu-show-planfile.md
apparentlymart: I think it's worth calling out explicitly that you'd still need to run `tofu show -plan=PLANFILE` with exactly the same version of OpenTofu that created the plan file, because we intentionally don't provide forward- and backward-compatibility of the plan file format between OpenTofu versions.

(This means that if you want to run `tofu show` on a different computer than the one where the plan was created then you do still need to "remember" which version of OpenTofu was used to create the plan file and use that same version to generate the JSON representation of the plan.)

rfc/20260605-speed-up-tofu-show-planfile.md
cam72cam: I'm curious how many unique resource types are typically used in a particular configuration. I was imagining that it would be stored top level in the state, so only one schema per resource type.

apparentlymart: My sense of it is that we learned from https://github.com/opentofu/opentofu/issues/1593 that some folks are particularly sensitive to the size of state snapshots, and we responded by switching to using compact JSON by default to reduce the storage size and data transfer. Adding schemas would be a different form of arguably-unnecessary bloat that could upset that same group.

Of course, if we think having state snapshots include the relevant subset of schema to make them self-contained would be valuable enough then we _could_ make that change. The arguments made in this RFC don't really seem to justify significant value in doing that though -- it argues in terms of use-cases that primarily concern plans rather than state snapshots -- so I'd suggest that we start by identifying specific situations that would benefit from having schema subsets embedded in state snapshots so that we can weigh that against the costs.

(Agreed that we should also actually _measure_ the real-world cost rather than just assuming that any increase is necessarily bad. But I'd prefer to start with understanding what benefit we're hoping to get from it to see if it's even worth spending the time to measure the cost.)

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+This RFC proposed storing a **minimal subset of provider schemas** that are required to render the planfile as json inside the plan file itsself, so that `tofu show` and other tooling built on it can decode and display the plan without requiring to launch providers.
+
+Beyond this being just a speed improvement, this turns the plan file in to a self-describing and portable artifact. Which means that the consumer of the planfile for analytics may not always need to be the same machine that created it.
+
+## Background - Why are schemas needed?
+
+A saved plan stores resource attributes as an opaque msgpack value blob. To turn this blobs back into a typed JSON OpenTofu needs to know the `cty.Type` for each resource's block, plus per-resource metadata such as the schema version and sensitivity marks.
+
+To get this `cty.Type` and metadata, OpenTofu needs to spawn the provider that manages the resource and ask for the schema over gRPC. This means spawning a seperate process, and in the case of large providers, transferring around megabytes of schema information.
```
yottta: > transferring around megabytes

This makes me wonder: this means that we would store megabytes of data in the plan. If we talk in context of CIs, this will still have to be transferred, right? Ok, we get rid of the time spent to start the providers processes, but are we ok with having the size of the plan increased heavily?

rfc/20260605-speed-up-tofu-show-planfile.md
yottta: Having this stored in a separate file inside the plan archive, seems the right approach IMO. Especially because we can have a different "schema loader" component that can work similarly (or be integrated into) with the get schema from the providers.

About storing this in the state, as others said, without having specific use cases, don't see the need to do so. Furthermore, when comparing the state with the remote system, there is still need to fetch the schema again, and relying on the state stored schema seems to add more cycles and more points of failure. And what if (rare situation), the schema version got updated? We still need to compare it, meaning that we still have to fetch the schema from the provider.

In other words, I am for storing this only in the plan archive and leave the state untouched.

rfc/20260605-speed-up-tofu-show-planfile.md
yottta: ~Yeah, while I was on the "megabytes" discussion above, I was thinking about this too. Technically, could get quite intense and computation heavy, especially when the resource types used from a provider is quite diverse.~

EDIT: does not apply anymore since I resolved the first conversation from this review round.
Though, I will leave this conversation unresolved, maybe we still want to consider compressing.

yottta: Or maybe, for a first iteration, we could just compress the content of the file using [gzip](https://pkg.go.dev/compress/gzip) or [snappy](https://github.com/golang/snappy/) 🤔

rfc/20260605-speed-up-tofu-show-planfile.md
yottta:
```suggestion
From this we can derive the minimal state of what resource schmeas need storing in the planfile. For each provider referenced by the plan, we can store a condensed version of the ProviderSchema that only contains:
```

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+To store this, we should pass the ProviderSchemas through to the logic that writes the planfile (where `planfile.Create` is called should be a good entrypoint). The trimming of the schemas can happen in here and we can re-use the same mechanism of writing to the archive that the other entries in the archive use.
+
+#### How to read
+
+... TODO
+
+### Open Questions
+
+- Should we have this as a default enabled feature? - It's nearly free at write time and there's lots of benefits for this, however we should be aware that some people want to keep their planfiles smaller.
```
yottta, line 89: At least for the start, I would add an environment variable to enable this. Who wants to use it, feel free to set an environment variable.
Later, after we will (hopefully) have some feedback, we can rollout it to be enabled by default, and switch the env var to disable it if needed.
But the env var should always stay.

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+- Encryption - Schemas are not sensitive information, but they live inside the encrypted blob like everything else in the planfile, we should check that that's okay.
+- Should we do the same for human readable `tofu show`, or only `-json` ? I think it's a freebie but requires some investigation.
+
+### Ideas for the future
+
+- It would be nice to have a global persistent provider schema cache on disk instead of embedding it into the provider. This could be done alongside existing providder caching.
```
yottta:
```suggestion
- It would be nice to have a global persistent provider schema cache on disk instead of embedding it into the plan. This could be done alongside existing providder caching.
```
Or you meant about the schema that is already defined inside the provider binary? But how would that work? How would that be versioned together with the provider? How would it be distributed?
Maybe I misunderstood the idea.

rfc/20260605-speed-up-tofu-show-planfile.md
apparentlymart: This proposal is making the assumption that by the time we are writing the plan file we already know exactly what subset of the schema information we would need to render the plan -- we need only the resource types of resources that appear in the plan -- and so the new plan file content should be significantly smaller than the full data transferred from a provider at the beginning of the plan process.

yottta: Thanks for poking.
> This RFC proposed storing a minimal subset of provider schemas that are required...

Probably went too fast over this part at the beginning 🤦

rfc/20260605-speed-up-tofu-show-planfile.md
Yantrio: Not as far as I'm aware.

rfc/20260605-speed-up-tofu-show-planfile.md
Yantrio: Im curious if maybe we add this as something inside the language block experimental flags. What do you think about this @opentofu/maintainers ? Im curious on other opinions here.

My gut feeling is that the plan size shouldn't increase too much, and so we should be fine enabling this by default and people get real performance gains with zero changes.

rfc/20260605-speed-up-tofu-show-planfile.md
Yantrio: You're right, I should've said "plan" here. Sorry for the confusion!

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+- Generic CI/CD users who execute tofu
+- Cost estimation tooling (Infracost/etc)
+- Tooling around policy engines (OPA/Conftest/Sentinel-style tools)
+- Increasingly, AI/LLM based agents
+
+For all of these, `tofu show -json <planfile>` is on the hot path, called once or many times per execution of `tofu plan`.
+
+This slow execution on the hot path hits end users in different ways, but I would like to focus on 2 primary ways with this RFC:
+
+- **Feedback loop latency** - This is the story for agentic development, allowing an agent to run a tight loop of `tofu plan` -> read the json -> reason, infer, fix, edit -> plan again`. The factor that matters here is what I will call "Time to actionable output". Launching even a single provider with a large schema taxes every cycle of this loop. Over time this adds up.
```
yottta:
```suggestion
- **Feedback loop latency** - This is the story for agentic development, allowing an agent to run a tight loop of `tofu plan` -> read the json -> reason, infer, fix, edit -> plan again. The factor that matters here is what I will call "Time to actionable output". Launching even a single provider with a large schema taxes every cycle of this loop. Over time this adds up.
```
Like that? If not, please fix the backticks pairing on this line.

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+To get this `cty.Type` and metadata, OpenTofu needs to spawn the provider that manages the resource and ask for the schema over gRPC. This means spawning a separate process, and in the case of large providers, transferring around megabytes of schema information.
+
+Because of this, **it is slow and it gets slower as you use more providers**. Each distinct provider in the plan means one more plugin to launch, one more gRPC call, one more teardown. Whilst this work is parallelized across providers, the cost of launching these providers is expensive.
+
+As mentioned above too, this **requires every provider to be installed locally**. If a provider is not present then OpenTofu is unable to determine the schema. There is no graceful degradation, its a hard fail. So if a plan file is copied from a CI runner to a reviewer's laptop, or an agent sandbox, this cannot display the planfile as json at all unless the environment is populated with providers correctly through `tofu init`.
```
yottta:
```suggestion
As mentioned above too, this **requires every provider to be installed locally**. If a provider is not present then OpenTofu is unable to determine the schema. There is no graceful degradation, it's a hard fail. So if a plan file is copied from a CI runner to a reviewer's laptop, or an agent sandbox, this cannot display the planfile as json at all unless the environment is populated with providers correctly through `tofu init`.
```

rfc/20260605-speed-up-tofu-show-planfile.md
```diff
@@ -0,0 +1,87 @@
...
+
+Provider schemas already have a protobuf representation as defined by the `GetProviderSchema` response in `tfplugin{5|6}`, and the codebase has existing logic to handle conversion in both directions. Reusing this protobuf representation simplifies the implementation of this feature and avoids inventing a new format that could drift over time.
+
+The new proposed zip entry should be a new protobuf message, a map of provider address to the trimmed per-provider schema.
+
+To store this, we should pass the ProviderSchemas through to the logic that writes the planfile (where `planfile.Create` is called should be a good entrypoint). The trimming of the schemas can happen in here and we can re-use the same mechanism of writing to the archive that the other entries in the archive use.
+
+#### How to read
+
+... TODO
```
yottta: Planning on having this filled in before merging this RFC?

rfc/20260605-speed-up-tofu-show-planfile.md
yottta: > My gut feeling is that the plan size shouldn't increase too much

Agree, especially because the plan is `gzip`ed.
But I would take a separate note and have this benchmarked after the implementation (before delivering it) and decide based on that.
Furthermore, if there will be a request in the future to have an option to opt-out of this, I would accept that issue without much pondering.

apparentlymart: Historically the plan file format has changed significantly over time, both growing and shrinking between versions, without any direct means to control those changes. I don't recall any significant complaint about these changes, though of course that could change if we made the files _significantly_ larger on average (which is why I raised that as a potential concern for https://github.com/opentofu/opentofu/pull/2049).

Personally I don't think we should make this runtime-configurable unless we expect this change will make plan files an order of magnitude larger. With filtering to only what the current configuration is using and general compression as part of the `.zip` container I am skeptical that this will make a significant difference for typical OpenTofu usage, since there are already various practical limits that discourage having a large number of resources in a single configuration and the number of distinct resources is the limit on the number of schema entries a particular plan file could contain.

All of that said: I'd suggest that we start with this as a mandatory change and then during implementation measure how much practical difference it makes with a selection of realistic configurations. If it turns out to make plan files significantly larger in practice then that calls into question whether we should do this at all, because the tradeoff being made by the proposal is aiming for a modest increase in filesize in return for a greater (but still modest) decrease in runtime of `tofu show -plan=...`. It doesn't seem to me that this modest decrease in runtime justifies the complexity of a user-facing option for it.

cam72cam (approved):

yottta (approved):

apparentlymart (approved):

lbordowitz (approved):
Side-note: when providing a rendered view, please refer to the branch rather than a commit. I was reading the old version of this RFC and was ready to make some comments, only to find they were already addressed.

PR #4250, opened 2026-06-10
Add depends_on support to the new engine

### LINKED ISSUE #4065: Implement `depends_on` for the new language runtime

cam72cam:
I tried to keep as much of this PR in the tofu package as possible, as the current understanding on "depends_on" is tied to assumptions throughout the current engine's dependency system.

I ended up taking the InstanceSelector approach discussed in the issue, as it seemed like the most straightforward place to implement it. The "weird" parts of this PR are in how we handle `module.name` implying depending on all resources in that branch of the module tree, as well as special cases to pull in marks from RepetitionData.

Open Questions:
* Is "AdditionalMarks" too vague?  Should we be more precise and say "DependsOn" instead?
* Are there any issues with utilizing this pattern via workgraph?
* Module depends_on does *not* apply to variables, locals, and outputs.
```hcl
# main.tofu

resource "simple" "dep" {}

module "mod" {
  source = "./mod"
  depends_on = [resource.simple.dep]
}

# Does NOT depend on simple.dep in the new engine
resource "simple" "outer" {
  value = module.mod.output
}


# mod/mod.tofu
resource "simple" "inner" {
}
# Does NOT depend on simple.dep in the new engine
output "output" {
  value = "some value"
}
```

Resolves #4065 

## Checklist

### Go checklist

 

### DIFF
--- a/internal/lang/eval/internal/configgraph/instances.go
+++ b/internal/lang/eval/internal/configgraph/instances.go
@@ -69,7 +69,12 @@ type InstanceSelector interface {
 	InstancesSourceRange() *tfdiags.SourceRange
 }
 
-type InstancesSeq = func(yield func(addrs.InstanceKey, instances.RepetitionData) bool)
+type InstanceSeq struct {
+	instances.RepetitionData
+	Marks cty.ValueMarks
+}
+
+type InstancesSeq = func(yield func(addrs.InstanceKey, InstanceSeq) bool)
 
 type compiledInstances[T any] struct {
 	KeyType addrs.InstanceKeyType
@@ -121,7 +126,7 @@ func compileInstances[T any](
 		if repData.CountIndex != cty.NilVal {
 			repData.CountIndex = repData.CountIndex.WithMarks(valueMarks)
 		}
-		obj := compileInstance(ctx, key, repData)
+		obj := compileInstance(ctx, key, repData.RepetitionData, repData.Marks)
 		ret.Instances[key] = obj
 	}
 	return ret, diags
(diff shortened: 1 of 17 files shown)


cam72cam:
Closing in favor of the more polished https://github.com/opentofu/opentofu/pull/4273

PR #4272, opened 2026-06-15
Start to wire in provider functions

### LINKED ISSUE #4061: Implement provider-defined functions in the new language runtime
OpenTofu's current language runtime has an intentionally-slightly-incorrect implementation of the Terraform plugin protocol for provider-defined functions, where it calls functions on configured providers rather than on unconfigured providers.

That strategy unfortunately violates the usual assumption that only resource instances can directly depend on provider instances, and instead any expression anywhere in the configuration could potentially depend directly on a provider instance. The new-style architecture currently relies on that assumption in its separation of concerns between the evaluator and the planning and applying engines: aside from provider-defined functions, the evaluator doesn't need to have any awareness of the concept of "configured providers" at all.

We need to decide first whether we want to retain this intentional behavior difference -- since only a few esoteric providers appear to be relying on it -- and then if so figure out a good compromise for how to support it in the new runtime.

cam72cam:
Configured provider functions makes this a bit more interesting than it should normally be.

Looking for some early feedback on this draft. Specifically on if this can get us into trouble dependency wise in the new engine, as well as if I've put things in the correct abstraction layers.

As a side effect, I've cleaned up ValidateProviderConfig in the new engine.

Resolves #4061

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_apply2_test.go
+++ b/internal/tofu/context_apply2_test.go
@@ -7057,7 +7057,6 @@ resource "test_resource" "res" {
 	}
 	plan, diags := ctx.Plan(context.Background(), m, states.NewState(), SimplePlanOpts(plans.NormalMode, nil))
 	assertNoErrors(t, diags)
-	assertState(t, plan.PlannedState)
 
 	state, diags := ctx.Apply(context.Background(), plan, m, nil)
 	assertNoErrors(t, diags)
--- a/internal/engine/applying/apply.go
+++ b/internal/engine/applying/apply.go
@@ -8,7 +8,9 @@ package applying
 import (
 	"context"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
 
 	"github.com/opentofu/opentofu/internal/addrs"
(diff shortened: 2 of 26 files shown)


yottta (approved):

apparentlymart (approved):
This seems like a reasonable approach!

I expect we're going to keep making adjustments to how we model providers as we continue here given how cross-cutting the provider functionality is, but based on what we know right now this seems like a good shape for the system.

internal/tofu/context_apply2_test.go
```diff
@@ -7057,7 +7057,6 @@ resource "test_resource" "res" {
 	}
 	plan, diags := ctx.Plan(context.Background(), m, states.NewState(), SimplePlanOpts(plans.NormalMode, nil))
 	assertNoErrors(t, diags)
-	assertState(t, plan.PlannedState)
```
lbordowitz, line 7060: Why is this removed?

lbordowitz (approved):
One small nit, but nothing that should be blocking.

internal/tofu/context_apply2_test.go
cam72cam: We do not output PlannedState in the new runtime, this is an atypical check for a test and does not accomplish anything.

PR #4273, opened 2026-06-15
tofu2024: Support for `depends_on` in the new language runtime

### LINKED ISSUE #4065: Implement `depends_on` for the new language runtime

apparentlymart:
This series of commits makes the explicit `depends_on` argument work in the new language runtime, for resource and module call blocks. This is based on earlier work that @cam72cam did in https://github.com/opentofu/opentofu/pull/4250, but slightly refactored so that more of the logic lives in `package tofu2024` instead of `package configgraph`.

As predicted in https://github.com/opentofu/opentofu/issues/4065, the handling of module-call-level dependencies was the trickiest part of this, in two ways:

- When a `module` block has `depends_on` in it, the current language treats that as if every resource instance declared in any descendant module depends on whatever was declared, regardless of whether any values passed into the module's input variables have dependencies or whether resource instance inside the module depend on those input variables.

    To handle that, the API that a parent module uses to compile a child module is extended to include the possibility for explicit dependency marks as another kind of input from the caller, alongside the existing input variables and "providers side-channel". The compilation process for the child module then arranges for those dependency marks to be attached to every resource instance compiled in the module, _and_ to any nested module calls.
- When a `depends_on` argument refers to a module call, as in `depends_on = [module.foo]`, the resource or module that argument appears in is considered to depend on every resource instance declared inside that module, regardless of whether the module's output values depend on those resource instances individually.

    To handle that, we have a special case in the code that compiles `depends_on` where it does a static check for whether the given traversal refers to a module call. If so, it enters a very special-cased codepath which just asks the child module call to enumerate all of the resource instances declared in it and announces dependencies on each of them, without performing any normal expression evaluation.

    This is something we intend to revisit later since it makes explicit dependencies on module calls act very inconsistently to everything else, but this is at least a starting point that approximates the (somewhat-questionable) behavior of the current language runtime.

Aside from those special cases, the strategy here is mainly to treat `depends_on` as if all of the expressions inside it were part of the main configuration of the object the dependencies are being declared for, augmenting the final object value to have additional dependency-marks so that the rest of the system will handle them in much the same way as it handles dependencies that are implied by normal references.

The way `depends_on` is "compiled" in `package tofu2024` is a little more complicated than it strictly needs to be for the current state of things, since `package configs` is currently forcing all `depends_on` items to be static traversals and so dynamic references to specific instances of other objects are not possible. However, we already identified that we want to be able to support arbitrary expressions in future, to allow for things like `depends_on = [aws_subnet.example[each.key]]` where this gets evaluated separately for each instance so they can all have a different `each.key`. The API here is overachieving to illustrate that this approach _can_ support that functionality, even though this initial implementation just has it stubbed out.

Closes https://github.com/opentofu/opentofu/issues/4065

---

### DIFF
--- a/internal/instances/instance_key_data.go
+++ b/internal/instances/instance_key_data.go
@@ -6,6 +6,8 @@
 package instances
 
 import (
+	"maps"
+
 	"github.com/zclconf/go-cty/cty"
 )
 
@@ -30,4 +32,36 @@ type RepetitionData struct {
 	// For correct operation, EachKey must always be either of type cty.String
 	// or cty.Number if not nil.
 	EachKey, EachValue cty.Value
+}
+
+// AllValueMarks returns a mark set containing the full set of marks across
+// all values inside the receiver.
+//
+// Use this instead of accessing each field in turn because future changes
+// might add new fields to [RepetitionData] which will be included in this
+// function's result without having to update every caller.
+func (rd *RepetitionData) AllValueMarks() cty.ValueMarks {
+	if rd == nil {
+		return nil
+	}
+	var ret cty.ValueMarks
+	addMore := func(v cty.Value) {
+		// We make some effort here to avoid creating additional temporary
+		// cty.ValueMarks values because [cty.Value.Marks] already makes it
+		// copy itself anyway, so we'd prefer to just write into the set
+		// it returned than to make yet another copy of it.
+		marks := v.Marks()
+		if len(marks) == 0 {
+			return
+		}
+		if ret == nil {
+			ret = marks
+			return
+		}
+		maps.Copy(ret, marks)
+	}
+	addMore(rd.CountIndex)
+	addMore(rd.EachKey)
+	addMore(rd.EachValue)
+	return ret
 }
--- a/internal/instances/instance_key_data.go
+++ b/internal/instances/instance_key_data.go
@@ -32,6 +32,31 @@ type RepetitionData struct {
 	// For correct operation, EachKey must always be either of type cty.String
 	// or cty.Number if not nil.
 	EachKey, EachValue cty.Value
+
+	// DecisionMarks describes any cty marks that were present on values that
+	// were directly involved in the decision to include the instance that this
+	// RepetitionData object belongs to.
+	//
+	// For example, in the case of for_each with a map this would include any
+	// marks from the map itself but not marks from the elements inside the map.
+	// The elements' own marks would appear on EachValue instead.
+	//
+	// NOTE: The old language runtime in "package tofu" does not populate or
+	// make use of this field. It is used only by the new runtime's evaluator,
+	// in the packages under "internal/lang/eval".
+	DecisionMarks cty.ValueMarks
+}
+
+// HasSymbolValues returns true if any of the fields that correspond to
+// additional symbols that would describe the current item in a repetition
+// are set to non-nil values.
+//
+// This is a somewhat-imprecise signal for "is using repetition at all", since
+// in the singleton or conditional enabled cases none of the value fields are
+// populated, but this should only be used for debug-related information like
+// logging and not for making any real behavioral decisions.
+func (rd *RepetitionData) HasSymbolValues() bool {
+	return rd != nil && rd.CountIndex != cty.NilVal || rd.EachKey != cty.NilVal || rd.EachValue != cty.NilVal
 }
 
 // AllValueMarks returns a mark set containing the full set of marks across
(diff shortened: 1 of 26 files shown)


internal/lang/eval/internal/configgraph/resource_instance_deps.go
```diff
@@ -130,3 +140,28 @@ func ResourceInstanceAddrs(insts iter.Seq[*ResourceInstance]) iter.Seq[addrs.Abs
 		}
 	}
 }
+
+// IsDependencyMark returns true if the given value is something that could
+// be used to mark a [cty.Value] to represent a "dependency".
+//
+// "Dependency" here means that some sort of externally-visble change must
```
yottta:
```suggestion
// "Dependency" here means that some sort of externally-visible change must
```

internal/lang/eval/internal/configgraph/module_call.go
```diff
@@ -95,7 +95,15 @@ var _ exprs.Valuer = (*ModuleCall)(nil)
 func (c *ModuleCall) Instances(ctx context.Context) map[addrs.InstanceKey]*ModuleCallInstance {
 	// We ignore the diagnostics here because they will be returned by
 	// the Value method instead.
-	result, _ := c.decideInstances(ctx)
+	result, diags := c.decideInstances(ctx)
+	if diags.HasErrors() && result == nil {
```
cam72cam, line 99: Should this be `||` instead? If we can have no diags, but a nil result, this will panic below?

cam72cam: (and similar comments on the same pattern below)

internal/lang/eval/internal/tofu2024/module_instance_scope.go
```diff
@@ -574,3 +578,107 @@ func (i *instanceLocalSymbolTable) ResolveAttr(ref hcl.TraverseAttr) (exprs.Attr
...
+	// rules: they have some characteristics in common with goroutines but
+	// are not capable of running asynchronously with the calling goroutine.
+	return evalglue.ResourceInstancesDeep(ctx, calledModuleInst)
+}
+
+// compiledModuleInstanceFromScope knows how to extract a
+// *CompiledModuleInstance from the various concrete [exprs.Scope]
+// implementations in this package.
+//
+// If the given scope is not know this function knows how to handle then it
```
yottta:
```suggestion
// If the given scope is not known this function knows how to handle then it
```

yottta (approved):
I just want to let this here since, personally, I had to test this to understand precisely "how" those points from the PR manifest in the current engine.

The relationships between the module's resources and the `depends_on` configured resource is transitive, and not direct. I was aware of the transitive nature, but the way _I understood_ the points from the description kind of made me think that I somehow missed the direct edges between the actors in question.

The same applies for the other point, when the module is configured with `depends_on`. The dependencies are expressed in a transitive way and not in a direct one.

cam72cam (approved):

internal/instances/instance_key_data.go
```diff
@@ -30,4 +32,65 @@ type RepetitionData struct {
...
+// all values inside the receiver.
+//
+// Use this instead of accessing each field in turn because future changes
+// might add new fields to [RepetitionData] which will be included in this
+// function's result without having to update every caller.
+//
+// Note that this only reports marks directly on the enclosed values and not
+// marks on values nested within them.
+func (rd *RepetitionData) AllValueMarks() cty.ValueMarks {
+	var ret cty.ValueMarks
```
diofeher, line 75: Should we add a nil-guard here like the one in **HasSymbolValues**?

```
func (rd *RepetitionData) HasSymbolValues() bool {
	return rd != nil ...
...
```

internal/lang/eval/internal/configgraph/module_call.go
apparentlymart: If we have no errors then the result being nil would be a bug in `decideInstances`, and so panicking in that case is intentional to draw attention to that bug instead of masking it.

This is intentionally allowing only the expected situation of it possibly (but not necessarily) being nil when there are errors, such as when the repetition or depends_on arguments cause a self-reference error.

internal/lang/eval/internal/tofu2024/module_instance_scope.go
apparentlymart: Thanks! This was actually intended to be "one" rather than "known" ... I'm not sure how I managed to type "know" instead of "one". :man_facepalming:

PR #4280, opened 2026-06-16
moved: change type with provider

### LINKED ISSUE #4271: Resources of the same type with a different provider are not migrated properly during moves
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.4
on darwin_arm64
```

```hcl
resource "random_pet" "dog" {

}

resource "random_pet" "cat" {

}

```

```hcl

resource "random_pet" "gerbil" {

}

moved {
    from = random_pet.dog
    to = random_pet.gerbil
}
```

### Debug Output

Not applicable

### Expected Behavior

When reproducing for resources of the same type but different provider (see steps below), it should detect that the resource types are different and act accordingly

### Actual Behavior

The move occurs as though the type did not change.

### Steps to Reproduce

I start with `cat` and `dog` in a config file. I then remove `dog`, add the `gerbil` and the `moved` statement, and crucially I change the provider.

### Additional Context

Internally, the function `isResourceMovedToDifferentType` does not check whether the resource was the same provider or not.

### References

_No response_

lbordowitz:
Resolves #4271

This fixes the case when the provider for a resource is different by dint of namespace or hostname. It is an unusual scenario, but a sharp corner nonetheless.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/tofu/context_plan2_test.go
+++ b/internal/tofu/context_plan2_test.go
@@ -35,6 +35,8 @@ import (
 	"github.com/opentofu/opentofu/internal/providers"
 	"github.com/opentofu/opentofu/internal/states"
 	"github.com/opentofu/opentofu/internal/tfdiags"
+
+	regaddr "github.com/opentofu/registry-address/v2"
 )
 
 func TestContext2Plan_removedDuringRefresh(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/tofu/context_plan2_test.go
```diff
@@ -1817,6 +1819,123 @@ func TestContext2Plan_movedResourceToDifferentType(t *testing.T) {
 		})
 	}
 }
+func TestContext2Plan_movedResourceToDifferentProvider(t *testing.T) {
+	// TODO determine whether this should just be "folded in" to the previous test...
```
lbordowitz: This test is heavily adapted from the previous one. Should I just fold it in, or keep it separate?

internal/tofu/node_resource_abstract.go
```diff
@@ -640,7 +640,15 @@ func (n *NodeAbstractResourceInstance) readResourceInstanceState(ctx context.Con
...
 	}
-	if isResourceMovedToDifferentType(addr, prevAddr) {
+	// temporarily set prevProviderAddr to the current one,
+	// and use it if prevRunState is not set in this context
+	providerAddr := n.Provider()
+	prevProviderAddr := providerAddr
+	prevRunState := evalCtx.PrevRunState()
+	if prevRunState != nil {
+		prevProviderAddr = prevRunState.ResourceProvider(addr.AffectedAbsResource()).Provider
+	}
```
lbordowitz: Should I fold all this into a private function or something? Also, it's so clunky and clumsy, because I kept getting nil dereference errors otherwise; is this really the best way to go about it?

yottta: For me this looks ok. I would extract it into a new method/function if doesn't add too much friction, but without trying I cannot really say.

internal/tofu/context_plan2_test.go
yottta: I would keep it this way and add a comment that this tests #4271 situation.

yottta (approved):
For me, this looks good. But please wait for others, maybe there is a better way to determine the old and new provider addr that I miss.

cam72cam (approved):

internal/tofu/node_resource_abstract.go
```diff
@@ -666,6 +667,18 @@ func (n *NodeAbstractResourceInstance) readResourceInstanceState(ctx context.Con
...
 }
 
+func (n *NodeAbstractResourceInstance) getResourceProviderAddrs(evalCtx EvalContext, addr addrs.AbsResourceInstance) (addrs.Provider, addrs.Provider) {
+	// temporarily set prevProviderAddr to the current one,
+	// and use it if prevRunState is not set in this context
+	providerAddr := n.Provider()
+	prevProviderAddr := providerAddr
+	prevRunState := evalCtx.PrevRunState()
+	if prevRunState != nil {
+		prevProviderAddr = prevRunState.ResourceProvider(addr.AffectedAbsResource()).Provider
```
lbordowitz, line 677: My final misgiving is that the `ResourceProvider` call puts a hold on the `SyncState`'s lock. Will this mess with performance in a significant way? I do not know.

yottta: Maybe profile this comparatively with the latest version of OpenTofu? 🤔

cam72cam: It's a lock, but a pretty minor one IMO. Agree with @yottta that it should be checked using either the benchmark tests in this package of opentofu-performance-testing.

apparentlymart: FWIW, I would be surprised if this makes any significant difference. These codepaths are dominated by RPC calls to provider plugins, and so as long as we're not holding the lock over any of those calls (which this new code is not) then a very brief additional lock on the previous run state ought to be immaterial.

We've merged code with much more potential for hurting performance without proactively running the performance tests before; we typically reserve that only for PRs that are significantly changing architecture or that are intentionally aiming to improve performance.

I would suggest that we go ahead and merge this now, without waiting to run the performance tests first, and then plan to do a one-time comparison of v1.12 vs v1.13 using the performance tests just after we publish v1.13.0-beta1, so we can consider the impact of all of the changes we made in this development period taken together.

lbordowitz: I ran the benchmark tests in the `tofu` package. They're nominal.

PR #4226, opened 2026-06-03
Better `create_before_destroy` handling for the new language runtime

apparentlymart:
(This is for https://github.com/opentofu/opentofu/issues/4057, but is not yet sufficient to close it.)

This series of commits implements a new execution subgraph shape for "create-then-destroy" replace operations, ensuring that the expected resource instance object addresses reach each of the execution graph nodes involved in the process.

This follows the new graph shape I discussed at the end of https://github.com/opentofu/opentofu/issues/4057#issuecomment-4578449311, which I'm replicating here for easier future reference:

```mermaid
graph TB

prior_state[Read old from prior state]
desired_state[Read new from config]
delete_plan[Plan to delete old]
create_plan[Plan to create new]
depose[Depose old]
create_apply[Create new]
delete_apply[Delete old]
reinterp_deposed([reinterpret as\ndeposed abc123])
style reinterp_deposed font-size:0.9em

prior_state -->|prior state| delete_plan
prior_state -->|prior state| depose
desired_state -->|desired state| create_plan

delete_plan -->|final plan| reinterp_deposed
reinterp_deposed -->|final plan| depose
reinterp_deposed -->|final plan| delete_apply
create_plan -->|final plan| create_apply
create_plan -.-> depose
depose -->|fallback| create_apply
create_apply -.-> delete_apply
```

Along the way I fixed some other assorted small problems that were blocking me from implementing and testing this. Therefore **it's probably best to [review this on a commit-by-commit basis](https://github.com/opentofu/opentofu/pull/4226/commits)**.

A small number of tests guarded with `ExperimentalFeatureCBD` are still failing, and although it seems that they are failing for reasons unrelated to `create_before_destroy` I've not completely confirmed that yet. This changeset is already too large so I'll deal with the remaining failures in a future PR.

I have one other known omission here that I'm intentionally skipping: there are various other kinds of marks that could potentially show up on a `create_before_destroy` argument once we generalize it to support arbitrary expressions, but right now in practice it can only be a constant value and so there isn't any way to test that additional complexity without writing some unit tests that are closer to `package configgraph` and `package planning`. I'll do that in a future PR.

---

- [x] I have read the [contribution guide](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md).
- [x] I have not used an AI coding assistant to create this PR.
- [x] I have written all code in this PR myself OR I have marked all code I have not written myself (including modified code, e.g. copied from other places and then modified) with a comment indicating where it came from.
[...]

PR #4282, opened 2026-06-16
resources: Compensate for providers' incorrect `RequiresReplace`

apparentlymart:
OpenTofu and its predecessor did not historically raise an error when providers returned `RequiresReplace` paths referring to attributes whose values haven't actually changed and instead just silently ignored them.

We need to mimic that behavior in the new runtime too, for compatibility. This is an extension of the previously-added rule of just throwing the entire `RequiresReplace` value away if the planned action is create or delete. For "update" we can't just throw the whole set away, but we do throw away any paths that refer to something that hasn't changed and thus continue the principle that `package resources` is the one responsible for checking for and dealing with provider misbehavior.

This new logic is based directly on the equivalent logic from the old runtime in `package tofu`, implementing exactly the same filtering rules.

---

This was ultimately the cause of the remaining "CBD-related" tests that were failing after https://github.com/opentofu/opentofu/pull/4226. As I suspected at the time, the failures were not directly related to the create-before-destroy handling and were instead just overrepresented in those particular tests because `create_before_destroy` tests always involve "replace" actions in one way or another.

### DIFF
--- a/internal/engine/planning/plan_managed.go
+++ b/internal/engine/planning/plan_managed.go
@@ -305,13 +305,7 @@ func (p *planGlue) planDesiredManagedResourceInstance(
 	plannedAction := plans.Update
 	if prevRoundState == nil {
 		plannedAction = plans.Create
-	} else if len(planResp.RequiresReplace) != 0 {
-		// FIXME: The mere presence of RequiresReplace items is not sufficient
-		// because existing providers often generate spurious entries in there
-		// for paths that aren't actually changing. We need to filter out
-		// any paths whose values are equal between prior and planned in order
-		// to get the expected behavior for real-world providers.
-
+	} else if !planResp.RequiresReplace.Empty() {
 		// For "replace" actions the execution graph will include two separate
 		// plan and apply operations, where one handles deletion and the other
 		// handles creation. There is therefore an implicit third intermediate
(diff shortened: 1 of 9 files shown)


apparentlymart:
Interesting... the test run here raised some errors that didn't appear when I ran the `package tofu` tests locally. I'm going to switch this back to draft while I try to understand what I'm doing differently to what the the automated check is doing. :thinking: 

**Edit:** my dev system's test cache was stale and so when I re-ran the whole suite it ended up running in the mode where the CBD tests were still being skipped. :man_facepalming:

apparentlymart:
Okay... I subsequently dug through all of the other failures and figured out what missing feature or bug was causing them and now my second commit includes a bunch of new `SkipExperimental` for that.

Along the way I noticed that the bug we've been calling `ExperimentalBugMissingProvider` is unfortunately intermittent, and I expect that actually any test that includes an apply phase with at least one resource instance change can _potentially_ trigger it, though for some reason it's more likely in some than others. I added the `SkipExperimental` to the ones that were triggering more often than not as I re-ran the test suite, but I'm sure there are other examples here that I didn't catch yet. It doesn't seem to have anything to do with create-before-destroy directly, so I'm not going to try to improve that situation in this PR.

yottta (approved):

cam72cam (approved):

PR #4284, opened 2026-06-17
Fix a variety of lingering provider issues within the new runtime

cam72cam:
## Checklist

### Go checklist

 

### DIFF
--- a/internal/lang/eval/internal/tofu2024/providers_sidechannel.go
+++ b/internal/lang/eval/internal/tofu2024/providers_sidechannel.go
@@ -35,17 +35,7 @@ import (
 
 // rootMissingProviders is used to ferry the dynamically injected missing providers
 // into the root module's ProviderConfig lookup
-type rootMissingProviders struct {
-	lock            sync.Mutex
-	providerConfigs map[addrs.LocalProviderConfig]*configgraph.ProviderConfig
-}
-
-func (r *rootMissingProviders) getOk(localAddr addrs.LocalProviderConfig) (*configgraph.ProviderConfig, bool) {
-	r.lock.Lock()
-	defer r.lock.Unlock()
-	ret, ok := r.providerConfigs[localAddr]
-	return ret, ok
-}
+type rootMissingProviders func(localAddr addrs.LocalProviderConfig) (*configgraph.ProviderConfig, tfdiags.Diagnostics)
 
 // compileProviderConfigRefMissingInRoot builds the "base" provider ref compiler used by the root module
 // This specifically handles the legacy edge cases where a resource references a provider that does not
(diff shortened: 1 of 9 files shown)


internal/tofu/context_apply_test.go
```diff
@@ -1069,7 +1069,7 @@ func TestContext2Apply_createBeforeDestroyUpdate(t *testing.T) {
 		}, nil),
 	})
 
-	SkipExperimental(t, ExperimentalBugMissingProvider)
+	SkipExperimental(t, ExperimentalBugMissingProvider, ExperimentalBugExecGraph2)
```
apparentlymart, line 1072: Note to me: this is guarding a confusing execution graph construction bug we observed while testing this changeset, where somehow the execution graph ended up having two "load the prior state" operations for the same resource instance, causing nondeterminisim.

I'm intending to look into this one more after this PR is merged.

apparentlymart (approved):

internal/lang/eval/internal/tofu2024/providers_sidechannel.go
```diff
@@ -124,6 +123,20 @@ func compileProviderConfigRefModule(
...
+		// Known BUG: this intentionally simplifed and does *not* implement some of the
+		// odd legacy behaviors before the required_providers block was introduced.
+		//
+		// More specifically, a "provider" block in the configuration with only an alias
+		// field set would simulate what we would now use required_providers for. This
+		// is an ambiguous and confusing pattern that has not been recommended for many
+		// years.
+		// This can been seen in the [tofu.TestContext2Apply_moduleInheritAlias] test.
+		//
+		// One possible way of fixing this would be to include the nessesary information
```
diofeher, line 135:
```suggestion
		// One possible way of fixing this would be to include the necessary information
```

internal/lang/eval/internal/tofu2024/providers_sidechannel.go
```diff
@@ -124,6 +123,20 @@ func compileProviderConfigRefModule(
 ) configgraph.CompileProviderConfigRef {
 	return func(ctx context.Context, providerInstAddr addrs.LocalProviderConfig) exprs.Valuer {
 
+		// Known BUG: this intentionally simplifed and does *not* implement some of the
```
diofeher, line 126:
```suggestion
		// Known BUG: this intentionally simplified and does *not* implement some of the
```

internal/lang/eval/internal/tofu2024/providers_sidechannel.go
```diff
@@ -108,13 +97,23 @@ func compileProviderConfigRefMissingInRoot(
...
+			if diags.HasErrors() {
+				return exprs.ForcedErrorValuer(diags)
+			}
+
+			return &sidechannelProviderInstanceRefValuer{
+				localAddr:  providerInstAddr,
+				mainValuer: missingConfig,
+			}
+		}, func(localAddr addrs.LocalProviderConfig) (*configgraph.ProviderConfig, tfdiags.Diagnostics) {
+			return providerConfig(context.Background(), localAddr)
```
diofeher, line 115: Shouldn't we pass the context from the caller?

cam72cam: We don't have that context here, it's also not used for any evaluation so background is fine.

PR #4285, opened 2026-06-17
internal/tracing: route OTel SDK logs through TF_LOG via custom logr.LogSink

### LINKED ISSUE #4102: OpenTelemetry Tracing: better feedback in TF_LOG
### Community note

> [!TIP]

### OpenTofu Version

```shell
Any version
```

I'm setting up OpenTofu's experimental OTel tracing (`OTEL_TRACES_EXPORTER=otlp`) and shipping spans to a backend (Datadog APM via the OTel Collector, and separately experimenting with backend-direct OTLP intakes). I cannot tell from the OpenTofu side whether spans are actually being emitted.

In [`internal/tracing/init.go`](https://github.com/opentofu/opentofu/blob/v1.11.7/internal/tracing/init.go), the OpenTelemetry Go SDK's internal logger is wired up via `stdr`:

```go
logger := stdr.New(log.New(os.Stdout, "", log.LstdFlags|log.Lshortfile))
otel.SetLogger(logger)
```

`stdr` defaults to verbosity `0`, which means errors only. The Go SDK does emit useful `INFO`-level diagnostics (batch flushes, exporter creation, queue saturation) at higher verbosities, but they are suppressed today. The only OpenTofu-side signals are:

| Signal | Source | Trigger |
| --- | --- | --- |
| `[TRACE] OpenTelemetry: OTEL_TRACES_EXPORTER not set...` | `log.Printf` | Tracing disabled |
| `[ERROR] OpenTelemetry error: <err>` | `otel.SetErrorHandler` → `log.Printf` | Export failure only |

The success path is completely silent. This causes two concrete problems:

1. **Intermittent failures are hard to diagnose.** process exits before `ForceFlush` ack, protocol/port mismatch (`http/protobuf` on `:4318` default vs gRPC on `:4317`)
2. **Collector-less / backend-direct setups have nowhere else to look.** When sending traces directly from OpenTofu to a SaaS OTLP intake, e.g. [Datadog's direct OTLP intake endpoint](https://docs.datadoghq.com/opentelemetry/setup/otlp_ingest/), [Honeycomb's OTLP endpoint](https://docs.honeycomb.io/send-data/opentelemetry/), Grafana Cloud, Lightstep, without a local OTel Collector, the operator has no intermediate vantage point. The collector-side `debug` exporter doesn't exist, and the SaaS backend gives no visibility into what was *received-but-rejected* vs. *never-received*. When traces don't appear, you can't tell whether (a) OpenTofu never emitted, (b) OpenTofu emitted but the export failed silently (auth, TLS, endpoint path), or (c) the spans were delivered but rejected/dropped on the backend side (wrong dataset, sampling, schema). Today I am debugging this blindly.

### Attempted Solutions

```shell
export TF_LOG=error
export OTEL_TRACES_EXPORTER=otlp
export OTEL_EXPORTER_OTLP_PROTOCOL=grpc
export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
export OTEL_EXPORTER_OTLP_INSECURE=true
export OTEL_SERVICE_NAME=opentofu

tofu plan 2>&1 | grep -i 'opentelemetry\|otel'
```

`TF_LOG=error` surfaces the `[ERROR] OpenTelemetry error: ...` line from the registered `otel.SetErrorHandler`, which catches export failures. But there is no equivalent confirmation that emission *succeeded*, when the run completes cleanly with nothing in the grep output, I still can't tell whether spans went out.

I also tried `OTEL_LOG_LEVEL=debug`, but the OTel Go SDK does not honor it so it has no effect.

### Proposal

Add an `OTEL_LOG_VERBOSITY` (or `OTEL_LOG_LEVEL`) env var that, when set to an integer, is passed to `stdr.SetVerbosity()` after the logger is registered change in `internal/tracing/init.go`

--
OR, set the verbosity as the same level as TF_DEBUG.

Happy to open a PR if this direction is acceptable.

- **Collector-side `debug` exporter.** Works when there *is* a collector, adding `debug` to the traces pipeline pretty-prints every received span to the collector's stdout, confirming OpenTofu's leg works. Does not help in backend-direct setups, and adds an entire OTel Collector deployment just for diagnostics.
  - But I don't want that. I'm currently just using a streamlined direct OTel endpoint with the tooling, e.g. Honeycomb https://docs.honeycomb.io/send-data/traces/opentelemetry#using-the-honeycomb-opentelemetry-endpoint
- **Bisect via `TF_LOG=error`.** Catches export failures, but not the silent-success path that's needed to rule OpenTofu in/out as the source of missing spans.

Compared to these, an env var would let an operator type one line in the shell to confirm "OpenTofu did its job," and then shift the investigation to the backend config (API key, dataset, endpoint path).

### References

- [`internal/tracing/init.go`](https://github.com/opentofu/opentofu/blob/v1.11.7/internal/tracing/init.go)
- #2028 
- [go-logr/stdr `SetVerbosity`](https://pkg.go.dev/github.com/go-logr/stdr#SetVerbosity)
- [OTLP exporter env-var spec](https://opentelemetry.io/docs/specs/otel/protocol/exporter/)

texasich:
Resolves #4102

## Checklist

### Go checklist


cam72cam:
@texasich please restore the original PR template.  We can also not review this PR until you read and sign the DCO.

texasich:
@cam72cam PR template restored and DCO sign-off added to the commit (amended with `Signed-off-by: texasich <texasich@gmail.com>`). 

The remaining unchecked boxes are done as well.

internal/tracing/init.go
```diff
@@ -156,12 +161,47 @@ func OpenTelemetryInit(ctx context.Context) (context.Context, error) {
...
+// with prefix-based log levels for compatibility with TF_LOG filtering.
+type otelLogSink struct{}
+
+func (s *otelLogSink) Init(info logr.RuntimeInfo) {}
+
+func (s *otelLogSink) Enabled(level int) bool {
+	return true
+}
+
+func (s *otelLogSink) Info(level int, msg string, keysAndValues ...interface{}) {
```
yottta, line 187: Could you try to make use of the `keysAndValues`? I didn't check the old logger to see what's the difference with this new sink.
But presumably, by not making use of the `keysAndValues` we loose some information from the log statements now.

Or is there another reason to have this implemented the way it is that I miss?

internal/tracing/init_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/go-logr/logr"
+)
+
+func TestOTelLogSink_ImplementsLogSink(t *testing.T) {
+	// Compile-time check: otelLogSink satisfies logr.LogSink
+	var sink logr.LogSink = &otelLogSink{}
```
yottta: Do you mind replacing this test with a specific declaration in `init.go`?
```go
var _ logr.LogSink = (*otelLogSink)(nil)
```

internal/tracing/init_test.go
```diff
@@ -0,0 +1,179 @@
...
+		t.Errorf("Error output = %q; want [ERROR] prefix", output)
+	}
+	if !strings.Contains(output, "cleanup") {
+		t.Errorf("Error output = %q; want message 'cleanup'", output)
+	}
+	// Should NOT contain "<nil>" when error is nil
+	if strings.Contains(output, "<nil>") {
+		t.Errorf("Error output = %q; should not contain '<nil>' when err is nil", output)
+	}
+}
```
yottta: I would put these 2 tests into a single top-level one, something like:
```go
func TestOTelLogSink_Error(t *testing.T) {
    t.Run("nil error", func(t *testing.T) {/*...*/}}
    t.Run("with error", func(t *testing.T) {/*...*/}}
}

**Comment by texasich** [comment]
Addressed all three. `keysAndValues` now formats through `formatKVs` as `key=value` pairs with `(MISSING)` for odd counts, swapped the interface test for a compile-time assertion in `init.go`, and merged the error tests into one with subtrees. Ready for another look whenever.

File: internal/tracing/init.go [hunk 4]
```diff
@@ -156,12 +162,79 @@ func OpenTelemetryInit(ctx context.Context) (context.Context, error) {
...
+	return b.String()
+}
+
+func (s *otelLogSink) WithValues(keysAndValues ...interface{}) logr.LogSink {
+	return s
+}
+
+func (s *otelLogSink) WithName(name string) logr.LogSink {
+	return s
+}
```
yottta, line 270: These 2 methods looks like NOOP.
Is this what we want?
Since this returns the same instance wouldn't be a better idea to create a new one by copying the old one and setting a private attribute with the given value and when the log is generated include these too?

yottta (commented):
Sorry for closing/opening, wrong button.

Please also fix the DCO issue.

internal/tracing/init.go
texasich: Good catch, thanks. WithValues/WithName were returning the same zero-size struct, all scoped logger context from the OTel SDK (processor names, component tags, key-value pairs) was getting silently dropped.

Fixed in the latest push:
- otelLogSink now carries name + values fields
- WithValues returns a new sink with a copy of the appended key-value pairs
- WithName returns a new sink with dot-separated hierarchical naming
- Info/Error thread accumulated name and values into output

Also fixed the DCO.

Ready for another look.

internal/tracing/init.go
```diff
@@ -98,13 +99,18 @@ func OpenTelemetryInit(ctx context.Context) (context.Context, error) {
...
 		return ctx, nil // By default, we just discard all telemetry calls
 	}
 
 	isTracingEnabled = true
 
-	log.Printf("[TRACE] OpenTelemetry: enabled")
+	// Wire OTel SDK internal logging through OpenTofu's global logger
+	// before initializing the tracer provider, so any SDK startup
+	// diagnostics are captured.
+	otel.SetLogger(logr.New(&otelLogSink{}))
```
apparentlymart, line 110: Perhaps we could skip this if OpenTofu's main logger is completely disabled, making the OpenTelemetry logging calls as close as possible to a no-op instead of sending everything through the prefix-based filter before it gets discarded. 🤔

We don't currently have a helper for testing whether logging is completely disabled or not -- the closest we have is `logging.IsDebugOrHigher` -- but I think we could add a similar function to `package logging` and then only call `otel.SetLogger` if it returns true:

```go
package logging

// IsEnabled returns true if any log level is enabled at all, regardless of level.
//
// Returns false if logging is completely disabled, which is the default.
func IsEnabled() bool {
	level, _ := globalLogLevel()
	return level != hclog.Off
}
```

However, from the implementation in the `otel` package it seems like the default if we don't call `SetLogger` is for OpenTelemetry's "error" logs to be written to stderr, and so we'd need an `else` clause that uses a completely-noop log sink instead. It seems like [`logr.Discard`](https://pkg.go.dev/github.com/go-logr/logr#Discard) returns a suitable logger for that case, right? (I'm not particularly familiar with logr, so I'm relying only on its documentation here.)

What do you think?

(Although it's not directly about OpenTofu, in suggesting this I'm thinking about https://github.com/hashicorp/terraform-plugin-framework/issues/1269 where the reporter found that a similar situation of generating and then immediately discarding logs caused a notable performance hit, suggesting that treating "completely disabled" as a special case might be helpful here too.)

internal/tracing/init.go
```diff
@@ -156,12 +162,116 @@ func OpenTelemetryInit(ctx context.Context) (context.Context, error) {
...
+	b.WriteString(msg)
+
+	if len(s.values) > 0 {
+		b.WriteString(" ")
+		writeLogValues(&b, s.values)
+	}
+	if len(keysAndValues) > 0 {
+		b.WriteString(" ")
+		writeLogValues(&b, keysAndValues)
+	}
```
yottta: Since there is already a pointer to `strings.Builder` given into `writeLogValues`, wouldn't be cleaner to move these `if` statements into that function?
And then we can easily use it like:
```go
writeLogValues(&b, s.values)
writeLogValues(&b, keysAndValues)
```

where its content would be something like:
```go
func writeLogValues(b *strings.Builder, kv []any) {
    if len(kv) == 0 {
        return
    }
    b.WriteString(" ")
	for i := 0; i < len(kv); i += 2 {
		if i > 0 {
			b.WriteString(" ")
		}
		fmt.Fprint(b, kv[i])
		if i+1 < len(kv) {
			b.WriteString("=")
			fmt.Fprint(b, kv[i+1])
		}
	}
}
```

internal/tracing/init_test.go
```diff
@@ -0,0 +1,284 @@
...
+	if len(rs.values) != 4 {
+		t.Errorf("WithValues: got %d values, want 4", len(rs.values))
+	}
+	if rs.values[0] != "key1" || rs.values[1] != "val1" {
+		t.Errorf("WithValues: values[0:2] = %v, %v; want key1, val1", rs.values[0], rs.values[1])
+	}
+	// Original sink should be unchanged
+	if len(sink.values) != 0 {
+		t.Errorf("WithValues: original sink was mutated, got %d values", len(sink.values))
+	}
```
yottta, line 125: Would be nice to have an assertion that will provide a glanceable understanding of the final output.

And maybe remove the cast to the actual type and the assertions of its attributes.
It's meant to be an opaque implementation of the `logr.LogSink`, so we should be interested in its final results.
The only issue that I have with this is that these tests will have to be updated in the future if the inner implementation will be changed, while we are more interested in its general way of working rather than _how_ it does its job.


_applicable to any unit test in this file that works with `WithValues` and `WithName`_

internal/tracing/init.go
texasich: Moved the guard into writeLogValues. Call sites in Info and Error are clean now.

internal/tracing/init_test.go
texasich: Refactored the WithValues/WithName tests to assert on log output instead of internal state. All use the logr.LogSink interface now.

apparentlymart:
Hi again @texasich!

Unfortunately your latest commit doesn't have the DCO sign-off line... annoyingly, we must repeat that for each new commit.

One way to quickly address this is to rebase your branch against `main` with the `--signoff` option, which automatically adds the suffix to any commit that doesn't already have it:

```
git rebase --signoff main
```

Because rebasing rewrites the history of your branch you would then need to "force-push" the result to update this PR, but after that the DCO check should pass.

internal/tracing/init.go
```diff
@@ -156,12 +162,116 @@ func OpenTelemetryInit(ctx context.Context) (context.Context, error) {
...
+}
+
+func (s *otelLogSink) Init(info logr.RuntimeInfo) {}
+
+func (s *otelLogSink) Enabled(level int) bool {
+	return true
+}
+
+func (s *otelLogSink) Info(level int, msg string, keysAndValues ...interface{}) {
+	prefix := "[INFO]"
```
apparentlymart: The linter seems correct here in noting that nothing accesses this value of `prefix` before it definitely gets reassigned by one of the cases of the switch statement below, so I suggest changing this to be just a declaration without also performing an assignment:

```suggestion
	var prefix string
```

Another way would be to change the initial value to `[TRACE]` and delete the `default` case from the switch statement, but I personally like how the switch statement is currently written were it's easy to see all of the possible values `prefix` could have.

internal/tracing/init.go
```diff
@@ -98,13 +99,18 @@ func OpenTelemetryInit(ctx context.Context) (context.Context, error) {
 	// that exporting should always be enabled and so will expect to find
 	// an OTLP server on localhost if no environment variables are set at all.
 	if os.Getenv(OTELExporterEnvVar) != "otlp" {
-		log.Printf("[TRACE] OpenTelemetry: %s not set, OTel tracing is not enabled", OTELExporterEnvVar)
+		// Silently disabled — log line removed per accepted plan
```
apparentlymart:
```suggestion
```

Let's just remove this completely since future readers of this code won't know what "accepted plan" this is talking about... it's only relevant to readers who were familiar with how this code was previously written.

apparentlymart (commented):
Thanks for your continued work on this, @texasich!

I just wanted to make sure you're aware that there are still some unresolved comments above that we'll need to deal with before merging this. If you have any questions or thoughts about those comments, please reply within those comment threads so that the person who started them can respond. Thanks!

internal/tracing/init.go
yottta: Sorry for resolving the conversation. Just took another look and the checks are still outside `writeLogValues`.
Could you please correct that?

internal/tracing/init_test.go
yottta: And please don't forget about this.

internal/tracing/init_test.go
yottta: What do you think about this?

internal/tracing/init.go
texasich: Good catch, this got lost in a rebase. writeLogValues now has the early return and the space separator inside the function, callers are clean. Included in the latest push.

internal/tracing/init_test.go
texasich: Done -- added var _ logr.LogSink = (*otelLogSink)(nil) in init.go and dropped the separate test.

internal/tracing/init_test.go
texasich: Agreed, merged them into TestOTelLogSink_Error with 'with error' and 'nil error' subtests.

internal/tracing/init.go
texasich: Removed.

internal/tracing/init.go
texasich: The no-op cost here is negligible. If TF_LOG is off, stdlib log.Write goes straight to /dev/null and never produces output. Happy to add the guard in a follow-up if you'd rather have it though.

yottta (approved):
Thanks for the fixes. Looks good now.

Not sure about that colon (`:`) after `OpenTelemetry*:*` in the `Info` and `Error` methods, but that's way to insignificant to be bothered by.
If others feel more strongly about this, feel free to start a new conversation on it.

internal/tracing/init.go
apparentlymart: All of the string-building work that `otelLogSink` does before calling `log.Print` is what I was intending to skip with this, but indeed I have not actually measured it so I have no evidence that it's a problem. I think it's reasonable to ship it like this first and then my comment above describes a change we could make later if we learn that this is a problem worth solving.

apparentlymart:
@yottta please resolve each of the comment threads you started too, since we can't merge with those still open. :grinning:

yottta:
> @yottta please resolve each of the comment threads you started too, since we can't merge with those still open. 😀

Sorry for this. Missed doing that 🤦

apparentlymart:
Thanks for working on this @texasich, and for your patience in the code review process!

This is now merged into the `main` branch, and so should be included in the forthcoming v1.13.0 release. Anyone who wants to try it out in an experimental capacity in the meantime could use [nightly builds](https://nightlies.opentofu.org/nightlies/), with this new behavior being available in builds from `20260710` onwards.

PR #4286, opened 2026-06-17
planning: Delete steps fall back to depending on create/update steps

apparentlymart:
This logic was previously missing an important rule for handling dependencies for a "delete" action: if something that's being deleted depends on something that is only being created or updated (not deleted or replaced) then that delete action is expected to wait until the create or update has completed.

This extra rule is subtle but helps deal with the rare bothersome situations where it's necessary to update one thing before another thing can be deleted, due to constraints enforced in a remote API. For example, it might be necessary to update a container to disable remote-enforced deletion protection before it's possible to delete anything nested inside that container.

This fixes a bug that we noticed earlier while preparing https://github.com/opentofu/opentofu/pull/4284.

---

The TODO comment added here is foreshadowing something we're going to need to deal with when we implement destroy-mode planning later: while it's tempting to think of destroy-mode planning as equivalent to planning with a prior state and an empty configuration, in practice it's treated quite differently because in that case all references from ephemeral objects to managed resource instances are expected to be resolved from the prior state before the dependencies are destroyed. (In normal-mode planning, a reference to an "orphan" resource instance would be forbidden.)

Therefore I'm anticipating that destroy-mode planning will be different enough to have its own logic for constructing the execution graph, but we'll wait until we get to implementing that to make a final decision about how to handle this case.

(The context tests covering that gap are already guarded by `ExperimentalFeatureDestroy`, so we're currently skipping them.)

---

### DIFF
--- a/internal/engine/planning/execgraph_resource.go
+++ b/internal/engine/planning/execgraph_resource.go
@@ -77,9 +77,7 @@ func (b *execGraphBuilder) AddResourceInstanceObjectSubgraphs(
 
 	// First we'll insert separate subgraphs for each resource instance object
 	// that has a planned action, without putting any explicit dependency
-	// edges between them yet. This loop also ensures that we have the
-	// operations needed for any provider instance at least one object
-	// belongs to.
+	// edges between them yet.
 	//
 	// We'll insert the explicit dependency edges between the subgraphs in a
 	// separate loop afterwards, along with any needed prior state operations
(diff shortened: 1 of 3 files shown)


internal/engine/planning/execgraph_resource.go
```diff
@@ -77,9 +77,7 @@ func (b *execGraphBuilder) AddResourceInstanceObjectSubgraphs(
 
 	// First we'll insert separate subgraphs for each resource instance object
 	// that has a planned action, without putting any explicit dependency
-	// edges between them yet. This loop also ensures that we have the
-	// operations needed for any provider instance at least one object
-	// belongs to.
+	// edges between them yet.
```
apparentlymart, line 80: I've deleted the end of this comment because we're no longer modelling provider dependencies in the execution graph at all since https://github.com/opentofu/opentofu/pull/4158. This was just a leftover we forgot to update in that PR.

yottta (approved):

diofeher (approved):

cam72cam (approved):

PR #4287, opened 2026-06-17
add `additional_authenticated_data` to `gcp_kms`

### LINKED ISSUE #4260: Add `AdditionalAuthenticatedData` support to `gcs_kms` key provider
The `gcs_kms` key provider is implemented in terms of the remote [`google.cloud.kms.v1.KeyManagementService.Encrypt`](https://docs.cloud.google.com/kms/docs/reference/rpc/google.cloud.kms.v1#google.cloud.kms.v1.KeyManagementService.Encrypt) and [`google.cloud.kms.v1.KeyManagementService.Decrypt`](https://docs.cloud.google.com/kms/docs/reference/rpc/google.cloud.kms.v1#google.cloud.kms.v1.KeyManagementService.Decrypt) RPC functions.

These functions both accept an optional argument `additional_authenticated_data` that must match across both calls for the original randomly-generated key to be recovered, but that is not currently exposed as part of the configuration schema for the `gcs_kms` key provider.

This issue is complete once the `gcs_kms` key provider supports an additional optional argument `additional_authenticated_data` that accepts a value of type `string` that must be base64 encoded using the standard base64 alphabet, so it would accept the results of the existing OpenTofu function [`base64encode`](https://opentofu.org/docs/language/functions/base64encode/)[^1], and sends the decoded `[]byte` form of it to both the `Encrypt` and `Decrypt` functions. Incorrect base64 encoding is a configuration validation error.

If the `additional_authenticated_data` argument is not set at all then the call to KMS must be equivalent to what would've been sent by previous versions of OpenTofu that did not support this argument, so that previously-created data keys are not invalidated just by upgrading OpenTofu.

No further abstraction is needed. The goal is just to directly expose the remote API's concept of additional authenticated data.

[^1]: This base64 encoding is the conventional way to represent arbitrary binary data in the OpenTofu language.

diofeher:
Resolves #4260

This PR implements the [`additional_authenticated_data`](https://docs.cloud.google.com/kms/docs/additional-authenticated-data) field for the Encrypt and Decrypt operations of Google KMS.

Errors raised when using invalid AADs:

```
│ Error: Unable to build encryption key data
│
│ key_provider.gcp_kms.with_aad failed with error: additional_authenticated_data must be valid standard base64: illegal base64 data at input
│ byte 16
```
```
╷
│ Error: Error acquiring the state lock
│
│ Error message: failed to write backup file: <nil>: Unable to fetch encryption key data; key_provider.gcp_kms.with_aad failed with error:
│ rpc error: code = InvalidArgument desc = Decryption failed.
│
│ OpenTofu acquires a state lock to protect the state from being written
│ by multiple users at the same time. Please resolve the issue above and try
│ again. For most commands, you can disable locking with the "-lock=false"
│ flag, but this is not recommended.
╵

```

## Checklist

### Go checklist

 

### DIFF
--- a/internal/encryption/keyprovider/gcp_kms/compliance_test.go
+++ b/internal/encryption/keyprovider/gcp_kms/compliance_test.go
@@ -6,6 +6,7 @@
 package gcp_kms
 
 import (
+	"encoding/base64"
 	"fmt"
 	"os"
 	"testing"
@@ -129,6 +130,24 @@ func TestKeyProvider(t *testing.T) {
 							}`,
 					ValidHCL:   true,
 					ValidBuild: true,
+				},
+				"with-aad": {
+					HCL: fmt.Sprintf(`key_provider "gcp_kms" "foo" {
+							kms_encryption_key = "%s"
+							key_length = 32
+							additional_authenticated_data = "bXktYWFkLXZhbA=="
+							}`, testKeyId),
+					ValidHCL:   true,
+					ValidBuild: true,
+				},
+				"invalid-aad-base64": {
+					HCL: fmt.Sprintf(`key_provider "gcp_kms" "foo" {
+							kms_encryption_key = "%s"
+							key_length = 32
+							additional_authenticated_data = "not-valid-base64!@#"
+							}`, testKeyId),
+					ValidHCL:   true,
+					ValidBuild: false,
 				},
 			},
 			JSONParseTestCases: map[string]compliancetest.JSONParseTestCase[*Config, *keyProvider]{
@@ -266,6 +285,36 @@ func TestKeyProvider(t *testing.T) {
 					ValidJSON:  true,
 					ValidBuild: true,
 				},
+				"with-aad": {
+					JSON: fmt.Sprintf(`{
+	"key_provider": {
+		"gcp_kms": {
+			"foo": {
+				"kms_encryption_key": "%s",
+				"key_length": 32,
+				"additional_authenticated_data": "bXktYWFkLXZhbA=="
+			}
+		}
+	}
+}`, testKeyId),
+					ValidJSON:  true,
+					ValidBuild: true,
+				},
+				"invalid-aad-base64": {
+					JSON: fmt.Sprintf(`{
+	"key_provider": {
+		"gcp_kms": {
+			"foo": {
+				"kms_encryption_key": "%s",
+				"key_length": 32,
+				"additional_authenticated_data": "not-valid-base64!@#"
+			}
+		}
+	}
+}`, testKeyId),
+					ValidJSON:  true,
+					ValidBuild: false,
+				},
 			},
 			ConfigStructTestCases: map[string]compliancetest.ConfigStructTestCase[*Config, *keyProvider]{
 				"success": {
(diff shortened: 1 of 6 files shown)


internal/encryption/keyprovider/gcp_kms/compliance_test.go
```diff
@@ -284,6 +333,38 @@ func TestKeyProvider(t *testing.T) {
...
+						KeyLength:                   32,
+						AdditionalAuthenticatedData: base64.StdEncoding.EncodeToString(make([]byte, 65*1024)),
+					},
+					ValidBuild: false,
+				},
+				"unpadded-aad-base64": {
+					Config: &Config{
+						KMSKeyName:                  testKeyId,
+						KeyLength:                   32,
+						AdditionalAuthenticatedData: "bXktYWFkLXZhbHVl",
```
apparentlymart: In case you want to write this as an `EncodeToString` like you did for the others, you could use [`base64.RawStdEncoding`](https://pkg.go.dev/encoding/base64#RawStdEncoding) which is the same as `base64.StdEncoding` except that it doesn't include padding.

(I don't feel strongly about changing it, though. It being written literally perhaps makes it easier to see that it's definitely unpadded. 🤷‍♂️)

internal/encryption/keyprovider/gcp_kms/compliance_test.go
apparentlymart, line 380: From what I can understand in the rest of the test code it seems like this is the only case that actually tries to call the GCP API through the `mockKMC` injected at the top. Is that right?

I'd like for there to be a test here which verifies that the `kmspb.EncryptRequest` and `kmspb.DecryptRequest` passed to the mock both have `AdditionalAuthenticatedData` populated correctly, to make sure we don't accidentally break one of them under future maintenance. I'm not sure how to achieve that with the current way the compliance test API is shaped, though. 🤔

internal/encryption/keyprovider/gcp_kms/compliance_test.go
diofeher: Oh, that's interesting! I think being consistent with the other tests is important as well, so let me try to improve this one

diofeher: Fixed that

internal/encryption/keyprovider/gcp_kms/compliance_test.go
diofeher: Added one. I removed the field to test if it is really working as expected:

```
--- FAIL: TestAADForwardedToRPC (0.00s)
    compliance_test.go:411: encrypt: AAD = "", want "my-aad-val"
    compliance_test.go:411: encrypt: AAD = "", want "my-aad-val"
    compliance_test.go:419: decrypt: AAD = "", want "my-aad-val"
FAIL
FAIL	github.com/opentofu/opentofu/internal/encryption/keyprovider/gcp_kms	0.776s
FAIL
```

I tried to add another mock to the TestKeyProvider, but I didn't like how it looked. So I just added a new test https://github.com/opentofu/opentofu/pull/4287/commits/8c8152eb8aca496813030b3592c8fb98ac512ff6

apparentlymart (approved):

internal/encryption/keyprovider/gcp_kms/config.go
```diff
@@ -148,10 +151,23 @@ func (c Config) Build() (keyprovider.KeyProvider, keyprovider.KeyMeta, error) {
...
 
+	var aad []byte
+	if c.AdditionalAuthenticatedData != "" {
+		var err error
+		aad, err = base64.StdEncoding.DecodeString(c.AdditionalAuthenticatedData)
+		if err != nil {
+			return nil, nil, &keyprovider.ErrInvalidConfiguration{Message: "additional_authenticated_data must be valid standard base64", Cause: err}
+		}
+		if len(aad) > 64*1024 {
+			return nil, nil, &keyprovider.ErrInvalidConfiguration{Message: "additional_authenticated_data must be less than the GCP limit of 64 KiB"}
```
yottta, line 164: Can you add this URL in a comment above this line?
https://docs.cloud.google.com/kms/docs/additional-authenticated-data

I know that it's in the encryption.mdx too, but would be easier for devs later to see the documentation inline.

yottta (approved):

internal/encryption/keyprovider/gcp_kms/config.go
diofeher: Sure!

PR #4289, opened 2026-06-18
Fix duplicated word in S3 backend key validation error

s3onghyun:
The S3 backend key validation error read "must not start or end with with \"/\"" — removed the duplicated "with". User-facing diagnostic.

### DIFF
--- a/internal/backend/remote-state/s3/backend.go
+++ b/internal/backend/remote-state/s3/backend.go
@@ -509,7 +509,7 @@ func (b *Backend) PrepareConfig(obj cty.Value) (cty.Value, tfdiags.Diagnostics) 
 		diags = diags.Append(tfdiags.AttributeValue(
 			tfdiags.Error,
 			"Invalid key value",
-			`The "key" attribute value must not start or end with with "/".`,
+			`The "key" attribute value must not start or end with "/".`,
 			cty.Path{cty.GetAttrStep{Name: "key"}},
 		))
 	}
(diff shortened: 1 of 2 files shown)


yottta (approved):

yottta:
Thanks for the fix.

Please reinstate the checkboxes from the [PR template](https://github.com/opentofu/opentofu/blob/main/.github/pull_request_template.md).

cam72cam (approved):

yottta:
@s3onghyun can you also fix the tests?

s3onghyun:
Thanks @yottta and @cam72cam for the review. After a closer look at the PR template, I should be upfront: I cannot honestly check the "I have not used an AI coding assistant to create this PR" box, so per the project's contribution policy the right thing is to close this rather than leave it in limbo or check that box inaccurately.

The underlying issue is still real and trivial if a maintainer wants to pick it up directly: in `internal/backend/remote-state/s3/backend.go` the key-validation diagnostic reads `must not start or end with with "/"` (duplicated "with"), and `backend_test.go` asserts the same duplicated string in the leading/trailing-slash cases. Apologies for the noise, and thanks for your time.

yottta:
@s3onghyun we greatly appreciate your openness on this! 🥇
Yes, that's the right thing to do indeed. 🙌 🤝 

Now because we reviewed this, myself and @cam72cam were exposed to possibly tainted code. 
Therefore, I suggest for both of us not to be involved anymore in this issue and leave this for the other @opentofu/opentofu-maintainers.

PR #4298, opened 2026-06-19
feat: added the aws kms encryption context

### LINKED ISSUE #4259: Add `EncryptionContext` support to `aws_kms` key provider
The `aws_kms` key provider is implemented in terms of [kms:GenerateDataKey](https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKey.html) and [kms:Decrypt](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html).

These operations both allow an optional argument `EncryptionContext` for providing [encryption context](https://docs.aws.amazon.com/kms/latest/developerguide/encrypt_context.html), but that is not currently exposed as part of the configuration schema for the `aws_kms` key provider.

This issue is complete once the `aws_kms` key provider supports an additional argument `encryption_context` that accepts a value of type `map(string)` and sends that map verbatim in both the `kms:GenerateDataKey` and `kms:Decrypt` calls.

If the `encryption_context` argument is not set at all then the call to KMS must be equivalent to what would've been sent by previous versions of OpenTofu that did not support this argument, so that previously-created data keys are not invalidated just by upgrading OpenTofu.

No further abstraction is needed. The goal is just to directly expose the remote API's concept of encryption context.

KrishnaSindhur:
Resolves #4259 

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ UPGRADE NOTES:
 ENHANCEMENTS:
 
 - The `gcp_kms` key provider now supports an optional `additional_authenticated_data` as part of the encryption and decryption operations. ([#4287](https://github.com/opentofu/opentofu/pull/4287))
+- The AWS KMS key provider for state encryption now supports an `encryption_context` field, allowing key-value string pairs to be passed to AWS KMS with every `GenerateDataKey` and `Decrypt` call. ([#4298](https://github.com/opentofu/opentofu/pull/4298))
 - The `cidrsubnets` function now supports prefix extensions greater than 32 bits when the base CIDR block uses an IPv6 address. ([#4042](https://github.com/opentofu/opentofu/pull/4042))
 - The `local-exec` provisioner now automatically sets the `TRACEPARENT` environment variable in child processes when OpenTelemetry tracing is active, following the W3C Trace Context specification. ([#4014](https://github.com/opentofu/opentofu/issues/4014))
 - When installing provider and module packages from OCI Distribution registries, OpenTofu now tracks separate transient credentials for each repository to support registry implementations that issue repository-scoped tokens.  ([#3316](https://github.com/opentofu/opentofu/issues/3316))
--- a/internal/encryption/keyprovider/aws_kms/mock_test.go
+++ b/internal/encryption/keyprovider/aws_kms/mock_test.go
@@ -32,24 +32,30 @@ func injectMock(m *mockKMS) {
 }
 
 func injectDefaultMock() {
+	injectCapturingMock("alias/my-mock-key")
+}
+
+func injectCapturingMock(keyId string) (capturedGenKeyContext *map[string]string, capturedDecryptContext *map[string]string) {
+	var genCtx, decCtx map[string]string
 	injectMock(&mockKMS{
 		genkey: func(params *kms.GenerateDataKeyInput) (*kms.GenerateDataKeyOutput, error) {
+			genCtx = params.EncryptionContext
 			keyData := make([]byte, 32)
 			_, err := rand.Read(keyData)
 			if err != nil {
 				panic(err)
 			}
-
 			return &kms.GenerateDataKeyOutput{
-				CiphertextBlob: append([]byte(*params.KeyId), keyData...),
+				CiphertextBlob: append([]byte(keyId), keyData...),
 				Plaintext:      keyData,
 			}, nil
-
 		},
 		decrypt: func(params *kms.DecryptInput) (*kms.DecryptOutput, error) {
+			decCtx = params.EncryptionContext
 			return &kms.DecryptOutput{
-				Plaintext: params.CiphertextBlob[:len(*params.KeyId)],
+				Plaintext: params.CiphertextBlob[len(keyId):],
 			}, nil
 		},
 	})
+	return &genCtx, &decCtx
 }
--- a/internal/encryption/keyprovider/aws_kms/mock_test.go
+++ b/internal/encryption/keyprovider/aws_kms/mock_test.go
@@ -37,11 +37,16 @@ func injectMock(t testing.TB, m *mockKMS) {
 	}
 }
 
+type capturedKMSCalls struct {
+	GenKeyContext  *map[string]string
+	DecryptContext *map[string]string
+}
+
 func injectDefaultMock(t testing.TB) {
 	injectCapturingMock(t, "alias/my-mock-key")
 }
 
-func injectCapturingMock(t testing.TB, keyId string) (capturedGenKeyContext *map[string]string, capturedDecryptContext *map[string]string) {
+func injectCapturingMock(t testing.TB, keyId string) capturedKMSCalls {
 	var genCtx, decCtx map[string]string
 	injectMock(t, &mockKMS{
 		genkey: func(params *kms.GenerateDataKeyInput) (*kms.GenerateDataKeyOutput, error) {
@@ -63,5 +68,8 @@ func injectCapturingMock(t testing.TB, keyId string) (capturedGenKeyContext *map
 			}, nil
 		},
 	})
-	return &genCtx, &decCtx
+	return capturedKMSCalls{
+		GenKeyContext:  &genCtx,
+		DecryptContext: &decCtx,
+	}
 }
--- a/internal/encryption/keyprovider/aws_kms/mock_test.go
+++ b/internal/encryption/keyprovider/aws_kms/mock_test.go
@@ -8,6 +8,7 @@ package aws_kms
 import (
 	"context"
 	"crypto/rand"
+	"testing"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/kms"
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ UPGRADE NOTES:
 
 ENHANCEMENTS:
 
+- The AWS KMS key provider for state encryption now supports an `encryption_context` field, allowing key-value string pairs to be passed to AWS KMS with every `GenerateDataKey` and `Decrypt` call. ([#4287](https://github.com/opentofu/opentofu/pull/4287))
```
```suggestion
- The AWS KMS key provider for state encryption now supports an `encryption_context` field, allowing key-value string pairs to be passed to AWS KMS with every `GenerateDataKey` and `Decrypt` call. ([#4298](https://github.com/opentofu/opentofu/pull/4298))
```

internal/encryption/keyprovider/aws_kms/mock_test.go
```diff
@@ -25,31 +26,42 @@ func (m *mockKMS) Decrypt(ctx context.Context, params *kms.DecryptInput, optFns
...
+				CiphertextBlob: append([]byte(keyId), keyData...),
 				Plaintext:      keyData,
 			}, nil
-
 		},
 		decrypt: func(params *kms.DecryptInput) (*kms.DecryptOutput, error) {
+			decCtx = params.EncryptionContext
 			return &kms.DecryptOutput{
-				Plaintext: params.CiphertextBlob[:len(*params.KeyId)],
+				Plaintext: params.CiphertextBlob[len(keyId):],
```
lbordowitz, line 67: Why this change?

internal/encryption/keyprovider/aws_kms/mock_test.go
```diff
@@ -25,31 +26,42 @@ func (m *mockKMS) Decrypt(ctx context.Context, params *kms.DecryptInput, optFns
...
 }
 
-func injectDefaultMock() {
-	injectMock(&mockKMS{
+func injectDefaultMock(t testing.TB) {
+	injectCapturingMock(t, "alias/my-mock-key")
+}
+
+func injectCapturingMock(t testing.TB, keyId string) (capturedGenKeyContext *map[string]string, capturedDecryptContext *map[string]string) {
+	var genCtx, decCtx map[string]string
```
lbordowitz, line 50: I like the concept behind the capture maps being returned. Returning them as a tuple is slightly sketchy; supposing we might want to add more capture pointers like these, it would work better if it were a struct instead. Adding more captures is unlikely, though, so treat this as optional and a nit.

lbordowitz (commented):
Left a couple comments. Overall looks good; I'll give the tests a spin tomorrow.

internal/encryption/keyprovider/aws_kms/mock_test.go
KrishnaSindhur: The mock stores the key like this!
[key ID][actual secret key]

Old code took the first part and returned the key ID.
New code takes the second part and returns the real secret key so we have to use the second part if i am not wrong!!

internal/encryption/keyprovider/aws_kms/mock_test.go
KrishnaSindhur: ya using struct will be neat!! fair point!!

internal/encryption/keyprovider/aws_kms/provider.go
```diff
@@ -70,8 +71,9 @@ func (p keyProvider) Provide(rawMeta keyprovider.KeyMeta) (keyprovider.Output, k
 	if inMeta.isPresent() {
 		// We have an existing decryption key to decrypt, so we should now populate the DecryptionKey
 		decryptedKeyData, decryptErr := p.svc.Decrypt(p.ctx, &kms.DecryptInput{
-			KeyId:          aws.String(p.KMSKeyID),
-			CiphertextBlob: inMeta.CiphertextBlob,
+			KeyId:             aws.String(p.KMSKeyID),
+			CiphertextBlob:    inMeta.CiphertextBlob,
+			EncryptionContext: p.EncryptionContext,
```
lbordowitz: Could you please put the encryption context into the meta, and use the EncryptionContext from the meta instead?

While I was testing this, I defined a key provider block, and then subsequently added an encryption context to the same key. I was left with this unhelpful error:

> Unable to fetch encryption key data; key_provider.aws_kms.basic failed with error: : operation error KMS: Decrypt, https response error StatusCode: 400, RequestID: ***OMITTED***,
> InvalidCiphertextException: 

The problem is that the encryption context from the config is immediately used for decryption. However, that wasn't the context used from last time! By saving the config in the meta, you'll be able to track which context is used when, and rotating context will be much smoother.

A side-note: you're gonna want to include `encryption_context` as part of the schema in the `tofu-ls` as well; I was getting squiggly red lines on VSCode while testing this (obviously, because it's not a feature yet!)

lbordowitz (commented):
One last thing before this gets approved. I tested everything else; the request looks good, everything is in working order when I did the fallback method, and the code is clean and maintainable. Great job!

internal/encryption/keyprovider/aws_kms/provider.go
KrishnaSindhur: I tested only for encrypt part so I missed that!! 
I ran the following to test this
applied small change using below key value 
`encryption_context = {
        Environment = "prod"
        Application = "opentofu"
      }`
after that changed to `dev` Environment from `prod` and ran again it decrypts and re-encrypted again with new context 
this test is good right?

KrishnaSindhur: Also on last part adding changes to schema @diofeher already raised the draft I might have to update same branch
https://github.com/opentofu/opentofu-schema/pull/62/changes

diofeher: Hey @KrishnaSindhur I just added the necessary changes (to also cover AWS) on my draft PR: https://github.com/opentofu/opentofu-schema/pull/62 so when this one is merged and next version is released, I'll move this one out of draft.

lbordowitz (approved):

apparentlymart (approved):
I reviewed the code in this PR but I did not proactively test it because the discussion above suggests that others already tested it against the real KMS API.

This looks good and matches what I expected based on how the existing code was written.

diofeher (approved):

yottta (changes requested):
Ran all of the tests in the `internal/encryption/keyprovider/aws_kms` against main and then against this PR:

### main
Image

### this PR
Image

```shell
=== RUN   TestKeyProvider/metadata/present-valid
    compliance.go:455: 
Build() did not return an error.
    compliance.go:461: 
Build() returned the correct key provider type of *aws_kms.keyProvider.
    compliance.go:475: 
Build() returned metadata and the declared metadata type is not an interface.
    compliance.go:481: 
Build() returned the correct metadata type of *aws_kms.keyMeta.
    compliance.go:528: 
The Provide() function returned an unexpected error: : operation error KMS: Decrypt, https response error StatusCode: 400, RequestID: 47b0e117-a1ad-44d6-8784-10a7f2acfc5d, InvalidCiphertextException: 
--- FAIL: TestKeyProvider/metadata/present-valid (0.15s)
```

Feel free to check this by using the following command:
```shell
AWS_REGION=<aws_region> TF_ACC=1 TF_AWS_KMS_KEY_ID=arn:aws:kms:<aws_region>:<account_id>:key/<key_id> TF_KMS_TEST=1 AWS_ACCESS_KEY_ID=<aws_access_key_id> AWS_SECRET_ACCESS_KEY=<aws_secret_access_key> go test github.com/opentofu/opentofu/internal/encryption/keyprovider/aws_kms -count=1 -ldflags=-linkmode=internal -v
```

KrishnaSindhur:
> Ran all of the tests in the `internal/encryption/keyprovider/aws_kms` against main and then against this PR:
> 
> ### main
> Image
> ### this PR
> Image
> ```shell
> === RUN   TestKeyProvider/metadata/present-valid
>     compliance.go:455: 
> Build() did not return an error.
>     compliance.go:461: 
> Build() returned the correct key provider type of *aws_kms.keyProvider.
>     compliance.go:475: 
> Build() returned metadata and the declared metadata type is not an interface.
>     compliance.go:481: 
> Build() returned the correct metadata type of *aws_kms.keyMeta.
>     compliance.go:528: 
> The Provide() function returned an unexpected error: : operation error KMS: Decrypt, https response error StatusCode: 400, RequestID: 47b0e117-a1ad-44d6-8784-10a7f2acfc5d, InvalidCiphertextException: 
> --- FAIL: TestKeyProvider/metadata/present-valid (0.15s)
> ```
> 
> Feel free to check this by using the following command:
> 
> ```shell
> AWS_REGION=<aws_region> TF_ACC=1 TF_AWS_KMS_KEY_ID=arn:aws:kms:<aws_region>:<account_id>:key/<key_id> TF_KMS_TEST=1 AWS_ACCESS_KEY_ID=<aws_access_key_id> AWS_SECRET_ACCESS_KEY=<aws_secret_access_key> go test github.com/opentofu/opentofu/internal/encryption/keyprovider/aws_kms -count=1 -ldflags=-linkmode=internal -v
> ```

ok this fails on KMS manual acceptance testing with real AWS, make test will not cover this.
Let me update!!

yottta (commented):
Oh, so it was using mocked meta information.
Great catch! Thanks for the fix.


EDIT: ran the tests again against the updated branch and everything works as expected.

yottta (approved):

PR #4304, opened 2026-06-22
Alternate approach to provider abstractions in the new runtime 

cam72cam:
Alternate approach to provider abstractions in the new runtime

When we originally integrated providers into the new runtime, it was done with the assumption that providers and ephemeral resources would be driven by the execution graph. In practice, we found this too restrictive and counter intuitive to our goals with ephemeral resources.

After we removed providers and ephemeral resources from the graph, it no longer made sense to keep most of the provider machinery at that level. Instead, we can consolidate most of that logic to the unexported
internals of the eval package.

This dramatically simplifies the code around managed providers and reduces the size of interfaces and integrations between engine and eval.

## Checklist

### Go checklist

 


apparentlymart (approved):
Overall, I think this is a good direction!

I think we may need to do some additional fiddling with forcing propagation of dependencies through provider instances to make sure we end up with a correct resource instance graph now that we've removed provider instances from that graph, but that was already not completely implemented before this change anyway so no harm in merging this and then we can iterate on the dependency details later.

---

To be specific, I think (**in a separate PR, later**) we should gather up all of the "dependency-related" marks from the configuration of each provider instance and then arrange for those marks to be applied to every resource instance belonging to a provider instance and to the result of every function called through that provider instance, so any indirect resource-instance-to-resource-instance dependencies will still get tracked.

(You previously asked me about this and I said I wasn't sure yet, but during some more recent work I've concluded that it's important because we ought to make sure these indirect dependencies get preserved in the state for use during destroy when we won't necessarily have the configuration available anymore.)

One interesting detail is in making sure the functions that require a configured provider still get their dependencies marked even when we're using stubs to skip calling the real functions, because in the new runtime's equivalent of early eval the current (partially-implemented) rule is that [values in early-eval locations are not allowed to be derived from resource instances](https://github.com/opentofu/opentofu/blob/79385bf5de231c0f5b985f5e778ce5137b68f4f6/internal/lang/eval/internal/configgraph/module_call.go#L137-L144) _even if the result is known_. Though perhaps we should also have a more direct mark for anything that comes from a configured provider instance regardless of whether its configuration depends on a resource instance, since calling configured providers in early-eval locations is problematic regardless of what they depend on. "Comes from a configured provider" could then be another kind of "dependency-related mark" that we preserve across module `depends_on` boundaries, etc.

PR #4305, opened 2026-06-23
Validate large count values before expansion

### LINKED ISSUE #4278: OpenTofu crashes when a resource uses a large count value
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.12.2
on linux_amd64
```

```hcl
resource "space_x_trades" "trades" {
  count = 9223372036854775807
  input = "x"
}
```

### Debug Output
```
tofu apply

OpenTofu crashed! This is always indicative of a bug within OpenTofu.
Please report the crash with OpenTofu[1] so that we can fix this.

When reporting bugs, please include your OpenTofu version, the stack trace
shown below, and any additional information which may help replicate the issue.

[1]: https://github.com/opentofu/opentofu/issues

runtime error: makeslice: len out of range
goroutine 87 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x5e
runtime/debug.PrintStack()
	runtime/debug/stack.go:18 +0x13
github.com/opentofu/opentofu/internal/logging.panicHandler({0x3d27ca0, 0x475bdb0}, {0x2616c1122800, 0x7ba, 0x800})
	github.com/opentofu/opentofu/internal/logging/panic.go:96 +0xd6
github.com/opentofu/opentofu/internal/tofu.(*Graph).walk.PanicHandlerWithTraceFn.func2()
	github.com/opentofu/opentofu/internal/logging/panic.go:82 +0x85
panic({0x3d27ca0?, 0x475bdb0?})
	runtime/panic.go:860 +0x13a
github.com/opentofu/opentofu/internal/instances.expansionCount.instanceKeys(...)
	github.com/opentofu/opentofu/internal/instances/expansion_mode.go:67
github.com/opentofu/opentofu/internal/instances.(*expanderModule).onlyResourceInstances(0x2616c1a5ef90?, {{}, 0x4d, {0x2616c18f0430, 0xe}, {0x481b508, 0x1}}, {0x2616c1730c80, 0x0, 0x4})
	github.com/opentofu/opentofu/internal/instances/expander.go:488 +0x89
github.com/opentofu/opentofu/internal/instances.(*expanderModule).resourceInstances(0x2616c10395e0?, {0x0?, 0x0?, 0xf0?}, {{}, 0x4d, {0x2616c18f0430, 0xe}, {0x481b508, 0x1}}, ...)
	github.com/opentofu/opentofu/internal/instances/expander.go:478 +0x2b1
github.com/opentofu/opentofu/internal/instances.(*Expander).ExpandResource(0x2616c1039640, {{}, {0x0, 0x0, 0x0}, {{}, 0x4d, {0x2616c18f0430, 0xe}, {0x481b508, ...}}})
	github.com/opentofu/opentofu/internal/instances/expander.go:228 +0x155
github.com/opentofu/opentofu/internal/tofu.(*nodeExpandPlannableResource).expandResourceInstances(0x2616c17a9c20, {0x47c18f8, 0x71daf80}, {0x47f65a0?, 0x2616c15030e0?}, {{}, {0x0, 0x0, 0x0}, {{}, ...}}, ...)
	github.com/opentofu/opentofu/internal/tofu/node_resource_plan.go:251 +0x23b
github.com/opentofu/opentofu/internal/tofu.(*nodeExpandPlannableResource).DynamicExpand(0x2616c17a9c20, {0x47f65a0, 0x2616c15030e0})
	github.com/opentofu/opentofu/internal/tofu/node_resource_plan.go:196 +0x1026
github.com/opentofu/opentofu/internal/tofu.(*Graph).walk.func1({0x445d0e0, 0x2616c17a9c20})
	github.com/opentofu/opentofu/internal/tofu/graph.go:95 +0x562
github.com/opentofu/opentofu/internal/dag.(*Walker).walkVertex(0x2616c10a4720, {0x445d0e0, 0x2616c17a9c20}, 0x2616c1033d00)
	github.com/opentofu/opentofu/internal/dag/walk.go:386 +0x2a5
created by github.com/opentofu/opentofu/internal/dag.(*Walker).Update in goroutine 23
	github.com/opentofu/opentofu/internal/dag/walk.go:309 +0x1048
With go-routine called from:
goroutine 23 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x5e
github.com/opentofu/opentofu/internal/logging.PanicHandlerWithTraceFn(...)
	github.com/opentofu/opentofu/internal/logging/panic.go:72
github.com/opentofu/opentofu/internal/tofu.(*Graph).walk(0x2616c1033a40, {0x47c1818, 0x2616c18f57d0}, {0x47c2cb0, 0x2616c1267300})
	github.com/opentofu/opentofu/internal/tofu/graph.go:50 +0x45
github.com/opentofu/opentofu/internal/tofu.(*Graph).Walk(...)
	github.com/opentofu/opentofu/internal/tofu/graph.go:39
github.com/opentofu/opentofu/internal/tofu.(*Context).walk(0x2616c13dd830, {0x47c1818, 0x2616c18f57d0}, 0x2616c1033a40, 0x2, 0x2616c1a63640)
	github.com/opentofu/opentofu/internal/tofu/context_walk.go:61 +0x131
github.com/opentofu/opentofu/internal/tofu.(*Context).planWalk(0x2616c13dd830, {0x47c1818, 0x2616c18f57d0}, 0x2616c12327e0, 0x2616c1a73d30, 0x2616c192b760)
	github.com/opentofu/opentofu/internal/tofu/context_plan.go:802 +0x5bb
github.com/opentofu/opentofu/internal/tofu.(*Context).plan(0x2616c13dd830, {0x47c1818, 0x2616c18f57d0}, 0x2616c12327e0, 0x2616c1a73d30, 0x2616c192b760)
	github.com/opentofu/opentofu/internal/tofu/context_plan.go:363 +0x4be
github.com/opentofu/opentofu/internal/tofu.(*Context).Plan(0x2616c13dd830, {0x47c1888, 0x2616c14bb110}, 0x2616c12327e0, 0x0, 0x2616c192b760)
	github.com/opentofu/opentofu/internal/tofu/context_plan.go:235 +0xd69
github.com/opentofu/opentofu/internal/backend/local.(*Local).opApply(0x2616c192b130, {0x47c1850, 0x2616c15eec30}, {0x47c1850, 0x2616c15eec80}, 0x2616c1200c40, 0x2616c1775640)
	github.com/opentofu/opentofu/internal/backend/local/backend_apply.go:127 +0x865
github.com/opentofu/opentofu/internal/backend/local.(*Local).Operation.func1()
	github.com/opentofu/opentofu/internal/backend/local/backend.go:344 +0xb5
created by github.com/opentofu/opentofu/internal/backend/local.(*Local).Operation in goroutine 1
	github.com/opentofu/opentofu/internal/backend/local/backend.go:337 +0x545
```

### Expected Behavior

Validation error or opentofu continue to work

### Actual Behavior

Opentofu crash
```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x332356eae320 stack=[0x332356eae000, 0x332376eae000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x451c042?, 0x442301?})
	runtime/panic.go:1229 +0x48 fp=0x33233300be98 sp=0x33233300be68 pc=0x48b6a8
runtime.newstack()
	runtime/stack.go:1178 +0x5fd fp=0x33233300bfc8 sp=0x33233300be98 pc=0x47025d
runtime.morestack()
	runtime/asm_amd64.s:681 +0x7d fp=0x33233300bfd0 sp=0x33233300bfc8 pc=0x491c1d

goroutine 1 gp=0x332332f261e0 m=3 mp=0x332332ff5008 [running]:
github.com/hashicorp/hcl/v2/hclsyntax.(*parser).parseBinaryOps(0x332376eacfa0, {0x3323333abce8?, 0x5?, 0x5?})
	github.com/hashicorp/hcl/v2@v2.20.1/hclsyntax/parser.go:546 +0xbb3 fp=0x332356eae330 sp=0x332356eae328 pc=0xb4e6d3
github.com/hashicorp/hcl/v2/hclsyntax.(*parser).parseBinaryOps(0x332376eacfa0, {0x3323333abce0?, 0x1?, 0x1bf9f?})
	github.com/hashicorp/hcl/v2@v2.20.1/hclsyntax/parser.go:563 +0x74 fp=0x332356eae548 sp=0x332356eae330 pc=0xb4db94
github.com/hashicorp/hcl/v2/hclsyntax.(*parser).parseTernaryConditional(0x332376eacfa0)
	github.com/hashicorp/hcl/v2@v2.20.1/hclsyntax/parser.go:495 +0x131 fp=0x332356eae890 sp=0x332356eae548 pc=0xb4cf11
```

### Steps to Reproduce

tofu apply

### Additional Context

Most probably, long values for count will not be required in real use cases, I found this bug when I was auto-generating tofu code and had a bug in the count value

### References

_No response_

goutamadwant:
Resolves #4278

## Checklist

### Go checklist

 

### DIFF
--- a/internal/lang/evalchecks/eval_count.go
+++ b/internal/lang/evalchecks/eval_count.go
@@ -21,10 +21,6 @@ type EvaluateFunc func(expr hcl.Expression) (cty.Value, tfdiags.Diagnostics)
 
 const maxCount = int64(math.MaxInt32)
 
-func maxCountValue() int64 {
-	return maxCount
-}
-
 // EvaluateCountExpression is our standard mechanism for interpreting an
 // expression given for a "count" argument on a resource or a module. This
 // should be called during expansion in order to determine the final count
@@ -117,11 +113,11 @@ func EvaluateCountExpressionValue(expr hcl.Expression, ctx EvaluateFunc) (cty.Va
 		return cty.UnknownVal(cty.Number), diags
 	}
 
-	if countVal.Type() == cty.Number && countVal.GreaterThan(cty.NumberIntVal(maxCountValue())).True() {
+	if countVal.Type() == cty.Number && countVal.GreaterThan(cty.NumberIntVal(maxCount)).True() {
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "Invalid count argument",
-			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: must be less than or equal to %d.`, maxCountValue()),
+			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: must be less than or equal to %d.`, maxCount),
 			Subject:  expr.Range().Ptr(),
 		})
 		return nullCount, diags
--- a/internal/lang/evalchecks/eval_count.go
+++ b/internal/lang/evalchecks/eval_count.go
@@ -7,6 +7,8 @@ package evalchecks
 import (
 	"fmt"
 	"runtime"
+	"strconv"
+	"unsafe"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/addrs"
@@ -17,6 +19,17 @@ import (
 )
 
 type EvaluateFunc func(expr hcl.Expression) (cty.Value, tfdiags.Diagnostics)
+
+func maxCountValue() int64 {
+	var instanceKey addrs.InstanceKey
+
+	// expansionCount later allocates a []addrs.InstanceKey, so count is
+	// bounded by the largest backing array Go can represent for that slice.
+	const heapAddrBits = 32 + 16*(strconv.IntSize/64)
+	const maxAllocBytes = (uint64(1) << heapAddrBits) - 1
+
+	return int64(maxAllocBytes / uint64(unsafe.Sizeof(instanceKey)))
+}
 
 // EvaluateCountExpression is our standard mechanism for interpreting an
 // expression given for a "count" argument on a resource or a module. This
@@ -126,6 +139,15 @@ func EvaluateCountExpressionValue(expr hcl.Expression, ctx EvaluateFunc) (cty.Va
 			Severity: hcl.DiagError,
 			Summary:  "Invalid count argument",
 			Detail:   `The given "count" argument value is unsuitable: must be greater than or equal to zero.`,
+			Subject:  expr.Range().Ptr(),
+		})
+		return nullCount, diags
+	}
+	if maxCount := maxCountValue(); int64(count) > maxCount {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid count argument",
+			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: must be less than or equal to %d.`, maxCount),
 			Subject:  expr.Range().Ptr(),
 		})
 		return nullCount, diags
--- a/internal/lang/evalchecks/eval_count.go
+++ b/internal/lang/evalchecks/eval_count.go
@@ -6,9 +6,8 @@ package evalchecks
 
 import (
 	"fmt"
+	"math"
 	"runtime"
-	"strconv"
-	"unsafe"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/addrs"
@@ -20,15 +19,10 @@ import (
 
 type EvaluateFunc func(expr hcl.Expression) (cty.Value, tfdiags.Diagnostics)
 
+const maxCount = int64(math.MaxInt32)
+
 func maxCountValue() int64 {
-	var instanceKey addrs.InstanceKey
-
-	// expansionCount later allocates a []addrs.InstanceKey, so count is
-	// bounded by the largest backing array Go can represent for that slice.
-	const heapAddrBits = 32 + 16*(strconv.IntSize/64)
-	const maxAllocBytes = (uint64(1) << heapAddrBits) - 1
-
-	return int64(maxAllocBytes / uint64(unsafe.Sizeof(instanceKey)))
+	return maxCount
 }
 
 // EvaluateCountExpression is our standard mechanism for interpreting an
@@ -123,6 +117,16 @@ func EvaluateCountExpressionValue(expr hcl.Expression, ctx EvaluateFunc) (cty.Va
 		return cty.UnknownVal(cty.Number), diags
 	}
 
+	if countVal.Type() == cty.Number && countVal.GreaterThan(cty.NumberIntVal(maxCountValue())).True() {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid count argument",
+			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: must be less than or equal to %d.`, maxCountValue()),
+			Subject:  expr.Range().Ptr(),
+		})
+		return nullCount, diags
+	}
+
 	var count int
 	err := gocty.FromCtyValue(countVal, &count)
 	if err != nil {
@@ -139,15 +143,6 @@ func EvaluateCountExpressionValue(expr hcl.Expression, ctx EvaluateFunc) (cty.Va
 			Severity: hcl.DiagError,
 			Summary:  "Invalid count argument",
 			Detail:   `The given "count" argument value is unsuitable: must be greater than or equal to zero.`,
-			Subject:  expr.Range().Ptr(),
-		})
-		return nullCount, diags
-	}
-	if maxCount := maxCountValue(); int64(count) > maxCount {
-		diags = diags.Append(&hcl.Diagnostic{
-			Severity: hcl.DiagError,
-			Summary:  "Invalid count argument",
-			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: must be less than or equal to %d.`, maxCount),
 			Subject:  expr.Range().Ptr(),
 		})
 		return nullCount, diags
--- a/internal/lang/eval/internal/tofu2024/instance_selector.go
+++ b/internal/lang/eval/internal/tofu2024/instance_selector.go
@@ -25,6 +25,8 @@ import (
 	"github.com/opentofu/opentofu/internal/tfdiags"
 )
 
+const maxCount = int64(math.MaxInt32)
+
 func compileInstanceSelector(ctx context.Context, declScope exprs.Scope, forEachExpr hcl.Expression, countExpr hcl.Expression, enabledExpr hcl.Expression, deps dependsOn) configgraph.InstanceSelector {
 	// We don't current verify that only one of the given expressions is set
 	// because we expect the configs package to check that.
@@ -107,11 +109,8 @@ func compileInstanceSelectorCount(_ context.Context, countValuer exprs.Valuer, d
 					err = errors.New("must be a whole number")
 				} else if bf.Cmp(big.NewFloat(0)) < 0 {
 					err = errors.New("must not be a negative number")
-				} else if v, acc := bf.Int64(); acc != big.Exact || v > math.MaxInt {
-					// This will eventually result in a Go slice of the
-					// requested length, so we are constrained by Go's maximum
-					// slice length on the current platform.
-					err = fmt.Errorf("must be between 0 and %d, inclusive", math.MaxInt)
+				} else if v, acc := bf.Int64(); acc != big.Exact || v > maxCount {
+					err = fmt.Errorf("must be between 0 and %d, inclusive", maxCount)
 				}
 			}
 			if err == nil {
--- a/internal/lang/eval/internal/tofu2024/instance_selector_test.go
+++ b/internal/lang/eval/internal/tofu2024/instance_selector_test.go
@@ -577,17 +577,18 @@ func TestCompileInstanceSelectorCount(t *testing.T) {
 				diagsHasError("must be a whole number."),
 			},
 			"very large number": {
-				// This number is definitely out of range on both 32-bit and
-				// 64-bit targets.
 				hcl.StaticExpr(cty.MustParseNumberVal("99999999999999999999"), rng),
 				dependsOn{},
 				nil,
 				nil,
-				// The exact upper bound in this error message differs between
-				// 32-bit and 64-bit targets, and so we only match the constant
-				// prefix here which is enough to distinguish it from all
-				// of the other errors this function could return.
-				diagsHasError("must be between 0 and "),
+				diagsHasError("must be between 0 and 2147483647, inclusive."),
+			},
+			"larger than maximum count": {
+				hcl.StaticExpr(cty.NumberIntVal(maxCount+1), rng),
+				dependsOn{},
+				nil,
+				nil,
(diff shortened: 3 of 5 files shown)


internal/lang/evalchecks/eval_count.go
```diff
@@ -130,6 +143,15 @@ func EvaluateCountExpressionValue(expr hcl.Expression, ctx EvaluateFunc) (cty.Va
 		})
 		return nullCount, diags
 	}
+	if maxCount := maxCountValue(); int64(count) > maxCount {
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid count argument",
+			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: must be less than or equal to %d.`, maxCount),
```
yottta:
```suggestion
			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable for the current machine architecture: must be less than or equal to %d.`, maxCount),
```
I am not sure if it's a good idea or not, but will leave this here for being debated.

yottta (approved):
Just one tiny nitpick.
Other than that, beautifully done! 🚀 
Thanks

internal/lang/evalchecks/eval_count.go
apparentlymart: I keep considering whether we should just limit this to the 32-bit limit on all platforms, for consistency.

A `cty.Value` on a 64-bit system is 16 bytes long (two Go interface values) and so a 2,147,483,648-element slice of them would immediately take 32GiB RAM even before considering the storage of the main object representing each instance. In practice using a `count` that large is likely to exceed the available memory on a typical system, and even if not OpenTofu would run unusably slowly because the algorithms it uses are not appropriate for that scale.

Specifying the maximum slice length on a 64-bit system would require 137,438,953,472 GiB RAM just for the main slice of `cty.Value`, before considering the individual objects. That is an utterly absurd amount of memory on today's practical systems, and so I don't think supporting that number of instances is useful. :grinning: 

Part of me wondered if we should set an even lower limit than 2<sup>31</sup> so that it isn't possible to specify something that is likely to make OpenTofu fail in more troublesome ways than panicking. For example, we could choose 2<sup>24</sup> as the limit, allowing up to 16,777,216 instances with a backing array of 256 MiB. But it's harder to argue for an arbitrary-seeming limit like that[^1] than to argue for the physical constraint of 32-bit numbers.

What do you all think? Should we just fix this at 2<sup>31</sup>-1 on all platforms, regardless of CPU architecture?

[^1]: With that said: I did previously successfully argue for setting the maximum length of the result of [the `range` function](https://opentofu.org/docs/language/functions/range/) to the far-smaller limit of 1024 elements, limiting the maximum slice storage size to 16kiB. That function tends to be used just like `count`, except in `for` expressions rather than in resource/module/provider declarations. :man_shrugging:

goutamadwant: makes sense to me.. I followed the earlier direction of using the maximum slice length, so the current limit ends up depending on the platform.

I am happy to switch this to a fixed cross platform cap if that is the preferred direction. `math.MaxInt32` seems reasonable since it keeps the behavior and diagnostic the same across platforms, avoids architecture specific wording, and is still far beyond any realistic `count` value.

If we go that route I can update the helper and tests to use the fixed cap and keep the diagnostic as a plain upper bound error. Let me know thoughts.

yottta: I totally agree @apparentlymart that the total possible number of a slice elements is absurdly high (TBH, in both cases) and mostly unusable IRL.
But I am still concerned of using that as the sole argument of adding a limitation that could could actually break somebody's workflow.
We have this PR because we want to fix _a panic_ by handling an edge case.

Even though the chances are slim for this to be hit by the users, by changing a long standing behaviour means a breaking change. And without clear visibility and the data points to support a decision like this, I wouldn't add a limit that maybe, somebody, somewhere on a 64bit system, could actually hit.

apparentlymart: Perhaps we could test how OpenTofu currently behaves when given `count = pow(2, 32)` on a 64-bit system, to see what behavior we'd be preserving compatibility with.

If it's possible to use that succesfully today, even if it's very slow, then I think it's reasonable to preserve it.

My worry was that it would cause a worse experience than the current panic -- an "out of memory" kill of the process on a Linux system, for example -- and so an error message we control could be better than that. But I agree that we shouldn't break it if we find that its actually possible to use it in today's OpenTofu.

(I'll give this a try on my Linux system soon, but I'd also be interested to see for it behaves on macOS and Windows where the memory management rules are different.)

apparentlymart: Here's the configuration I tried on my `linux_amd64` system:

```hcl
resource "terraform_data" "example" {
  count = pow(2, 32)
}
```

When I ran `tofu plan` for this configuration, it immediately crashed with a different unrecoverable runtime error:

```
fatal error: runtime: out of memory
```

Since this is a fatal error, we would not be able to intercept it using `recover` and then return a customized error message. This result doesn't seem particularly better than the `makeslice: len out of range` error we were trying to fix.

My system has 64GiB RAM total, so just allocating the slice already requires half of my system's available RAM. However, I don't have swap enabled on my system and so overcommit is not possible. This could potentially have worked and just been very slow if I had a sufficiently-large swap partition.

I'd still be curious to see how this behaves on a macOS or Windows system.

---

**Edit:** Out of curiosity I created a 64GiB swap file and enabled it and tried again. It spun for 15 minutes gradually allocating more and more memory until it had saturated all of my available RAM _and_ all of my available swap space, and then the Linux kernel OOM-killed my terminal, my text editor, and finally the `tofu` process itself. :upside_down_face:

In practice OpenTofu tracks quite a lot of additional information on a per-resource-instance basis, and so that 32GiB of `cty.Value` I previously quoted is just a bare minimum and the true memory usage for those instances would be considerably higher.

I can't see how anyone could be usefully depending on a `count` that's greater than the largest positive `int32` value in any practical situation. The real practical limit seems to be considerably lower than that.

yottta: Great! Thanks for taking your time to test this.
Considering the data that was provided in the previous reply, I would be ok with this suggestion, so @goutamadwant, we are all in line with this suggestion so you can apply it:
> What do you all think? Should we just fix this at 2<sup>31</sup>-1 on all platforms, regardless of CPU architecture?

goutamadwant: Thanks @yottta @apparentlymart for checking this and for testing the larger value.
I will update the patch to use a fixed `2^31 - 1` limit across platforms and adjust the tests around that.

apparentlymart:
I notice that the tests for the new language runtime are failing here.

Our original goal had been for new runtime development to stay out of the way of other work with the assumption that if other work breaks the new runtime then those working on the new runtime would be responsible for fixing it separately later. As far as I know we didn't intentionally change that policy but we have _effectively_ changed it by adding this extra test run as a mandatory check. 😖

In this particular case I expect it shouldn't be too hard to add a matching fix to the new runtime and that would hopefully be enough to make the experimental test run work. I'll take a look at this during my work day tomorrow unless someone else gets there first.

goutamadwant:
@apparentlymart @yottta Pushed a follow-up fix for the experimental runtime path too. The workflows are waiting for maintainer approval.. could someone trigger them when you get a chance?

internal/lang/evalchecks/eval_count.go
```diff
@@ -20,15 +19,10 @@ import (
...
-	var instanceKey addrs.InstanceKey
-
-	// expansionCount later allocates a []addrs.InstanceKey, so count is
-	// bounded by the largest backing array Go can represent for that slice.
-	const heapAddrBits = 32 + 16*(strconv.IntSize/64)
-	const maxAllocBytes = (uint64(1) << heapAddrBits) - 1
+const maxCount = int64(math.MaxInt32)
 
-	return int64(maxAllocBytes / uint64(unsafe.Sizeof(instanceKey)))
+func maxCountValue() int64 {
```
yottta: Is this needed anymore?

internal/lang/evalchecks/eval_count.go
```diff
@@ -123,31 +117,32 @@ func EvaluateCountExpressionValue(expr hcl.Expression, ctx EvaluateFunc) (cty.Va
...
-	if count < 0 {
+
+	var count int
+	err := gocty.FromCtyValue(countVal, &count)
+	if err != nil {
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "Invalid count argument",
-			Detail:   `The given "count" argument value is unsuitable: must be greater than or equal to zero.`,
+			Detail:   fmt.Sprintf(`The given "count" argument value is unsuitable: %s.`, err),
```
yottta, line 132: Why not keeping the order of the checks as it was before? 🤔 
Because `FromCtyValue` will return an error if the type of the pointer is different than the type of the `countVal`.
And then we can keep the other 2 validations as before.

apparentlymart: I think this is a similar situation as in the new runtime's code in `instance_selector.go` where using `gocty.FromCtyValue` as the main check can cause the error message to be confusing.

If `gocty.FromCtyValue` finds that the given number is not in range for the target type -- for example, if it were too large -- then `gocty` would generate its own error message that mentions the range of `int`:

```
must be between -9223372036854775808 and 9223372036854775807, inclusive
```

...but `count` is actually required to be greater than or equal to zero, so this default error message would be incorrect for count. Performing the range check first means that we can customize the error message to always mention the correct range for this argument.

Even with this updated code I think that weird behavior would still be observable if `count` were set to a very _negative_ number, like `-9223372036854775809`, because that would fail the conversion to `int` before we reach the test for the value being negative. But I don't think we should worry about fixing that here because it's a pretty unlikely edge case and the new runtime is already handling that situation better anyway so this'll be fixed once we adopt that new codepath.

goutamadwant: Thanks @apparentlymart @yottta . The current ordering keeps the upper-bound check before `gocty.FromCtyValue`, so large values get the count-specific diagnostic rather than the generic int conversion error. Hope that's okay.

The only remaining cleanup I see is the now-redundant `maxCountValue()` helper. I can remove that in a small follow-up if you prefer. Let me know.

yottta: Thanks @goutamadwant too.
Yeah, let's go with that and let's see if the next review round will be the last. 🚀

apparentlymart (approved):

internal/lang/evalchecks/eval_count.go
apparentlymart: Hi @goutamadwant!

I just wanted to make sure you're aware that this comment is blocking us from moving forward here.

I think what @yottta was referring to[^1] is that since `maxCountValue` is just immediately returning the value of `maxCount` without any further computation, presumably all of the calls to this function could be replaced with direct references to `maxCount` instead.

[^1]: Please correct me if I'm wrong!

goutamadwant: sorry this got missed @yottta @apparentlymart addressed this now.  I removed the redundant maxCountValue helper and switched the remaining validation and also test references to use maxCount directly.

yottta (approved):

internal/lang/evalchecks/eval_count.go
yottta: Looks good. Thanks for handling this.

PR #4306, opened 2026-06-23
Resolve symlinks in NormalizePath to fix tofu fmt with absolute paths

### LINKED ISSUE #3879: `tofu fmt` fails to find file if given an absolute path from a symlinked directory in some cases
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.11.5
on linux_amd64
```

Not applicable, it doesnt' really matter what the hcl is, it's a bug in the CLI

### Debug Output

https://gist.github.com/tmccombs/5b29dfb1816daa023262b8c915dba6b7

### Expected Behavior

Passing an absolute path to `tofu fmt` should format the file at that location.

### Actual Behavior

I get an error that the file doesn't exist, showing an incorrect relative path instead of the absolute path I gave it:

```
│ Error: No file or directory at ../a/b/c/test.tf
│ 
│ 
```

when passed a path like `/home/redacted/a/b/c/test.tf`

### Steps to Reproduce

From some directory on linux, run the following commands:

```bash
$ mkdir -p a/b/c
$ touch a/b/c/test.tf
$ ln -s $PWD/a/b z
$ f=$PWD/a/b/c/test.tf
$ cd z
$ tofu fmt $f
╷
│ Error: No file or directory at ../a/b/c/test.tf
│ 
│ 
```

### Additional Context

Interestingly, if the symlink is to a folder at the same level as the source of the symlink, then this doesn't happen. You need to have run `tofu fmt` from a directory that is a symlink to a folder that is at least one level deeper or shallower than the current directory.

### References

_No response_

ricardbejarano:
## Summary

Fixes #3879.

`tofu fmt /absolute/path/to/file.tf` fails with `No file or directory` when the command is run from a directory that was reached through a symlink (e.g. `cd` into a symlink that points one or more levels deeper/shallower).

### Root cause

`Dir.NormalizePath` derives the working directory with `filepath.Abs(d.mainDir)`, which for the usual `"."` main dir calls `os.Getwd()`. On Unix `os.Getwd()` prefers the `$PWD` environment variable when it names the current directory, so it returns the **symlinked** path. The kernel, however, resolves relative paths against the **real** directory. `filepath.Rel` then produces a relative path against the symlinked location (e.g. `../a/b/c/test.tf`) that `os.Stat` resolves against the real cwd, pointing at a non-existent file.

This was diagnosed in the issue thread by the reporter and a contributor.

### Fix

Canonicalize the main directory with `filepath.EvalSymlinks` before computing the relative path, keeping the result consistent with how the filesystem resolves it. Relative and `"."` inputs are unaffected: the canonicalization cancels out in the subsequent `filepath.Rel` computation, so only absolute inputs under a symlinked working directory change behavior.

## Testing

Added `TestNormalizePathSymlinkedMainDir`, which reproduces the issue's scenario (a symlinked main directory plus an absolute target path) and asserts the normalized result both equals the expected relative path and resolves back to the original file. The test is skipped on Windows, where the symlink semantics under test don't apply.

## Target Release

1.13.0

## CHANGELOG entry

Added under BUG FIXES.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ BUG FIXES:
 - Fixed TRACESTATE log message incorrectly printing the TRACEPARENT value instead. ([#4168](https://github.com/opentofu/opentofu/issues/4168))
 - Fix rendering of plans where a nested block's replacement is unknown. ([#4256](https://github.com/opentofu/opentofu/issues/4256))
 - `errored.tfstate` is now produced during a go runtime panic. This file will be a partial state and is intended for aiding in recovery from a hard crash. ([#4064](https://github.com/opentofu/opentofu/pull/4064))
+- `tofu fmt` no longer fails with "No file or directory" when given an absolute path while running from a directory reached through a symlink. ([#3879](https://github.com/opentofu/opentofu/issues/3879))
 
 ## Previous Releases
 
--- a/internal/command/workdir/normalize_path.go
+++ b/internal/command/workdir/normalize_path.go
@@ -40,6 +40,20 @@ func (d *Dir) NormalizePath(given string) string {
 		return filepath.Clean(given)
 	}
 
+	// filepath.Abs relies on os.Getwd, which on Unix prefers the $PWD
+	// environment variable when it names the current directory. If OpenTofu
+	// was launched from a directory reached through a symlink, $PWD keeps the
+	// symlinked path while the kernel still resolves relative paths against
+	// the real directory. Computing a relative path against the symlinked
+	// path then yields a result that no longer resolves to the intended file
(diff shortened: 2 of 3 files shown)


yottta:
Hi @ricardbejarano,

Thanks a lot for the submission, but I will have to close this for the following reasons:
* The issue it fixes is not accepted since there is no consensus yet on what the final fix for this would be;
* You are not assigned to the issue, which is [one of the rules](https://github.com/opentofu/opentofu/blob/main/CONTRIBUTING.md#get-started) that help maintainers to keep things organised;
* Your PR is missing the required checklist from the [PR template](https://github.com/opentofu/opentofu/blob/main/.github/pull_request_template.md)

Thanks for the understanding.

Issue #4308, opened 2026-06-23
Upgrade to Go 1.27


apparentlymart:
We plan for the OpenTofu v1.13 series to be based on Go 1.27. This issue therefore represents the work of making OpenTofu use a non-prerelease version of Go 1.27.

As part of this we will need to also complete https://github.com/opentofu/opentofu/issues/3951, because Go 1.27 adopts Unicode 17.

apparentlymart:
Completing this will also serve as the equivalent of https://github.com/opentofu/opentofu/pull/4472 and https://github.com/opentofu/opentofu/pull/4473 for the `main` branch, because those same upstream fixes have already been incorporated into go1.27rc3 and so will presumably be in the final release too.

Since there have not yet been any stable releases in the OpenTofu v1.13 series, the v1.13 changelog need not mention these changes.

apparentlymart:
The changes needed for us to begin using Unicode 17 once we upgrade to Go 1.27 are already merged from resolving https://github.com/opentofu/opentofu/issues/3951, and we're already claiming Unicode 17 support in our changelog.

However, we won't _actually_ be using Unicode 17 until we complete this issue, so the changelog is currently technically wrong.

PR #4309, opened 2026-06-24
go.mod: Upgrade to Go 1.27rc1

apparentlymart:
For the forthcoming OpenTofu v1.13.x series we intend to use Go 1.27. The first stable release in that series is not available yet, but in the meantime we'll adopt its first release candidate so we can get early notice if there are any unexpected changes we need to deal with.

This commit deals with the first significant upstream change: the standard library implementation of DEFLATE compression has been optimized in this release in a way that changes the exact bytes it produces, and since DEFLATE is the basis of both the gzip and zip packages this has two knock-on effects for OpenTofu:

- Our `base64gzip` function now produces equivalent but non-equal results compared to previous releases.

    This is a user-facing change that is therefore noted in the changelog.

- The helper we use to dynamically generate installable provider packages for testing now generates equivalent but non-equal .zip archives, meaning that the expected `zh:` checksums have changed but the `h1:` checksums have not.

    This is NOT a user-facing change, because OpenTofu only generates provider package archives inside its own test suite as throwaway test fixtures.

The updated changelog also confirms the change in which macOS versions are supported, which we pre-announced in the changelog for the v1.12 series.

There are other changes we will need to make before we can consider ourselves to have fully adopted Go 1.27 -- not least of which is actually using a stable release from that series, once available! -- but this deals with the immediate, mandatory changes we need to make to keep the tests working and make the change information visible to those who might try to test using our nightly prereleases.

This is the first step for https://github.com/opentofu/opentofu/issues/4308.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ UPGRADE NOTES:
 
     [Modern Windows versions now support OpenSSH](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse), and so we suggest that anyone currently relying on WinRM plan to migrate to using SSH instead.
 
+- The `base64gzip` function now generates results that are equivalent to _but not equal to_ the results from previous releases, as a result of a new optimized DEFLATE compression implementation.
+
+    If you use this function as part of an argument to a managed resource then OpenTofu is likely to propose to update or replace the instances of that resource, depending on how the provider responds to the differing base64 data. The new compressed form should nonetheless still decompress to the same sequence of bytes.
+
+- OpenTofu on macOS now requires macOS 13 Ventura or later. Earlier versions are no longer supported.
+
 ENHANCEMENTS:
 
 - The `gcp_kms` key provider now supports an optional `additional_authenticated_data` as part of the encryption and decryption operations. ([#4287](https://github.com/opentofu/opentofu/pull/4287))
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/opentofu/opentofu
 
-go 1.26.4
+go 1.27
+
+toolchain go1.27rc1
 
 // At the time of adding this configuration, the new Go feature introduced here https://github.com/golang/go/issues/67061,
 // was having a good amount of issues linked to, affecting AWS Firewall, GCP various services and a lot more.
--- a/internal/command/init_test.go
+++ b/internal/command/init_test.go
@@ -1852,7 +1852,7 @@ func TestInit_providerSource(t *testing.T) {
 			getproviders.MustParseVersionConstraints("= 1.2.4"),
 			[]getproviders.Hash{
 				getproviders.HashScheme1.New("vEthLkqAecdQimaW6JHZ0SBRNtHibLnOb31tX9ZXlcI="),
-				getproviders.HashSchemeZip.New("ec7c3fd6eb575c06f0e6957e1ee8531a588805c4eeb8abb5e4156911e080eb31"),
+				getproviders.HashSchemeZip.New("4115ba79e1745c9236f24ce18e643cfc1893140cbc963066348d2dde9c7c55c5"),
 			},
 		),
 		addrs.NewDefaultProvider("test"): depsfile.NewProviderLock(
@@ -1861,7 +1861,7 @@ func TestInit_providerSource(t *testing.T) {
 			getproviders.MustParseVersionConstraints("= 1.2.3"),
 			[]getproviders.Hash{
 				getproviders.HashScheme1.New("8CjxaUBuegKZSFnRos39Fs+CS78ax0Dyb7aIA5XBiNI="),
-				getproviders.HashSchemeZip.New("6f85a1f747dd09455cd77683c0e06da647d8240461b8b36b304b9056814d91f2"),
+				getproviders.HashSchemeZip.New("ac90a1419dd4648431af93b6695b6d6efcad87950ee608de4198df8e6debff23"),
 			},
 		),
 		addrs.NewDefaultProvider("source"): depsfile.NewProviderLock(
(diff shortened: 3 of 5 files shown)


apparentlymart:
Hmm... it seems that the `setup-go` GitHub Action doesn't understand how to install Go 1.27rc1, even though the Go toolchain itself understands this syntax. :thinking: 

I'll mark this as draft for now and see if I can find a way to get this prerelease to work with how our test jobs are implemented.

**Edit:** Based on some initial investigation upstream, it seems like we may need to temporarily add an explicit `go-version` setting to each of the GitHub Actions workflow configurations, specifying a semver-based form of this version number, instead of letting the action automatically decide the version based on `go.mod`. Ugh, if so... the whole point of having it in `go.mod` was to have only one place to update. I'll look for some better options soon.

apparentlymart:
I can at least see what's causing the failure in the upstream source code: [the `go.mod` parsing code](https://github.com/actions/setup-go/blob/924ae3a1cded613372ab5595356fb5720e22ba16/src/installer.ts#L660-L674) is using a regular expression to parse the `go` directive and that regexp happens to match the `go 1.27` part of `go 1.27rc1`, so it's trying to install 1.27 final and that doesn't exist yet.

There's another pattern there which attempts to match a `toolchain` directive and that one _does_ have special support for matching an `rc\d+` suffix. Unfortunately the Go toolchain itself does not accept that syntax in `toolchain`:

```
go: invalid toolchain "1.27rc1" in go.mod
```

According to [the `go.mod` documentation](https://go.dev/ref/mod#go-mod-file-go) the logic in this action is backwards: the `go` directive _does_ support this `1.27rc1` syntax, but the `toolchain` directive does not. The current behavior seems to have been implemented in https://github.com/actions/setup-go/pull/460 but there's no discussion there which explains why it doesn't match the way the Go toolchain interprets these directives. I assume it was just an oversight. :man_shrugging: 

Unfortunately though I think this means we can't use Go 1.27rc1 unless we temporarily switch back to having our selected version number duplicated in both `go.mod` and in a separate legacy file like `.go-version`, which I would like to avoid since we switched away from that style intentionally to make sure we're always testing with the same version we're building with.

**Edit:** After some more fiddling I realized what I was missing: the `toolchain` directive _does_ support selecting a release candidate, but it also requires the toolchain version to begin with `go`, like this:

```gomod
toolchain go1.27rc1
```

This syntax (including the support for release candidates) is confirmed in the related documentation [Go toolchain names](https://go.dev/doc/toolchain#name).

The Go toolchain seems happy with this and so I'm going to try submitting it and see if setup-go also supports it. (The source code linked above suggests that it should.)

apparentlymart:
Okay, it now matches the release candidate as expected but still doesn't work:

```
Setup go version spec 1.27rc1
Attempting to download 1.27rc1...
matching 1.27rc1...
Not found in manifest.  Falling back to download directly from Go
Error: Unable to find Go version '1.27rc1' for platform linux and architecture x64.
```

Referring again to the action source code, it appears to first try its own local manifest, and then fall back to fetching `https://go.dev/dl/?mode=json&include=all` and try to find the selected version in there, and that fallback is what apparently happened in this case.

[The version-matching code](https://github.com/actions/setup-go/blob/924ae3a1cded613372ab5595356fb5720e22ba16/src/installer.ts#L580-L599) transforms the toolchain names like `go1.27rc1` from that endpoint into an NPM-style variant syntax something like `1.27.0-rc.1`.

I dug in a little to the `semver` library that this is relying on for the "satisfies" check and I think firstly this codepath doesn't seem to support prereleases at all (it isn't setting the `includePrerelease` option when calling `satisfies`), but even if it did I think it also isn't applying the same rewriting to the version constraint and so I think it would be trying to match `1.27.0-rc.1` with `1.27-rc1` and so fail to find a match anyway.

So again I'm left with the conclusion that we can't select this release candidate using directives in the `go.mod` file, though for a different reason than I initially thought. :confounded: 

---

**Edit:** I re-ran one the test jobs with debug logging enabled, and indeed it seems to be trying to compare the semver-transformed version with the raw string from the `go.mod` file:

```
##[debug]check 1.27.0-rc.1 satisfies 1.27rc1
```

The fact that this call isn't allowing the semver library to match prereleases anyway probably makes this a moot point, but it seems like this wouldn't match even if prereleases were enabled because `rc1` and `rc.1` are two separate prerelease versions as far as the semver rules are concerned.

I'm going to give up on this for now. I don't think we can actually use Go release candidates in our `go.mod` with the current implementation of `actions/setup-go`.

apparentlymart:
I'm not sure why I didn't find this the first time I looked, but looking again today I noticed that there's already an open issue and a potential PR open for this upstream:

- https://github.com/actions/setup-go/issues/525
- https://github.com/actions/setup-go/pull/701

Neither has had much activity lately, though. I don't think we should count on this being fixed in the near future.

apparentlymart:
When revisiting this again we should add the following "ENHANCEMENTS" changelog entries to match changes in upstream Go:

> - OpenTofu now respects the `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables on macOS and Windows, in the same situations where they were previously respected on Linux and other Unix systems.
> - OpenTofu language features that are based on Unicode specifications now use the algorithms and character database from Unicode 17.

The second of these changelog entries represents that upgrading to Go 1.27 is the final step of upgrading to Unicode 17, but note that we also need to make sure we're running cty 1.19.0 or later, `golang.org/x/text` v0.34.0 or later, and a HCL release that includes https://github.com/hashicorp/hcl/pull/814 (which we may have to pull into our fork if upstream doesn't merge it and release it before we need to merge this PR).

apparentlymart:
Perhaps during the next release period we'll consider ways we can adopt the Go release candidate as soon as it becomes available, but at this point the final Go 1.27 release will presumably land in the next week or two anyway so I'm going to abandon this PR and we'll just upgrade directly to Go 1.27 once it's available.

PR #4313, opened 2026-06-25
Better integration of providers and modules installation in new runtime

### LINKED ISSUE #4060: Reimplement `tofu init`'s behavior in terms of the new language runtime
Our initial work on the new language runtime (from https://github.com/opentofu/opentofu/issues/3414) treated it as an experimental opt-in dispatched from inside the old runtime, and so in particular it's only used for operations that run through the old runtime: validation, planning, and applying.

`tofu init` in particular continues to work directly with the old config loader and its "early evaluation" mechanism, but we'd like to unify that so that in the long run everything is using the shared evaluator and "early evaluation" becomes just normal evaluation with more unknown values in scope and restrictions on where unknown values are allowed to appear.

- [ ] Some mechanism for dynamic opt-in to the new-style `tofu init` behavior, ideally reacting to the same signals that the opt-in shim inside the old language runtime reacts to so that the new runtime is either activated for everything or not activated at all.
- [ ] In the new codepath, use the new evaluator to find the backend configuration, the module dependencies, and the provider dependencies.

    In particular this means that the module installer would run concurrently with the evaluator and install new modules gradually throughout evaluation as `source` and `version` argument values become known. We may wish to defer this until after we've reworked our caching strategy for remote dependencies, since a shared cache that's turned on by default is likely a better fit for the way the evaluator wants this to work. We'll see, though.
- [ ] Change the module-loading code used during the validation, planning, and applying phases to fully support what the new-style `tofu init` produces, vs. the current situation where it has only basic support for local module sources and can't interact with remote modules at all.

cam72cam:
This is a partial solution to #4060. It swaps out the module and provider installer, as well as wires in support for remote modules into the legacy path.

There is a lot in this PR that is not ideal, but it aims to be as close as possible to a 1-1 representation of how we init and run tofu now. In the long term, I expect much of this code will go away when we remove the "split-path" state that tofu is in currently.

Resolves #4060

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/meta_backend.go
+++ b/internal/command/meta_backend.go
@@ -172,6 +172,8 @@ func (m *Meta) Backend(ctx context.Context, opts *BackendOpts, enc encryption.St
 		}
 	}
 	if !m.NewRuntimeEnabled() {
+		// The new runtime does not need a pre-run validation pass as it is already baked in
+		// The old engine however does need the pre-run validation pass
 		cliOpts.Validation = true
 	}
 
--- a/internal/command/meta_backend.go
+++ b/internal/command/meta_backend.go
@@ -171,7 +171,9 @@ func (m *Meta) Backend(ctx context.Context, opts *BackendOpts, enc encryption.St
 			return nil, diags
 		}
 	}
-	cliOpts.Validation = true
+	if !m.NewRuntimeEnabled() {
+		cliOpts.Validation = true
+	}
 
 	// If the backend supports CLI initialization, do it.
 	if cli, ok := b.(backend.CLI); ok {
--- a/internal/configs/version_constraint.go
+++ b/internal/configs/version_constraint.go
@@ -111,6 +111,9 @@ func (v VersionConstraint) HasRequirements() bool {
 
 func (v VersionConstraint) Check(ver *version.Version) bool {
 	if !isUnsetOrAllVersions(v.RequiredSet) {
+		// This conversion between version constraint types/packages is
+		// not ideal, but is a best effort for the new engine. Eventually,
+		// we should remove v.Required in a subsequent refactoring pass.
 		nver, err := versions.ParseVersion(ver.String())
 		if err != nil {
 			// This should not be possible as the two version packages have the same supported formats
--- a/internal/configs/version_constraint.go
+++ b/internal/configs/version_constraint.go
@@ -8,6 +8,7 @@ package configs
 import (
 	"fmt"
 
+	"github.com/apparentlymart/go-versions/versions"
 	version "github.com/hashicorp/go-version"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/lang/marks"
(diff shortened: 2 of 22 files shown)


internal/command/meta_backend.go
```diff
@@ -171,7 +171,9 @@ func (m *Meta) Backend(ctx context.Context, opts *BackendOpts, enc encryption.St
 			return nil, diags
 		}
 	}
-	cliOpts.Validation = true
+	if !m.NewRuntimeEnabled() {
+		cliOpts.Validation = true
+	}
```
apparentlymart, line 178: I expect it'll be helpful to add a comment here explaining why this is necessary, so that when we return to clean this up later we'll be able to evaluate whether to keep this or not.

(And probably the same for the various other little conditional branches based on `m.NewRuntimeEnabled()`, for the same reason.)

internal/configs/version_constraint.go
```diff
@@ -98,3 +100,31 @@ func decodeVersionConstraintValue(attr *hcl.Attribute, val cty.Value) (VersionCo
...
+func (v VersionConstraint) HasRequirements() bool {
+	return len(v.Required) > 0 || !isUnsetOrAllVersions(v.RequiredSet)
+}
+
+func (v VersionConstraint) Check(ver *version.Version) bool {
+	if !isUnsetOrAllVersions(v.RequiredSet) {
+		nver, err := versions.ParseVersion(ver.String())
+		if err != nil {
+			// This should not be possible as the two version packages have the same supported formats
+			panic(err)
```
apparentlymart, line 120: Unfortunately we've observed that this isn't quite true, with some notes about that in https://github.com/opentofu/opentofu/issues/2147.

However, I think I'm understanding that this particular branch can only run when the new runtime is in use since only then would `v.RequiredSet` be populated. Is that right? If that were true then I think we could probably accept treating certain unlikely input as now being invalid -- this is less likely and easier to fix than some of the other differences we already provisionally accepted for the new runtime.

I think it's also true that the most significant differences between `version` and `versions` are taken care of by round-tripping through `version.Version` and then back to its string representation. For example, that would cause any user-authored `v` prefix to be trimmed off, avoiding a repeat of the problem discussed in https://github.com/opentofu/opentofu/issues/2117.

So all of that is to say: I don't think this is problematic in practice, but just wanted to draw attention to some potential hazards in case we end up referring back to this in future after finding our assumptions were wrong. 😀

apparentlymart (approved):
Hmm I just came here today and found these unsubmitted comments I wrote a while ago 🤦‍♂️

Sorry for the delay.

internal/configs/version_constraint.go
cam72cam: Added some additional comments in [b6334ed](https://github.com/opentofu/opentofu/pull/4313/commits/b6334eddfd3631f98585f61e74e4ce97c09d07fe)

internal/command/meta_backend.go
cam72cam: Added comments in [b8a7a88](https://github.com/opentofu/opentofu/pull/4313/commits/b8a7a88c7cdff551589e57b3b50b63f8c55a06cd)

PR #4315, opened 2026-06-25
Add support for tofu.* and path.* attributes in new runtime

cam72cam:
Depends on https://github.com/opentofu/opentofu/pull/4313

## Checklist

### Go checklist

 

### DIFF
--- a/internal/lang/eval/internal/evalglue/context.go
+++ b/internal/lang/eval/internal/evalglue/context.go
@@ -51,6 +51,21 @@ type EvalContext struct {
 	// but leaves OriginalWorkingDir unchanged.
 	RootModuleDir, OriginalWorkingDir string
 
+	// Applying currently wired in via the tofu context shims and does not
+	// represent the final state of how this information should be passed
+	// into the evaluator.
+	//
+	// TODO When we remove the shim, we should try to find a way to prevent
+	// this from disagreeing with the current operation the evaluator is
+	// being asked to perform.
+	Applying bool
+
+	// Workspace is the current tofu workspace passed in from the command
+	// layer
+	Workspace string
+
+	// PlanTimestamp is the time at which the plan was created. This is
+	// used to provide a consistent result for the plantimestamp function.
 	PlanTimestamp time.Time
 }
 
--- a/internal/lang/eval/internal/tofu2024/module_instance_scope.go
+++ b/internal/lang/eval/internal/tofu2024/module_instance_scope.go
@@ -9,14 +9,17 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"path/filepath"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/opentofu/opentofu/internal/addrs"
+	"github.com/opentofu/opentofu/internal/didyoumean"
 	"github.com/opentofu/opentofu/internal/instances"
 	"github.com/opentofu/opentofu/internal/lang/eval/internal/configgraph"
 	"github.com/opentofu/opentofu/internal/lang/eval/internal/evalglue"
 	"github.com/opentofu/opentofu/internal/lang/exprs"
+	"github.com/opentofu/opentofu/internal/lang/marks"
 	"github.com/opentofu/opentofu/internal/tfdiags"
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/function"
@@ -43,6 +46,15 @@ type moduleInstanceScope struct {
 	inst              *CompiledModuleInstance
 	coreFunctions     map[string]function.Function
 	providerFunctions func(context.Context, addrs.ProviderFunction, hcl.Range) (function.Function, tfdiags.Diagnostics)
+
+	// tofu attrs
+	applying  bool
+	workspace string
+
+	// path attrs
+	workingDir string
+	rootDir    string
+	sourceDir  string
 }
 
 var _ exprs.Scope = (*moduleInstanceScope)(nil)
@@ -108,7 +120,7 @@ func (m *moduleInstanceScope) ResolveAttr(ref hcl.TraverseAttr) (exprs.Attribute
 	var diags tfdiags.Diagnostics
 	switch ref.Name {
 
-	case "var", "local", "module":
+	case "var", "local", "module", "path", "terraform", "tofu":
 		// For various relatively-simple cases where there's just one level of
 		// nested symbol table we use a single shared [exprs.SymbolTable]
 		// implementation which then just delegates back to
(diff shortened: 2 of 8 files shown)


internal/lang/eval/internal/evalglue/context.go
```diff
@@ -51,6 +51,9 @@ type EvalContext struct {
 	// but leaves OriginalWorkingDir unchanged.
 	RootModuleDir, OriginalWorkingDir string
 
+	Applying  bool
```
apparentlymart: This one is interesting, because the the evaluator currently knows from the entrypoint used whether we're applying or not, but that information isn't visible to `tofu2024` during compilation.

I think having this information represented in two cases seems okay for now -- no need to overthink it -- but at the later step where we start calling the new runtime directly instead of shimming through `package tofu` hopefully we can find a different way to arrange this where it's harder to accidentally make `terraform.applying`'s result disagree with what operation the evaluator thinks it's currently performing. 🤔

I don't have any big ideas about this yet, but maybe we could put a `TODO` comment here to remind us to consider this again when we reach that point?

internal/lang/eval/internal/tofu2024/module_instance_scope.go
```diff
@@ -233,7 +246,98 @@ func (m *moduleInstanceScope) resolveSimpleChildAttr(topSymbol string, ref hcl.T
...
+		case "env":
+			// Prior to Terraform 0.12 there was an attribute "env", which was
+			// an alias name for "workspace". This was deprecated and is now
+			// removed.
+			diags = diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  fmt.Sprintf("Invalid %q attribute", topSymbol),
+				Detail:   fmt.Sprintf(`The %s.env attribute was deprecated in v0.10 and removed in v0.12. The "state environment" concept was renamed to "workspace" in v0.12, and so the workspace name can now be accessed using the %s.workspace attribute.`, topSymbol, topSymbol),
+				Subject:  &ref.SrcRange,
+			})
```
apparentlymart: This has been obsolete for long enough now that I think we could stand to stop treating it as a special case and just let the generic error handling case below handle it, but I do see that it doesn't add much complexity to keep using this special error message so I don't feel very strongly about it.

internal/lang/eval/internal/tofu2024/module_instance_scope.go
```diff
@@ -233,7 +246,98 @@ func (m *moduleInstanceScope) resolveSimpleChildAttr(topSymbol string, ref hcl.T
...
+		var val cty.Value
+
+		// Mostly copied from the tofu/evaluate.go
+		switch ref.Name {
+
+		case "cwd":
+			wd := m.workingDir
+			var err error
+			if wd == "" {
+				wd, err = os.Getwd()
```
apparentlymart: Although I can see that I apparently originally set this up so that the field populated by the called in `EvalContext` is `OriginalWorkingDir` and thus it's implied that it should only be set when the working directory was overridden, now I see how this ended up I think I'd prefer a model where the caller is required to _always_ provide an absolute working directory path -- handling any errors itself -- so that the logic here is literally just to return that provided path every time instead of doing its own lookup.

Do you think that's a reasonable way to shape this? Generally I just prefer the idea of doing this work once up front so we can get any strange failure cases out of the way immediately, instead of performing a separate fallible operation for each reference to `path.cwd`, since failure to determine the current working directory is an environmental problem rather than a problem with the expression itself.

apparentlymart (commented):
Overall this seems reasonable to me! Some initial reactions inline, but these are all opinions weakly held.

internal/lang/eval/internal/evalglue/context.go
cam72cam: Added a TODO and part of your explanation.

internal/lang/eval/internal/tofu2024/module_instance_scope.go
cam72cam: Removed, but left a comment explaining why.

internal/lang/eval/internal/tofu2024/module_instance_scope.go
cam72cam: Done. It did duplicate a call to filepath.Abs as we construct the context in two places currently, but that's pretty minor.

apparentlymart (approved):

diofeher (approved):

internal/lang/eval/internal/tofu2024/module_instance_scope.go
```diff
@@ -233,7 +245,63 @@ func (m *moduleInstanceScope) resolveSimpleChildAttr(topSymbol string, ref hcl.T
...
+
+		case "workspace":
+			val = cty.StringVal(m.workspace)
+
+		// case "env": intentionally removed as it has been deprecated for a *long* time.
+		default:
+			diags = diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  fmt.Sprintf("Invalid %q attribute", topSymbol),
+				Detail:   fmt.Sprintf(`The %q object does not have an attribute named %q. The only supported attributes are %s.workspace and %s.applying`, topSymbol, ref.Name, topSymbol, topSymbol),
```
yottta, line 264:
```suggestion
				Detail:   fmt.Sprintf(`The %q object does not have an attribute named %q.`, topSymbol, ref.Name),
```
I wouldn't hardcode existing options because will be easy later when adding new attributes to miss updating this error message.

We either remove that altogether, or we do a Levenshtein distance to provide suggestion or move the attributes to a `map[string]func()` and use the keys to generate the suggestion.

I would say that not adding the suggestions at all would be a good start. And this will be consistent with the next topLevel case (ie: `"path"`) where there is no suggestion.

internal/tofu/context_temp_runtime.go
```diff
@@ -107,6 +125,8 @@ func (c *Context) newEngineShim(ctx context.Context, config *configs.Config, inp
 		Providers:          plugins,
 		Provisioners:       plugins,
 		PlanTimestamp:      planTimestamp,
+		Applying:           false,
```
yottta:
```suggestion
		Applying:           applying,
```
Uh, mistake?

yottta (changes requested):

internal/lang/eval/internal/tofu2024/module_instance_scope.go
cam72cam: This was designed to be as close as possible to the existing code, which uses this approach.  The message is updated once every 5 years or so.

yottta (approved):

PR #4318, opened 2026-06-26
fix - apply progress reporting stalling after first heartbeat in -json output

### LINKED ISSUE #4107: tofu apply -json emits only one apply_progress message per resource (heartbeat goroutine blocks on a test-only channel)
### OpenTofu Version

```
OpenTofu v1.10.5
on darwin_arm64
```

The relevant code is unchanged on `v1.10.x`, `v1.11.x`, `v1.12.0`, and `main`, so all currently supported versions are affected.

```hcl
terraform {
  required_providers {
    time = {
      source = "hashicorp/time"
    }
  }
}

# Stays in the "applying" state for 60s, so the per-resource apply
# heartbeat should fire at 10s, 20s, 30s, 40s and 50s.
resource "time_sleep" "wait" {
  create_duration = "60s"
}
```

### Debug Output

With `-json`, exactly one `apply_progress` message is emitted for the resource (at ~10s elapsed), and then no further progress messages appear for the remaining ~50s of the apply:

```
{"@level":"info","@message":"time_sleep.wait: Still creating... [10s elapsed]","@module":"tofu.ui","type":"apply_progress","hook":{"resource":{...},"action":"create","elapsed_seconds":10}}
```

Running the same configuration **without** `-json` produces the expected repeating output:

```
time_sleep.wait: Still creating... [10s elapsed]
time_sleep.wait: Still creating... [20s elapsed]
time_sleep.wait: Still creating... [30s elapsed]
time_sleep.wait: Still creating... [40s elapsed]
time_sleep.wait: Still creating... [50s elapsed]
```

### Expected Behavior

In `-json` mode, `apply_progress` messages should be emitted periodically (every 10s) for each resource that is still applying, for as long as it is applying — matching both the non-JSON output and the [machine-readable UI documentation](https://opentofu.org/docs/internals/machine-readable-ui/), which describes `apply_progress` as being emitted "periodically, showing elapsed time output".

### Actual Behavior

In `-json` mode, only the **first** `apply_progress` message (at ~10s) is emitted per resource. No further progress messages are produced, regardless of how long the resource takes. The heartbeat goroutine also leaks for the duration of the apply.

### Steps to Reproduce

1. `tofu init`
2. `tofu apply -json -auto-approve` and observe the streamed JSON — only one `apply_progress` for `time_sleep.wait`.
3. For comparison, `tofu destroy -auto-approve` then `tofu apply -auto-approve` (no `-json`) — observe `Still creating...` printed every 10s.

### Additional Context

The cause is in the JSON view's per-resource heartbeat goroutine, `internal/command/views/hook_json.go`.

`applyProgress` carries an `elapsed` channel that — per its own comment — exists only for tests:

```go
// elapsed is used to allow tests to safely check for heartbeat executions
elapsed chan time.Duration
```

It is created **unbuffered** in `PreApply`:

```go
elapsed: make(chan time.Duration),
```

and the heartbeat goroutine performs a **blocking send** on it after each tick:

```go
func (h *jsonHook) applyingHeartbeat(progress applyProgress) {
	defer close(progress.heartbeatDone)
	defer close(progress.elapsed)
	for {
		select {
		case <-progress.done:
			return
		case <-h.timeAfter(heartbeatInterval):
		}

		elapsed := h.timeNow().Round(time.Second).Sub(progress.start)
		h.view.Hook(json.NewApplyProgress(progress.addr, progress.action, elapsed))
		progress.elapsed <- elapsed   // <-- blocks forever in production
	}
}
```

The only readers of `progress.elapsed` anywhere in the codebase are in `internal/command/views/hook_json_test.go`. In a real run nothing receives from it, so after the first tick the goroutine emits one `apply_progress` via `h.view.Hook(...)` and then blocks permanently on `progress.elapsed <- elapsed`. It never returns to the `select`, so:

- no further heartbeats are emitted for that resource, and
- it never observes `progress.done`, so the goroutine leaks (and `heartbeatDone` is never closed) until the process exits.

#### Proposed fix

Decouple production behaviour from the test-only channel by buffering it (cap 1) and making the send non-blocking:

```go
// in PreApply
elapsed: make(chan time.Duration, 1),
```

```go
// in applyingHeartbeat, replace `progress.elapsed <- elapsed` with:
select {
case progress.elapsed <- elapsed:
default:
}
```

- **Production:** the goroutine never blocks on the send, so heartbeats continue to fire every `heartbeatInterval`, and the goroutine returns promptly once `progress.done` is closed (no leak).
- **Tests:** `hook_json_test.go` drains `elapsed` once per simulated tick before advancing the mocked clock, so with a cap-1 buffer the buffer is always empty when the goroutine sends — the non-blocking send always succeeds and the existing `<-elapsedChan` synchronisation still works deterministically.

I'm happy to open a PR with this fix if it would help.

### References

- `internal/command/views/hook_json.go` — `applyProgress.elapsed`, `PreApply`, `applyingHeartbeat`
- `internal/command/views/hook_json_test.go` — the only readers of `progress.elapsed`
- [Machine-Readable UI documentation — `apply_progress`](https://opentofu.org/docs/internals/machine-readable-ui/)

Jougan-0:
When running apply with `-json`, progress messages stopped after the first one. The goroutine that sends progress was getting stuck on a channel that only the tests read, so in real runs it blockedforever. I changed it to pass that channel in only when needed (nil in production) as suggested under the issue by @apparentlymart , so progress keeps coming. Tests now create their own channel through a small `preApply` helper.

This PR also fixes an unrelated flaky test, TestVarsParsing/console, that was intermittently failing CI on Windows. The console command is read-only but still acquired a state lock; on slow runners the "Acquiring state lock…" message exceeded the 400ms threshold and leaked into stdout, breaking the test's exact-output assertion. Since console is read-only, the test now runs with -lock=false, making it deterministic.

Resolves #4107 

## Checklist

### Go checklist

 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 
-- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one. ([#4107](https://github.com/opentofu/opentofu/issues/4107))
+- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one. ([#4107](https://github.com/opentofu/opentofu/pull/4318))
 - The built-in function `contains` now accepts `null` as its second argument, to test whether a collection contains any null values. ([#4043](https://github.com/opentofu/opentofu/issues/4043))
 - The built-in function `merge` no longer fails when its only argument is a null value of an object type. ([#4043](https://github.com/opentofu/opentofu/issues/4043))
 - The built-in function `cidrhost` no longer returns a "panic" error when called with an out-of-range host number represented in more than 64 bits. ([#4056](https://github.com/opentofu/opentofu/pull/4056))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 
-- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one.([#4107](https://github.com/opentofu/opentofu/issues/4107))
+- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one. ([#4107](https://github.com/opentofu/opentofu/issues/4107))
 - The built-in function `contains` now accepts `null` as its second argument, to test whether a collection contains any null values. ([#4043](https://github.com/opentofu/opentofu/issues/4043))
 - The built-in function `merge` no longer fails when its only argument is a null value of an object type. ([#4043](https://github.com/opentofu/opentofu/issues/4043))
 - The built-in function `cidrhost` no longer returns a "panic" error when called with an out-of-range host number represented in more than 64 bits. ([#4056](https://github.com/opentofu/opentofu/pull/4056))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 
+- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one.([#4107](https://github.com/opentofu/opentofu/issues/4107))
 - The built-in function `contains` now accepts `null` as its second argument, to test whether a collection contains any null values. ([#4043](https://github.com/opentofu/opentofu/issues/4043))
 - The built-in function `merge` no longer fails when its only argument is a null value of an object type. ([#4043](https://github.com/opentofu/opentofu/issues/4043))
 - The built-in function `cidrhost` no longer returns a "panic" error when called with an out-of-range host number represented in more than 64 bits. ([#4056](https://github.com/opentofu/opentofu/pull/4056))
--- a/internal/command/command_test.go
+++ b/internal/command/command_test.go
@@ -1250,7 +1250,7 @@ func TestVarsParsing(t *testing.T) {
 			},
 		}
 
-		args := append([]string{"-no-color"}, varArgs...)
+		args := append([]string{"-no-color", "-lock=false"}, varArgs...)
 		code := c.Run(args)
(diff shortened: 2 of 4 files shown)


Jougan-0:
Taking a look at the failing test case

Jougan-0:
Fixed the flaky test and updated the description. It was a one-line change so I bundled it here instead of a separate PR, let me know if you'd prefer it split out.
Root cause: the test acquired a state lock that isn't needed (console is read-only). On slow CI the "Acquiring state lock…" message leaked into stdout and broke the exact-output assertion; -lock=false makes it deterministic.

CHANGELOG.md
```diff
@@ -16,10 +16,10 @@ ENHANCEMENTS:
...
 - The `local-exec` provisioner now automatically sets the `TRACEPARENT` environment variable in child processes when OpenTelemetry tracing is active, following the W3C Trace Context specification. ([#4014](https://github.com/opentofu/opentofu/issues/4014))
 - When installing provider and module packages from OCI Distribution registries, OpenTofu now tracks separate transient credentials for each repository to support registry implementations that issue repository-scoped tokens.  ([#3316](https://github.com/opentofu/opentofu/issues/3316))
-- The `providers lock` command now supports the argument `-oci-mirror`. The functionality mimics that of the field `repository_template` of `oci_mirror`-block in [`provider_installation`](https://opentofu.org/docs/cli/config/config-file/#provider-installation) with the exception of using a URI template instead of a HCL one.
 
 BUG FIXES:
 
+- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one. ([#4107](https://github.com/opentofu/opentofu/issues/4107))
```
```suggestion
- `tofu apply -json` now emits periodic `apply_progress` heartbeat messages for the full duration of a resource operation, instead of stopping after the first one. ([#4107](https://github.com/opentofu/opentofu/pull/4318))
```
I know that we are inconsistent with this (as it can be seen in the other entries), but I am trying lately to head towards uniformity.

yottta (approved):
This looks good to me.
Left just a small correction in the changelog.

lbordowitz (approved):

yottta:
Just tested this to be sure and everything looks good when compared with OpenTofu v1.12.2.
image

Thanks @Jougan-0 for the contribution! 🙌

PR #4325, opened 2026-06-30
Just enough new-runtime tracing to shim the `tofu.Hook` API

apparentlymart:

This PR introduces a new general-purpose API for tracing events in the new planning and applying engines, and then uses that API to shim the `tofu.Hook` API just enough to get a bunch of new context tests running and passing.

The tracing API here is designed in the same modern-ish shape we've used for other tracers recently while we were implementing OpenTelemetry support, where the "Start" event is allowed to return a new `context.Context` annotated with tracing-related metadata and then that context is used for all operations related to that Start/End pair, including the eventual call to "End" that can then terminate the trace. In the long run I expect us to use this to implement OpenTelemetry tracing as a more centralized concern, living in the CLI layer instead of spread awkwardly over the entire runtime.

However, the `tofu.Hook` API long predates OpenTelemetry tracing or `context.Context` and so that particular aspect of the tracing API is unused for now. Instead, this adds some partial stubs of the hook API that _approximate_ how the original language runtime would've called those hooks.

I prioritized this now because there are several tests which use the `tofu.Hook` API as a side concern for testing the behavior of something else, and so having this minimal implementation in place unblocks us for working on the various other problems those tests are representing. This is not yet intended to be complete enough to fully drive the CLI layer's UI hooks and so I expect we'll do another pass for that in a later PR but we have more fundamental concerns to deal with first.

There are a few tests that continue to fail even with these changes in place, and so I've added some more granular flags to represent those. Note that only one of them is classified as a bug to be fixed, because the other two are from tests of behaviors that are intentionally different in the new runtime and we have no current intention of changing that.

---

### DIFF
--- a/internal/engine/applying/apply.go
+++ b/internal/engine/applying/apply.go
@@ -16,6 +16,7 @@ import (
 	"github.com/opentofu/opentofu/internal/lang/eval"
 	"github.com/opentofu/opentofu/internal/lang/grapheval"
 	"github.com/opentofu/opentofu/internal/plans"
+	"github.com/opentofu/opentofu/internal/shared"
 	"github.com/opentofu/opentofu/internal/states"
 	"github.com/opentofu/opentofu/internal/tfdiags"
 )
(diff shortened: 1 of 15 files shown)


cam72cam (approved):
I'm curious what the impact of putting lots of concerns in the context.Context will be over time, but don't object to the experiment.

apparentlymart:
I do agree with the concern about overusing `context.Context` values, though to me _this_ use of it seems the least controversial of what we've been doing lately because it seems aligned with the use-cases that motivated that API: cross-cutting concerns like logging and tracing. If we eventually use this for OpenTelemetry as I suggested here then this will complement how OpenTelemetry trace metadata gets passed around inside Go programs, making sure that the trace metadata and the callbacks that use it always propagate together to wherever they are needed.

The one I'm keeping my eye on the most is using it to track the workgraph workers in the evaluator, since that's definitely pushing beyond "cross-cutting concern" and into faking goroutine-local storage. But as before I lack any better idea for how to handle that given how many different parts of the system need to collaborate on planning and applying. :thinking:

PR #4326, opened 2026-06-30
Implement moved block statements in the new engine

### LINKED ISSUE #4229: `moved` support for the new language runtime
In the OpenTofu language, `moved` blocks serve as a record of how resource instance objects have moved between different addresses during the maintenance of a module. OpenTofu can use this information for a number of purposes in principle, but the main behavior authors rely on is OpenTofu's ability to automatically rebind an existing object to a new resource instance address instead of planning to destroy it at its old address and create a replacement object at the new address.

The current language runtime handles this as a preprocessing step where it directly modifies the prior state to pretend as if the objects had _always_ been at their new addresses, and only then begins the normal planning phase.

We have not yet decided in detail how to handle this in the new language runtime, but one possible approach is to preload all of the moved statements into a central object that encapsulates them all and allows us to ask questions like "what other resource instance addresses could potentially have an object that needs to move into this one?", and then ask those questions as part of the normal planning process, instead of faking it by rewriting the prior state before planning even begins. The new runtime is in theory designed to be able to support objects having different addresses at different phases of the process, but we've not yet put that into practice and we may find gaps as we try to implement this.

lbordowitz:
Resolves #4229 

This PR implements move statements for the experimental runtime. The main two files of this change are `config_plan_move.go`, which lets the plan oracle handle move statement and move results, and `plan_managed.go`, which uses those methods to implement moves for managed resources. There's a lot to say about this implementation, and I hope this description helps with reviewing the code!

In short: the oracle has been updated to have move statements and record moves. Moves are detected in "unwanted" resources, which are not deleted but instead ignored and handled in "desired." In addition to picking up the previously-orphaned state, "desired" also handles provider requests for state upgrades.

There's a subtle aspect of moves that bears mentioning: a move only occurs when there is state for an address, but no configuration at that address; if there's no state, there's nothing to move, and if there's configuration, any move statement that might occur would error given that configuration matches some `from` address. Because we're looking for state without configuration, it's most logical to have move logic in `planUnwantedManagedResourceInstanceObject`, and I recommend reviewers start from that function. This is also the reason I put `announcePlanOrphans` and `checkAll` in sequence in `config_plan.go`; the moves made in "unwanted" are subsequently picked up in "desired"[^1].

[^1]: I suppose a future version might have concurrent "adoption of orphaned state", but it would have to be some fancy footwork to manage blocking between candidate state and configuration.

I found it interesting how there are similarities between the current runtime and the experimental runtime. With `NoOp`, the current set of circumstances determining which action to take resembles [a switch statement in `NodeAbstractResourceInstance.plan`](https://github.com/opentofu/opentofu/blob/2ea7fa46c78ab9ba0c5e888234b8db07445359dc/internal/tofu/node_resource_abstract_instance.go#L1505-L1518). You may find it helpful to also compare the decision of whether to run a `MoveResourceState` or `UpgradeResourceState` to its corresponding decision point [in the current runtime](https://github.com/opentofu/opentofu/blob/2ea7fa46c78ab9ba0c5e888234b8db07445359dc/internal/tofu/node_resource_abstract.go#L646-L650). The diff is somewhat inscrutable because of the changed control structure, but the `Upgrade` part of the code has not changed much. Actually, the only change made to the `Upgrade` part of the code involves implicit moves: if it's an implied move, we change the place we save state over to the previous address. This prevents a [panic](https://github.com/opentofu/opentofu/blob/2ea7fa46c78ab9ba0c5e888234b8db07445359dc/internal/states/state.go#L545) later in the [exec operations](https://github.com/opentofu/opentofu/blob/2ea7fa46c78ab9ba0c5e888234b8db07445359dc/internal/engine/applying/operations_resource_managed.go#L478). I'm... not entirely sure about that part of the PR...

That diff in `planDesiredManagedResourceInstance` is so difficult to understand... let me try to simplify. If the state is non-nil, we'll try to upgrade it. Using the same decision logic, we choose which branch to go down. Note that implied moves are always upgraded and moves are only upgraded if the resource type hasn't changed. Also note that both upgrades are very similar; we may want to do a refactor equivalent to what it looks like in the current runtime.

Speaking of implicit moves: their handling is somewhat different than what happens in the current runtime. Rather than creating ersatz statements if no such statement exists, we look in the configuration for a "reciprocal address", that is, one that is equivalent to the given address in state, with the exception of swapping `NoKey` and `IntKey(0)` in one location[^3]. Another difference is that implicit moves only occur if there are no explicit moves for that address in state. I think it might technically be possible to mix implicit and explicit moves, but it seems a bit difficult or wasteful resource-wise[^2].

[^3]: I've also done an investigation about nested implicit moves, where multiple counts are changed in both modules and resource. It doesn't look like our current engine does more than one implicit move at a time; I do not believe the experimental engine's implicit moves do more than one implicit move, either, however, I was not able to test this as I encountered a strange bug when I tried!

[^2]: The wasteful part is the address lookup. I wish there were some sort of tree to look up absolute resource addresses, and also a "trie" data structure that we could access all the instance keys for a given absolute resource address... although see "pyrrhic move" for why that still isn't sufficient. As it stands, looping through instances to find the right one every time seems like a bad idea, and we might want a (non-sync, read-only) map to cache said results once expansion is complete.

I'll quickly mention the "pyrrhic move." Occasionally, a resource changes from `NoKey` to `count = 0`. The current runtime records this as an implicit move, but then deletes it anyway because there's no corresponding instance. The pyrrhic move exists to match this quirk in functionality. See the `TestContext2Apply_scaleInMultivarRef` test, which gave me no small amount of grief.

Part of this PR involved implementing the `NoOp` operation. This simplified a bit of logic in `planDesiredManagedResourceInstance`. You'll also notice that some tests have been readjusted to put them back in line with how the current runtime operates.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/planning/execgraph_managed.go
+++ b/internal/engine/planning/execgraph_managed.go
@@ -47,7 +47,7 @@ func (b *execGraphBuilder) ManagedResourceInstanceSubgraph(
 	if (plannedChange.Action == plans.Delete || plannedChange.Action == plans.Forget) && !plannedChange.After.IsNull() {
 		panic(fmt.Sprintf("change for %s has action %s but non-null planned new value", plannedChange.PrevRunAddr, plannedChange.Action))
 	}
-	if plannedChange.Action != plans.Create && plannedChange.Action != plans.Delete && plannedChange.Action != plans.Forget && plannedChange.Action != plans.NoOp && (plannedChange.Before.IsNull() || plannedChange.After.IsNull()) {
+	if plannedChange.Action != plans.Create && plannedChange.Action != plans.Delete && plannedChange.Action != plans.Forget && (plannedChange.Before.IsNull() || plannedChange.After.IsNull()) {
 		panic(fmt.Sprintf("change for %s has action %s but does not have both a before and after value", plannedChange.PrevRunAddr, plannedChange.Action))
 	}
 
@@ -291,13 +291,11 @@ func (b *execGraphBuilder) managedResourceInstanceSubgraphNoOp(
 	plannedChange *plans.ResourceInstanceChange,
 ) resourceInstanceObjectSubgraph {
 	_, addCreateDep := b.lower.MutableWaiter()
-	_, addDeleteDep := b.lower.MutableWaiter()
 
 	_, _, priorStateRef, _ := b.managedResourceInstanceChangeInputs(plannedChange)
 	return resourceInstanceObjectSubgraph{
 		valueRef:      priorStateRef,
 		addDesiredDep: addCreateDep,
-		addOrphanDep:  addDeleteDep,
 	}
 }
 
--- a/internal/engine/planning/execgraph_managed.go
+++ b/internal/engine/planning/execgraph_managed.go
@@ -47,7 +47,7 @@ func (b *execGraphBuilder) ManagedResourceInstanceSubgraph(
 	if (plannedChange.Action == plans.Delete || plannedChange.Action == plans.Forget) && !plannedChange.After.IsNull() {
 		panic(fmt.Sprintf("change for %s has action %s but non-null planned new value", plannedChange.PrevRunAddr, plannedChange.Action))
 	}
-	if plannedChange.Action != plans.Create && plannedChange.Action != plans.Delete && plannedChange.Action != plans.Forget && (plannedChange.Before.IsNull() || plannedChange.After.IsNull()) {
+	if plannedChange.Action != plans.Create && plannedChange.Action != plans.Delete && plannedChange.Action != plans.Forget && plannedChange.Action != plans.NoOp && (plannedChange.Before.IsNull() || plannedChange.After.IsNull()) {
 		panic(fmt.Sprintf("change for %s has action %s but does not have both a before and after value", plannedChange.PrevRunAddr, plannedChange.Action))
 	}
 
@@ -72,10 +72,7 @@ func (b *execGraphBuilder) ManagedResourceInstanceSubgraph(
 	case plans.CreateThenDelete:
 		return b.managedResourceInstanceSubgraphCreateThenDelete(plannedChange)
 	case plans.NoOp:
-		// TODO: We need to handle this because it can occur if the
-		// configuration hasn't changed but the object will move to a new
-		// resource instance address during the apply phase.
-		panic("plans.NoOp execution graph not yet implemented")
+		return b.managedResourceInstanceSubgraphNoOp(plannedChange)
 	default:
 		// We should not get here: the cases above should cover every action
 		// that [planGlue.planDesiredManagedResourceInstance] can possibly
@@ -285,6 +282,20 @@ func (b *execGraphBuilder) managedResourceInstanceSubgraphCreateThenDelete(
 	return resourceInstanceObjectSubgraph{
 		valueRef:      createResultRef,
 		deletionRef:   deletionRef,
+		addDesiredDep: addCreateDep,
+		addOrphanDep:  addDeleteDep,
+	}
+}
+
+func (b *execGraphBuilder) managedResourceInstanceSubgraphNoOp(
+	plannedChange *plans.ResourceInstanceChange,
+) resourceInstanceObjectSubgraph {
+	_, addCreateDep := b.lower.MutableWaiter()
+	_, addDeleteDep := b.lower.MutableWaiter()
+
+	_, _, priorStateRef, _ := b.managedResourceInstanceChangeInputs(plannedChange)
+	return resourceInstanceObjectSubgraph{
+		valueRef:      priorStateRef,
 		addDesiredDep: addCreateDep,
 		addOrphanDep:  addDeleteDep,
 	}
--- a/internal/engine/planning/plan_managed.go
+++ b/internal/engine/planning/plan_managed.go
@@ -133,146 +133,246 @@ func (p *planGlue) planDesiredManagedResourceInstance(
 	var prevRoundVal cty.Value
 	var prevRoundPrivate []byte
 	prevRoundState := p.planCtx.prevRoundState.SyncWrapper().ResourceInstanceObjectFull(inst.Addr.CurrentObject())
-	if prevRoundState != nil {
-		// While we know prevRoundState is non-nil, let's upgrade state, too.
-		// Let's do a schema version comparison before upgrade
 
-		if prevRoundState.SchemaVersion > uint64(schema.Version) {
-			return nil, diags.Append(tfdiags.Sourceless(
-				tfdiags.Error,
-				"Resource instance managed by newer provider version",
-				// This is not a very good error message, but we don't retain enough
-				// information in state to give good feedback on what provider
-				// version might be required here. :(
-				// Or maybe we do. I dunno, I just copied the comment+diag from
-				// upgrade_resource_state.go:upgradeResourceStateTransform :P
-				fmt.Sprintf("The current state of %s was created by a newer provider version than is currently selected. Upgrade the %s provider to work with this state.", inst.Addr, inst.Provider.Type),
-			))
-		}
+	prevRunAddr := inst.Addr
+	diags = diags.Append(p.oracle.CheckMovesFromAddr(inst.Addr))
+	if diags.HasErrors() {
+		return ret, diags
+	}
 
-		upgradeCtx := ctx
-		if cb := tracer.StartManagedResourceInstanceObjectUpgrade; cb != nil {
-			upgradeCtx = cb(ctx, inst.Addr.CurrentObject())
-		}
-		upgradeReq := providers.UpgradeResourceStateRequest{
-			TypeName: inst.Addr.Resource.Resource.Type,
-
-			// TODO: The internal schema version representations are all using
-			// uint64 instead of int64, but unsigned integers aren't friendly
-			// to all protobuf target languages so in practice we use int64
-			// on the wire. In future we will change all of our internal
-			// representations to int64 too.
-			Version: int64(prevRoundState.SchemaVersion),
-
-			// We'll make the same assumption as [ResourceInstanceObjectFullSrc] and
-			// assume we'll never encounter a legacy state snapshot that uses AttrsFlat.
-			RawStateJSON: prevRoundState.Value.ValueJSON,
-		}
-		upgradeResp := providerClient.UpgradeResourceState(upgradeCtx, upgradeReq)
-		diags = diags.Append(upgradeResp.Diagnostics)
-		if cb := tracer.EndManagedResourceInstanceObjectUpgrade; cb != nil {
-			upgradedVal := cty.DynamicVal
-			if upgradeResp.UpgradedState != cty.NilVal {
-				// TODO: Should apply "sensitive" marks here where appropriate in
-				// case the tracer is reporting events in the UI.
-				upgradedVal = upgradeResp.UpgradedState
-			}
-			cb(upgradeCtx, inst.Addr.CurrentObject(), upgradedVal, diags)
-		}
+	impliedMove := false
+	moved := false
+	if prevRoundState == nil {
+		// check for ambiguous moves
+		_, moveDiags := p.oracle.FindAddressesMovedToHere(ctx, inst.Addr)
+		diags = diags.Append(moveDiags)
 		if diags.HasErrors() {
+			// More than one address this could move to.
+			// "Ambiguous move statements": many From, one To
 			return ret, diags
 		}
-		newState := upgradeResp.UpgradedState
-
-		// After upgrading, the new value must conform to the current schema. When
-		// going over RPC this is actually already ensured by the
-		// marshaling/unmarshaling of the new value, but we'll check it here
-		// anyway for robustness, e.g. for in-process providers.
-		if errs := newState.Type().TestConformance(schema.Block.ImpliedType()); len(errs) > 0 {
-			providerType := inst.Addr.Resource.Resource.ImpliedProvider()
-			for _, err := range errs {
-				diags = diags.Append(tfdiags.Sourceless(
-					tfdiags.Error,
-					"Invalid resource state transformation",
-					fmt.Sprintf("The %s provider changed the state for %s, but produced an invalid result: %s.", providerType, inst.Addr, tfdiags.FormatError(err)),
-				))
+		if moveInfo, ok := p.oracle.MovedAddress(inst.Addr); ok {
+			// we tracked a move in "Unwanted", manage state upgrades and plans here.
+			moved = true
+			prevRunAddr = moveInfo.From
+			prevRoundState = p.planCtx.prevRoundState.SyncWrapper().ResourceInstanceObjectFull(prevRunAddr.CurrentObject())
+			impliedMove = moveInfo.Implied
+		}
+	}
+	// only run MoveResourceState or UpgradeResourceState if prevRoundState is non-nil at this point.
+	if prevRoundState != nil {
+		if moved && !impliedMove && (resourceType.ResourceTypeName() != prevRoundState.ResourceType || !inst.Provider.Equals(prevRoundState.ProviderInstanceAddr.Config.Config.Provider)) {
+			moveCtx := ctx
+			if cb := tracer.StartManagedResourceInstanceObjectMove; cb != nil {
+				moveCtx = cb(ctx, inst.Addr.CurrentObject())
 			}
-			return nil, diags
-		}
 
-		src, err := ctyjson.Marshal(newState, schema.Block.ImpliedType())
-		if err != nil {
-			// We just checked for type conformance above, so getting into this
-			// codepath is probably a bug.
-			diags = diags.Append(tfdiags.Sourceless(
-				tfdiags.Error,
-				"Failed to encode result of resource state transformation",
-				fmt.Sprintf("Failed to encode state for %s after resource schema upgrade: %s.", inst.Addr, tfdiags.FormatError(err)),
-			))
-		}
+			// log.Printf("[TRACE] moveResourceStateTransform: new address: %s, previous address: %s", inst.Addr, prevRunAddr)
+			req := providers.MoveResourceStateRequest{
+				SourceProviderAddress: prevRunAddr.Resource.Resource.ImpliedProvider(),
+				SourceTypeName:        prevRunAddr.Resource.Resource.Type,
+				SourceSchemaVersion:   prevRoundState.SchemaVersion,
+				// We'll make the same assumption as [ResourceInstanceObjectFullSrc] and
+				// assume we'll never encounter a legacy state snapshot that uses AttrsFlat.
+				SourceStateJSON: prevRoundState.Value.ValueJSON,
+				// SourceStateFlatmap:    prevRoundState.AttrsFlat,
+				SourcePrivate:  prevRoundState.Private,
+				TargetTypeName: inst.Addr.Resource.Resource.Type,
+			}
+			resp := providerClient.MoveResourceState(moveCtx, req)
+			diags = diags.Append(resp.Diagnostics)
+			// TODO this tracer bit is copypasta for upgrade instance, IDK if it's actually legit...
+			if cb := tracer.EndManagedResourceInstanceObjectMove; cb != nil {
+				upgradedVal := cty.DynamicVal
+				if resp.TargetState != cty.NilVal {
+					// TODO: Should apply "sensitive" marks here where appropriate in
+					// case the tracer is reporting events in the UI.
+					upgradedVal = resp.TargetState
+				}
+				cb(moveCtx, inst.Addr.CurrentObject(), upgradedVal, diags)
+			}
+			if diags.HasErrors() {
+				return
+			}
 
-		upgradedPrevState := &states.ResourceInstanceObjectFullSrc{
-			Value: states.ValueJSONWithMetadata{
-				ValueJSON:      src,
-				SensitivePaths: prevRoundState.Value.SensitivePaths,
-			},
-			Private:              prevRoundState.Private,
-			Status:               prevRoundState.Status,
-			ProviderInstanceAddr: prevRoundState.ProviderInstanceAddr,
-			ResourceType:         prevRoundState.ResourceType,
-			SchemaVersion:        uint64(schema.Version),
-			Dependencies:         prevRoundState.Dependencies,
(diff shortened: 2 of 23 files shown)


internal/lang/eval/config_plan_oracle.go
```diff
@@ -11,15 +11,24 @@ import (
...
-	root      evalglue.CompiledModuleInstance
-	providers *managedProviders
+	root           evalglue.CompiledModuleInstance
+	providers      *managedProviders
+	moveStatements []refactoring.MoveStatement
+	moveResults    moveResults
+}
+
+// HasAddress queries the root module instance to determine if the address is
+// present in the configuration.
```
lbordowitz, line 29: Maybe there should be some comment about `ResourceInstance` blocking until instances are expanded?

internal/lang/eval/config_plan_move.go
```diff
@@ -0,0 +1,350 @@
...
+			// TODO determine DeclRange, probably by adding some config information from the caller
+			declaredAt := ""
+
+			diags = diags.Append(&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Moved object still exists",
+				Detail: fmt.Sprintf(
+					"This statement declares a move from %s, but that %s is still declared%s.\n\nChange your configuration so that this %s will be declared as %s instead.",
+					// TODO: these are wrong, see above
+					addr, addr.Noun(), declaredAt, addr.ShortNoun(), to,
```
lbordowitz: Any ideas on how to do this right? I'm not obtaining the equivalent diagnostic present in [ValidateMoves](https://github.com/opentofu/opentofu/blob/2ea7fa4/internal/refactoring/move_validate.go#L105-L118)

internal/engine/planning/plan_managed.go
```diff
@@ -525,16 +634,76 @@ func (p *planGlue) planUnwantedManagedResourceInstanceObject(
...
+
+					// We also undo the address, but keep that we "found a move"
+					// so we correctly remove this blocked state
+					currentRunAddr = addr.InstanceAddr
+					break
+				}
+				currentRunAddr = nextAddr
+
+				// We continue the loop, although technically, we should not
+				// find any more addresses in configuration.
```
lbordowitz, line 685: Should I just break here instead, to save potential cycles?

internal/engine/planning/plan_managed.go
```diff
@@ -133,146 +133,238 @@ func (p *planGlue) planDesiredManagedResourceInstance(
...
+				// We'll make the same assumption as [ResourceInstanceObjectFullSrc] and
+				// assume we'll never encounter a legacy state snapshot that uses AttrsFlat.
+				SourceStateJSON: prevRoundState.Value.ValueJSON,
+				// SourceStateFlatmap:    prevRoundState.AttrsFlat,
+				SourcePrivate:  prevRoundState.Private,
+				TargetTypeName: inst.Addr.Resource.Resource.Type,
+			}
+			resp := providerClient.MoveResourceState(moveCtx, req)
+			diags = diags.Append(resp.Diagnostics)
+			// TODO this tracer bit is copypasta for upgrade instance, IDK if it's actually legit...
```
lbordowitz, line 184: Am I doing tracing right?

internal/engine/planning/plan_managed.go
```diff
@@ -133,146 +133,238 @@ func (p *planGlue) planDesiredManagedResourceInstance(
...
-				ret.StateDependencies.Add(instAddr.CurrentObject())
+			stateSaveObj := inst.Addr.CurrentObject()
+			if impliedMove {
+				// due to a quirk in how moves are handled,
+				// if it's an implied move, we save state
+				// in the prevAddr instead of current.
+				// TODO: is this true?
+				// TODO: is this actually a "current object"?
+				stateSaveObj = prevRunAddr.CurrentObject()
 			}
```
lbordowitz, line 334: I could use a bit of advice for this, as mentioned in the description.

internal/lang/eval/config_plan_move.go
```diff
@@ -0,0 +1,350 @@
...
+// FindAddressesMovedToHere returns all of the addresses that might be moved to this address,
+// that is, has a series of move statements with From addresses that eventually lead to
+// addr as a To address.
+//
+// Importantly, this function assumes the underlying move statement graph
+// has no cycles. Be sure to check for that before calling this method, or
+// it may take a while to return.
+//
+// A flag is returned and set to true if the move was ambiguous.
+func (o *PlanningOracle) FindAddressesMovedToHere(ctx context.Context, addr addrs.AbsResourceInstance) ([]addrs.AbsResourceInstance, tfdiags.Diagnostics) {
```
lbordowitz, line 61: This function is unused! I think it needs to be used in "desired" to run the diagnostic check...

lbordowitz:
I don't understand the [current failures](https://github.com/opentofu/opentofu/actions/runs/30647038732/job/91210862414?pr=4326#step:5:1445), which appear to be related to the execgraph and not anything that I've changed on my end.

apparentlymart:
There was already some code in the execution graph builder trying to handle situations that would occur with `moved` blocks, but of course those code paths were previously mostly untested because the relevant tests in `package tofu` were being skipped.

The two failures I'm seeing seem like they're for tests that don't involve `moved` blocks though:

```
=== RUN   TestContext2Apply_forcedCBD
    context_apply_test.go:12452: apply errors: Invalid execution graph in plan: Failed to unmarshal the execution graph: invalid waiter in element 20: refers to later element 19.
--- FAIL: TestContext2Apply_forcedCBD (0.00s)
```

```
=== RUN   TestContext2Apply_removeReferencedResource
    context_apply_test.go:12514: apply errors: Invalid execution graph in plan: Failed to unmarshal the execution graph: invalid waiter in element 12: refers to later element 11.
--- FAIL: TestContext2Apply_removeReferencedResource (0.00s)
```

Running each one of those tests separately with `TF_LOG=trace` enabled would print out the execution graph and so might give a clue as to what's going on. I'll give that a try on my system here using your branch from this PR and see if I can quickly explain what's going wrong here.

---

**EDIT:** Focusing just on `TestContext2Apply_forcedCBD` to start, the plan phase of that test produces the following execution graph:

```
  r[0] = ResourceInstancePrior(test_instance.a);
  r[1] = ResourceInstanceCurrentMeta(test_instance.a, r[0]);
  r[2] = ResourceInstanceDesired(r[1]);
  r[3] = ManagedFinalPlan(r[1], r[2], nil, v[0]);
  r[4] = ManagedFinalPlan(r[1], nil, r[0], v[1]);
  r[5] = ManagedPrepareDepose(r[4], "364119e4");
  r[6] = ManagedPerformDepose(r[0], r[5], await(r[3]));
  r[7] = ManagedApply(r[3], r[6], await());
  r[8] = ManagedApply(r[5], nil, await(r[7], r[9], nil));
  r[9] = ResourceInstancePrior(test_instance.b);
  r[10] = ResourceInstanceCurrentMeta(test_instance.b, r[9]);
  r[11] = ResourceInstanceDesired(r[10]);
```

I manually translated that into Mermaid graph syntax:

```mermaid
graph BT

r0

r1 --> r0

r2 --> r1

r3 --> r1
r3 --> r2

r4 --> r1
r4 --> r0

r5 --> r4

r6 --> r0
r6 --> r5
r6 -.-> r3

r7 --> r3
r7 --> r6

r8 --> r5
r8 -.-> r7
r8 -.-> r9

r10 --> r9

r11 --> r10
```

It doesn't seem like there's actually a cycle in this graph, so I'm guessing this is actually a bug in the graph marshaling/unmarshaling code rather than in the actual graph builder.

I'm going to continue debugging this, but since neither of these tests seem directly relevant to `moved`-related functionality I suggest adding an `ExperimentalBugCircularReference` skip to each of these tests for now so that this doesn't block your work, and then I'll try to fix whatever's going on here in a separate PR.

**EDIT AGAIN:** The fix is in https://github.com/opentofu/opentofu/pull/4425, which is now merged.

internal/engine/planning/execgraph_managed.go
```diff
@@ -47,7 +47,7 @@ func (b *execGraphBuilder) ManagedResourceInstanceSubgraph(
 	if (plannedChange.Action == plans.Delete || plannedChange.Action == plans.Forget) && !plannedChange.After.IsNull() {
 		panic(fmt.Sprintf("change for %s has action %s but non-null planned new value", plannedChange.PrevRunAddr, plannedChange.Action))
 	}
-	if plannedChange.Action != plans.Create && plannedChange.Action != plans.Delete && plannedChange.Action != plans.Forget && (plannedChange.Before.IsNull() || plannedChange.After.IsNull()) {
+	if plannedChange.Action != plans.Create && plannedChange.Action != plans.Delete && plannedChange.Action != plans.Forget && plannedChange.Action != plans.NoOp && (plannedChange.Before.IsNull() || plannedChange.After.IsNull()) {
```
apparentlymart, line 50: I'm curious about this change because the behavior I'm accustomed to is that `NoOp` only appears for something that already exists (and therefore its current value is not null), and that it gets populated with both `Before` and `After` set to the same value (and so neither should be null). In other words, `NoOp` is traditionally modeled the same as `Update` except that the before and after values are equal.

By this I'm not meaning to say that this change is wrong, but just that it seems to be allowing for something that was impossible in the traditional runtime as far as I know, and so I'm wondering what is different in the new runtime that causes this situation (NoOp with at least one of its values being null) to now be possible.

internal/engine/planning/execgraph_managed.go
```diff
@@ -290,6 +287,20 @@ func (b *execGraphBuilder) managedResourceInstanceSubgraphCreateThenDelete(
...
+) resourceInstanceObjectSubgraph {
+	_, addCreateDep := b.lower.MutableWaiter()
+	_, addDeleteDep := b.lower.MutableWaiter()
+
+	_, _, priorStateRef, _ := b.managedResourceInstanceChangeInputs(plannedChange)
+	return resourceInstanceObjectSubgraph{
+		valueRef:      priorStateRef,
+		addDesiredDep: addCreateDep,
+		addOrphanDep:  addDeleteDep,
+	}
```
apparentlymart, line 301: As noted in my previous comment I'm accustomed to `NoOp` being treated in a similar way as `Update` except that the before/after values are equal and therefore no actually change gets made.

The answer to my curiosity in the previous comment might indirectly answer this question, but: do you think we actually need `addOrphanDep` here, given that `managedResourceInstanceSubgraphCreateOrUpdate` doesn't populate it? My expectation would be that the special cases we have in the graph builder to handle the interactions between deletes and updates would handle interactions between deletes and noops too, but the fact that you needed to make this codepath accept `NoOp` changes with nil values suggests that something about the new runtime is slightly different and that difference might explain why we need to model noop differently from update in this way. 🤔

internal/engine/planning/plan_managed.go
apparentlymart: I don't think the old runtime actually had tracing directly around this operation, but it does seem like a good idea to model it here since it'll mean that when we use this API to drive OpenTelemetry tracing later the `MoveResourceState` provider request would get associated with a span representing the move operation.

The pattern you followed here seems consistent to me: the "Start" function gets an opportunity to create a child `context.Context` if it needs to, which we then use for all of the requests up to and including the corresponding `End`.

Since I don't think we previously modeled this operation as a directly-traceable event it isn't clear to me yet exactly what arguments we'll want for these two events when we eventually hook this up to OpenTelemetry in later work, but of course we can always change the signatures later if we need to since this is an internal API. 😀

internal/engine/planning/plan_managed.go
apparentlymart: As things are currently set up the system is expecting to finally resolve the move by changing the address of the object during the apply phase using the `ManagedChangeAddr` operation, and so what you did here seems intuitive to me: we're effectively upgrading/refreshing the prior state object _in-place_, while leaving it bound to its original address, and then waiting until the apply phase to rebind that upgraded/refreshed object to its final address, along with any other changes we'll be making to that object.

This setup takes a position that is perhaps somewhat arbitrary but I think consistent: during the planning phase we do work to make the prior state match what the provider reports as the real remote system state before we compare with the desired state, and then during the apply phase we try to update both the real remote system state and our own local copy of that information match the desired state.

The address at which an object ought to be tracked is being treated as part of the desired state here, and so changes to that are treated as something the apply phase should handle. The plan phase is just making sure that the prior state object we'd be starting with (before we change either its address or its actual state data) matches the remote system as closely as possible.

This treatment is intentionally a little different to what the original runtime did, since it treated rebinding things to different addresses as an entirely separate phase that happened even _before_ the main planning phase, working around the fact that the overall system was designed on the assumption that resource instance addresses remain constant throughout the plan and apply phases. The new runtime is intentionally designed to expect that addresses might change, meaning that we can now handle address changes in a similar way to how we deal with various other kinds of configuration changes.

Does that reasoning resonate with you?

internal/engine/planning/plan_managed.go
apparentlymart: I think I'm not understanding exactly what case would cause us to get here. Can you say a little more about what situation causes control to reach this point, so I can think more about the consequences of continuing or not continuing here?

internal/engine/planning/plan_managed.go
```diff
@@ -525,16 +642,76 @@ func (p *planGlue) planUnwantedManagedResourceInstanceObject(
...
+
+				// We continue the loop, although technically, we should not
+				// find any more addresses in configuration.
+			}
+		}
+		if !foundAddr && !blockedMove {
+			// no address found. Try an implicit move
+			// TODO a logical question: do we check these implicit moves for EVERY candidate address above?
+			// I'd prefer it if we didn't... but I'm afraid that might match what we're expecting...
+			// Except! Who in the world is actually combining moves like that???
```
apparentlymart, line 692: This is probably just a follow-on uncertainty from the previous comment, but: I'm not sure what "moves like that" this is referring to. Can you say more about what potential situation this code isn't handling? I'd again like to think a little more about what the consequences of this decision might be.

internal/lang/eval/internal/tofu2024/module_instance.go
```diff
@@ -376,3 +379,18 @@ func (c *CompiledModuleInstance) AnnounceAllGraphevalRequests(announce func(work
 		n.AnnounceAllGraphevalRequests(announce)
 	}
 }
+
+// GetMoveStatements implements evalglue.GetMoveStatements.
+func (c *CompiledModuleInstance) GetMoveStatements(ctx context.Context) []refactoring.MoveStatement {
```
apparentlymart, line 384: This is not super important with how the code is currently structured, so I don't mind leaving this as-is if you prefer, but I'd consider making this return `iter.Seq[refactoring.MoveStatement]` and then letting the caller make the decision about whether to collect the items into a slice (using `slices.Collect`) instead of forcing a slice allocation here.

I'm okay with ignoring this for now because I can see that the only real caller of this function _does_ need it to be a slice anyway, and so we wouldn't actually be changing the allocation behavior in any way unless the caller changes to treat the result differently in future.

Mainly I'm just raising this as a general style thing, where in general in this new runtime we've been preferring to use `iter.Seq` instead of a materialized slice in cases like this because in the _original_ runtime there are various cases where it ends up making temporary slices just because that's how an API was designed and those throwaway allocations can unfortunately add up, whereas using `iter.Seq` as our default approach means that the decision to allocate is moved to the specific callers that need it instead of forcing it on all callers. As things currently stand though, it'd be relatively easy to change this to `iter.Seq` later if we need to, and so this isn't urgent.

internal/lang/eval/config_plan_oracle.go
apparentlymart: I think we're effectively making a global assumption that any method which takes a `context.Context` argument is one that can potentially block waiting for another goroutine to do some work, or for an external network request to complete, or both.

So I don't personally feel the need to call this particular case out individually, but I'm curious what other maintainers think since I may be biased due to having spent a lot of time in this code lately and am assuming this general assumption is clearer than it actually is. 🤔 

One related detail that is probably worth documenting is exactly what callers should expect if the instances for the resource cannot be decided at all, such as if a `count`/`for_each`/`enabled` value somewhere in the chain fails to evaluate. It seems like that would cause this to return `false`, which seems fine (because a failure to resolve instances means that we won't plan any actions for the associated instances anyway, and so we just want to return whatever result prevents generating spurious additional errors) but still important for a potential new caller of this function to know and take into account.

apparentlymart (commented):

Some of my comments here are questions to make sure I'm understanding correctly before I do a more detailed study of how this is all fitting together, but I also tried to address some of the questions you'd asked and make some initial observations.

internal/engine/planning/plan_managed.go
lbordowitz: I've thought hard about this, and I don't think any possible situation could bring us here that wouldn't be an error. If `HasAddress` returns true twice across the move chain, it means we would be moving `from` something that exists in configuration. I've amended the comment and code accordingly.

internal/engine/planning/plan_managed.go
lbordowitz: When I say "moves like that", it would be someone both moving their configuration to a new name and changing the `count` or `for_each` in the destination resource. When I wrote that comment, I was expecting this behavior in the current runtime, but it does not happen. It's a bit convoluted, but I'll try to give the situation.

In the current engine, we convert implicit moves into move statements that are then applied like any other move statement. What I'm thinking about is this situation. In the configuration:

```tf
resource "random_pet" "b" {
    count = 1
}

moved {
    from = random_pet.a
    to = random_pet.b
}
```

And in the state is `test_resource.a`, with `NoKey`. However, I do not know why the current engine does not do that!

There's also an issue on whether nested implicit moves work, where someone might change the count across multiple nested modules, but I could not test that appropriately.

internal/engine/planning/execgraph_managed.go
lbordowitz: Now that I think about it, it probably does not. I'll remove it.

internal/engine/planning/execgraph_managed.go
lbordowitz: I had a panic in earlier testing, for which I added this exception. That panic appears to no longer be there; I've removed `NoOp` from this condition.

internal/engine/planning/plan_managed.go
```diff
@@ -364,25 +462,12 @@ func (p *planGlue) planDesiredManagedResourceInstance(
...
-		if cb := tracer.EndManagedResourceInstanceObjectPlanChanges; cb != nil {
-			cb(planChangesCtx, inst.Addr.CurrentObject(), plans.NoOp, refreshedVal, refreshedVal, diags)
-		}
-		return ret, diags
-	}
-
 	plannedAction := plans.Update
 	if prevRoundState == nil {
 		plannedAction = plans.Create
+	} else if eq, _ := planResp.Planned.Value.Equals(refreshedVal).Unmark(); eq.IsKnown() && eq.True() && !forceReplace {
```
cam72cam: Does this skip the `!planResp.RequiresReplace.Empty()` check below?  Even if the plannedVal == refreshedVal, we still care if RequiresReplace is set.

lbordowitz: I'll pop this condition after the "replace" check, that'll make the logic a little simpler, too.

internal/tofu/context_apply_test.go
```diff
@@ -12376,6 +12378,7 @@ resource "test_resource" "a" {
 }
 
 func TestContext2Apply_forcedCBD(t *testing.T) {
+	SkipExperimental(t, ExperimentalBugCircularReference)
```
cam72cam: This is a odd interesting new failure mode

lbordowitz: Let me try to remove it; it's been so long, I forgot why I added it in the first place.

apparentlymart: This is one of the things added to unblock this while I was working on https://github.com/opentofu/opentofu/pull/4326#issuecomment-5145405857, but the fix for that is merged now so this shouldn't be needed after rabasing this onto `main`.

internal/engine/planning/plan_managed.go
apparentlymart: I must admit I'm not 100% sure on this but IIRC the intended behavior for the original implementation is that "implicit" move statements can only be generated for resources that don't have any explicit move statements referring to them, with the idea being that if the author went to the trouble of writing down an explicit rule then they probably know something special is going on and expect the move to be resolved in a special way.

The example you shared is an interesting example though, because if we apply the "author knows that something special is going on" thinking to that then the "special way" this would get resolved is to first rebind any existing instances of `random_pet.a` into `random_pet.b` without changing their instance keys and _only then_ compare that with the desired state. Which means:

- If there's a no-key instance of `random_pet.a` in the prior state, it would move to `random_pet.b` and then OpenTofu would propose to destroy it because that instance address isn't in the desired state.
- If there's already a `random_pet.a[0]` instance then that would become `random_pet.b[0]` and then get planned as normal, producing either a no-op, update, or replace plan depending on how its configuration relates to the prior object.
- If there's a string-based key like `random_pet.a["foo"]` then just like the no-key instance that would first move to `random_pet.b["foo"]` and then OpenTofu would propose to destroy it because that instance address isn't in the desired state.

Whether the first and last items in the above list are likely to match author intent is debatable, but I think if the original runtime _does_ behave how I just described then we should preserve that behavior as closely as possible to start. If either preserving this behavior is tricky _or_ if I'm misremembering the original runtime's behavior then we can think about what might be a good compromise here. :grinning: 

(A subtlety here is that if both `from` and `to` in a `moved` block are no-instance keys then as a special case OpenTofu interprets that as "all instances moved" instead of just "the no-key instance moved". But if the author had written `to = random_pet.b[0]` instead then OpenTofu treats that as a single-instance move and assumes that `from = random_pet.a` refers _only_ to the no-key instance of that resource. This is awkward and we probably ought to have required writing `random_pet.a[*]` to explicitly say "all instances" but I think it's too late to change the meaning of this now.)

apparentlymart (approved):
Thanks for persisting on this! This behavior is quite cross-cuttingly awkward in both the original and now the new runtime so it's inevitably quite hard to keep track of the interactions between everything but I tried to follow through the codepaths here and overall it makes sense to me.

I suspect we'll probably continue fiddling with details of this as we introduce other features since we're intentionally handling moves in a slightly different sequence with everything else in this new implementation and that'll probably have some knock-on effects on how we implement other features, but overall this looks like a good foundation to build from and we can keep iterating on the details as we go.

cam72cam (approved):

PR #4330, opened 2026-07-02
Escape the script_path on the communicator and remote-exec provisioner

yottta:
This fixes a bug in the `remote-exec` provisioner that allowed the `script_path` field to include also additional commands to be executed together (or after) running the script that OpenTofu builds, uploads and execute on the remote host.

ℹ️ Tried to add an e2e test but there is not really a way to do so since this requires an ssh connection to a remote host. If any of the reviewers have a good idea on how to do it, please let me know down below.

## Checklist

### Go checklist

 


cam72cam (approved):

apparentlymart (approved):
This seems like a good solution to me.

One small hazard here is that we might find that someone has actually been depending on this bug in some strange way to cause another command to run, but based on what behavior we saw while experimenting yesterday it doesn't seem like anyone should be doing that without realizing they are doing something very strange and unintended, so I think we should make this change and then we can find a workaround for the unlikely person who was doing something strange with `script_path`.

It would technically be possible to start a test-only SSH server using the functionality from [`golang.org/x/crypto/ssh`](https://pkg.go.dev/golang.org/x/crypto/ssh) and then tell the provisioner to connect to it in a unit test, but I don't think it's really worth it given how infrequently this code gets changed and how small this change is otherwise.

Remember to add a changelog entry before merging!

PR #4332, opened 2026-07-02
New Engine: Root Module Outputs

cam72cam:
This adds full support for root module outputs to the new runtime, including generated changes.

The only question I have is if we should include NoOp changes for outputs. I see that we omit them as unnecessary for resources in the new engine and suspect we should do the same for outputs.

## Checklist

### Go checklist

 


internal/engine/planning/plan.go
```diff
@@ -219,5 +226,113 @@ func buildPlanChanges(
...
+		// removed from both the config and the state.
+		sensitiveChange := sensitiveBefore || valueSensitive
+
+		// strip any marks here just to be sure we don't panic on the True comparison
+		unmarkedVal, _ := value.UnmarkDeep()
+
+		action := plans.Update
+		switch {
+		case value.IsNull() && before.IsNull() &&
+			valueDeprecated == deprecatedBefore:
```
cam72cam, line 264: This seems like it might be a bug from the original deprecated handling, not sure...

internal/lang/eval/config_plan_oracle.go
```diff
@@ -41,6 +46,22 @@ func (o *PlanningOracle) ProviderInstance(ctx context.Context, addr addrs.AbsPro
 	return o.providers.ProviderInstance(ctx, addr)
 }
 
+// RootOutputs produces the root output values and set resource dependencies
+func (o *PlanningOracle) RootOutputs(ctx context.Context) (cty.Value, addrs.Set[addrs.AbsResourceInstance]) {
```
cam72cam: Consider using adapting PlanningResult instead or removing it.

internal/lang/eval/config_apply.go
```diff
@@ -235,6 +235,13 @@ func (o *ApplyOracle) AnnounceAllGraphevalRequests(announce func(workgraph.Reque
...
+func (o *ApplyOracle) RootOutputs(ctx context.Context) cty.Value {
+	ctx = grapheval.ContextWithNewWorker(ctx)
+
+	val, _ := o.root.ResultValuer(ctx).Value(ctx)
+	return configgraph.PrepareOutgoingValue(val)
+}
+
 func (o *ApplyOracle) Close(ctx context.Context) tfdiags.Diagnostics {
-	return o.providers.Close(ctx)
+	return checkAll(ctx, o.root).Append(o.providers.Close(ctx))
```
yottta, line 245: The removed `closeStack` from `internal/engine/applying/operations.go`, held also ephemeral resources. Right before this PR, did it store only providers?
Asking because I see here logic that seems to be related to closing the providers but I see no ephemeral resources closing logic in this PR. Just want to be sure that by removing `closeStack` will not impact the ephemeral resources closing.

cam72cam: The close provider + ephemeral logic was moved out of here a few weeks ago. This was the last vestige of the removed code.

yottta (approved):

apparentlymart (approved):

internal/engine/planning/plan.go
cam72cam: Ah, I misread the code NVM

Issue #4310, opened 2026-06-25
RFC tracker: Linting support


yottta:
This issue is an RFC Tracker of the [Built-in linter RFC](https://github.com/opentofu/opentofu/pull/3999) ([rendered version](https://github.com/opentofu/opentofu/blob/main/rfc/20260406-linting.md)).

---

The intention of this tracker is to be the first iteration of the built-in linter implementation as requested in https://github.com/opentofu/opentofu/issues/2213.
It will handle the initial structure, wiring and basic core linting rules.

---

The list below contains the work items that reflect the agreed actions from the RFC:
- [x] Basic layers and integration #4337
  - Introduce a new diagnostic type
    - Based on the content of **[Lint message rendering](https://github.com/opentofu/opentofu/blob/rfc-linting/rfc/20260406-linting.md#lint-message-rendering)** and of **[Lint diagnostics](https://github.com/opentofu/opentofu/blob/rfc-linting/rfc/20260406-linting.md#lint-message-rendering)**, we will not handle the new type in any specific way right now, but will be handled exactly the same as we handle the warning diagnostics.
    - The only difference will be in how we exclude/include diagnostics rendering if are of the new type and do not match the `-lint=` flag rules.
  - Create a linting rules layer (engine) that can create a generic implementation to run and generate lint diagnostics
    - This part should be also responsible with enforcing the [rule name format](https://github.com/opentofu/opentofu/pull/3999#discussion_r3105900102)
  - Wire everything together (flag + opted-in creation and execution of the rule(s) + rendering)
- [ ] Core rules implementation
  - Create the rest of the proposed basic core rules and create the groups for them:
      - depends_on (https://github.com/opentofu/opentofu/pull/4384)
      - no-type-variable (#4337)
      - quoted reference
      - ineffective equality
      - impure function usage in resource argument
      - count == 0 instead of enabled (#4393)
      - unused inputs or local values (#4466)
- [x] Selective running of linting rules #4337
  - Create the rules inclusion/exclusion logic
      - Use also **[Lint-enabled Hints](https://github.com/opentofu/opentofu/blob/rfc-linting/rfc/20260406-linting.md#lint-message-rendering)** to avoid running lint rules if those will be ignored anyway from the rendering
- [x] Testing
  - Add unit/integration/e2e tests (#4537)
- [x] Documentation (#4499)
    - Document the rules in a new website page
[...]

PR #4337, opened 2026-07-07
Linting - Initial skeleton and the necessary structures to build rules on top

yottta:
Part of #4310

This PR provides the structure and the layers necessary for linting.

Even though the changes should be quite easy to follow I structured the commits in this PR by scope so you can review it commit by commit.

---

This followed as close as possible the requirements agreed in the [RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20260406-linting.md), with a small divergence:
* see the TODO from tfdiags/lint.go - not having the context hints disables the entire linting instead of enabling it [as agreed in the RFC](https://github.com/opentofu/opentofu/blob/main/rfc/20260406-linting.md#lint-enabled-hints)

---

Another thing that I would like to share that I am not sure that it's the way to go with,  it's the changes on the consolidation logic, where additional bitmask options has been added to be able to skip collecting the count of the consolidated diagnostics.
I did it this way to be able to use the same logic to deduplicate linting diagnostics without showing the count since a linting rule could generate multiple diagnostics (during validate, plan and apply phases) but we want to show the diagnostic of a particular rule and a particular source only once without any indication that it was executed multiple times.

The other possible approach that is not explored in this PR and I thought about, is to have a collector/executor propagated through the `context.Context` and have it say "I have no information stored for this combination of ruleID(groupIDs)+source" and "I have the information for this ruleID(groupIDs)+source and I will skip executing it again".
This way we will execute the logic only once (regardless of the phase it is first executed on) which will also improve the performance.

---

For reference, having the changes here executed on the following configuration:
```hcl
variable "in" {
  default = "string"
}
```
results in the following output:
Screenshot 2026-07-07 at 18 22 24

---

Most probably, the rule implemented here for testing will need to be reimplemented since I didn't put much effort on confirming its validity in terms of various use cases so if you have any takes on how that should be implemented, would appreciate your suggestion.

Each of the other rules listed in the [RFC tracker](https://github.com/opentofu/opentofu/issues/4310) will have their own PRs after we have a general architecture accepted.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/command/arguments/lint.go
+++ b/internal/command/arguments/lint.go
@@ -48,23 +48,9 @@ func ParseLintingRules(rules []string) (collections.Set[linting.RuleAddr], colle
 		}
 		t[la] = struct{}{}
 	}
-	for _, r := range collectionsIntersection(include, exclude) {
+	for _, r := range include.Intersection(exclude) {
 		log.Printf("[WARN] Linting rule %q included and excluded in the same time. This might create unwanted behavior", r)
 	}
 
 	return include, exclude
-}
-
-// collectionsIntersection returns all the common elements between two sets.
-func collectionsIntersection(s1, s2 collections.Set[linting.RuleAddr]) collections.Set[linting.RuleAddr] {
-	result := collections.NewSet[linting.RuleAddr]()
-	if len(s1) > len(s2) {
-		s1, s2 = s2, s1
-	}
-	for item := range s1 {
-		if _, exists := s2[item]; exists {
-			result[item] = struct{}{}
-		}
-	}
-	return result
 }
--- a/internal/command/arguments/lint.go
+++ b/internal/command/arguments/lint.go
@@ -1,0 +1,70 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package arguments
+
+import (
+	"log"
+	"strings"
+
+	"github.com/opentofu/opentofu/internal/collections"
+	"github.com/opentofu/opentofu/internal/linting"
+)
+
+// ParseLintingRules is a utility function that coordinates the parsing of a raw string into structured linting rules.
+//
+// The lintingRulesRaw string argument is expected to contain a string separated list of linting rules
+// to be executed or skipped.
+//
+//	Format: rule1,rule2,!rule3,!rule4
+//
+// which will translate to "include rule1 and rule2" and "exclude rule3 and rule4".
+//
+// Any parsing failure will be only logged. Due to the nature of the linting feature, we want to be
+// permissive in this layer since migrating from one OpenTofu version to another some rules might be
+// changed or removed which should not break the existing operators configuration.
+//
+// This function allows for the same rule to be included and excluded but will warn about it. Due to the
+// filtering rules in the internal/tfdiags/lint.go, in situations where the same rule is included and excluded,
+// the inclusion will take precedence.
+func ParseLintingRules(rules []string) (collections.Set[linting.RuleAddr], collections.Set[linting.RuleAddr]) {
+	include, exclude := collections.NewSet[linting.RuleAddr](), collections.NewSet[linting.RuleAddr]()
+	for _, rawRule := range rules {
+		rawRule = strings.TrimSpace(rawRule)
+		if rawRule == "" {
+			continue
+		}
+		t := include
+		rule, found := strings.CutPrefix(rawRule, "!")
+		if found {
+			t = exclude
+		}
+		la, err := linting.ParseRuleAddr(rule)
+		if err != nil {
+			log.Printf("[WARN] Linting rule %q ignored since it's parsing failed: %s", rawRule, err)
+			continue
+		}
+		t[la] = struct{}{}
+	}
+	for _, r := range collectionsIntersection(include, exclude) {
+		log.Printf("[WARN] Linting rule %q included and excluded in the same time. This might create unwanted behavior", r)
+	}
+
+	return include, exclude
+}
+
+// collectionsIntersection returns all the common elements between two sets.
+func collectionsIntersection(s1, s2 collections.Set[linting.RuleAddr]) collections.Set[linting.RuleAddr] {
+	result := collections.NewSet[linting.RuleAddr]()
+	if len(s1) > len(s2) {
+		s1, s2 = s2, s1
+	}
+	for item := range s1 {
+		if _, exists := s2[item]; exists {
+			result[item] = struct{}{}
+		}
+	}
+	return result
+}
--- a/internal/linting/addr.go
+++ b/internal/linting/addr.go
@@ -1,0 +1,85 @@
+// Copyright (c) The OpenTofu Authors
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) 2023 HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package linting
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+const (
+	ruleNamespaceSeparator = ":"
+	ruleCoreNamespace      = "core"
+	ruleDefaultNamespace   = ruleCoreNamespace
+)
+
+// AllRulesGroupID represents the inclusion of all rule and group linting IDs.
+var AllRulesGroupID = RuleAddr{
+	Name: "all",
+}
+
+// RuleAddr represents the identifier for an individual lint rule or for a
+// group of lint rules.
+type RuleAddr struct {
+	Namespace string
+	Name      string
+}
+
+func (ra RuleAddr) String() string {
+	if ra.Namespace == "" {
+		return ra.Name
+	}
+	return fmt.Sprintf("%s:%s", ra.Namespace, ra.Name)
+}
+
+// addrMatchingReg is used to validate the format of the lint rule ID.
+// To summarise this:
+// * both, the namespace and name, must start with a lower-case letter or a digit.
+// * both, the namespace and name, can contain lower-case letters, digits, dashes and underscores.
+// * because the namespace could point to a provider later, it should be able to contain also forward slashes.
+var addrMatchingReg = regexp.MustCompile(`^([a-z0-9]+[a-z0-9_\-/]*:)?[a-z0-9]+[a-z0-9_\-]*$`)
+
+// ParseRuleAddr parses a string into a [RuleAddr] struct.
+// This is used to validate the given identifier and if all the validation passes, it will return the parsed [RuleAddr].
+// The validations applied against the given rule id are explained in details on [addrMatchingReg], but in short, here are
+// some valid formats:
+//   - core:depends_on
+//   - 0namespace:lint_rule_id
+//   - namespace/full/path:0lines
+func ParseRuleAddr(raw string) (RuleAddr, error) {
+	if !addrMatchingReg.Match([]byte(raw)) {
+		return RuleAddr{}, fmt.Errorf("invalid lint rule id %q. It does not match the required format or character restriction", raw)
+	}
+	parts := strings.Split(strings.TrimSpace(raw), ruleNamespaceSeparator)
+	// since the regex above is meant to catch spaces and multiple namespace separators, there is no need to do
+	// additional validations here
+
+	ns := ruleDefaultNamespace
+	ruleName := parts[len(parts)-1] // name is always the last one
+	if len(parts) == 2 {
+		ns = parts[0]
+	}
+	// special case for the "all" value. That is meant to include all of the existing linting rules so that is unnamespaced.
+	if ruleName == AllRulesGroupID.Name {
+		return AllRulesGroupID, nil
+	}
+
+	return RuleAddr{
+		Namespace: ns,
+		Name:      ruleName,
+	}, nil
+}
+
+// MustParseRuleAddr is the same as ParseRuleAddr but it panics if a validation error occurs.
+// This function is recommended to be used in tests and at init time.
+func MustParseRuleAddr(raw string) RuleAddr {
+	res, err := ParseRuleAddr(raw)
+	if err != nil {
+		panic(err)
+	}
+	return res
+}
--- a/internal/linting/addr.go
+++ b/internal/linting/addr.go
@@ -13,8 +13,6 @@ import (
 
 const (
 	ruleNamespaceSeparator = ":"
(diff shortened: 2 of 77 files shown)


internal/tfdiags/lint.go
```diff
@@ -0,0 +1,238 @@
...
+// linting diagnostics.
+// The summary entry for a linting diagnostic will have the following format:
+//
+//	<summary> (<rule_id>[, <groupID1>, <groupID2>, ...)
+func (lm lintMessage) Description() Description {
+	s := strings.TrimSpace(lm.diagnosticBase.summary)
+	if s == "" {
+		s = "<missing summary>"
+	}
+	all := make([]string, len(lm.groupIDs)+1)
```
apparentlymart: I was expecting that we'd include only the primary rule ID in the summary, and not show all of the groups, because I worry this will make the summary lines considerably longer than they typically have been in other situations. 🤔 

I acknowledge that leaving this out means that there won't be any obvious indication of which groups each lint rule belongs to in the UI output, but that felt like the best initial compromise to me with the assumption that we'd have documentation describing what each lint rule does and what groups it belongs to and that would be the main reference for now.

(In the long run I imagined extending the JSON representation of diagnostics to include the lint rule ID and group IDs as separate JSON properties so that alternative UI implementations can choose to show them if they want, but I'm not proposing we do that in the initial release since I want to minimize the amount of new compatibility promises we're making while our understanding of the needs for built-in linting is still very early.)

internal/tfdiags/diagnostic.go
```diff
@@ -36,9 +36,20 @@ type Severity rune
 const (
 	Error   Severity = 'E'
 	Warning Severity = 'W'
+	// LintingWarning is not meant to be used directly. This is used internally in this package for generating linting
+	// diagnostics.
+	LintingWarning Severity = 'L'
```
apparentlymart: Introducing a new severity wasn't part of what I'd been intending to propose in my early draft of the RFC, because I don't think we know yet whether thinking of this as a different severity is the best model.

For example, we previously discussed that some folks might want to treat certain lint failures as errors rather than warnings, at which point we might decide to represent that either as just normal `Error`, or as a new severity `LintingError`, or in some other completely-different way.

For the small amount of code that needs to distinguish between linting warnings and other warnings, type-asserting the diagnostic to the special `tfdiags.LintMessage` type ought to be enough, right?

(My concern about this is mainly in making this distinction visible in the JSON output where it therefore becomes a compatibility promise, and so I'd also be okay with the smaller compromise of making the `jsonentities` package return `"warning"` as the JSON representation of this severity since that means we'd be free to change how we represent lint diagnostics later without breaking backward-compatibility. My concern here is only in keeping as many design options as possible open for future development as we learn more from feedback, not to say that using a severity for this is _definitely wrong_.)

internal/linting/addr.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return fmt.Sprintf("%s:%s", ra.Namespace, ra.Name)
+}
+
+// addrMatchingReg is used to validate the format of the lint rule ID.
+// To summarise this:
+// * both, the namespace and name, must start with a lower-case letter or a digit.
+// * both, the namespace and name, can contain lower-case letters, digits, dashes and underscores.
+// * because the namespace could point to a provider later, it should be able to contain also forward slashes.
+var addrMatchingReg = regexp.MustCompile(`^([a-z0-9]+[a-z0-9_\-/]*:)?[a-z0-9]+[a-z0-9_\-]*$`)
```
apparentlymart, line 42: I feel a little conflicted about treating `foo` as an alias for `core:foo` right now, while we're not yet 100% sure how the naming of will evolve.

Even with the initial functionality this creates an ambiguity between `core:all` and `all`, which means a special case for that in the `ParseRuleAddr` function below.

But thinking forward more: one possible extension later is to allow the author of a root module to include a configuration file that defines their own aliases for various lint rules they commonly use, at which point we might prefer to make the syntax with no colons refer to the top-level rules defined in that file if that turns out to the most common usage pattern. I'm not meaning to say this is _definitely_ something we will do, but I'd prefer to make it mandatory to include a prefix for now so we don't constrain ourselves while we're focused on only a small part of the overall problem. (We can always introduce `core:` as the default prefix later if that still feels right after more of the rest of the design is settled.)

apparentlymart (commented):
The broad shape of this seems good to me!

My comments inline are all about perhaps being more conservative for the first round so that we leave more design options open for later work.

internal/tfdiags/lint.go
yottta: It felt a little bit off to have only the ruleIDs without groupIDs, and having those "invisible from the tool" and visible only through documentation doesn't sound like a good idea to me. That creates a docs dependency on how to use the "tool" (ie: `tofu`).
But I agree with the compatibility promises part so I will remove that for the moment.

yottta: Fixed in 4837425bcc2eb80321fc3b210e47247fce8cba85

internal/tfdiags/diagnostic.go
yottta: Applied this in 08588900b580e01e2ad530244edcacd06d2aa7bf.

One thing that I kinda don't like about this is around lint related diagnostics consolidation logic, which is changed in the same commit.
Because the current implementation generates duplicates of the same rule (because it's executed by all 3 phases: validate, plan, apply), we want to skip the counting during consolidation of the linting diagnostics (to not show that `(and X more similar warnings elsewhere)`).
Because of that, by removing the linting severity, we need to extract the linting diagnostics out of the initial slice to be able to run consolidation separately for them and another consolidation for the regular warning diagnostics.

It works ok and look acceptable but I kinda liked it more before 🤷 

Maybe the whole consolidation thing is the questionable part here? 🤔  
I have another idea on how to do it, and it's described in this PR description:
> The other possible approach that is not explored in this PR and I thought about, is to have a collector/executor propagated through the context.Context and have it say "I have no information stored for this combination of ruleID(groupIDs)+source" and "I have the information for this ruleID(groupIDs)+source and I will skip executing it again".
This way we will execute the logic only once (regardless of the phase it is first executed on) which will also improve the performance.

But I am not yet sure of the possible future side effects or limitations such an approach would introduce 🤔

internal/linting/addr.go
yottta: I really like the idea. Let me give it a shot.

yottta: Changed in 18a577512708c18b3ab2a5f85ba87802efed4ea7

internal/tfdiags/diagnostic.go
yottta: Ok, I put some time into exploring how would it look like to have the above quoted idea implemented. [Here](https://github.com/opentofu/opentofu/compare/linting-first-phase...lint-executor?expand=1) it is.
That commit has a long comment and I recommend to whoever comes here to read it carefully. It's more of an exploration than a finalised and well shaped solution.

internal/plugin/convert/diagnostics.go
```diff
@@ -60,6 +60,8 @@ func ProtoToDiagnostics(ds []*proto.Diagnostic) tfdiags.Diagnostics {
 	for _, d := range ds {
 		var severity tfdiags.Severity
 
+		// TODO linting - at some point, when we will add linting diagnostics to the provider protocol we will have
+		//  to convert that accordingly here. For now, this will be untouched
```
apparentlymart: Since we're just using the warning severity for lint rules for now, I don't think this comment is needed anymore and may cause confusion if we choose a different representation of lint warnings later.

(For example, one design I would consider is to extend the `Diagnostic` message type with fields `LintRuleID` and `LintGroupIDs` and then say that a lint diagnostic is a warning with a non-empty `LintRuleID`. I'm not saying that is necessarily what we will do, but is an example of one way we could design it that could make this comment confusing.)

internal/plugin6/convert/diagnostics.go
```diff
@@ -60,6 +60,8 @@ func ProtoToDiagnostics(ds []*proto.Diagnostic) tfdiags.Diagnostics {
 	for _, d := range ds {
 		var severity tfdiags.Severity
 
+		// TODO linting - at some point, when we will add linting diagnostics to the provider protocol we will have
+		//  to convert that accordingly here. For now, this will be untouched
```
apparentlymart: ...and this one too 😀

apparentlymart (approved):

internal/plugin/convert/diagnostics.go
yottta: Good point. Removed in ee87558b5e4c6931549d8b7ba8c540553dce3ada

internal/plugin6/convert/diagnostics.go
yottta: Good point. Removed in ee87558b5e4c6931549d8b7ba8c540553dce3ada

yottta:
Thanks a lot for your review @apparentlymart 🙌 
Added the `-lint` flag into the help text of the validate, plan and apply commands in 1e5d594189d094a90ce071b10e43be55939048d7 which solved another TODO.

internal/tfdiags/diagnostic.go
yottta: _Unresolved this conversation because I would still want to have this explored. It is not urgent, since the implementation in the current PR works as expected and the only thing that it's questionable is the multiple times execution and the generation of multiple diagnostics for the same (linting rule + configuration source) which leads to specific deduplication before rendering the diagnostics. But this can be improved separately into a future iteration_.

As I managed to work on some rules in the meantime, the other approach shared in the comment above has some limitations that I seen also when I implemented it, but I wanted to have it as a POC for one idea: the context will not carry only the linting rules enabled via the CLI flag, but will carry also a "container" that will be able to know when a particular rule for a particular configuration source has been executed successfully.

The approach shared above, allows a single **check** of the rule, but does not guarantee **at least one successful execution of a particular rule on a particular configuration source**.
Therefore, _if people agrees with storing some dynamic information into the context_ (which I kinda hate TBH), I would change that approach to be more of a "reporting container" where the execution of the linting rules will have to report back the status of the check. For example, during the `validate` phase, a resource might not have all the necessary information to have the linting rule executed, so it could report an error. But during the `plan`/`apply` phase, it might have it, so it should be allowed to run again.
The current implementation in that POC does not care of the status of the linting rule execution, it takes care if it was attempted to be executed only.

internal/command/arguments/lint.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(s1) > len(s2) {
+		s1, s2 = s2, s1
+	}
+	for item := range s1 {
+		if _, exists := s2[item]; exists {
+			result[item] = struct{}{}
+		}
+	}
+	return result
+}
```
lbordowitz: This might be better off as a function in the `collections` package.

internal/linting/core/variable_with_no_type.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/opentofu/opentofu/internal/configs"
+	"github.com/opentofu/opentofu/internal/linting"
+	"github.com/opentofu/opentofu/internal/tfdiags"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// RootVariableWithNoType is a linting rule that checks any root "variable" block for a defined
+// "type". If none is present, then it issues a new linting diagnostic.
+func RootVariableWithNoType(ctx context.Context, vc *configs.Variable) tfdiags.Diagnostic {
+	if !tfdiags.LintRuleEnabled(ctx, variableWithNoTypeRuleID, GroupIDConfusing) {
```
lbordowitz, line 21: It appears that linting rules are filtered twice; once at the point where a linting rule might be added, and a second time whenever diagnostics are returned.

lbordowitz (approved):
The testing is phenomenal, and I actually had to go and delete some of my comments because they were a handled case in the tests.

I like how linting rules are handled with this set up. It looks like the inclusion/exclusion rules are implemented correctly, and I see the path forward for implementing more core rules or even implementing lint rules from a provider.

I have not done any manual testing with this.

internal/linting/core/variable_with_no_type.go
yottta: > ... whenever diagnostics are returned

More like ["whenever diagnostics are rendered"](https://github.com/opentofu/opentofu/blob/linting-first-phase/internal/command/views/view.go#L177-L179). That's why the code linked here has a comment specifically about this.
Theoretically, the rendering time filtering is redundant, but I left it just in case a rule misses to do the filtering before running its logic.
If you think that we should skip doing the rendering time filtering, I would remove it. Maybe use the rendering time filtering to panic or return an error in case a diagnostic for a not included rule is found? This way we can signal an implementation error (ie: forgot about the `LintRuleEnabled` call in the rule implementation) 🤔

internal/command/arguments/lint.go
yottta: Good point! Fixed in aea0d0abb34cf86291e9a4f00798f745698c0c25

internal/linting/core/variable_with_no_type.go
lbordowitz: I'm not perturbed by it, and I hardly think it'll be a bottleneck. This isn't a practical problem, but it bugged me enough to at least comment on it.

internal/tfdiags/diagnostic.go
yottta: After some exchanges with @cam72cam, I decided to go forward with the other approach. Can be seen in https://github.com/opentofu/opentofu/pull/4371.

With that, doesn't need linting diagnostics consolidation and removes the possible performance penalty of running the same linting rule multiple times.

ℹ️ _That PR targets this one so if approved, it will land here and afterwards in `main`_

Yantrio (approved):

internal/command/arguments/lint.go
```diff
@@ -0,0 +1,56 @@
...
+			continue
+		}
+		t := include
+		rule, found := strings.CutPrefix(rawRule, "!")
+		if found {
+			t = exclude
+		}
+		la, err := linting.ParseRuleAddr(rule)
+		if err != nil {
+			log.Printf("[WARN] Linting rule %q ignored since it's parsing failed: %s", rawRule, err)
```
lbordowitz, line 46:
```suggestion
			log.Printf("[WARN] Linting rule %q ignored since its parsing failed: %s", rawRule, err)
```

lbordowitz (commented):
Small typo, nothing crazy.

internal/command/arguments/lint.go
yottta: [PR](https://github.com/opentofu/opentofu/pull/4524) created to fix it. Thanks for catching this.

PR #4338, opened 2026-07-07
Support Provisioners in the new runtime

cam72cam:
Provisioners are a very old concept and therefore do not integrate well into the new engine. I've done what I can to keep the changes here as unintrusive as possible.

The difference in this implementation path (when compared to the rest of the new engine) is that we need to interact with resources directly, not resource instances for the destroy path. Only resource instances that exist within the configuration will have create provisioners, there is no current "config entity" to attach destroy resource instance provisioners to. Therefore we expose configgraph.Resource and the corresponding destroy provisioners.

This exposes a gap in the new engine, where we don't have a good or consistent mechanism for orphaned instances to have configuration attributes attached. This is a larger item to consider that is outside the scope of what is intended to be addressed in this PR.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/applying/tracing.go
+++ b/internal/engine/applying/tracing.go
@@ -52,6 +52,11 @@ type Tracer struct {
 	StartManagedResourceInstanceObjectApply func(ctx context.Context, addr addrs.AbsResourceInstanceObject, priorVal, plannedVal cty.Value) context.Context
 	EndManagedResourceInstanceObjectApply   func(ctx context.Context, addr addrs.AbsResourceInstanceObject, resultVal cty.Value, diags tfdiags.Diagnostics)
 
+	// Provisioner Hooks for Managed Resources
+	StartProvisionInstanceStep func(ctx context.Context, addr addrs.AbsResourceInstance, typeName string) context.Context
+	ProvisionOutput            func(ctx context.Context, addr addrs.AbsResourceInstance, typeName string, line string, configMarks cty.ValueMarks)
+	StopProvisionInstanceStep  func(ctx context.Context, addr addrs.AbsResourceInstance, typeName string, diags tfdiags.Diagnostics)
+
 	////////// Data Resource Applying Events
 
 	// StartDataResourceInstanceRead and EndDataResourceInstanceRead mark the
--- a/internal/lang/eval/internal/tofu2024/compile_resources.go
+++ b/internal/lang/eval/internal/tofu2024/compile_resources.go
@@ -31,20 +31,21 @@ func compileModuleInstanceResources(
 	moduleProviders configgraph.CompileProviderConfigRef,
 	moduleInstanceAddr addrs.ModuleInstance,
 	providers evalglue.ProvidersSchema,
+	provisioners evalglue.ProvisionersSchema,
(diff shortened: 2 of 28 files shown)


internal/engine/applying/tracing.go
```diff
@@ -52,6 +52,11 @@ type Tracer struct {
 	StartManagedResourceInstanceObjectApply func(ctx context.Context, addr addrs.AbsResourceInstanceObject, priorVal, plannedVal cty.Value) context.Context
 	EndManagedResourceInstanceObjectApply   func(ctx context.Context, addr addrs.AbsResourceInstanceObject, resultVal cty.Value, diags tfdiags.Diagnostics)
 
+	// Provisioner Hooks for Managed Resources
+	PreProvisionInstanceStep  func(ctx context.Context, addr addrs.AbsResourceInstance, typeName string) context.Context
+	PostProvisionInstanceStep func(ctx context.Context, addr addrs.AbsResourceInstance, typeName string, diags tfdiags.Diagnostics)
+	ProvisionOutput           func(ctx context.Context, addr addrs.AbsResourceInstance, typeName string, line string, configMarks cty.ValueMarks)
```
apparentlymart:
```suggestion
	StartProvisionInstanceStep  func(ctx context.Context, addr addrs.AbsResourceInstance, provisionerType string) context.Context
	ProvisionOutput           func(ctx context.Context, addr addrs.AbsResourceInstance, provisionerType string, line string, configMarks cty.ValueMarks)
	EndProvisionInstanceStep func(ctx context.Context, addr addrs.AbsResourceInstance, provisionerType string, diags tfdiags.Diagnostics)
```

(Just to follow the same naming scheme as the other start/end fields, and using ordering to imply that `ProvisionOutput` happens between the start and the end events.)

apparentlymart (approved):
As you mentioned in the writeup, there is a more general problem to solve here in modelling the difference between the "desiredness" of a resource instance and the various aspects of its configuration that can appear regardless of whether it's desired (e.g. in `removed` blocks or in resource-wide settings applied to "orphan" instances). 

I've been about a more general solution to that which I expect will rework some of the changes you made to `package applying` here, but I'd like to land this first regardless because it'll be useful to have the provisioner-related tests working in order to help validate that alternative approach, and the main machinery for how we model provisioners in the configuration is unlikely to change significantly.

internal/lang/eval/internal/tofu2024/compile_resources.go
```diff
@@ -88,14 +90,45 @@ func compileModuleInstanceResource(
...
+	var destroyProvisioners []compiledProvisioner
+	if config.Managed != nil {
+		var baseConn hcl.Body
+		if config.Managed.Connection != nil {
+			baseConn = config.Managed.Connection.Config
+		}
+
+		for _, prov := range config.Managed.Provisioners {
+			provComp, provDiags := compileProvisioner(ctx, prov, baseConn, provisioners)
+			diags = diags.Append(provDiags)
```
yottta, line 112: A little bit unrelated with the actual provisioner changes: I see that we append to `diags` but where are these returned/rendered? I see no return, no rendering call. Do I miss something?

yottta (approved):

internal/lang/eval/internal/tofu2024/compile_resources.go
cam72cam: Excellent catch, I missed that.  I'm so used to every function's signature being `return result, diags`

PR #4339, opened 2026-07-07
`exprs.FromValue` instead of `configgraph.Maybe`

apparentlymart:
We've repeatedly run into situations where we want to use a normal Go struct type but still have some way to represent a value being unknown and/or having marks.

Over in `configgraph` we had the interface type `Maybe` that dealt with the unknown-values part of this problem, but it didn't deal with marks and also using an interface type where `nil` represents unknown makes the usage pretty different to our common usage patterns with `cty.Value` and so we've made mistakes with it several times already. That type can also be used only inside the internals of `package eval`, since it's not reachable from the rest of the system.

Therefore this PR introduces a new variant of `configgraph.Maybe` as `exprs.FromValue`, now as a struct type instead of an interface type because that means it can more easily follow familiar usage patterns from `cty.Value`. In addition to representing unknown values it also supports marks, with method names and usage patterns that intentionally match how we handle marks on `cty.Value`.

The second commit then removes the now-redundant `configgraph.Maybe` and updates all of its callers to use `exprs.FromValue` instead. Doing this also allows us to remove some separate `cty.ValueMarks` return values where we previously used that to compensate for `configgraph.Maybe` not being able to represent those. We can now send the marks along with the values they apply to just as we would with `cty.Value`, thereby making it less likely that we'll fail to handle them.

This PR intentionally stops at the boundary of `package eval` and its internal subpackages, and so the exported `eval.DesiredResourceInstance` is still using a pointer to `addrs.AbsProviderInstanceCorrect` which can be `nil` to represent unknown. In a future commit we will switch that to using `exprs.FromValue` too, as the first example of using this in the broader system outside of the evaluator internals.

---


yottta (approved):
Other than the missing unit tests for the `FromValue`, it looks good.

_sorry for the noise with Copilot... tried to retry a failing check and clicked accidentally on a new button GH introduced which started a Copilot session_ 🤦

cam72cam (approved):

PR #4344, opened 2026-07-09
Implement replace_triggered_by in the new runtime

cam72cam:
This functionally reproduces the logic that exists in the current engine, but could be made a lot smarter down the line.

The biggest downside of the current approach is that I had to duplicate evaluate_triggers*.go into the tofu2024 package from the tofu package.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/planning/plan_eval_glue.go
+++ b/internal/engine/planning/plan_eval_glue.go
@@ -11,11 +11,13 @@ import (
 	"iter"
 	"log"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
 
 	"github.com/opentofu/opentofu/internal/addrs"
 	"github.com/opentofu/opentofu/internal/collections"
 	"github.com/opentofu/opentofu/internal/lang/eval"
+	"github.com/opentofu/opentofu/internal/plans"
 	"github.com/opentofu/opentofu/internal/providers"
 	"github.com/opentofu/opentofu/internal/states"
 	"github.com/opentofu/opentofu/internal/tfdiags"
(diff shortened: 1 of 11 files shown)


apparentlymart (approved):
This seems like a reasonable first pass to me!

w.r.t the concern about "duplicating code from `tofu`", I'm personally not concerned about that at all... temporarily duplicating the code is effectively what we're doing for this entire project, even though some of the new code is more similar to the old code than in other cases. If this project is successful then we'll hopefully be deleting the entire `tofu` package eventually anyway, and we have a shared context test suite to help us keep them both matching in the meantime. 🤷‍♂️

internal/engine/planning/plan_eval_glue.go
```diff
@@ -341,3 +343,66 @@ func resourceInstancesFilter(state *states.State, want func(addrs.AbsResourceIns
...
+	var changes []*plans.ResourceInstanceChange
+
+	instance, ok := p.planCtx.resourceInstObjs.Get(ref.ResourceInstance.CurrentObject())
+	if !ok {
+		// This should not happen!
+		diags = diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  `Reference to undeclared resource`,
+			Detail:   fmt.Sprintf(`A resource %s has not been declared`, ref.ResourceInstance),
+			//Subject:  expr.Range().Ptr(),
```
yottta: Do we want a TODO on this? 🤔

internal/engine/planning/plan_managed.go
yottta, line 440: Is this TODO still needed?

internal/lang/eval/internal/tofu2024/compile_resources.go
```diff
@@ -191,6 +192,8 @@ func compileModuleInstanceResource(
 		// allowing us to finalize all of the details for a specific instance
 		// of this resource.
 		CompileResourceInstance: func(ctx context.Context, key addrs.InstanceKey, repData instances.RepetitionData) *configgraph.ResourceInstance {
+			var diags tfdiags.Diagnostics
```
yottta, line 195: Do we return/render this? Doesn't look like we do it.

yottta (approved):

internal/engine/planning/plan_managed.go
cam72cam: Yes, there are a lot more options here, this is just one of the 20 or so enum values.

internal/lang/eval/internal/tofu2024/compile_resources.go
cam72cam: We do, there is an explicit comment in the config valuer

internal/engine/planning/plan_eval_glue.go
cam72cam: In practice, this code should never be hit. All of these attributes should be validated well in advance. This diag exists solely to handle a programming bug.

yottta (approved):

PR #4350, opened 2026-07-10
Implement prevent_destroy in the new runtime

cam72cam:
This is implemented as a post-processing step once we have built the full set of changes.  This may change in the future as we rework the internals of the engine.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/engine/planning/plan_context.go
+++ b/internal/engine/planning/plan_context.go
@@ -7,10 +7,14 @@ package planning
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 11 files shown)


yottta (approved):

apparentlymart (approved):
This seems reasonable to me!

I think it's a little unfortunate that the logic depends on the old `plans.Changes` model which previously we'd been producing primarily for UI use rather than for internal functionality. But since this is a self-contained separate step at the end of the planning process that seems fine for now and we can revisit it when we're ready to decide what might replace the `plans.Plan` and `plans.Changes` model as the new runtime gets integrated more directly in later work.

PR #4353, opened 2026-07-12
fix: preserve non-VCS dot directories in CopyDir

### LINKED ISSUE #4335: Module caching optimization (`CopyDir`) skips dot-prefixed directories
### Community note

> [!TIP]

### OpenTofu Version

```shell
OpenTofu v1.12.2
on windows_amd64
```

https://github.com/skether/opentofu-dotfiles-issue

### Debug Output

https://gist.github.com/skether/2d8acbbca8da760e3aa8cee5b2e2e22f

### Expected Behavior

Both modules should be successfully initialized. The second module should successfully reuse the downloaded repository from the first module and properly locate its subdirectory inside the `.modules/example_b` directory.

### Actual Behavior

The first module (`example_a`) succeeds because it performs a full `git clone` which naturally includes the `.modules/` directory.

The second module (`example_b`) fails with the following error because the `.modules` directory is silently skipped during the internal cache copying process:

```text
Downloading git::https://github.com/skether/opentofu-dotfiles-issue.git?depth=1&ref=main for example_b...
╷
│ Error: Failed to expand subdir globs
│
│ subdir ".modules/module_b" not found
╵
```

### Steps to Reproduce

1. Clone the remote Git repository `https://github.com/skether/opentofu-dotfiles-issue`.
2. Run `tofu init -backend=false`.

### Additional Context

We could reproduce this issue on both developer workstations (Windows x64) and CI runners (Linux x64).

Looking into the `.terraform/modules` directory, we can see that `example_a` contains the whole git repository as expected. While `example_b` only contains a partial copy of the repository. Specifically, it is missing every file and directory whose name starts with a dot.
This matches up with the finding of the root cause analysis below.

**Root Cause Analysis**
*(Note: An LLM (Antigravity - Gemini) was used to help identify the root cause of this problem.)*

The issue originates within the local module caching mechanism:
1. In `internal/getmodules/installer.go` and `internal/getmodules/getter.go` (`reusingGetter`), OpenTofu caches downloaded module repositories by their package address.
2. For the second module, it encounters a cache hit and uses `copy.CopyDir(instPath, prevDir)` to copy the downloaded repository locally instead of running another `git clone`.
3. In `internal/copy/copy_dir.go`, the `CopyDir` function has logic that explicitly skips any file or directory starting with `.`:

```go
// internal/copy/copy_dir.go
if strings.HasPrefix(filepath.Base(path), ".") {
    // Skip any dot files
    if info.IsDir() {
        return filepath.SkipDir
    }
    return nil
}
```

This logic arbitrarily prevents `.modules/` (or any other custom dot-prefixed directory) from being copied, which causes the subdirectory expansion later in the process to fail.

### References

_No response_

SebTardif:
## Summary

`copy.CopyDir` skipped **all** names starting with `.`, including legitimate module package paths such as `.modules`. That breaks module-cache reuse when a remote module uses a dot-prefixed subdirectory: the first install (`git clone`) keeps `.modules`, but a later `CopyDir` from the cache drops it (#4335).

## Change

- Skip only known VCS metadata directories: `.git`, `.hg`, `.svn`, `.bzr`
- Preserve other dot-prefixed paths
- Regression test for `.modules` preserved and `.git` still skipped

Fixes #4335

## Test plan

- [x] `go test ./internal/copy/ -count=1`

### DIFF
--- a/internal/copy/copy_dir.go
+++ b/internal/copy/copy_dir.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"golang.org/x/sync/errgroup"
 )
@@ -34,9 +33,11 @@ import (
 // File and directory modes are not preserved exactly, but the executable
 // flag is preserved for files on operating systems where it is significant.
 //
-// Any "dot files" it encounters along the way are skipped, even on platforms
-// that do not normally ascribe special meaning to files with names starting
-// with dots.
+// Version-control metadata directories (.git, .hg, .svn, .bzr) are skipped so
+// module cache copies do not recurse into nested VCS trees. Other
+// dot-prefixed names (for example a module subdirectory named ".modules")
+// are preserved so remote module layouts that use them still work when
+// CopyDir reuses a previously downloaded tree (see issue #4335).
 //
 // Callers may rely on the above details and other undocumented details of
 // this function, so if you intend to change it be sure to review the callers
@@ -58,8 +59,8 @@ func CopyDir(dst, src string) error {
 			return nil
 		}
 
-		if strings.HasPrefix(filepath.Base(path), ".") {
-			// Skip any dot files
+		base := filepath.Base(path)
+		if isVCSMetaName(base) {
 			if info.IsDir() {
 				return filepath.SkipDir
 			}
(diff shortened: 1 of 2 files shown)


cam72cam:
@SebTardif and @opentofu/maintainers  I've marked this PR as blocked for a few reasons.

1. You have ignored the PR template and have been instructed to keep this on multiple occasions.
2. You have started working on an issue that is still under consideration due to complex legacy decisions
    - You were explicitly asked to not do this as part of your last PR: https://github.com/opentofu/opentofu/pull/4121#issuecomment-4509460555

Yantrio:
I'm closing this, and I need to be direct: We will not be able to accept further contributions from you.

This isn't a first warning. In previous pull requests we have been attempting to discuss with you the importance of working collaboratively with the OpenTofu maintainers and doing so under our contributing guidelines.

These guidelines exist to keep OpenTofu maintainable and to let us coordinate work properly; they aren't optional, and we can't keep making exceptions.

For that reason, we will no longer be accepting pull requests from you. We genuinely appreciated the earlier fixes, and we're sorry it had to end this way.

PR #4354, opened 2026-07-13
Fix validate issues from tests in new runtime

cam72cam:
This fixes a bunch of "null" accessors in the configuration and replaces them with "unknowns", similar to how the new engine functions.

It also implements "depends_on" for output values.

## Checklist

### Go checklist

 

### DIFF
--- a/internal/lang/eval/config_prepare.go
+++ b/internal/lang/eval/config_prepare.go
@@ -104,7 +104,7 @@ func (v *preparationGlue) ResourceInstanceValue(ctx context.Context, ri *configg
 	// reduction on a dag, etc. The main problem seems to be that it allocates
 	// a _lot_ of temporary objects, and so there's lots of GC pressure.
 	proposed := objchange.ProposedNew(
-		schema.Block, cty.NullVal(schema.Block.ImpliedType()), configVal,
+		schema.Block, cty.UnknownVal(schema.Block.ImpliedType()), configVal,
 	)
 
 	// Handle provisioners validate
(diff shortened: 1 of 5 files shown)


internal/lang/eval/config_prepare.go
```diff
@@ -104,7 +104,7 @@ func (v *preparationGlue) ResourceInstanceValue(ctx context.Context, ri *configg
 	// reduction on a dag, etc. The main problem seems to be that it allocates
 	// a _lot_ of temporary objects, and so there's lots of GC pressure.
 	proposed := objchange.ProposedNew(
-		schema.Block, cty.NullVal(schema.Block.ImpliedType()), configVal,
+		schema.Block, cty.UnknownVal(schema.Block.ImpliedType()), configVal,
```
apparentlymart, line 107: I'm curious to know what sort of behavior differences you noticed after making this change.

When I originally wrote this my thinking was that we'd be pretending as though we're validating the result of "creating" the resource instance. In the traditional runtime I don't think there's any situation were we'd ask `objchange.ProposedNew` to treat the state as a completely unknown object, and so I'm not sure how exactly that behaves so I'm not saying this is definitely wrong but I'm also not sure what effect this change would have in practice. 🤔

cam72cam: The computed fields were being set to cty null and causing issues in several validate computed tests.

apparentlymart: Looking more closely at the `objchange.ProposedNew` code I realize that your change here is creating the same effect as `objchange.PlannedUnknownObject`, so I was wrong that there was no situation where we set prior as entirely unknown before -- I forgot that we had that alternative entrypoint relying on it.

But there is at least one test covering this situation, so I guess we can assume it'll do something reasonable even though we're using this in a slightly different way than the traditional runtime would use it:

https://github.com/opentofu/opentofu/blob/3cc4b302082885582662d6287bc2ecf8d8bb91d7/internal/plans/objchange/objchange_test.go#L2001-L2049

apparentlymart (approved):

PR #4361, opened 2026-07-15
planning: Inter-resource-instance dependencies only on `ManagedApply` operations

apparentlymart:

The original concept for the execution graph was that it existed primarily to control the order of externally-visible side effects involving providers, like what we do in `ManagedFinalPlan`, `ManagedApply`, and `DataRead`.

The execution graph _does not_ need to constrain expression evaluation order: whereas the traditional runtime just silently returns the wrong answer if you ask it to evaluate an expression too early, the new architecture's evaluator is built to block until the necessary information is available and so it already ensures that expression evaluation happens in a reasonable order, and so the execution graph can focus only on capturing what the expression graph doesn't: the ordering of separate create/destroy steps when replacing, and change ordering information that might've been more conservative during the planning phase due to unknown values.

Old habits die hard though, and so in recent evolution of the execution graph builder (e.g. in https://github.com/opentofu/opentofu/pull/4316) I started to forget about that distinction and started trying to build the execution graph around expression evaluation instead of focusing only on provider-based side effects. That's caused things to get pretty complicated and caused some bugs where the execution graphs are not correct for various interactions between resource instance subgraphs for different actions (https://github.com/opentofu/opentofu/pull/4342).

This PR is an initial step back toward the original focus, and in fact takes an even stronger position than the original design: it'll be primarily focused on forcing the ordering of `ManagedApply` and `DataRead` in particular, with `ManagedFinalPlan` instead running as soon as expression evaluation provides the data needed to do it. We already treat planning in that way during the planning phase, so it ought to be acceptable to do the same during the apply phase as long as we take care to apply the generated plans in the correct sequence. This simplification will give us the flexibility to wait until we've planned both the create and delete legs of a replace before applying either of them, although this PR doesn't go all the way to that outcome yet.

Already this is enough to make chained `DeleteThenCreate` actions plannable again, but there's still some more work to do for interactions between "replace" subgraphs and update-only subgraphs, which are still not quite right as of this PR but I don't want to change too much at once.

---

There is a small side-quest here to fill a gap that previously existed in the interaction between execgraph and the evaluator, which the old execgraph shape was hiding:

The main execution graph machinery has explicit support for an operation to report that it hasn't completed enough for its dependents to be executed, causing execution to end early. However, resource instance final values travel indirectly through the evaluator and the evaluator can't directly propagate that signal, so we were previously just returning a
generic `cty.DynamicVal` and losing the "stop executing downstream" signal and thus attempting to execute more execgraph operations than we ought to have.

Now we'll approximate that "stop executing downstream" signal by marking the dynamic value as an evaluation error, and then having the `ResourceInstanceDesired` operation at the other end check for that. This therefore propagates the "stop" signal across into the subgraph of any resource instance whose configuration is derived from a failed result, translating it back into the explicit `ok` flag that the execution graph machinery is expecting so that we can halt execution as expected, instead of failing in a confusing way.

This uses the general-purpose `exprs.EvalError` mark instead of a specialized mark intentionally because that means we'll also stop if the downstream config is derived from something whose evaluation failed in some other way too. `exprs.EvalError` is our general way to say "this value is just a placeholder for something that failed", used for other similar purposes inside the evaluator.

---

### DIFF
--- a/internal/engine/planning/execgraph_managed.go
+++ b/internal/engine/planning/execgraph_managed.go
@@ -93,9 +93,8 @@ func (b *execGraphBuilder) managedResourceInstanceSubgraphCreate(
 	// Per the conventions in the old engine, After contains a marked value
 	unmarkedAfter, _ := plannedChange.After.UnmarkDeep()
 	plannedValRef := b.lower.ConstantValue(unmarkedAfter)
-	waitFor, addConfigDep := b.lower.MutableWaiter()
-	desiredInstRef := b.lower.ResourceInstanceDesired(instAddrRef, waitFor)
-	valueRef := b.managedResourceInstanceSubgraphPlanAndApply(
+	desiredInstRef := b.lower.ResourceInstanceDesired(instAddrRef, b.lower.Waiter())
+	valueRef, addConfigDep := b.managedResourceInstanceSubgraphPlanAndApply(
 		desiredInstRef,
 		execgraph.NilResultRef[*exec.ResourceInstanceObject](),
 		plannedValRef,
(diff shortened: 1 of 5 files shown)


internal/engine/planning/execgraph_managed.go
```diff
@@ -93,44 +93,41 @@ func (b *execGraphBuilder) managedResourceInstanceSubgraphCreate(
...
+		valueRef:      valueRef,
+		addDesiredDep: addConfigDep,
 	}
 }
 
 func (b *execGraphBuilder) managedResourceInstanceSubgraphUpdate(
 	plannedChange *plans.ResourceInstanceChange,
 ) resourceInstanceObjectSubgraph {
-	instAddrRef, priorStateRef, addStateDep := b.managedResourceInstanceChangeAddrAndPriorStateRefs(plannedChange)
+	instAddrRef, priorStateRef, _ := b.managedResourceInstanceChangeAddrAndPriorStateRefs(plannedChange)
```
cam72cam: It appears that the third returned value (addStateDep) from `managedResourceInstanceChangeAddrAndPriorStateRefs` is no longer used and could probably be removed.

cam72cam (approved):

PR #4365, opened 2026-07-15
Openbao associated data

### LINKED ISSUE #4261: Add `associated_data` support to `openbao` key provider
The `openbao` key provider is implemented in terms of the OpenBao transit secret engine's [Generate data key](https://openbao.org/api-docs/secret/transit/#generate-data-key) and [Decrypt data](https://openbao.org/api-docs/secret/transit/#decrypt-data) operations.

These endpoints both accept an optional argument `associated_data` that must match across both calls for the original randomly-generated key to be recovered, but that is not currently exposed as part of the configuration schema for the `openbao` key provider.

This issue is complete once the `openbao` key provider supports an additional optional argument `associated_data` that accepts a value of type `string` that must be base64 encoded using the standard base64 alphabet, so it would accept the results of the existing OpenTofu function [`base64encode`](https://opentofu.org/docs/language/functions/base64encode/)[^1], and sends it to the underlying API[^2]. Incorrect base64 encoding is a configuration validation error.

If the `associated_data` argument is not set at all then the call to OpenBao must be equivalent to what would've been sent by previous versions of OpenTofu that did not support this argument, so that previously-created data keys are not invalidated just by upgrading OpenTofu.

No further abstraction is needed. The goal is just to directly expose the remote API's concept of associated data.

[^1]: This base64 encoding is the conventional way to represent arbitrary binary data in the OpenTofu language.
[^2]: The OpenBao documentation does not specify which base64 alphabet and padding convention is expected, but [in the transit backend source code](https://github.com/openbao/openbao/blob/06c839f985817a919253912036dd2da7daa707be/builtin/logical/transit/path_encrypt.go#L68-L70) we can see that it's using [`base64.StdEncoding`](https://pkg.go.dev/encoding/base64#StdEncoding) which implements the same [RFC 4648 section 4](https://tools.ietf.org/html/rfc4648#section-4) parameters used by OpenTofu's base64 functions. Therefore we presumably don't need to decode and then reencode the author's provided value before sending it, but we should still check for correct base64 encoding while validating the configuration.

KrishnaSindhur:
Resolves #4261 

## Checklist

### Go checklist

 


internal/encryption/keyprovider/openbao/provider.go
```diff
@@ -12,17 +12,19 @@ import (
 )
 
 type keyMeta struct {
-	Ciphertext []byte `json:"ciphertext"`
+	Ciphertext     []byte `json:"ciphertext"`
+	AssociatedData string `json:"associated_data,omitempty"`
```
lbordowitz, line 16: Nice!

website/docs/language/state/encryption.mdx
```diff
@@ -289,6 +289,7 @@ This key provider uses the [OpenBao Transit Secret Engine](https://openbao.org/d
 | address                  | OpenBao server address to access the API. OpenTofu can read it from the `BAO_ADDR` environment variable as well. Your system must trust the TLS certificate of the server.  | N/A  | https://127.0.0.1:8200             |
 | transit_engine_path      | Path at which the Transit Secret Engine is enabled in OpenBao. Customize this if you changed the transit engine path.                                                       | N/A  | /transit                           |
 | key_length               | Number of bytes to generate as a key. Available options are `16`, `32` or `64` bytes.                                                                                       | 16   | 32                                 |
+| associated_data          | Base64-encoded sent with data key generation and decryption.                                                                                                                | N/A  | -                                  |
```
lbordowitz: This documentation is unclear: is the user doing the encoding or is OpenTofu?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ ENHANCEMENTS:
 - When OpenTelemetry trace collection is active, OpenTofu now copies any log lines generated by the OpenTelemetry libraries into its own debug log stream that you can activate using the `TF_LOG` environment variable. ([#4285](https://github.com/opentofu/opentofu/issues/4285))
 - When installing provider and module packages from OCI Distribution registries, OpenTofu now tracks separate transient credentials for each repository to support registry implementations that issue repository-scoped tokens.  ([#3316](https://github.com/opentofu/opentofu/issues/3316))
 - The `providers lock` command now supports the argument `-oci-mirror`. The functionality mimics that of the field `repository_template` of `oci_mirror`-block in [`provider_installation`](https://opentofu.org/docs/cli/config/config-file/#provider-installation) with the exception of using a URI template instead of a HCL one.
+- The OpenBao key provider for state encryption now supports an `associated_data` field, allowing a base64-encoded value to be passed to OpenBao Transit with every data key generation and decryption call. ([#4365](https://github.com/opentofu/opentofu/pull/4365)
```
lbordowitz: You're missing the closing paren around the link.

lbordowitz (commented):
This looks great, nicely done!

I was on the fence about whether or not we should expect the user to have a valid Base64 string for their additional data. It looks like that's what the GCP KMS does for its Additional Authenticated Data, so I'm thinking it's fine here, too.

I'll approve once I've run through testing.

KrishnaSindhur:
@lbordowitz there is no test file for `provider` for `openbao` should i add test for provider in this PR?

lbordowitz:
> @lbordowitz there is no test file for `provider` for `openbao` should i add test for provider in this PR?

No need. I'll run compliance tests and also do some manual testing.

website/docs/language/state/encryption.mdx
```diff
@@ -289,6 +289,7 @@ This key provider uses the [OpenBao Transit Secret Engine](https://openbao.org/d
 | address                  | OpenBao server address to access the API. OpenTofu can read it from the `BAO_ADDR` environment variable as well. Your system must trust the TLS certificate of the server.  | N/A  | https://127.0.0.1:8200             |
 | transit_engine_path      | Path at which the Transit Secret Engine is enabled in OpenBao. Customize this if you changed the transit engine path.                                                       | N/A  | /transit                           |
 | key_length               | Number of bytes to generate as a key. Available options are `16`, `32` or `64` bytes.                                                                                       | 16   | 32                                 |
+| associated_data          | Must be a base64-encoded string in config (use `base64encode()` for plain text) sent to OpenBao Transit with data key generation and decryption.                            | N/A  | -                                  |
```
yottta:
```suggestion
| associated_data          | Base64-encoded string sent to OpenBao Transit with data key generation and decryption providing authenticity protection.                          | N/A  | -                                  |
```
Considering the tone used for the other attributes, I would say that this would be more in line.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ ENHANCEMENTS:
 - When OpenTelemetry trace collection is active, OpenTofu now copies any log lines generated by the OpenTelemetry libraries into its own debug log stream that you can activate using the `TF_LOG` environment variable. ([#4285](https://github.com/opentofu/opentofu/issues/4285))
 - When installing provider and module packages from OCI Distribution registries, OpenTofu now tracks separate transient credentials for each repository to support registry implementations that issue repository-scoped tokens.  ([#3316](https://github.com/opentofu/opentofu/issues/3316))
 - The `providers lock` command now supports the argument `-oci-mirror`. The functionality mimics that of the field `repository_template` of `oci_mirror`-block in [`provider_installation`](https://opentofu.org/docs/cli/config/config-file/#provider-installation) with the exception of using a URI template instead of a HCL one.
+- The OpenBao key provider for state encryption now supports an `associated_data` field, allowing a base64-encoded value to be passed to OpenBao Transit with every data key generation and decryption call. ([#4365](https://github.com/opentofu/opentofu/pull/4365))
```
```suggestion
- The OpenBao key provider accepts a new `associated_data` (known as AAD) argument, allowing a base64-encoded value to be passed to OpenBao on every data key generation and decryption call. ([#4365](https://github.com/opentofu/opentofu/pull/4365))
```

yottta (approved):
Besides 2 small nitpicks, this looks clean.

Also ran the tests as described in [`compliance_test.go`](https://github.com/opentofu/opentofu/blob/ba4dca8092f45a7c35864adc1c5d976514a25881/internal/encryption/keyprovider/openbao/compliance_test.go#L21-L47) and everything works as expected.

Thanks for the help on this @KrishnaSindhur! 👏

lbordowitz (approved):
I manually tested this against an OpenBao dev server running on my laptop. Switching associated data worked and was able to decrypt and encrypt appropriately. Well done, @KrishnaSindhur !

